PR #6, opened 2016-11-21
Remove regexp character class ambiguity

dbenhur:
A literal dash '-' in regexp character class must be at beginning or end or escaped to avoid meaning character range.


sanathkr:
Thanks for the PR. I will merge it

PR #17, opened 2016-11-22
Fixed typo as in 'FunctionArn' to 'Arn'

heitorlessa:
Quick fix in the docs that is explained in this Issue: https://github.com/awslabs/serverless-application-model/issues/16


sanathkr:
Thanks for fixing

PR #29, opened 2016-11-29
API Backend sample based on Swagger

awsed:
Simple API based on swagger template without proxy resource that returns API call and SAM Resources information from stage invoke URL


sanathkr:
@awsed Thanks for the PR. I am going to drop this one in favor of #30 because it has a few other elements like CORS, Simple Proxy integration which had been an ask from the community.

PR #62, opened 2017-01-18
Swagger-With-CORS example works as-is

nicknovitski:
Instead of telling people to edit the example with their region and account id, why not automatically pass those values in as stage variables?

### DIFF
--- a/examples/2016-10-31/api_swagger_cors/swagger.yaml
+++ b/examples/2016-10-31/api_swagger_cors/swagger.yaml
@@ -19,8 +19,7 @@ paths:
         responses:
           default:
             statusCode: 200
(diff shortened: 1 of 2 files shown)


sanathkr:
Thanks for the PR! 

Unfortunately, this will not work. When you try to deploy the template, the API endpoint will fail to create because APIGW expects the integration ARN to be in ARN format before stage variables get substituted.

nicknovitski:
Woah!  So you couldn't pass the complete ARN as a stage variable either?  That's what I was going to suggest next.

I came on this because I was trying to make a template with connected `Serverless::Function` and `Serverless::Api` resources.  Is having a mostly-hard-coded ARN really the only way to do that currently?

PR #72, opened 2017-01-25
Indicate API definitions must be on S3

tlvince:
`'DefinitionUri' is not a valid S3 Uri of the form "s3://bucket/key" with optional versionId query parameter.` is returned otherwise


sanathkr:
Thanks for the PR, but in the context of this HowTo this change isn't necessary. The example asks you to run `aws cloudformation package` to upload the code for you and set the S3 URI there.

tlvince:
Ah, didn't spot that. Thanks!

PR #86, opened 2017-02-05
DefinitionUri supports a relative path

JoeMcGuire:
Clarified documentation that DefinitionUri can support a relative path which is less steps than needing uploading the swagger definition to S3 first.


sanathkr:
Thanks for the PR.

So not quite. If you use the AWS CLI command `aws cloudformation package` it will do the uploading job for you. But if you are passing a SAM template directly to CloudFormation, DefinitionUri has to be a S3 URI. Technically, the spec accepts only S3 URI for this property.

Would it be better to just add a note below the table, explaining that if you use a relative path, you should use the CLI to package first?

sanathkr:
We want to keep this doc as source of truth for the SAM Spec which only supports S3 URI. So not taking this PR.

PR #174, opened 2017-08-13
Fixes #173

### LINKED ISSUE #173: simplify examples to use --s3-bucket when running package
As the how-to describes (https://github.com/awslabs/serverless-application-model/blob/master/HOWTO.md) the examples should reflect that the --s3-bucket argument can be used. 

This simplifies the usage.

kimsaabyepedersen:


sanathkr:
Thanks for the PR! But we want to leave it as S3 paths because this examples is focused on true SAM spec which only supports S3 URIs. Feel free to re-open if you disagree.

PR #175, opened 2017-08-14
Update nodejs version in examples to 6.10.

rowanu:
So I can continue to copy and paste examples.


sanathkr:
Perfect! Thanks for doing it.

PR #177, opened 2017-08-16
fix template for helloworld to pass validate in sam-local

teknogeek0:


sanathkr:
Thanks! This is a temporary change, because we will fix the validate command in SAM Local. I'll pull this anyway because it will prevent people from tripping over the validate command in meantime

PR #197, opened 2017-09-14
Fix CORS template to use '200' instead of 200

kienpham2000:
An error occurred (ValidationError) when calling the CreateChangeSet operation: Template format error: [/Resources/ApiGatewayApi/Properties/DefinitionBody/paths//api/networth/options/responses] map keys must be strings; received numeric [200] instead


willyg302:
Verified both the error and that the fix does not produce the error. Thanks for the contribution!

PR #204, opened 2017-09-29
docs: add syntax highlighting to HOWTO.md

shimont:


iph:
Looks awesome. Minor nit: In the git commit, can it be `add` instead of `adding` ? We don't follow our own advice, but it's better to be consistent with our verbage :)

shimont:
@iph Done :) you could squash merge and change the message

PR #246, opened 2018-01-07
Enhance Docs for Safe Deploy

georgmao:
Adding instructions and examples for Lambda Safe Deployments.


sanathkr:
@jfuss is reviewing..

docs/safe_lambda_deployments.rst
```diff
@@ -102,35 +103,43 @@ resource:
...
--  After traffic shifting completes, CodeDeploy will invoke the
-   **PostTraffic Hook** Lambda Function. This is similar to PreTraffic
-   Hook where the function must callback to CodeDeploy to report a
-   Success or Failure. PostTraffic hook is a great place to run
-   integration tests or other validation actions.
--  If everything went well, the Alias will be pointing to the new Lambda
-   Version.
+- Cloudformation pubilshes a new Lambda Version from the new code
+- Since a deployment preference is set, CodeDeploy takes over the job of actually shifting traffic from old version to new version.
+- Before traffic shifting starts, CodeDeploy will invoke the **PreTraffic Hook** Lambda Function. This Lambda function must call back to CodeDeploy with an explicit status of Success or Failure, via the PutLifecycleEventHookExecutionStatus_ API. On Failure, CodeDeploy will abort and report a failure back to CloudFormation. On Success, CodeDeploy will proceed with the specified traffic shifting. Here_ is a sample Lambda Hook function.
```
jfuss: Can you add a note here about making sure the version of the SDK has the `PutLifecycleEventHookExecutionStatus` API. Under the Pre and Post hook section is probably the best place for this data. It would be nice to call out the min version of SDK for Java, Python (version 1.4.8), and Node.

docs/safe_lambda_deployments.rst
```diff
@@ -102,35 +103,43 @@ resource:
...
+- ``Type: Linear10PercentEvery10Minutes`` instructs CodeDeploy to start with 10% traffic on new version and add 10% every 10 minutes. It will complete traffic shifting in 100 minutes.
+- During traffic shifting, if any of the CloudWatch Alarms go to *Alarm* state, CodeDeploy will immediately flip the Alias back to old version and report a failure to CloudFormation.
+- After traffic shifting completes, CodeDeploy will invoke the **PostTraffic Hook** Lambda Function. This is similar to PreTraffic Hook where the function must callback to CodeDeploy to report a Success or Failure. PostTraffic hook is a great place to run integration tests or other validation actions.
+- If everything went well, the Alias will be pointing to the new Lambda Version.
+
+.. _PutLifecycleEventHookExecutionStatus: https://docs.aws.amazon.com/codedeploy/latest/APIReference/API_PutLifecycleEventHookExecutionStatus.html
+
+.. _Here: https://docs.aws.amazon.com/codedeploy/latest/userguide/reference-appspec-file-structure-hooks.html#reference-appspec-file-structure-hooks-section-structure-lambda-sample-function
```
jfuss: I would prefer if this links to the example you added in the PR

docs/safe_lambda_deployments.rst
```diff
@@ -102,35 +103,43 @@ resource:
...
 
 Traffic Shifting Configurations
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 In the above example ``Linear10PercentEvery10Minutes`` is one of several preselected traffic shifting configurations 
-available in CodeDeploy. You can pick the configuration that best suits your application. Here is the complete list:
+available in CodeDeploy. You can pick the configuration that best suits your application. See docs_ for the complete:
+
+.. _docs: https://docs.aws.amazon.com/lambda/latest/dg/automating-updates-to-serverless-apps.html
```
jfuss: Let's point it to the documentation to ours: https://github.com/awslabs/serverless-application-model/blob/master/docs/safe_lambda_deployments.rst#traffic-shifting-configurations

docs/safe_lambda_deployments.rst
```diff
@@ -180,20 +189,39 @@ Hooks are extremely powerful because:
   you can directly invoke the version to run integration tests or even pre-warm the Lambda containers before exposing
   to production traffic.
 
+    NOTE: The event payload delivered to the Hook function will not contain the Lambda ARN to be tested.
+    We recommend adding an Envrionment variable to the Hook function that maintains the current version of the function requiring safe deployments
+
+.. code:: yaml
+
+  CurrentVersion: !Ref MySafeLambdaFunction.Version
```
jfuss: Add a little context here for clarity:

```
Environment:
  Variables:
    CurrentVersion: !Ref MySafeLambdaFunction.Version
```

docs/safe_lambda_deployments.rst
```diff
@@ -180,20 +189,39 @@ Hooks are extremely powerful because:
...
+
 - **Hooks are executed per-function**: Each Lambda function gets its own PreTraffic and PostTraffic hook (technically
   speaking hooks are executed once per DeploymentGroup, but in this case the DeploymentGroup contains only one Lambda
   Function). So you can customize the hooks logic to the function that is being deployed.
 
-Checkout the examples folder for an example for how to create a hook function.
-
     NOTE: If the Hook functions are created by the same SAM template that is deployed, then make sure to turn off
-    traffic shifting deployments for the hook functions. 
+    traffic shifting deployments for the hook functions. Also, you should not rely on SAM to create the default function Role, since it
```
jfuss: This is very misleading. Best practice is to let SAM create the default Role. It is true that the Generated role will not have the 'codedeploy:PutLifecycleEventHookExecutionStatus' but this should be added through the Policies block. We should not be suggesting to complete manage a Role for a function. There are very few use-cases to have customers create the Role and attach it to the function. Please update this section accordingly to use Policies blocks. I would rather just leave out the Role completely.

docs/safe_lambda_deployments.rst
```diff
@@ -180,20 +189,39 @@ Hooks are extremely powerful because:
...
+    will not contain the necessary permissions to call the CodeDepoloy APIs or Invoke your new Lambda function for testing.
+    Instead, either use the Role_ property to specify your own custom Role or use the Policies_ attribute to provide the permissions needed. Finally, use the ``FunctionName`` property to control the exact name of the Lambda function Cloudformation creates. Otherwise, Cloudformation will create your Lambda function with the Stack name and a unique ID added as part of the name.
```
jfuss: Should be suggesting Policies not Role

docs/safe_lambda_deployments.rst
```diff
@@ -180,20 +189,39 @@ Hooks are extremely powerful because:
...
+
+.. _Role: https://github.com/awslabs/serverless-application-model/blob/master/versions/2016-10-31.md#awsserverlessfunction
+.. _Policies: https://github.com/awslabs/serverless-application-model/blob/master/versions/2016-10-31.md#resource-types
+
+.. code:: yaml
+
+    FunctionName: 'CodeDeployHook_preTrafficHook'
+    DeploymentPreference:
+        Enabled: false
+    Role: [your Custom RoleArn Here]
```
jfuss: Same as comment above. Use Policies

docs/safe_lambda_deployments.rst
```diff
@@ -180,20 +189,39 @@ Hooks are extremely powerful because:
...
+    Policies:
+     - [YourPolicyTemplateHere]
```
jfuss: Add the correct Policy that is needed here

examples/2016-10-31/lambda_safe_deployments/template.yaml
```diff
@@ -0,0 +1,33 @@
...
+      DeploymentPreference:
+          Type: Linear10PercentEvery1Minute
+          Hooks:
+            PreTraffic: !Ref preTrafficHook
+
+  preTrafficHook:
+    Type: AWS::Serverless::Function
+    Properties:
+      Handler: preTrafficHook.handler
+      Role: [arn To IAM Role With Code DeployAndInvoke Permissions]
```
jfuss: Use Policies to make this a complete example

docs/safe_lambda_deployments.rst
```diff
@@ -102,35 +103,43 @@ resource:
           PreTraffic: !Ref PreTrafficLambdaFunction
           PostTraffic: !Ref PostTrafficLambdaFunction
 
+  PreTrafficLambdaFunction:
+    Type: AWS::Serverless::Function
+    Properties:
+      Handler: preTrafficHook.handler
+      Role: [arn To IAM Role With Code DeployAndInvoke Permissions]
```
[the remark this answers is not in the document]
georgmao: Agreed. Done.

docs/safe_lambda_deployments.rst
```diff
@@ -102,35 +103,43 @@ resource:
...
+      Runtime: nodejs6.10
+      Tracing: Active
```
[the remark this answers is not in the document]
georgmao: Removed.

docs/safe_lambda_deployments.rst
georgmao: Done. Policies added.

docs/safe_lambda_deployments.rst
georgmao: Done. Policies added. Role removed.

docs/safe_lambda_deployments.rst
georgmao: Done. Policies added.

docs/safe_lambda_deployments.rst
georgmao: Done. Policies added.

examples/2016-10-31/lambda_safe_deployments/template.yaml
georgmao: Done. Policies added.

docs/safe_lambda_deployments.rst
georgmao: Note added

docs/safe_lambda_deployments.rst
georgmao: now linked to the sample function in PR

docs/safe_lambda_deployments.rst
georgmao: Changed

docs/safe_lambda_deployments.rst
georgmao: Updatd

docs/safe_lambda_deployments.rst
```diff
@@ -61,7 +61,7 @@ property to your ``AWS::Serverless::Function`` resource:
 This will:
 
 - Create an Alias with ``<alias-name>`` 
-- Create & publishe a Lambda version with the latest code & configuration 
+- Create & publishes a Lambda version with the latest code & configuration
```
jfuss: Nit: Creates

docs/safe_lambda_deployments.rst
jfuss: "Also, you should not rely on SAM to create the default function Role, since it will not contain the necessary permissions to call the CodeDepoloy APIs or Invoke your new Lambda function for testing."

They should rely on the Role SAM creates. I understand what you are trying to get across here but I think it can be clearer. How about something like:
The Role SAM generates for a Lambda Execution Role does not include all permissions needed for Per and Post hook functions. To achieve this, use the Policies attribute ....

docs/safe_lambda_deployments.rst
```diff
@@ -184,20 +204,49 @@ Hooks are extremely powerful because:
...
+    FunctionName: 'CodeDeployHook_preTrafficHook'
+    DeploymentPreference:
+        Enabled: false
+    Policies:
+        - Version: "2012-10-17"
+          Statement:
+          - Effect: "Allow"
+            Action:
+              - "codedeploy:PutLifecycleEventHookExecutionStatus"
+            Resource: "*"
```
jfuss: same

docs/safe_lambda_deployments.rst
```diff
@@ -106,35 +106,55 @@ resource:
...
+    Type: AWS::Serverless::Function
+    Properties:
+      Handler: preTrafficHook.handler
+      Policies:
+        - Version: "2012-10-17"
+          Statement:
+          - Effect: "Allow"
+            Action:
+              - "codedeploy:PutLifecycleEventHookExecutionStatus"
+            Resource: "*"
```
jfuss: Is this scope-able at all? We may need to use Conditions here to restrict this to the correct Resource.

examples/2016-10-31/lambda_safe_deployments/template.yaml
```diff
@@ -0,0 +1,41 @@
...
+    Type: AWS::Serverless::Function
+    Properties:
+      Handler: preTrafficHook.handler
+      Policies:
+        - Version: "2012-10-17"
+          Statement:
+          - Effect: "Allow"
+            Action:
+              - "codedeploy:PutLifecycleEventHookExecutionStatus"
+            Resource: "*"
```
jfuss: same

docs/safe_lambda_deployments.rst
georgmao: Agreed - updated.

docs/safe_lambda_deployments.rst
georgmao: Updated!

docs/safe_lambda_deployments.rst
georgmao: I don't think CodeDeploy supports any service-specific context keys. I've restricted the Resource from * to -->

arn:aws:codedeploy:[region]:[accountid]:deploymentgroup:*/*

georgmao: SAM/CF generates something like this
arn:aws:codedeploy:us-west-2:[accountId]:deploymentgroup:[stackname]-ServerlessDeploymentApplication-[uniqueId]/[stackName]-[lambdaFunctionName]DeploymentGroup-[uniqueId]

I don't think we should hardcode these things. What do you think?

docs/safe_lambda_deployments.rst
georgmao: Same comment above.

examples/2016-10-31/lambda_safe_deployments/template.yaml
georgmao: Same comment above.

.gitignore
```diff
@@ -1 +1,2 @@
 .DS_Store
+serverless-application-model.code-workspace
```
jfuss: What is this for?

georgmao: Accidentally checked this in. It's removed now.

PR #268, opened 2018-01-26
Python encryption sample: bugfix in encryption function

1123:
See: https://github.com/awslabs/serverless-application-model/issues/267

### DIFF
--- a/examples/2016-10-31/encryption_proxy/encryption.py
+++ b/examples/2016-10-31/encryption_proxy/encryption.py
@@ -31,14 +31,12 @@ def encrypt(key, message):
 
 def post(event, context):
 
-    try:
-        key_id = os.environ['keyId']
-        message = event['body']
-        if message is None:
-            raise ValueError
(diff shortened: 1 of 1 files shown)


examples/2016-10-31/encryption_proxy/encryption.py
```diff
@@ -31,14 +31,12 @@ def encrypt(key, message):
...
-        message = event['body']
-        if message is None:
-            raise ValueError
-    except KeyError:
+    key_id = os.environ.get('keyId')
+    if key_id is None:
         raise Exception("KMS Key ID not found.")
-    except ValueError:
+
+    message = event.get('body')
```
sanathkr: this needs to be `event.get('body', None)` to prevent a KeyError from being raised with the dictionary does not have the key body

sanathkr: My mistake! This is correct as is

PR #286, opened 2018-02-02
Fix docs reference about hooks naming convention

lauraglasu:
The IAM role CodeDeployServiceRole is created with this default policy that allows running the hook functions:
```
        {
            "Action": [
                "lambda:InvokeFunction"
            ],
            "Resource": "arn:aws:lambda:*:*:function:CodeDeployHook_*",
            "Effect": "Allow"
        }
```
Naming the hook functions as per the current docs prefixed just with`CodeDeploy_` would not match the policy so the CodeDeploy Deployment would not be able to run the hook. I tested and, as expected, it works fine when the hook functions are prefixed with `CodeDeployHook_`.


lauraglasu:
@sanathkr tiny PR when you have a moment

sanathkr:
The example is correct, just the docs is wrong - https://github.com/awslabs/serverless-application-model/blob/master/examples/2016-10-31/lambda_safe_deployments/template.yaml#L37

PR #290, opened 2018-02-07
Add Documentation page for CloudFormation Compatibility

jfuss:


docs/cloudformation_compatibility.rst
```diff
@@ -0,0 +1,97 @@
...
+.. code:: yaml
+
+  LambdaFunction:
+    DependsOn: SomeOtherResources
+    Type: AWS::Serverless::Function
+    ...
+
+CloudFormation Intrinsic Funtions
+-------------------
+Currently, we do not support all Intrinsic Functions for all Property Values. Please see below tables for a details on which Intrinsic Functions can be used on a given field.
```
sanathkr: Can you mention that intrinsics support is partial in AWS::Serverless::* resources, but fully available in other CloudFormation resources

docs/cloudformation_compatibility.rst
```diff
@@ -0,0 +1,97 @@
...
+
+The Condition Function is not currently supported on any ``AWS::Serverless::*`` Resource type
+
+.. code:: yaml
+
+  LambdaFunction:
+    Condition: SomeCondition
```
sanathkr: I would skip the example because Conditions is *not* supported. Examples are useful for things we actually support

docs/cloudformation_compatibility.rst
```diff
@@ -0,0 +1,97 @@
...
+    ...
+
+
+AWS::Serverless::Function
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+======================== ======================== ========================
+     Property Name        Intrinsic(s) Supported          Reaons
+======================== ======================== ========================
+Handler                          Ref Only
```
sanathkr: Not true. Handler, Runtime, Description, MemorySize, Timeout - We support all intrinsics.

docs/cloudformation_compatibility.rst
```diff
@@ -0,0 +1,97 @@
...
+Policies                           All
+Environment                        All
+VpcConfig                          All
+Events                             All
+Tags                               All
+Tracing                            All
+KmsKeyArn                          All
+DeadLetterQueue                    All
+DeploymentPreference               All
+AutoPublishAlias                 Ref Only
```
sanathkr: To be precise: Ref of a CloudFormation Parameter
Reason: Alias resource created by SAM uses a LogicalID <FunctionLogicalId+AliasName>. So SAM either needs a string for alias name, or a Ref to template Parameter that SAM can resolve into a string.

docs/cloudformation_compatibility.rst
```diff
@@ -0,0 +1,97 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+======================== ======================== ========================
+     Property Name        Intrinsic(s) Supported          Reaons
+======================== ======================== ========================
+Name                     All
+StageName                All
+DefinitionUri            All
+DefinitionBody           All
+CacheClusterEnabled      Ref Only
```
sanathkr: All for CacheClusterEnabled & CacheClusterSize

docs/cloudformation_compatibility.rst
```diff
@@ -0,0 +1,97 @@
...
+
+AWS::Serverless::Api
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+======================== ======================== ========================
+     Property Name        Intrinsic(s) Supported          Reaons
+======================== ======================== ========================
+Name                     All
+StageName                All
+DefinitionUri            All
```
sanathkr: DefinitionUri also has Bucket/Key variant like CodeUri. Can you mention that as well?

docs/cloudformation_compatibility.rst
jfuss: Will update

docs/cloudformation_compatibility.rst
jfuss: Fair. Will remove

docs/cloudformation_compatibility.rst
jfuss: I thought I tested this, but must have been doing something else. Will update to reflect what is actually supported

docs/cloudformation_compatibility.rst
jfuss: Updated

docs/cloudformation_compatibility.rst
jfuss: Updated

docs/cloudformation_compatibility.rst
jfuss: Updated

jfuss:
Updated based on feedback.

PR #312, opened 2018-02-22
Add description about another region SNS topic

kazufumi-nishida:


sanathkr:
Actually, can you add a description about this problem to the FAQ docs? - https://github.com/awslabs/serverless-application-model/blob/master/docs/faq.rst

You should be able to grab this info from the Github Issue you referenced. FAQs are a good place to document such quirks

PR #341, opened 2018-03-22
Lambda edge sample

heitorlessa:
*Issue #, if available:*

*Description of changes:*

This is the first step towards making SAM Samples ready to consume for those who already have a S3 bucket. 

This PR primarily fixes DomainName issue spotted by Eric Hammond and adds additional info for SAM first time users on how to package/deploy this one up including S3 bucket creation.

### DIFF
--- a/examples/2016-10-31/lambda_edge/README.md
+++ b/examples/2016-10-31/lambda_edge/README.md
@@ -1,0 +1,40 @@
+## Lambda@Edge sample
+
+This example leverages [SAM Safe Deployments feature](https://github.com/awslabs/serverless-application-model/blob/master/docs/safe_lambda_deployments.rst) in order to ease Lambda@Edge deployments by automatically publishing a new version upon deployments. Since SAM supports standard Cloudformation resources Cloudfront Distribution configuration will be automatically updated as soon as a new Lambda function version is available.
+
+The following Lambda function snippet uses ``AutoPublishAlias`` property which provides an additional property named `<LogicalName>.Version`:
+
+```yaml
+LambdaEdgeFunctionSample:
+    Type: AWS::Serverless::Function
+    Properties:
+        CodeUri: s3://<bucket>/lambda_edge.zip
+        Runtime: nodejs6.10
+        Handler: app.handler
+        Timeout: 5
+        # More info at https://github.com/awslabs/serverless-application-model/blob/master/docs/safe_lambda_deployments.rst
+        AutoPublishAlias: live 
+```
+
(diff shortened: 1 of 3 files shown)


heitorlessa:
@jfuss for a quick review

examples/2016-10-31/lambda_edge/README.md
```diff
@@ -0,0 +1,40 @@
...
+        AutoPublishAlias: live 
+```
+
+We can now configure our [Cloudfront Distribution Lambda Association Property](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-cloudfront-distribution-lambdafunctionassociation.html) to always reference the latest available Lambda Function Version ARN:
+
+```yaml
+CFDistribution:
+    Type: AWS::CloudFront::Distribution
+    Properties:
+    DistributionConfig:
```
jfuss: The DistributionConfig key should be one more indention in.

```
CFDistribution:
  Type: AWS::CloudFront::Distribution
  Properties:
    DistributionConfig:
      Enabled: 'true'
      ...
```

examples/2016-10-31/lambda_edge/template.yaml
```diff
@@ -0,0 +1,79 @@
...
+                - Origin
+              Cookies:
+                Forward: none
+            ViewerProtocolPolicy: allow-all
+
+    LambdaEdgeFunctionSample:
+        Type: AWS::Serverless::Function
+        Properties:
+          CodeUri: s3://<bucket>/lambda_edge.zip
+          Role: !GetAtt LambdaEdgeFunctionRole.Arn
```
jfuss: This is totally valid but unsure if you are aware that SAM can generate this role for you. If will give nearly the same role you have below (-edgelambda.amazonaws.com service). You can then append additional Policies to the generated Role through the `Policies` Property on the `AWS::Serverless::Function`. 

Personally, I like having SAM generate this resource. For this example, I don't think it matters as much but at least letting you know the other option in case you were not aware.

examples/2016-10-31/lambda_edge/README.md
heitorlessa: good catch! I was trying to find out why that was the case as I deployed everything to test, but... that was in the README ;)

Fixed in a new commit!

examples/2016-10-31/lambda_edge/template.yaml
heitorlessa: Yep I'm aware of that and I'd have loved to use except for the Trust Policy :(. 

You can append Policies but you cannot append a Trust Policy - I tried all approaches before keeping on this one. This could've been possible if we allow customers to supply that perhaps through Globals or as an additional property.

If that were to work then I could simply use `LambdaEdgeFunctionSampleRole`.

heitorlessa:
Opening up a new PR as I messed up my branch and it's faster to open a PR from the correct branch at this stage

PR #342, opened 2018-03-22
Replaces s3 placeholder to relative path

heitorlessa:
*Issue #, if available:*

*Description of changes:*

This PR changes SAM Templates under examples from S3 Placeholders to a relative path. All of those that have API Gateway as an example now also have the API under Outputs for the correct URL to trigger Lambda.

This should improve first time experience with SAM when using the examples and all you need now is a S3 bucket to use when packaging.


jfuss:
Sorry, this got lost in the shuffle.

I know we have some documentation around deploying and packaging [here](https://github.com/awslabs/serverless-application-model/blob/master/HOWTO.md#packing-artifacts) but curious if you think it is enough to help people get started with the examples? Should we add a README.md in the examples folder that explains how to use the examples in general?

I know there are README.md's in all the examples but I like uniformity and if all follow this model (the one you add in this PR) then it is less (what does this example do differently from the others I have used). 

Curious on your thoughts.

examples/2016-10-31/api_backend/template.yaml
```diff
@@ -7,7 +7,7 @@ Resources:
     Properties:
       Handler: index.get
       Runtime: nodejs6.10
-      CodeUri: s3://<bucket>/api_backend.zip
+      CodeUri: src/
```
jfuss: I am assuming you tested that these work with the `package` and `deploy` commands?

Could you explain how you verified? Did you deploy and package from the repo or pull them our of here and deploy/package that way?

michaelj-smith: I want to add my support for this PR. When first learning the tools, I modified these files to have:
```
CodeUri: ./
```
...in order to use the `sam local package` and `sam local deploy` commands for deploying the examples. Having an S3 link in the example template adds a complexity that makes it more difficult to follow the examples without prior experience. I like what @heitorlessa has done here.

michaelj-smith: The other nice thing about this usage of `CodeUri` is that it will work out-of-the-box with the SAM Local tools, such as `sam local start-api`

heitorlessa: @jfuss I packaged and deployed each of those when testing them after these changes. Also made me notice that encryption_proxy had a bug as KMS changed the return to Bytes and there wasn't any decode in there -- Added an extra commit to fix that too.

I'll try to create a separate PR to include README with a given format that we can discuss as I agree with that idea too

jfuss: AWESOME!!!!

PR #352, opened 2018-03-31
Update template.yml

sorengoyal:
*Issue #, if available:*
When i tried to build this template, CodeBuild returned the error "YAML_FILE_ERROR Message: wrong number of container tags, expected 1" 

*Description of changes:*
version is required yaml field.
Reference - https://docs.aws.amazon.com/codebuild/latest/userguide/build-spec-ref.html


sanathkr:
This is the SAM template YAML, and not buildspec file which requires version. You have probably configured CodeBuild wrong. 

I am going to close this PR because it’s not relevant

PR #363, opened 2018-04-04
feat(event-source): add Alexa Smart Home as an event source

sanathkr:
*Issue #, if available:*

*Description of changes:*
Live pull request from Twitch. @Twitch Audience: Please review my code and give me feedback.

### DIFF
--- a/samtranslator/model/eventsources/push.py
+++ b/samtranslator/model/eventsources/push.py
@@ -484,6 +484,23 @@ class AlexaSkill(PushEventSource):
         return resources
 
 
+class AlexaSmartHome(PushEventSource):
+    resource_type = 'AlexaSmartHome'
+    principal = 'alexa-connectedhome.amazon.com'
+
+    property_types = {}
+
+    def to_cloudformation(self, **kwargs):
+        function = kwargs.get('function')
+
+        if not function:
+            raise TypeError("Missing required keyword argument: function")
+
+        resources = []
+        resources.append(self._construct_permission(function))
+
+        return resources
+
 class IoTRule(PushEventSource):
     resource_type = 'IoTRule'
     principal = 'iot.amazonaws.com'
--- a/tests/translator/input/alexa_smart_home.yaml
+++ b/tests/translator/input/alexa_smart_home.yaml
@@ -1,0 +1,21 @@
+# File: sam.yml
+# Version: 0.9
+
+AWSTemplateFormatVersion: '2010-09-09'
+Parameters: {}
+Resources:
+  AlexaSkillFunc:
+    Type: AWS::Serverless::Function
+    Properties:
+      CodeUri: s3://sam-demo-bucket/hello.zip
+      Description: Created by SAM
+      Events:
+
+        AlexaSkillEvent:
+          Type: AlexaSmartHome
+
+      Handler: index.handler
+      MemorySize: 1024
+      Runtime: nodejs4.3
+      Timeout: 3
+
--- a/tests/translator/output/alexa_smart_home.json
+++ b/tests/translator/output/alexa_smart_home.json
@@ -1,0 +1,66 @@
+{
+  "AWSTemplateFormatVersion": "2010-09-09",
+  "Parameters": {},
+  "Resources": {
+    "AlexaSkillFuncRole": {
+      "Type": "AWS::IAM::Role",
+      "Properties": {
+        "ManagedPolicyArns": [
+          "arn:aws:iam::aws:policy/service-role/AWSLambdaBasicExecutionRole"
+        ],
+        "AssumeRolePolicyDocument": {
+          "Version": "2012-10-17",
+          "Statement": [
+            {
+              "Action": [
+                "sts:AssumeRole"
+              ],
+              "Effect": "Allow",
+              "Principal": {
+                "Service": [
+                  "lambda.amazonaws.com"
+                ]
+              }
+            }
+          ]
+        }
+      }
+    },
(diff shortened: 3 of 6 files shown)


satblip:
Live pull request, nice :-)

bvanneerven:
This misses unit tests IMHO ;) ;)

tests/translator/input/alexa_smart_home.yaml
```diff
@@ -0,0 +1,21 @@
+# File: sam.yml
+# Version: 0.9
```
brettstack: Remove?

samtranslator/model/eventsources/push.py
```diff
@@ -484,6 +484,23 @@ def to_cloudformation(self, **kwargs):
         return resources
 
 
+class AlexaSmartHome(PushEventSource):
+    resource_type = 'AlexaSmartHome'
+    principal = 'alexa-connectedhome.amazon.com'
+
+    property_types = {}
```
jfuss: Shouldn't this have an 'EventSourceToken' Property?

samtranslator/model/eventsources/push.py
```diff
@@ -484,6 +484,23 @@ def to_cloudformation(self, **kwargs):
         return resources
 
 
+class AlexaSmartHome(PushEventSource):
```
jfuss: We are lacking pretty hard in the unit tests. I love the functional tests but we can do better.

This file is also getting fairly large. Can we split this out into its own class? What are your thoughts around that?

brettstack: Move this to `alexa-smart-home.py` (or `alexasmarthome.py` - need to decided on naming convention)

brettstack: We did this for CloudWatchLogs event and should continue with this approach.

brettstack:
Missing example, documentation, and spec update. I don't think we need to tackle unit tests for event sources in this PR but we should get that going soon #364

jfuss:
I think we also need the validator schema updated too right?

We also didn’t add any tests for intrinsic functions. Are we supporting all or a subset and why?

samtranslator/model/eventsources/push.py
brettstack: Can you update the existing Alexa Skill with this also?

ScriptAutomate:
Any updates on this, as it has been >8 months since the last comment update on here? This is a blocker of mine, as detailed in #613, while I use the following workaround after a `sam-cli deploy`:

* [[Gist] Adding an Alexa Skills Kit Permission w/ InvokeFunction Permissions After Deploying w/ SAM CLI](https://gist.github.com/ScriptAutomate/54ab687ff560e85191db1ff8c732e7f8)

ScriptAutomate:
Additional Comment:

After testing today, it looks like CFN already may support this functionality, though it can only currently be done outside of the serverless transform via standard CFN resources.

```yaml
Description: Alexa App Hello World Lambda Endpoint

Mappings:
    Variables:
        AlexaSkillKit:
            Id: amzn1.ask.skill.xxxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx

Globals:
    Function:
        Timeout: 3

Resources:

    HelloWorldFunction:
        Properties:
            CodeUri: hello_world/
            Handler: app.lambda_handler
            Runtime: python3.6

    HelloWorldFunctionAskPermission:
        Type: AWS::Lambda::Permission
        DependsOn: HelloWorldFunction
        Properties:
            Action: lambda:InvokeFunction
            EventSourceToken: !FindInMap
                - Variables
                - AlexaSkillKit
                - Id
            FunctionName: !GetAtt HelloWorldFunction.Arn
            Principal: alexa-appkit.amazon.com

Outputs:

    HelloWorldFunction:
        Description: "Alexa Hello World Lambda Function ARN"
        Value: !GetAtt HelloWorldFunction.Arn

    HelloWorldFunctionIamRole:
        Description: "Implicit IAM Role created for Alexa Hello World function"
        Value: !GetAtt HelloWorldFunctionRole.Arn

    HelloWorldFunctionAlexaSkillKitId:
        Description: "Alexa Skill Permitted Lambda Invokation Permissions"
        Value: !FindInMap
            - Variables
            - AlexaSkillKit
            - Id
```

I haven't created Alexa Smart Home permissions before, but it looks like the principal would be `alexa-connectedhome.amazon.com` instead of `alexa-appkit.amazon.com`.

So, the above should work in `template.yaml` configurations with `sam-cli` without need of added capability in the serverless transform (for those who have wanted to figure out how to do this without a transform shorthand support).

jlhood:
@kienpham2000 Thanks for reaching out. This PR hasn't been worked on since it was initially open, but we'd love to have this feature in SAM. If you want to take it on, you can open your own PR against the `develop` branch using this PR as an example.

If you're new to contributing to SAM, [this](https://github.com/awslabs/serverless-application-model/blob/develop/DEVELOPMENT_GUIDE.rst) is a good place to start.

ragurney:
Thanks @ScriptAutomate that worked for me! Also @sanathkr was that live stream recorded by any chance? I would love to see your process for contributing to this repo.

ShreyaGangishetty:
@ragurney Please checkout the [this link](https://www.youtube.com/watch?v=uxv1dOExq5U&feature=youtu.be)

jfuss:
Closing this as this has become inactive for quite some time. If this feature is useful to you, I would encourage you to submit a new PR with the changes.

Issue #368, opened 2018-04-05
CodePipelineReadOnlyPolicy is requiring a pipeline name


cplee:
**Description:**

Unable to use the `CodePipelineReadOnlyPolicy` policy template since it requires a single pipeline name via the parameter `PipelineName` and the schema for parameters doesn't allow `*`

**Steps to reproduce the issue:**
1. Create a template with:
``` 
  - CodePipelineReadOnlyPolicy:
       PipelineName: '*'
```

**Observed result:**
> Invalid Serverless Application Specification document. Number of errors found: 1. Errors: Resource with id [PipelineDashboardEventHandler] is invalid. 'CodePipelineReadOnlyPolicy' policy template contains unsupported value for a parameter 'PipelineName'.

**Expected result:**
Should work

sanathkr:
'*' is definitely a valid value and is allowed. I think this might be something to do with YAML formatting.

Can you convert your YAML to JSON (you can do it online here https://www.json2yaml.com/), and post the relevant JSON sections? We want to ensure it looks like `{"CodePipelineReadOnlyPolicy": {"PipelineName": "*"}}`

cplee:
@sanathkr thanks for the quick response.  I converted to JSON and still get the same error.  Complete template is available below:

```
{
  "AWSTemplateFormatVersion": "2010-09-09",
  "Transform": "AWS::Serverless-2016-10-31",
  "Description": "Create pipeline-dashboard application.",
  "Resources": {
    "PipelineDashboardEventHandler": {
      "Type": "AWS::Serverless::Function",
      "Properties": {
        "Description": "Create CloudWatch metrics from CodePipeline events",
        "Handler": "index.handlePipelineEvent",
        "Runtime": "nodejs6.10",
        "CodeUri": "s3://pipeline-dashboard-us-east-1/bc131ad7fd38fd97ecd0387124bb2bda",
        "Events": {
          "PipelineEventRule": {
            "Type": "CloudWatchEvent",
            "Properties": {
              "Pattern": {
                "source": [
                  "aws.codepipeline"
                ],
                "detail-type": [
                  "CodePipeline Pipeline Execution State Change",
                  "CodePipeline Stage Execution State Change",
                  "CodePipeline Action Execution State Change"
                ]
              }
            }
          }
        },
        "Policies": [
          {
            "CloudWatchPutMetricPolicy": {
            }
          },
          {
            "CodePipelineReadOnlyPolicy": {
              "PipelineName": "*"
[...]

PR #369, opened 2018-04-05
Allow wildcard as a valid character in parameters

cplee:
Issue: #368 

Allow `*` in parameters


sanathkr:
Hey @cplee, thanks a lot for the PR! But unfortunately it is not correct

The schema is used to define the structure of policy_templates.json file and not used to validate your SAM template. There might be something else going on with your template and why it didn't work with '*'. 

I am going to close this PR because it is not relevant and work with you on the Issue to resolve it

PR #370, opened 2018-04-06
Update DESIGN.md to include an Overview and entry point for SAM

jfuss:
*Issue #, if available:*

*Description of changes:*
This will add an Overview section to the DESIGN.md file that explains how SAM works and the entry point into the code base. This is useful to help new people in being able to trace through the code base.

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -1,3 +1,11 @@
+# Overview
+
+SAM is invoke/called by the CloudFormation Service. CloudFormation recognises the `Transform: AWS::Serverless-2016-10-31` header and invokes the SAM translator. This will then take your SAM template and expand it
+into a full fledged CloudFormation Template. The CloudFormation Template that is produced from SAM is the template that is executed by CloudFormation to create/update/delete AWS resources.
+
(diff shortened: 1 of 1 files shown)


DESIGN.md
```diff
@@ -1,3 +1,11 @@
+# Overview
+
+SAM is invoke/called by the CloudFormation Service. CloudFormation recognises the `Transform: AWS::Serverless-2016-10-31` header and invokes the SAM translator. This will then take your SAM template and expand it
```
mwarkentin: invoke -> invoked? Or maybe just drop and do “SAM is called by...”

Nr18:
Oh this was what i was looking for but the reason why was because i wanted to know how i can help and add some things like a property LogRetention on the `AWS::Serverless::Function` resource that will created a log group.

A generic description how you can add things like that would help me and others to contribute to SAM.

DESIGN.md
jfuss: Updated in next commit.

jfuss:
@PaulMaddox Will be updated in the next commit.

@Nr18 Been discussing on #sam-dev slack channel on adding the support but this also got me thinking of how to add more docs. Will comment in slack my idea around more docs and surfacing this data.

mwarkentin:
👍

PR #375, opened 2018-04-09
Added validation on make setup target

Nr18:
Added validation to the setup target and raise a clear error when pyenv is not found on your system.

*Issue #, if available:*

*Description of changes:*
Added a check if pyenv could be found on the developers system, if not display a better error describing the problem.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,14 +35,6 @@ activate:
 	$(info [*] Activate virtualenv $(NAME)...)
 	$(shell eval "$$(pyenv init -)" && eval "$$(pyenv virtualenv-init -)" && pyenv activate $(NAME) && pyenv local $(NAME))
 	
-clean:
-	$(info [*] Cleaning up...)
-	@find . | grep -E "(__pycache__|\.pyc|\.pyo$$)" | xargs rm -rf
-	@test -f .coverage && rm .coverage || true
-	@test -f .python-version && rm .python-version || true
-	@test -d ~/.pyenv/versions/$(NAME) && rm -r ~/.pyenv/versions/$(NAME) || true
-	@test -d ~/.pyenv/versions/$(PYTHON_VERSION) && rm -r ~/.pyenv/versions/$(PYTHON_VERSION) || true
-
 init:
 	$(info [*] Install requirements...)
 	@pip install -r requirements-dev.txt -r requirements.txt
@@ -75,6 +67,5 @@ TARGETS
 	dev         Run all development tests after a change.
 	build-docs  Generate the documentation.
 	pr          Perform all checks before submitting a Pull Request.
-	clean       Cleans the virtualenv and all the cached files and test generated files.
 
 endef
--- a/Makefile
+++ b/Makefile
@@ -1,17 +1,42 @@
+NAME=samtranslator27
+PYTHON_VERSION=2.7.14
 PYENV := $(shell command -v pyenv 2> /dev/null)
+UNAME := $(shell uname -s)
 
-install:
-	curl -L https://github.com/pyenv/pyenv-installer/raw/master/bin/pyenv-installer 2> /dev/null | bash
-
-setup: install
-ifndef PYENV
-    $(error "pyenv was not found, please make sure that pyenv is accessible by adding it to your PATH variable in your bash_profile.")
+ifeq ($(UNAME),Darwin)
+	# Mac OS X https://github.com/pyenv/pyenv/wiki/Common-build-problems#build-failed-error-the-python-zlib-extension-was-not-compiled-missing-the-zlib
+	CFLAGS="-O2 -I$$(xcrun --show-sdk-path)/usr/include"
+else
+	# See: https://github.com/pyenv/pyenv/wiki/Common-build-problems#make-your-pythons-a-little-faster
+	CFLAGS="-O2"
 endif
-	pyenv install '2.7.14' 
-	pyenv local 2.7.14 
-	pyenv virtualenv 2.7.14 samtranslator27 
-	pyenv activate samtranslator27 
-	pyenv local samtranslator27 
+
+# Make sure that pyenv is configured properly and that we can use it in our setup target.
+validation:
+ifndef PYENV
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -1,10 +1,17 @@
...
-	pyenv activate samtranslator27
-	pyenv local samtranslator27
+PYENV := $(shell command -v pyenv 2> /dev/null)
+
+install:
+	curl -L https://github.com/pyenv/pyenv-installer/raw/master/bin/pyenv-installer 2> /dev/null | bash
+
+setup: install
+ifndef PYENV
+    $(error "pyenv was not found, please make sure that pyenv is accessible by adding it to your PATH variable in your bash_profile.")
```
sanathkr: Can we generalize this a bit by saying: `make sure pyenv is accessible in your path, (usually by adding to PATH variable in bash_profile, zshrc, or other locations based on your platform)`

I don't want someone to think bash_profile is a requirement when they are using a Z-Shell or Windows

jfuss: Would it be worth add the url for pyenv here as well? So that way, customers can jump right to the pyenv install guide?

Makefile
```diff
@@ -1,10 +1,17 @@
...
-	pyenv virtualenv 2.7.14 samtranslator27
-	pyenv activate samtranslator27
-	pyenv local samtranslator27
+PYENV := $(shell command -v pyenv 2> /dev/null)
+
+install:
+	curl -L https://github.com/pyenv/pyenv-installer/raw/master/bin/pyenv-installer 2> /dev/null | bash
+
+setup: install
+ifndef PYENV
```
jfuss: Will this exit the command at all? If not should we be? We know that it isn't installed and therefore can't use pyenv at the moment.

Makefile
Nr18: I will change it and add a link to the docs, good feedback

Makefile
Nr18: Interesting point now it will install it on the first attempt and then report that it's not configured. I was working on a implementation to get installed properly on my mac but that makes the Makefile complex. Shall i remove the installation and just message that pyenv should be installed properly before you can use it?

Nr18:
So i have removed the installation part from the Makefile and added a note to the `DEVELOPMENT_GUIDE.rst`.

From my Macbook the environment didn't got activated and i needed to place it in one shell in order to get it working:

```$(shell eval "$$(pyenv init -)" && eval "$$(pyenv virtualenv-init -)" && pyenv activate $(NAME) && pyenv local $(NAME))```

Do you have any concerns that this will not work on another device/os?

**Comment by Nr18** [comment]
@jfuss regarding the change from master to develop, i used master due to the following in `CONTRIBUTING.md`

```1. You are working against the latest source on the *master* branch.```

Would you like me to update that to develop?

**Comment by jfuss** [comment]
> Would you like me to update that to develop?

That would be awesome.

> Do you have any concerns that this will not work on another device/os?

For me, no but my shell scripting/makefile abilities are very limited. I will be leaning on @sanathkr on this one.

**Comment by Nr18** [comment]
Branch name has been updated, @sanathkr do you see any concerns regarding my shell activation changes?

File: Makefile [hunk 3]
```diff
@@ -1,10 +1,42 @@
...
+	pyenv local $(PYTHON_VERSION) 
+	pyenv virtualenv $(PYTHON_VERSION) $(NAME)
+	make activate
+	
+# Activate the environment
+activate:
+	$(shell eval "$$(pyenv init -)" && eval "$$(pyenv virtualenv-init -)" && pyenv activate $(NAME) && pyenv local $(NAME))
+	
+# Remove all pyenv versions to start fresh	
+clean:
```
jfuss: Can we add pyenv to this command? I can imagine a clean command that cleans all .pyc files or tox. It will make this command a little clearer on what will happen.

Makefile
```diff
@@ -1,10 +1,42 @@
...
+	rm -rf ~/.pyenv/versions/*
```
jfuss: This is a little scary to me. What is the reason we need to -f this?

Nr18: I removed it but it will ask for a are you sure confirmation on: /.pyenv/versions/2.7.14/lib/libpython2.7.a?

Makefile
Nr18: Not sure how to clean tox things but added the pyc and python cache folders

Nr18:
@jfuss i have added some documentation in the Makefile that will be displayed when you just enter `make` (copied the idea from the https://github.com/aws-samples/cookiecutter-aws-sam-python repo)

I also included the install target again for those who want to use it.

jfuss:
@Nr18 This looks like it was mostly taken from the cookiecutter project right? Generally, this looks good to me.

@heitorlessa Would you mind taking a look at this PR? I am not a makefile magician like you :)

Nr18:
@jfuss yes

Makefile
```diff
@@ -1,25 +1,80 @@
...
+	@pyenv local $(PYTHON_VERSION) 
+	$(info [*] Create virtualenv $(NAME) using python $(PYTHON_VERSION)...)
+	@pyenv virtualenv $(PYTHON_VERSION) $(NAME)
+	@$(MAKE) activate
+	
+activate:
+	$(info [*] Activate virtualenv $(NAME)...)
+	$(shell eval "$$(pyenv init -)" && eval "$$(pyenv virtualenv-init -)" && pyenv activate $(NAME) && pyenv local $(NAME))
+	
+clean:
```
sanathkr: I know this is convenient to clean things, but can we actually remove it from here? This might work for someone setting up SAM Translator with pyenv and other things, but they are by no means required to. I am concerned that we are going beyond the scope.

Thoughts?

Nr18:
So i love having a way to clean everything and start fresh but i have removed it so that the PR can go forward.

sanathkr:
Appreciate making the changes @Nr18. Looks good to me.

sanathkr:
@jfuss can you take a quick look and approve?

jfuss:
Looks good to me. 

@Nr18, I would agree but I think we need to think about what that actually means. Do we include nuking everything or just cleaning .pyc files, etc. When using pyevn nuking is pretty easy and I think I would rather have nuking be explicit by the dev than something that could easily go wrong. A deeper discussion is worth having to see what we can/ should do here though.

jfuss:
@sanathkr can you merge and to the conventional commit? I’m on mobile which makes this somewhat difficult.

PR #394, opened 2018-04-18
Update packaging configuration in setup.py and MANIFEST.in

digioak:
*Description of changes:*

Refactoring packaging configuration, removing `package_data` in `setup.py` and adding JSON files to MANIFEST.in


sanathkr:
Can you update the PR with a better title and description of your problem? I can't evaluate the change without knowing why it is required

digioak:
@brettstack Not according to https://docs.python.org/2/distutils/sourcedist.html.  Files declared in `package_data` is pulled in by default.

heitorlessa:
my 2c: For simplicity I would prefer **not to have** `package_data` and only rely on `MANIFEST` file.

PR #397, opened 2018-04-19
Add FirehoseWritePolicy and FirehoseCrudPolicy

lafiosca:
*Issue #, if available:* no issue

*Description of changes:* I've created two new policy templates in support of Kinesis Firehose use. For my own use case, I just need the FirehoseWritePolicy, which I am currently constructing manually in my SAM template. As this is my first real contribution to the repo, I am open to any changes you think are appropriate to make it fit.

### DIFF
--- a/docs/policy_templates_data/policy_templates.json
+++ b/docs/policy_templates_data/policy_templates.json
@@ -1132,6 +1132,68 @@
           "Resource": "*"
         }]
       }
+    },
+    "FirehoseWritePolicy": {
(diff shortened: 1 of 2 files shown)


lafiosca:
Additionally, I have no idea how to test this behavior locally.

salmanap:
Agreed we need integration tests that should be open sourced as well. Will try to get those out soon. In the meantime, it could be useful to transform your SAM templates with these policies as CFN templates and call "aws cloudformation deploy" to create the stack and test. Will get these reviewed by our security engineers in the meantime.

PR #407, opened 2018-04-25
Update SAM Translator version to 1.5.2

sanathkr:
*Issue #, if available:*

*Description of changes:*
Just released https://pypi.org/project/aws-sam-translator/1.5.2/


brettstack:
This should probably be against master directly

sanathkr:
Nope. Always `develop -> master` and never the other way around. This is like promoting Gamma -> Prod (you would never do the other way around)

PR #412, opened 2018-04-30
[Small fix] SAM Translator install fails on Py3.2.3 or higher

heitorlessa:
… or else ignore it

*Issue #, if available:*

*Description of changes:*

**Functools32** dependency is only required for Python versions older than 3.2.3 otherwise it will fail installation - This PR adds a marker to prevent that from happening


brettstack:
Thanks! @jfuss is working on Py3 support in #428. I think we can close this one in favor of that.

PR #415, opened 2018-05-02
docs(example): improve microservice-http-endpoint-python3 example

bedge:
1) Remove superfluous index.js file (looks to be accidentally duplicated from another example)
2) Include the dynamo DB in the template so one does not need to manually create it, and, sppecify it as the 1st parameter of every API call.
3) Fix the "no params" GET case that was failing trying to interpret None as a list.


examples/apps/microservice-http-endpoint-python3/template.yaml
```diff
@@ -3,17 +3,31 @@ Transform: 'AWS::Serverless-2016-10-31'
...
     Type: 'AWS::Serverless::Function'
     Properties:
       Handler: lambda_function.lambda_handler
-      Runtime: python3.6
       CodeUri: .
       Description: >-
         A simple backend (read/write to DynamoDB) with a RESTful API endpoint
         using Amazon API Gateway.
-      MemorySize: 512
+      #MemorySize: 512
```
brettstack: Remove

examples/apps/microservice-http-endpoint-python3/test.sh
```diff
@@ -0,0 +1,19 @@
+#/bin/bash -e
+
+# Test invocation using httpie (http) in place of curl for syntax simplicity
+#
+# Args: $1: API GW path
+#       $2 json payload data file
+#
+# eg: ./test.sh https://vxldx8ck9i.execute-api.us-west-2.amazonaws.com/Prod/MyResource ./test-payload.json
```
brettstack: Replace your API ID with `YOUR_API_ID` and region with `YOUR_REGION`

Issue #414, opened 2018-05-02
Feature: Kinesis Event Source Mapping - 'Enabled' property as per the console/cloudF.


kurtmaile:
Hi,

Can an 'enabled' property of the kinesis event source mapping be added inline with console and raw cloud formation:

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-lambda-eventsourcemapping.html

At least based on the documentation for SAM this does not appear to be supported?

Thanks

Nr18:
Looks like it's in there: https://github.com/awslabs/serverless-application-model/blob/0bb8ef767eff54640297e52115fd84b0e2b85165/samtranslator/model/lambda_.py#L57 

It's not mentioned in https://github.com/bedge/serverless-application-model/blob/f1d2a879b1bacc52c1f8a31c21f4a5eabb62bd51/versions/2016-10-31.md#kinesis but the following should work:

```
Type: Kinesis
Properties:
  Stream: arn:aws:kinesis:us-east-1:123456789012:stream/my-stream
  StartingPosition: TRIM_HORIZON
  BatchSize: 10
  Enabled: True
```

kurtmaile:
Awesome thanks will close this then

PR #418, opened 2018-05-03
docs: Add Enabled property to the kinesis event object

Nr18:
*Issue #, if available:* #414

*Description of changes:*
- #414 pointed out that the `Enabled` property was missing in the documentation
- Made the file a bit more compatible with the markdown format

### DIFF
--- a/versions/2016-10-31.md
+++ b/versions/2016-10-31.md
@@ -16,7 +16,6 @@ AWS SAM is based on [AWS CloudFormation](https://aws.amazon.com/cloudformation/)
 
 AWS SAM defines a set of objects which can be included in a CloudFormation template to describe common components of serverless applications easily.
 
-
 ## Specification
 
 ### Format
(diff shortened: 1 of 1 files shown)


versions/2016-10-31.md
```diff
@@ -319,6 +321,7 @@ Property Name | Type | Description
 Stream | `string` | **Required.** ARN of the Amazon Kinesis stream.
 StartingPosition | `string` | **Required.** One of `TRIM_HORIZON` or `LATEST`.
 BatchSize | `integer` | Maximum number of stream records to process per function invocation.
+Enabled | `bool` | Indicates whether Lambda begins polling the event source.
```
Nr18: This line was added, the rest are style changes

brettstack: This isn't explicitly defined in the translator https://github.com/awslabs/serverless-application-model/blob/develop/samtranslator/model/eventsources/pull.py#L51. I may be wrong, but I don't think it'll copy over any additional values you define unless they're explicitly defined in the translator.

brettstack:
@Nr18 do you want to make the changes to https://github.com/awslabs/serverless-application-model/blob/develop/samtranslator/model/eventsources/pull.py#L18 or should I close out this PR? It should just be a matter of copying how we do `BatchSize` for instance and updating tests to include Enabled.

RMSD:
@brettstack I'm actually really interested in Enabled myself for Rate/Scheduled lambdas and it feels like Enabled should be supported on all applicable events. Is there a good place I could start and understand how to contribute this change there, or at a higher abstraction level for all events?

I've been hunting for this feature for months and I just stumbled upon https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-lambda-eventsourcemapping.html and this thread today; was going to make a feature request if I hadn't see this PR first 😄.

brettstack:
@RMSD this is just a documentation update. If you want to send a PR with this feature I'd be happy to merge it in. Closing this PR as it's been inactive.

PR #428, opened 2018-05-09
feat: add support for Python 3

jfuss:
*Issue #, if available:*

*Description of changes:*
This is a work in progress. Py3 is not fully supported but sending this to get early feedback while I try to get the rest of the tests running in Py (3.6, specifically).

Details of the commit:
Most of this was drop in replacements for methods:
* .iteritems() -> .items()
* .itervalues() -> values()
* .keys() returns a view into the dictionary in Py3 (list in Py2). So where needed, wrapped the .keys() in a list().
* Updated imports to be relative for modules that were in the same folder.
* Needed to do some trickery with strings due to differences in Py3 and Py2. This only impacts the LogicalIdGenerator but is the most concerning (imo) and biggest potential to cause deployments for Api Gateway when customers don't expect. Please look over this part with a fine tooth comb! 

All tests pass in Py3 except for the majority of the 'functional tests' in the `test_translator.py`. This is caused by:
1. The deepsort is comparing elements of different types which is causing issues in Py3. 
1. The strings we generate for the exceptions are non-deterministic between Py2 and Py3. 

Ways to resolve what is left:
1. Move away from deepsort. This was written to use pytest and asserts. All new code is written using `unittest` where you can compare dict by `self.assertEquals()`. 
1. I need to spend more time and dig a little deeper here. At the moment I don't have a clean way of solving. Suggestions always welcome!


brettstack:
LGTM. Py2 tests are passing which is the main thing. I think we can merge this and address the remaining Py3 test issues in a separate PR.

brettstack:
@sanathkr can we get an approval for this and merge it, knowing that there is still work to be done to get us all the way to Py3?

samtranslator/translator/logical_id_generator.py
```diff
@@ -55,7 +56,8 @@ def get_hash(self, length=HASH_LENGTH):
 
         data_hash = ""
         if self.data_str:
-            data_hash = hashlib.sha1(bytes(self.data_str)).hexdigest()
+            utf_encoded_data_str = str(self.data_str).encode("utf8")
```
jfuss: I think Py2 default encoding is latin1. This may need to be updated to that but unsure.

samtranslator/model/preferences/deployment_preference_collection.py
```diff
@@ -1,9 +1,9 @@
-from deployment_preference import DeploymentPreference
+from .deployment_preference import DeploymentPreference
```
eric-wieser: You should add `from __future__ import absolute_import`, which will make this fail on python 2 as well

tests/translator/test_function_resources.py
```diff
@@ -177,7 +177,7 @@ def test_sam_function_with_disabled_deployment_preference_does_not_add_update_po
         self.intrinsics_resolver_mock.resolve_parameter_refs.assert_called_with(enabled)
         aliases = [r.to_dict() for r in resources if r.resource_type == LambdaAlias.resource_type]
 
-        self.assertTrue("UpdatePolicy" not in aliases[0].values()[0])
+        self.assertTrue("UpdatePolicy" not in list(aliases[0].values())[0])
```
eric-wieser: `list(x.values())[0]` is possibly faster as `next(iter(x.values()))`

eric-wieser:
If you run python2 with the `-3` argument, it will give warnings about any parts of the code that change behavior in python3

samtranslator/model/exceptions.py
```diff
@@ -61,6 +61,9 @@ def __init__(self, logical_id,  message):
         self._logical_id = logical_id
         self._message = message
 
+    def __lt__(self, other):
```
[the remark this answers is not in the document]
adamn: Yeah, that's a magic method in Python-speak: https://rszalski.github.io/magicmethods/

samtranslator/translator/logical_id_generator.py
eric-wieser: What is the type of `self.data_str`?

* If it's `bytes` (`data_str = b'test'`), then `str(data_str)` gives "b'test'" on python 3
* If it's `unicode` (`data_str = u'snowman: \u2603'`), then `str(data_str)` raises a `UnicodeEncodeError`

I think what you actually want here is:

```
unicode = type(u'')   # or from six import unicode
s = self.data_str
if isinstance(s, unicode):
    s = s.encode('utf-8')
```


Calling `str` here is likely to do bad things -

sanathkr: `self.data_str` is a string data - so in py2 it is instance of basestring and in py3 it is unicode. Your solution is correct. We should encode only if it is unicode and skip otherwise.

PR #443, opened 2018-05-22
Correcting import errors when installed with aws-sam-cli 0.3.0

brysontyrrell:
*Issue #, if available:*

I encountered numerous ImportErrors when installing `aws-sam-cli` 0.3.0 and worked through all of them when just trying to invoke the utility to display help text.

*Description of changes:*

Removed relative imports and replaced with absolute imports. Also moved two class variables to global variables in `globals.py` and updated accordingly.


jfuss:
Those are most of the differences that are handled in the Py3 Support PR I linked. Not sure what you mean by cross compatible. We are working on getting Py2 and Py3 support for both SAM CLI, which in turn means the SAM Translator is Py2 and Py3 compatible

Issue #248, opened 2018-01-08
Support API Gateway Logging Metrics, Usage Plans, EndpointConfiguration, and CORS


jfuss:
Parent Tracking Issue for Logging, Usage Plans, CORS etc on API resources (Implicit and Explicit). As part of this, we will move to a model where all API features will be supported in both Implicit & Explicit definitions consistently.

- [x] Logging
- [x] Usage Plans [queued for v1.21.0](https://github.com/awslabs/serverless-application-model/pull/1179)
- [x] Custom Authorizers (AWS::ApiGateway::Authorizer and AWS Cognito) (#49) (#512)
- [x] AWS_IAM 
- [x] Custom Domains [v1.20.0](https://github.com/awslabs/serverless-application-model/releases/tag/v1.20.0)
- [ ] ACM 
- [x] Base Path Mapping
- [x] Regional Endpoints

For deeper details on the different issues, see the appropriate issue that is referenced below.

rabowskyb:
Regarding the cors:true issue:  could you please prioritize?  Many users would like to utilize SAM to build web apps -- it is a huge percentage of use cases, many serveless app in fact are NOT "mobile first".  Without direct CORS support, it is difficult to use SAM for web apps.  Using Swagger with SAM to get around this can be painful and buggy.

jfuss:
@rabowskyb All of these (in the issue) are being prioritize with high priority in the team and are going to be the first set of things we will address this year. The team is actively talking about how to move forward and we are trying to do this as quickly as we can. We don't have timelines to share, as of now though. I hope to share more soon.

samwan:
Does this also mean it will be more straightforward to declare a AWS::ApiGateway::Authorizer resource with Type: "Cognito_user_pool" within a SAM template? There are a few unanswered requests for advice about this topic in the aws-sam-local issue 137 (https://github.com/awslabs/aws-sam-local/issues/137)

steveparker-kcom:
Can you confirm if this issue covers the built in **AWS_IAM** authorizer? It's a bit unclear from reading it but was referred to in the closure of https://github.com/awslabs/serverless-application-model/issues/49

jfuss:
@samwan Yes, we will be supporting AWS::ApiGateway::Authorizer

@steveparker-kcom Yes, this will cover the AWS_IAM authorizer.

I updated the list above to call this explicitly out. (Feel free to update or add more details as well)

johnbest:
[...]

PR #444, opened 2018-05-22
feat(implicit-api): add support for ApiKeyRequired

tylersouthwick:
*Issue #, if available:*
addresses part of #248

*Description of changes:*
Adds ApiKeyRequired to the API Event Properties. If this approach looks good, it might be useful to also add this at the global level.

### DIFF
--- a/samtranslator/plugins/api/implicit_api_plugin.py
+++ b/samtranslator/plugins/api/implicit_api_plugin.py
@@ -190,6 +190,11 @@ class ImplicitApiPlugin(BasePlugin):
         editor = SwaggerEditor(swagger)
         editor.add_path(path, method)
 
+        if "ApiKeyRequired" in event_properties:
+            if event_properties["ApiKeyRequired"]:
+                editor.add_api_key_required(path, method)
+            del event_properties["ApiKeyRequired"]
+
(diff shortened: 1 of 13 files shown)


brettstack:
This is awesome. Can you describe how you tested your changes?

tylersouthwick:
@brettstack I validated it by generating a simple lambda template and deploying it and validating the setting in API gateway.

brettstack:
@tylersouthwick we want to do a deeper design around this before we merge. Specifically want to think about an API that will allow specifying different kinds of auth strategies (API Key, Custom Auth, Cognito Auth, IAM, etc.). If you have thoughts on this we'd love to hear them. Otherwise I'll create an RFC issue around the design when this reaches the top of my list.

brettstack:
Going to merge this into a separate branch for now

bramkoot:
@brettstack any chance this would make it into the 'normal' release now that the auth spec is clear?

brettstack:
We'll need to revise this to make it work with the new Auth support

deleugpn:
Any ETA for when this would be available? Looks like the branch is already in conflict with `develop`. Why wasn't this merged to `develop` instead?

PR #445, opened 2018-05-24
fix: Fixing unit tests to work in Python 3

sanathkr:
*Issue #, if available:*
No issue in this repo but related to https://github.com/awslabs/aws-sam-cli/issues/387 

*Description of changes:*
This change is complimentary to the earlier PR (#428) to make unit tests run in Python 3. There is a lot of interesting nuances in this PR that might not be obvious at start. 

* Translator computes a SHA hash of Swagger dictionary for API Gateway resources. This hash is appended to API Gateway Deployment resource's LogicalID to trigger a new APIGW deployment when Swagger changes. This hash is computed by stringifying swagger dictionary using `str(swagger_dict)`. The string value changes between Python2 and Python3 because of difference between the underlying dictionary implementation in the two versions. Hence the computed hash also changes. In this PR, we update the test expectation by re-computing the SHA within the test for Python3 only. 

* The test used deepsort to sort the output CFN template dictionary and compare against the expectation. In Python2, this worked because you could sort a list with complex values just based on the data type of the value. This behavior is not there in Python3. Because the lists get sorted, they are by definition compared as an unordered list. This is wrong because lists like `Fn::Sub: []` are ordered. To keep the PR focused on py3, we choose to ignore this problem. Instead, we used a custom comparator to mimic Py2's list sorting behavior for Py3.

* Updated Tox and Travis to run tests against both Python versions. All subsequent PRs must support both versions of Python before they can be merged.

* Tox was setting PYTHONHASHSEED to random value. In Py2, this caused the order of elements within a dictionary to change resulting in APIGW Deployment Hash to change between every test run. Changed this to a fixed value in `.tox.ini` file

#### NOTE (only if you run Translator locally)
SAM Translator will work without any issues in Python 3. But if you already had a CloudFormation stack created using SAM Translator Py2, then updating your stack using Py3 translator will re-deploy your API Gateway endpoints. This is due to the first issue described above. If you never ran SAM Translator locally and always relied on CloudFormation, you don't have to worry about this problem. 


tests/translator/test_translator.py
```diff
@@ -29,10 +31,34 @@ def deep_sorted(value):
...
     else:
         return value
 
+def py2_compare_order(obj1, obj2):
+    """
+    This is a comparison function that mimics the results of Py2 Comparisons.
+    """
+    if isinstance(obj1, dict) and isinstance(obj2, dict):
+        obj1 = json.dumps(obj1, sort_keys=True)
+        obj2 = json.dumps(obj2, sort_keys=True)
```
eric-wieser: No way does this mimic the python 2 comparison order

tests/translator/test_translator.py
```diff
@@ -139,8 +165,82 @@ def test_transform_success(self, testcase, partition_with_region):
 
         print(json.dumps(output_fragment, indent=2))
 
+        # Only update the deployment Logical Id hash in Py3.
+        if sys.version_info[0] >= (3):
```
eric-wieser: Parens here do nothing. `[0]` can be better spelt `.major`

tests/translator/test_translator.py
eric-wieser: I think the code you're trying to emulate is:

https://github.com/python/cpython/blob/2.7/Objects/object.c#L767-L807

jfuss: @eric-wieser Rewriting the complete sort order is not the goal here. We wanted to get mimic the functionality of Py2 until we can make a deeper pass at the tests to assert not only the template but the order on which values are in lists (some lists in a template must be ordered). This is called out in the second bullet point in the description of the PR.

sanathkr:
Pushed a change that addresses Eric's feedback

samtranslator/translator/logical_id_generator.py
```diff
@@ -55,9 +56,19 @@ def get_hash(self, length=HASH_LENGTH):
...
+        encoded_data_str = self.data_str
+        if sys.version_info.major == 2:
+            # In Py2, only unicode needs to be encoded.
+            if isinstance(self.data_str, unicode):
+                encoded_data_str = self.data_str.encode('utf-8')
+        else:
+            # data_str should always be unicode on python 3
+            encoded_data_str = self.data_str.encode('utf-8')
+
+        data_hash = hashlib.sha1(bytes(encoded_data_str)).hexdigest()
```
eric-wieser: There is no need for `bytes` here - you've already ensured the data is `bytes` by using `encode` above.

tests/translator/test_translator.py
```diff
@@ -139,7 +190,81 @@ def test_transform_success(self, testcase, partition_with_region):
...
+            output_resources[tuple_to_add[0]] = tuple_to_add[1]
+            del output_resources[logical_id_to_remove]
+
+        # Update any Output References in the template
+        for output_key, output_value in resources.get("Outputs", {}).items():
+            if output_value.get("Ref") in deployment_logical_id_dict:
+                output_value["Ref"] = deployment_logical_id_dict[output_value.get("Ref")]
+
+    def _generate_new_deployment_hash(self, logical_id, dict_to_hash, rest_api_to_swagger_hash):
+        data_bytes = bytes(json.dumps(dict_to_hash, separators=(',', ':'), sort_keys=True).encode("utf8"))
```
eric-wieser: `bytes` here is pointless too - `encode` always returns bytes

fatbasstard:
This is a blocker to enable "good" SAM template linting: https://github.com/awslabs/cfn-python-lint/issues/81

Issue #287, opened 2018-02-04
Allow Inline Lambda Function Definitions


talawahtech:
It would be be great if we could define the code for Lambda functions inline. This functionality already exists for AWS::Lambda::Function resources and it is very convenient for things like Custom Resources that execute as part of the CloudFormation template or really simple Lambda@Edge functions.

Of course it would be understood if there needs to be a limit to the number of characters.

E.g. of current usage in a regular CloudFormation template

```
Resources:
  StaticSiteHelper: 
    Type: "AWS::Lambda::Function"
    Description: "Implements index.html support for directories &redirects to add a trailing slash"
    Properties: 
      Handler: "index.handler"
      Role: !GetAtt LambdaRole.Arn
      Code:
        ZipFile: |
          exports.handler = (event, context, callback) => {
            const request = event.Records[0].cf.request;
            const matchArray = request.uri.match(/[^/]+$/);
              
            if (matchArray === null){
              request.uri = request.uri.replace(/\/$/, '\/index.html');        
              return callback(null, request);
            }	
            else if (matchArray[0].includes('.') === false){
              const response = {
                status: '302',
                headers: { location: [{ key: 'Location',value: request.uri + '/' }] }
              };
              return callback(null, response);
            }

            return callback(null, request);
          };
      Runtime: "nodejs6.10"
      Timeout: "30"
 ```

brettstack:
Added in #447

PR #447, opened 2018-05-25
add support for InlineCode in AWS::Serverless::Function

tylersouthwick:
*Issue #, if available:*
#287 

*Description of changes:*
Adds support for ZipFile lambda definitions

Due to the nature of 'allOf', the schema cannot enforce 'additionalProperties: false' with the current schema approach.

### DIFF
--- a/samtranslator/model/sam_resources.py
+++ b/samtranslator/model/sam_resources.py
@@ -34,7 +34,8 @@ class SamFunction(SamResourceMacro):
         'FunctionName': PropertyType(False, one_of(is_str(), is_type(dict))),
         'Handler': PropertyType(True, is_str()),
         'Runtime': PropertyType(True, is_str()),
-        'CodeUri': PropertyType(True, one_of(is_str(), is_type(dict))),
+        'CodeUri': PropertyType(False, one_of(is_str(), is_type(dict))),
+        'ZipFile': PropertyType(False, one_of(is_str(), is_type(dict))),
         'DeadLetterQueue': PropertyType(False, is_type(dict)),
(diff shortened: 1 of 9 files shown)


brettstack:
Thanks @tylersouthwick. What's the benefit of this over letting SAM just zip the directory for you? I'm not strongly opposed to this, but want to understand customer value/frustration first before adding complexity.

tylersouthwick:
Letting SAM zip the directory requires an s3 bucket to be setup and using the cloudformation cli. The specific use case where I use ZipFile is for really small lambdas such as cloudformation custom resources. I've also used it for small lambdas needed in step function workflows.

parsnips:
I'd like to see this merged as well.  Makes inlined lambda's even shorter.

brettstack:
Can you rename to InlineCode? ZipFile is confusing (I know it's what CloudFormation calls it).

tylersouthwick:
@brettstack I agree that ZipFile was super confusing (when I first saw it I was wondering if it needed to be a base64 encoded zip file). I've updated it to be InlineCode

teknogeek0:
+1 on this. Would for example streamline CFN Custom Resources a bit

timoschilling:
when will this be released?

Kharkovlanok:
Hi, although the commit has been merged, it seems the support in CF isn't effective yet, option InlineCode isn't recognized.

image

timoschilling:
@Kharkovlanok It's only merged in develop not in master

cdugga:
Great, thank you!

Issue #59, opened 2017-01-11
Issue with Lambda-Permission for API-Gateway


gpuettmann:
All,

it seems to me that there is a bug with the created LAMBDA-Policy when creating a  "ANY"- API-Gateway-Mapping to a Lambda-Function, see the template here:

[app_spec.txt](https://github.com/awslabs/serverless-application-model/files/699546/app_spec.txt)

The created Lambda policy includes (for Resource analytics the following condition), and therefore a Lambda permission Error occurs when testing the API-Gateway:

"Condition":{"ArnLike":{"AWS:SourceArn":"arn:aws:execute-api:us-east-1:310073746091:yi1i6eckj4/\*/ANY/analytics/\*"}}

instead of (resource manual has been created manually afterwards via API-Gateway-Gui):

"Condition":{"ArnLike":{"AWS:SourceArn":"arn:aws:execute-api:us-east-1:310073746091:yi1i6eckj4/\*/\*/manual/*"}}

Any Idea?

Best Regards, 

Gregor
gpuettmann@tecracer.de

sanathkr:
Where is the actual error? Does it happen when you make HTTP request to the API, say via Curl?

gpuettmann:
...yes, the error occurs, when I test the function, Error-Message:
"Tue Jan 31 14:52:20 UTC 2017 : Execution failed due to configuration error: Invalid permissions on Lambda function
Tue Jan 31 14:52:20 UTC 2017 : Method completed with status: 500"

sanathkr:
This error is on the APIGateway Web Console UI right? If so, I think this is a known problem

Squareomatic:
Hi, yes it seams that permissions for api gateway to call laambdas are not given to certain endpoints, I have seen it randomly in some resources of mine too

ryanlitalien:
Hi, I'm experiencing the same problems as the others here. Hopefully I can provide some more info.
I have a SAM with an API and 3 lambdas. Everything is deployed successfully and connected using stage variables as well. My API cannot talk to the lambda because of the permissions issue mentioned above.
I was able to click to change the lambda, and then this is the modal that popped:

![image](https://cloud.githubusercontent.com/assets/1479756/25341213/21bdc6c2-28d6-11e7-9b02-d0225666ba36.png)

After I said "OK", the lambda was reachable via API Gateway.

In addition, since I'm using a stage variable, it provided an extra message:

[...]

PR #449, opened 2018-05-29
Fix ANY method ARN generation using wildcard

villasv:
*Issue #, if available:* #59 

*Description of changes:* Convert the `ANY` method to a wildcard `*` when generating permission ARNs

I did not add test templates because this change is already covered by existing tests.

### DIFF
--- a/samtranslator/model/eventsources/push.py
+++ b/samtranslator/model/eventsources/push.py
@@ -429,11 +429,12 @@ class Api(PushEventSource):
             raise RuntimeError("Could not add permission to lambda function.")
 
         path = self.Path.replace('{proxy+}', '*')
+        method = '*' if self.Method == 'any' else self.Method.upper()
 
         api_id = self.RestApiId
 
         # RestApiId can be a simple string or intrinsic function like !Ref. Using Fn::Sub will handle both cases
-        resource = '${__ApiId__}/' + '${__Stage__}/' + self.Method.upper() + path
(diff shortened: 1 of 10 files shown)


samtranslator/model/eventsources/push.py
```diff
@@ -429,11 +429,12 @@ def _get_permission(self, resources_to_link, stage, suffix):
             raise RuntimeError("Could not add permission to lambda function.")
 
         path = self.Path.replace('{proxy+}', '*')
+        method = '*' if self.Method == 'any' else self.Method.upper()
```
jfuss: This may work but can be easily broken. I am not sure if self.Method is normalized (all upper or lower) but we should just force one of these here. It will also help in readability.

brettstack: It's not obvious, but this is already being forced to lower here https://github.com/awslabs/serverless-application-model/blob/develop/samtranslator/model/eventsources/push.py#L399

jfuss: Let's make it obvious here then :)

villasv: So I should make it like `self.Method.lower() == 'any'`?

jfuss: `.lower()` or `.upper()`. Which one doesn't explicitly matter, just as long as it matches what you compare it to (`any` or `ANY`).

villasv:
Updated, sorry for deleting my previous question. I realized later it was kinda already answered in the initial comment.

PR #451, opened 2018-05-30
feat(event-source): add SQS as Lambda event source

brettstack:
*Issue #, if available:*

*Description of changes:*

Adds support for SQS as an event source. This feature is not yet supported by AWS (pre-announced earlier this year) and will not be merged to master until it is available.


docs/internals/generated_resources.rst
```diff
@@ -220,6 +220,33 @@ AWS::Lambda::Permissions           MyFunction\ **MyTrigger**\ Permission
...
+
+  MyFunction:
+    Type: AWS::Serverless::Function
+    Properties:
+      ...
+      Events:
+        MyTrigger:
+          Type: SQS
+          Properties:
+            Queue: arn:aws:sqs:us-east-1:123456789012:my-queue
```
sanathkr: should we call this property `QueueArn` instead? I wonder if users might get confused and specify a queue name here

tests/translator/input/streams.yaml
```diff
@@ -25,3 +25,15 @@ Resources:
             Stream: arn:aws:dynamodb:us-west-2:012345678901:table/TestTable/stream/2015-05-11T21:21:33.291
             BatchSize: 200
             StartingPosition: LATEST
+  SQSFunction:
```
sanathkr: Consider moving SQS test to a separate file. This file is already getting overloaded.

docs/internals/generated_resources.rst
brettstack: I went with Queue to be consistent with Stream. Though I agree, QueueArn is more descriptive. If we want in the future we can add QueueArn (and StreamArn) leave Queue as an alias (and possibly deprecate?).

sanathkr: Fair enough. Let's stick with Queue to keep the language consistent.

samtranslator/model/eventsources/pull.py
```diff
@@ -8,16 +8,17 @@
 class PullEventSource(ResourceMacro):
     """Base class for pull event sources for SAM Functions.
 
-    The pull events are the streams--Kinesis and DynamoDB Streams. Both of these correspond to an EventSourceMapping in
-    Lambda, and require that the execution role be given to Kinesis or DynamoDB Streams, respectively.
+    The pull events are Kinesis Streams, DynamoDB Streams, and SQS Queues. All of these correspond to an EventSourceMapping in
+    Lambda, and require that the execution role be given to Kinesis Streams, DynamoDB Streams, or SQS Streams, respectively.
```
muthiah90: I suppose we shouldn't be calling SQS as streams. SQS in itself should do. Thoughts ?

brettstack: Good catch; that was a miss. I refer to it as SQS Queues in the first sentence.

tests/translator/output/aws-cn/streams.json
```diff
@@ -1,120 +1,166 @@
...
             }
-          ]
+          }]
+        }
+      }
+    },
+    "SQSFunctionMySqsQueue": {
+      "Type": "AWS::Lambda::EventSourceMapping",
+      "Properties": {
+        "BatchSize": 200,
```
muthiah90: Batch size for SQS is between 1 and 10

tests/translator/output/aws-us-gov/streams.json
```diff
@@ -1,120 +1,166 @@
...
             }
-          ]
+          }]
+        }
+      }
+    },
+    "SQSFunctionMySqsQueue": {
+      "Type": "AWS::Lambda::EventSourceMapping",
+      "Properties": {
+        "BatchSize": 200,
```
muthiah90: BatchSize between 1 and 10

tests/translator/output/streams.json
```diff
@@ -1,120 +1,166 @@
...
             }
-          ]
+          }]
+        }
+      }
+    },
+    "SQSFunctionMySqsQueue": {
+      "Type": "AWS::Lambda::EventSourceMapping",
+      "Properties": {
+        "BatchSize": 200,
```
muthiah90: Batch size between 1 and 10

versions/2016-10-31.md
```diff
@@ -371,6 +372,26 @@ Properties:
...
+---|:---:|---
+Queue | `string` | **Required.** ARN of the SQS queue.
+BatchSize | `integer` | Maximum number of messages to process per function invocation.
+
+##### Example: SQS event source object
+
+```yaml
+Type: SQS
+Properties:
+  Stream: arn:aws:sqs:us-west-2:012345678901:my-queue # NOTE: FIFO SQS Queues are not yet supported
```
muthiah90: This should be Queue: and not Stream:

samtranslator/model/lambda_.py
```diff
@@ -61,7 +61,7 @@ class LambdaEventSourceMapping(Resource):
             'Enabled': PropertyType(False, is_type(bool)),
             'EventSourceArn': PropertyType(True, is_str()),
             'FunctionName': PropertyType(True, is_str()),
-            'StartingPosition': PropertyType(True, is_str())
+            'StartingPosition': PropertyType(False, is_str())
```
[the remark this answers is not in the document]
brettstack: JSON Schema validation will pick this up once we turn it on (it's currently not doing anything with failed JSON Schemas as we're not yet ready). Would prefer to not add additional validation right now.

samtranslator/model/eventsources/pull.py
```diff
@@ -8,16 +8,17 @@
...
     """
     resource_type = None
     property_types = {
-            'Stream': PropertyType(True, is_str()),
-            'BatchSize': PropertyType(False, is_type(int)),
-            'StartingPosition': PropertyType(True, is_str())
+        'Stream': PropertyType(False, is_str()),
+        'Queue': PropertyType(False, is_str()),
+        'BatchSize': PropertyType(False, is_type(int)),
+        'StartingPosition': PropertyType(False, is_str())
```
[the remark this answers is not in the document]
sanathkr: Good point. We do need to validate them separately now that these properties are not required

samtranslator/model/lambda_.py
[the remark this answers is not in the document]
sanathkr: What if someone forgot to set the Queue property? SAM Translator will crash instead of giving them a validation error. After turning on JSON Schema, we do need a deeper scrub of code. This extra validation won't increase the surface area of the scrub significantly IMO

samtranslator/model/eventsources/pull.py
```diff
@@ -48,6 +49,10 @@ def to_cloudformation(self, **kwargs):
         except NotImplementedError:
             function_name_or_arn = function.get_runtime_attr("arn")
 
+        if not self.Stream and not self.Queue:
```
sanathkr: Can you instead add a validate functino to each subclass that gets called here? Each subclass can then validate and return a customized error message. What do you think?

brettstack: I deliberately went with this simpler approach since this is temporary code and I think the error message is clear enough even when they are combined for both cases.

PR #456, opened 2018-06-04
updating safe_lambda_deployment.rst to include example Cloudwatch alarms for safe deployment usage

bundyfx:
Show example of Cloudwatch alarms in Cloudformation that relate the safe deployments. These are not easy to find online and will help people jumpstart the process of getting this going.


Nr18:
Just thinking out load this looks like a very good default should this or something like this be generated by SAM? (Enable it via a option in the AWS::Serverless::Function resource)

Issue #366, opened 2018-04-05
Move Tests to unittest and not just assert


jfuss:
**Description:**
Most of our unit tests are written with [`unittest`](https://docs.python.org/2/library/unittest.html) but we have some lingering assert statements in our tests. Unittest gives us a little more flexibility and readability (we don't have have to do `deepsort` or other things to make the assert work). This will also reduce the `print()` statements we have in our tests that were used for debugging tests when the assert failed.

Relevant test files this needs to be addressed in:

* `tests/translator/test_translator.py`
* `tests/translator/validator/test_validator.py`
* `tests/translator/test_managed_policies_translator.py`
* `tests/translator/test_api_resource.py`
* `tests/test_types.py`
* `tests/test_model.py`

PR #462, opened 2018-06-08
Migrate to unittest

tylersouthwick:
*Issue #, if available:*
#366

*Description of changes:*
Convert tests to be in TestCase classes and to use self.assert*

### DIFF
--- a/tests/test_model.py
+++ b/tests/test_model.py
@@ -1,11 +1,14 @@
-import pytest
-
 from unittest import TestCase
-from mock import Mock, call, ANY
-from samtranslator.model.exceptions import InvalidResourceException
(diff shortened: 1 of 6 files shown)


jlhood:
Really appreciate the contribution! What's the benefit of converting tests to use TestCase classes? We're using pytest and my understanding is pytest supports TestCase subclasses to ease migration from unittest to pytest.

https://docs.pytest.org/en/latest/unittest.html

From reading that, I thought we'd want to convert away from TestCase subclasses, not towards them.

jlhood:
@tylersouthwick Ah I see, thanks! Yeah go ahead and update the branch to resolve merge conflicts and I'll work to get this merged in. Thanks so much!

PR #464, opened 2018-06-09
Feature/cors allow credentials

cast:
Relates to issue https://github.com/awslabs/serverless-application-model/issues/463

Added support for CORS headers Access-Control-Expose-Headers and Access-Control-Allow-Credentials in the CORS Configuration block at the Serverless::Api resource.


tests/translator/output/api_with_cors_and_only_credentials.json
```diff
@@ -0,0 +1,154 @@
...
+                    "application/json": "{\n  \"statusCode\" : 200\n}\n"
+                  }, 
+                  "responses": {
+                    "default": {
+                      "statusCode": "200", 
+                      "responseTemplates": {
+                        "application/json": "{}\n"
+                      }, 
+                      "responseParameters": {
+                        "method.response.header.Access-Control-Allow-Origin": "'*'",
```
chrisoverzero: This is invalid output. A-C-A-O isn't allowed to have the wildcard value when A-C-A-C is set.

brettstack: Good catch! @cast could you add some logic to throw `InvalidDocument` when a user tries to specify `Allow-Credentials: true` and `Allow-Origin: *`?

brettstack:
Thanks for raising this @chrisoverzero. After reviewing the spec and other documentation more thoroughly it does appear `Expose-Headers` is not intended for preflight responses.

brettstack:
@cast if you can provide a repro/use-case for including this in preflight I can take another look. Otherwise the `Allow-Credentials` part of this PR is still very useful.

cast:
Thanks @chrisoverzero for the spec review. Apart from a now fixed side effect leading me to think the allowed exposed headers was an issue in our internal development I added it mainly because other api management solutions often allow or force to take a liberal approach to preflight CORS (ApiMan, Axway,...). 

I'm not aware of specific browsers and browser versions that require the exposed header in the preflight response. I'm assuming if there is an issue it would be in old versions due to the current living standard.

I'll remove the exposed headers parts.

JenningsWu:
Any updates? This is really helpful for our project : )

PR #466, opened 2018-06-11
fix: log a warning on invalid schema validation

brettstack:
*Issue #, if available:*

*Description of changes:*
Log a warning on invalid schema validation.

### DIFF
--- a/samtranslator/parser/parser.py
+++ b/samtranslator/parser/parser.py
@@ -2,6 +2,7 @@ from samtranslator.model.exceptions import InvalidDocumentException, InvalidTemp
 from samtranslator.validator.validator import SamTemplateValidator
 from samtranslator.model import ResourceTypeResolver, sam_resources
 from samtranslator.plugins import LifeCycleEvents
+import logging
 
 class Parser:
     def __init__(self):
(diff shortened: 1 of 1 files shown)


samtranslator/parser/parser.py
```diff
@@ -2,6 +2,7 @@
 from samtranslator.validator.validator import SamTemplateValidator
 from samtranslator.model import ResourceTypeResolver, sam_resources
 from samtranslator.plugins import LifeCycleEvents
+import logging
```
jfuss: This uses the root logger, which will log in any one using SAM to translate without a way to turn it off. 

logging should be done by adding a `__name__`.

`logger = logging.getLogger(__name__)` and then log by doing `logger.warning("Logging a Warning")`

brettstack: Can you send a PR? How do other consumers then turn that off?

jfuss: https://stackoverflow.com/a/35325280

Should magically work. You can also manipulate the libraries logger too. For example, SAM CLI could then turn the SAM Translators logging on for a given level if they so wanted to.

fatbasstard: Based on https://stackoverflow.com/questions/24750523/disable-output-of-root-logger we're handling it like this now:

https://github.com/awslabs/cfn-python-lint/pull/117/files#diff-6d7631d1ad3e87ab0c24664abb771bc4R79

PR #468, opened 2018-06-11
Added API Gateway multi-origin CORS example

loganzartman:
*Issue #, if available:*

*Description of changes:*
Created an example of using Lambda to provide CORS headers on API Gateway endpoints given a list of allowed origins. The Lambda function checks the origin of incoming requests against a list, and only provides CORS headers if there is a match. 


examples/2016-10-31/api_multiple_origin_cors/template.yaml
```diff
@@ -0,0 +1,37 @@
...
+          Type: 'Api'
+          Properties:
+            Path: '/'
+            Method: 'get'
+        OptionsTest:
+          Type: 'Api'
+          Properties:
+            Path: '/test'
+            Method: 'options'
+        PostTest:
```
jackrk: I see there is some special case for DELETE methods in the function. Is it useful to have a DELETE method in the example API? Up to you.

loganzartman: I chose DELETE arbitrarily to demonstrate how to handle preflight requests, since it is one of the methods that is guaranteed to trigger one. So I think it is more a specific example of a general case. 
I do notice now that the event is still named "PostTest", whereas it should maybe be something like "DeleteTest".

examples/2016-10-31/api_multiple_origin_cors/cors-util.js
```diff
@@ -0,0 +1,65 @@
...
+    "X-Amz-Security-Token"
+];
+
+/**
+ * Return an object literal that contains an Access-Control-Allow-Origin header
+ * if the request origin matches a pattern for an allowed origin.
+ * Otherwise, return an empty object literal.
+ * @param event Lambda event event associated with a request
+ * @param allowedOrigins A list of strings or regexes representing allowed origin URLs
+ * @return {Object} an object containing a header => value mapping, or nothing
```
brettstack: "containing allowed headers and their value" and remove ", or nothing"

examples/2016-10-31/api_multiple_origin_cors/cors-util.js
```diff
@@ -0,0 +1,65 @@
...
+ */
+exports.makeOriginHeader = (event, allowedOrigins) => {
+    // determine origin
+    const origin = event.headers.Origin || event.headers.origin;
+    if (typeof origin === "undefined")
+        return {}; // no CORS headers necessary; browser will load resource
+
+    // look for origin in list of allowed origins
+    const allowedPattern = allowedOrigins.find(pattern => origin.match(pattern));
+    if (typeof allowedPattern === "undefined")
```
brettstack: `if (allowedPattern === undefined)`; same for `origin` check above

examples/2016-10-31/api_multiple_origin_cors/cors-util.js
```diff
@@ -0,0 +1,65 @@
...
+        return {}; // return no CORS headers; browser will not load resource
+                   // we do not return a "null" origin because this is exploitable
+    
+    allowedOrigin = origin.match(allowedPattern)[0]; // determine exact match
```
brettstack: Missing `const`.  This also looks wrong... Won't `allowedPattern` be an array? Did you mean `origin.match(allowedPattern[0])[0]`? And won't that just yield allowedPattern[0]?

examples/2016-10-31/api_multiple_origin_cors/cors-util.js
```diff
@@ -0,0 +1,65 @@
...
+
+    // console.log("origin: " + origin);
```
brettstack: Remove comments

examples/2016-10-31/api_multiple_origin_cors/cors-util.js
```diff
@@ -0,0 +1,65 @@
...
+    // console.log("allowed by: " + allowedPattern);
+    
+    // produce origin header
```
brettstack: Not useful. Remove.

examples/2016-10-31/api_multiple_origin_cors/cors-util.js
```diff
@@ -0,0 +1,65 @@
...
+    return {"Access-Control-Allow-Origin": allowedOrigin};
```
brettstack: You mention `null` is exploitable. Maybe add a final `if (!allowedOrigin) return {}` to be safe.

examples/2016-10-31/api_multiple_origin_cors/cors-util.js
```diff
@@ -0,0 +1,65 @@
...
+/**
+ * Return an object literal that contains all headers necessary for a CORS
+ * preflight.
+ * @param event Lambda event associated with a request
+ * @param {Array} allowedOrigins see documentation for makeOriginHeader() 
+ * @param {Array} allowedMethods a list of strings representing allowed HTTP methods
+ * @param {Array} allowedHeaders (optional) a list of strings representing allowed headers
+ * @return {Object} an object containing several header => value mappings
+ */
+exports.makeHeaders = (event, allowedOrigins, allowedMethods, allowedHeaders = DEFAULT_ALLOWED_HEADERS) => {
```
brettstack: This should also just be `origin`

examples/2016-10-31/api_multiple_origin_cors/cors-util.js
```diff
@@ -0,0 +1,65 @@
...
+
+/**
+ * Compiles a path containing wildcards into a regular expression.
+ * Using this is optional.
+ * @param {String} path the path to compile
+ * @return {RegExp} compiled regular expression
+ */
+exports.compileURLWildcards = (path) => {
+    const parts = path.split("*");
+    const escapeRegex = str => str.replace(/([.?*+^$(){}|[\-\]\\])/g, "\\$1");
```
brettstack: Add comment to make this clearer; maybe use an example.

examples/2016-10-31/api_multiple_origin_cors/cors-util.test.js
```diff
@@ -0,0 +1,82 @@
+const cors = require("./cors-util");
+
+const makeMockEvent = origin => {
+    return {"headers": {"origin": origin}};
+};
+
+// makeOriginHeader
```
brettstack: remove

examples/2016-10-31/api_multiple_origin_cors/index.js
```diff
@@ -0,0 +1,67 @@
...
+};
+
+/**
+ * Demonstrates an endpoint that accepts DELETE requests.
+ * 
+ * In this case, the browser will perform a CORS preflight request, so we must
+ * handle OPTIONS requests and provide the CORS headers.
+ * When the browser makes the DELETE request, we only need to provide the origin.
+ */
+const handleTest = (event, context, callback) => {
```
brettstack: With nodejs8.10 you should be able to define handlers like

```
async () => {
  return { headers, statusCode }
}
```

examples/2016-10-31/api_multiple_origin_cors/cors-util.js
```diff
@@ -0,0 +1,65 @@
+const DEFAULT_ALLOWED_HEADERS = [
```
brettstack: Is there a library that exists that might be better than rolling your own?

examples/2016-10-31/api_multiple_origin_cors/template.yaml
brettstack: It says 'PostTest' but has `Method: 'delete`. To really test CORS you should have an OPTIONS method on the /test (or other) path also. Won't this fail a real world CORS scenario as it currently stands? Has you confirmed this works?

examples/2016-10-31/api_multiple_origin_cors/index.js
```diff
@@ -0,0 +1,67 @@
+// Demonstrates a Lambda function that provides CORS headers for many origins.
+
+const cors = require("./cors-util");
+
+// define a whitelist of allowed origins
+const allowedOrigins = [
+    "http://127.0.0.1",
+    "https://*.example.com",
+    "https://*.amazon.com"
+].map(cors.compileURLWildcards);
```
brettstack: Simpler interface would be `cors.compileURLWildcards([])`. Even better if you just do this for them behind the scenes (inside `cors.makeHeaders`)

loganzartman: My goal was to isolate the wildcard support from the rest of the implementation, in case a user wants to use some other library or provide their own regex for the origins.

brettstack: They can easily modify themselves. cors-utils is baked in; not a library.

examples/apps/api-gateway-multiple-origin-cors/cors-util.js
```diff
@@ -0,0 +1,70 @@
...
+ * any number of URL characters in place of each wildcard (*) character.
+ * http://*.example.com matches http://abc.xyz.example.com but not http://example.com
+ * http://*.example.com does not match http://example.org/.example.com
+ * @param {String} url the url to compile
+ * @return {RegExp} compiled regular expression
+ */
+exports.compileURLWildcards = (url) => {
+    // unreserved characters as per https://tools.ietf.org/html/rfc3986#section-2.3
+    const url_unreserved_pattern = "[A-Za-z0-9\-._~]";
+    const wildcard_pattern = url_unreserved_pattern + "*";
```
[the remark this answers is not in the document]
loganzartman: oops

examples/apps/api-gateway-multiple-origin-cors/index.js
```diff
@@ -0,0 +1,57 @@
+// Demonstrates a Lambda function that provides CORS headers for many origins.
+
+const cors = require("./cors-util");
+
+// define a whitelist of allowed origins
+const allowedOrigins = [
+    "http://127.0.0.1",
+    "https://*.example.com",
+    "https://*.amazon.com"
+].map(cors.compileURLWildcards);
```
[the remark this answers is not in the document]
loganzartman: I'll change that

examples/apps/api-gateway-multiple-origin-cors/index.js
```diff
@@ -0,0 +1,57 @@
...
+ * When the browser makes the DELETE request, we only need to provide the origin.
+ */
+exports.handleTest = async (event, context) => {
+    const origin = event.headers.Origin || event.headers.origin;
+    const allowedMethods = ["OPTIONS", "DELETE"];
+
+    if (event.httpMethod === "OPTIONS") {
+        // return an empty response, with all CORS headers
+        return {
+            headers: cors.makeHeaders(origin, allowedOrigins, allowedMethods),
```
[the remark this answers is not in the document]
loganzartman: A higher level of abstraction seems reasonable. I'm not sure if there would be a case where a user might want to add extra headers to the response. It might be sufficient to add a parameter for `Access-Control-Max-Age`, which seems to be the only other header relevant to a preflight response.

examples/apps/api-gateway-multiple-origin-cors/cors-util.js
```diff
@@ -0,0 +1,70 @@
...
+ * Return an object literal that contains an Access-Control-Allow-Origin header
+ * if the request origin matches a pattern for an allowed origin.
+ * Otherwise, return an empty object literal.
+ * @param {String} origin the origin to test against the allowed list
+ * @param {Array} allowedOrigins A list of strings or regexes representing allowed origin URLs
+ * @return {Object} an object containing allowed header and its value
+ */
+exports.makeOriginHeader = (origin, allowedOrigins) => {
+    if (origin === undefined || origin === null)
+        return {}; // no CORS headers necessary; browser will load resource
```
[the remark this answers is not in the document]
loganzartman: Firefox, for example, will not send an Origin header in a same-origin request, but will always send one in a cross-origin request. If there is no origin, I don't think there's anything else to be done here.

examples/apps/api-gateway-multiple-origin-cors/index.js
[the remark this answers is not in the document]
loganzartman: I've gone ahead and done that

brettstack:
Looking good. The only other things (as communicated offline) is to split this into multiple Lambda Functions instead of implementing routing. Alternatively, if you prefer to use a single function, use aws-serverless-express (and in this case you could probably just use the cors middleware)

loganzartman:
Improved the routing

examples/apps/api-gateway-multiple-origin-cors/template.yaml
```diff
@@ -2,31 +2,38 @@ AWSTemplateFormatVersion: '2010-09-09'
...
       Runtime: 'nodejs8.10'
       Events:
-        DeleteTest:
+        Delete:
           Type: 'Api'
           Properties:
             Path: '/test'
             Method: 'delete'
-        OptionsTest:
+  ExampleTestPreflight:
```
brettstack: This should go in the same Lambda function as delete.

PR #498, opened 2018-07-05
Added remark about CodeUri property on Globals

glnds:
Currently CodeUri on Globals is not yet supported by CloudFormation package.

*Issue #, if available:* As i was unaware of this it took me quite a while to figure this out...

*Description of changes:* Added just as simple comment to point out the fact the property is not yet supported by CloudFormation package.


brettstack:
Makes sense since package is run pre-transformation. Might be a feature request for when SAM CLI implements its own `package` command @jfuss @sanathkr

PR #408, opened 2018-04-26
feat(policy-template): add mobile analytics targeting policy templates

simalexan:
Issue #, if available:
N/A

Description of changes:

- Adding the MobileAnalyticsWriteOnlyAccessPolicy PolicyTemplate with tests.
- Adding the new PinpointEndpointAccessPolicy PolicyTemplate with tests.
- Updated PolicyTemplate Docs to include all new Templates that can be consumed.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #523, opened 2018-07-25
Add FilterLogEvents policy to policy templates.

jbasden:
This commit essentially copies the pattern of commit #408
to add a policy template that permits calling the
filter log events cloudwatch API.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/docs/policy_templates_data/policy_templates.json
+++ b/docs/policy_templates_data/policy_templates.json
@@ -1243,7 +1243,9 @@
     "FilterLogEventsPolicy": {
       "Description": "Gives permission to filter Log Events from any Log Group",
       "Parameters": {
-
+        "LogGroupName": {
+          "Description": "Name of the Log Group"
+        }
       },
       "Definition": {
         "Statement": [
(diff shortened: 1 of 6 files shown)


docs/policy_templates_data/policy_templates.json
```diff
@@ -1239,6 +1239,34 @@
           }
         ]
       }
+    },
+    "FilterLogEventsPolicy": {
+      "Description": "Gives permission to filter Log Events from any Log Group",
```
ylynn: Update the description since you change the Resource to scope down to one log group.

docs/policy_templates_data/policy_templates.json
```diff
@@ -1239,6 +1239,34 @@
...
+      "Definition": {
+        "Statement": [
+          {
+            "Effect": "Allow",
+            "Action": [
+              "logs:FilterLogEvents"
+            ],
+            "Resource": {
+              "Fn::Sub": [
+                "arn:${AWS::Partition}:logs:${AWS::Region}:${AWS::Account}:log-group:${logGroupName}*",
```
ylynn: Do you need the "*" at the end? You can try the policy definition in your own template to figure out.

docs/policy_templates_data/policy_templates.json
jbasden: Good catch. Thanks

docs/policy_templates_data/policy_templates.json
jbasden: I put it there because of the lambda invoke example I was following. I don't believe it is needed though.

PR #525, opened 2018-07-25
Add AWSSecretsManagerGetSecretValueAccess Policy

willtong1234:
*Description of changes:*

This change adds a AWSSecretsManagerGetSecretValueAccess policy which grants access to the Secrets Manager GetSecretValue API for a single secret ARN.

### DIFF
--- a/docs/policy_templates_data/policy_templates.json
+++ b/docs/policy_templates_data/policy_templates.json
@@ -931,34 +931,34 @@
         ]
       }
     },
-	  "AWSSecretsManagerGetSecretValueAccess": {
-	    "Description": "Grants permissions to GetSecretValue for the specified AWS Secrets Manager secret",
-	    "Parameters": {
-	      "SecretArn": {
-	        "Description": "The ARN of the secret to grant access to"
-	      }
-	    },
-	    "Definition": {
-	      "Statement": [
-	        {
-	          "Effect": "Allow",
-	          "Action": [
-	            "secretsmanager:GetSecretValue"
-	          ],
-	          "Resource": {
-	            "Fn::Sub": [
-	              "${secretArn}",
-	              {
-	                "secretArn": {
-	                  "Ref": "SecretArn"
-	                }
-	              }
-	            ]
-	          }
-	        }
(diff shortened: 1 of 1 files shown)


docs/policy_templates_data/policy_templates.json
```diff
@@ -931,6 +931,34 @@
...
+      "Definition": {
+        "Statement": [
+          {
+            "Effect": "Allow",
+            "Action": [
+              "secretsmanager:GetSecretValue"
+            ],
+            "Resource": {
+              "Fn::Sub": [
+                "${secretArn}",
```
jlhood: Our policy templates follow the pattern of forcing scoping the resource to the current account and region for tighter security. Will following this pattern work for your use case? For example, you could change the SecretArn parameter to SecretName. Then do this:

```
"Resource": {
  "Fn::Sub": [
    "arn:${AWS::Partition}:secretsmanager:${AWS::Region}:${AWS::AccountId}:secret:${secretName}",
    {
      "secretName": {
        "Ref": "SecretName"
      }
    }
  ]
}
```

docs/policy_templates_data/policy_templates.json
```diff
@@ -931,6 +931,34 @@
         ]
       }
     },
+    "AWSSecretsManagerGetSecretValueAccess": {
```
jlhood: The other policy templates end in Policy. How about `AWSSecretsManagerGetSecretValuePolicy`?

docs/policy_templates_data/policy_templates.json
jlhood: Discussed use case offline. There is a use case for needing to specify the full ARN.

brettstack:
This file no longer exists on the develop branch as it was redundant. Please rebase and make your changes to https://github.com/awslabs/serverless-application-model/blob/develop/samtranslator/policy_templates_data/policy_templates.json instead.

willtong1234:
This is eclipsed by https://github.com/awslabs/serverless-application-model/pull/539. Closing this one out.

PR #530, opened 2018-07-29
added cloudwatch event to MS Teams

MattTunny:
*Issue #, if available:* 
no cloudwatch events or teams example
*Description of changes:*
added example for cloudwatch event to MS teams with sample events


ilyash:
Semantics ignored leading to poor code quality.

* Function called `get_security_group` should not send Teams message.
* Function called `lookup_security_group` should not send Teams message.
* Group ID should not be stored in a variable named `cidr_address`

PR #541, opened 2018-08-07
feat(plugins): added DefaultDefinitionBody plugin

brettstack:
This makes DefinitionBody and DefinitionUri optional, and if not provided then the API will use the Swagger generated by SAM. This makes it easy for users who rely on Implicit API to use the existing Cors feature and upcoming Authorizers feature by defining an explicit API resource without needing to redefine their entire Swagger

### DIFF
--- a/samtranslator/plugins/api/default_definition_body_plugin.py
+++ b/samtranslator/plugins/api/default_definition_body_plugin.py
@@ -1,0 +1,37 @@
+from samtranslator.plugins import BasePlugin
+from samtranslator.swagger.swagger import SwaggerEditor
+from samtranslator.public.sdk.resource import SamResourceType
+from samtranslator.public.sdk.template import SamTemplate
+
+
+class DefaultDefinitionBodyPlugin(BasePlugin):
+    """
+    If the user does not provide a DefinitionBody or DefinitionUri
+    on an AWS::Serverless::Api resource, the Swagger constructed by
+    SAM is used. It accomplishes this by simply setting DefinitionBody
+    to a minimum Swagger definition and sets `__MANAGE_SWAGGER: true`.
+    """
+
+    def __init__(self):
+        """
+        Initialize the plugin.
+        """
+
+        super(DefaultDefinitionBodyPlugin, self).__init__(DefaultDefinitionBodyPlugin.__name__)
+
+    def on_before_transform_template(self, template_dict):
+        """
+        Hook method that gets called before the SAM template is processed.
+        The template has pass the validation and is guaranteed to contain a non-empty "Resources" section.
+
+        :param dict template_dict: Dictionary of the SAM template
+        :return: Nothing
+        """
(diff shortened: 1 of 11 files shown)


samtranslator/plugins/api/default_definition_body_plugin.py
```diff
@@ -0,0 +1,37 @@
...
+        """
+        Initialize the plugin.
+        """
+
+        super(DefaultDefinitionBodyPlugin, self).__init__(DefaultDefinitionBodyPlugin.__name__)
+
+    def on_before_transform_template(self, template_dict):
+        """
+        Hook method that gets called before the SAM template is processed.
+        The template has pass the validation and is guaranteed to contain a non-empty "Resources" section.
```
keetonian: nit: grammar

tests/plugins/api/test_default_definition_body_plugin.py
```diff
@@ -0,0 +1,51 @@
+from unittest import TestCase
+from mock import Mock, patch
```
keetonian: You could optionally fix this import order here- keep them in alphabetical order.

samtranslator/translator/translator.py
```diff
@@ -198,6 +199,7 @@ def prepare_plugins(plugins):
     """
 
     required_plugins = [
+        DefaultDefinitionBodyPlugin(),
```
keetonian: Do we have a default naming convention? Half of these are camel case, the other half use underscores. Do we need to update our documentation for naming conventions?

tests/plugins/api/test_default_definition_body_plugin.py
```diff
@@ -0,0 +1,51 @@
...
+        # Name is the class name
+        expected_name = "DefaultDefinitionBodyPlugin"
+
+        self.assertEquals(self.plugin.name, expected_name)
+
+    def test_plugin_must_be_instance_of_base_plugin_class(self):
+        self.assertTrue(isinstance(self.plugin, BasePlugin))
+
+
+class TestDefaultDefinitionBodyPlugin_on_before_transform_template(TestCase):
```
keetonian: I think this (and your test updates in general) misses the case where a ```DefinitionBody``` or ```DefinitionUri``` exists in part of a template but doesn't in another.  Tests exist that target the old SAM requirements (forcing either a ```DefinitionBody``` or ```DefinitionUri``` to be present) and you wrote new tests for the case where neither one exists at all. It might be worth it to have a test where both cases are present to enforce the desired behavior.

samtranslator/translator/translator.py
brettstack: This is Python convention. `DefaultDefinitionBodyPlugin` is a class; others are functions/methods.

keetonian: Ok, thank you for the clarification.

PR #543, opened 2018-08-08
feat: hard-code list of managed policies and use that by default

brettstack:
This was the only AWS Service call in SAM. By removing this it allows for the SAM transformer to be run offline.

Maintained the initial code with an option so that in the future we an expose a way to grab the authoritative list from IAM. This also allows us to easily grab an updated list periodically.

*Issue #, if available:*

*Description of changes:*


tests/translator/test_managed_policies_translator.py
```diff
@@ -7,6 +7,22 @@ def create_page(policies):
...
+    iam.get_paginator.return_value = paginator
+
+    actual = ManagedPolicyLoader(iam).load()
+
+    assert not iam.get_paginator.called
+    assert isinstance(actual, dict)
+    assert actual.get('AmazonMechanicalTurkCrowdFullAccess')
+
+    # NOTE: this is a moving target that must be updated when ManagedPolicyLoader._local_managed_policies_map is updated
+    assert len(actual) == 401
```
ombozek: This test doesn't really tell us anything. If we want a test around the list of managed policies we should actually download the list of managed policies from IAM and do a shallow (or maybe even deep...) comparison to our list. The downside there is that we'll be forced to update our list every time a modification to the list of managed policy arises.

brettstack: It tells us that at least one of the expected policies is there, and that the number of entries is as expected. I think this is reasonable.

samtranslator/policy_templates_data/managed_policies.json
```diff
@@ -0,0 +1,403 @@
+{
+  "SecurityAudit": "arn:aws:iam::aws:policy/SecurityAudit",
```
jfuss: This is not deployable to all regions. This needs to be partition aware. Do you know if the names are the same across all partitions? Some services or features might be missing from other partitions as well, so some of these could be invalid there as well.

brettstack: Good catch. I'm going to shelve this for now until Authorizers are done and I'll look into making this partition-aware.

fatbasstard:
Since both [SAM CLI](https://github.com/awslabs/aws-sam-cli) and [cfn-lint](https://github.com/awslabs/cfn-python-lint) already implement the same feature, this is a really nice change from a consumer point of view

jfuss:
This seemed to be shelved in Aug 2018 and not update since. Closing as this is stale.

PR #546, opened 2018-08-09
feat(auth): add support for API Gateway Authorizers

brettstack:
This is a WIP. Remaining:

- [x] Tests
- [x] Documentation
- [x] Examples
- [x] Additional end-to-end  manual verification

*Issue #, if available:*

*Description of changes:*


samtranslator/model/apigateway.py
```diff
@@ -85,3 +86,142 @@ def make_auto_deployable(self, stage, swagger=None):
...
+                                   {'__FunctionArn__': self.function_arn})
+
+            swagger[APIGATEWAY_AUTHORIZER_KEY]['authorizerUri'] = authorizer_uri
+            reauthorize_every = self._get_reauthorize_every()
+            function_invoke_role = self._get_function_invoke_role()
+
+            if reauthorize_every:
+                swagger[APIGATEWAY_AUTHORIZER_KEY]['authorizerResultTtlInSeconds'] = reauthorize_every
+
+            if function_invoke_role:
```
brettstack: Do we need to explicitly test for truthiness here? i.e. if this ends up getting set to `None` (`null`?) in Swagger, will it be stripped out automatically?

samtranslator/model/apigateway.py
```diff
@@ -85,3 +86,142 @@ def make_auto_deployable(self, stage, swagger=None):
...
+            partition = ArnGenerator.get_partition_name()
+            authorizer_uri = fnSub('arn:{0}:apigateway:${{AWS::Region}}:lambda:path/2015-03-31/functions/'
+                                   '${{__FunctionArn__}}/invocations'.format(partition),
+                                   {'__FunctionArn__': self.function_arn})
+
+            swagger[APIGATEWAY_AUTHORIZER_KEY]['authorizerUri'] = authorizer_uri
+            reauthorize_every = self._get_reauthorize_every()
+            function_invoke_role = self._get_function_invoke_role()
+
+            if reauthorize_every:
```
brettstack: Remove this check otherwise `0` won't be allowed.

samtranslator/model/api/api_generator.py
```diff
@@ -208,6 +213,73 @@ def _add_cors(self):
         # Assign the Swagger back to template
         self.definition_body = editor.swagger
 
+    def _add_auth(self):
```
brettstack: Note that additional auth settings such as ResourcePolicy and IAM will be set here also when they're added.

samtranslator/model/api/api_generator.py
```diff
@@ -208,6 +213,73 @@ def _add_cors(self):
...
+            raise InvalidResourceException(self.logical_id, INVALID_ERROR)
+
+        if not SwaggerEditor.is_valid(self.definition_body):
+            raise InvalidResourceException(self.logical_id, "Unable to add Auth configuration because "
+                                                            "'DefinitionBody' does not contain a valid Swagger")
+        swagger_editor = SwaggerEditor(self.definition_body)
+        auth_properties = AuthProperties(**self.auth)
+        authorizers = self._get_authorizers(auth_properties.Authorizers)
+
+        if authorizers:
```
brettstack: TODO: Verify no errors when a template specifies `Auth: {}`

samtranslator/model/api/api_generator.py
```diff
@@ -208,6 +213,73 @@ def _add_cors(self):
...
+    def _get_authorizers(self, authorizers_config):
+        if not authorizers_config:
+            return None
+
+        authorizers = {}
+
+        for authorizerName, authorizer in authorizers_config.items():
+            authorizers[authorizerName] = ApiGatewayAuthorizer(
+                api_logical_id=self.logical_id,
+                name=authorizer.get('Name'),
```
brettstack: Remove this and other instances. `Name` is no longer in the spec.

samtranslator/model/apigateway.py
```diff
@@ -85,3 +86,142 @@ def make_auto_deployable(self, stage, swagger=None):
...
+
+        if authorizer_type == 'COGNITO_USER_POOLS' or (authorizer_type == 'LAMBDA' and self._get_function_payload_type() == 'TOKEN'):
+            identity_validation_expression = self._get_identity_validation_expression()
+
+            if identity_validation_expression:
+                swagger[APIGATEWAY_AUTHORIZER_KEY]['identityValidationExpression'] = identity_validation_expression
+
+        return swagger
+
+    def _construct_role(self, function_invoke_role):
```
brettstack: Rename to `_get_function_invoke_role`. Initially planned on allowing users to specify just a role name and we would then construct a full ARN if one wasn't provided. I don't plan on doing this for initial release; best practice is to define a role in the template (or send in via Parameters, etc.)

samtranslator/model/apigateway.py
```diff
@@ -85,3 +86,142 @@ def make_auto_deployable(self, stage, swagger=None):
...
+        return 'LAMBDA'
+
+    def _get_identity_header(self):
+        if not self.identity or not self.identity.get('Header'):
+            return 'Authorization'
+
+        return self.identity.get('Header')
+
+    def _get_reauthorize_every(self):
+        if not self.identity or not self.identity.get('ReauthorizeEvery'):
```
brettstack: `0` is a valid value but won't work because of this

samtranslator/model/apigateway.py
```diff
@@ -85,3 +86,142 @@ def make_auto_deployable(self, stage, swagger=None):
...
+
+        return self.function_invoke_role
+
+    def _get_swagger_authtype(self):
+        authorizer_type = self._get_type()
+
+        return 'cognito_user_pools' if authorizer_type == 'COGNITO_USER_POOLS' else 'custom'
+
+    def _get_function_payload_type(self):
+        return 'TOKEN' if not self.function_payload_type else self.function_payload_type
```
brettstack: Add some validation (higher up than this) to check payload type is either TOKEN or REQUEST

samtranslator/plugins/api/implicit_api_plugin.py
```diff
@@ -232,6 +232,7 @@ def __init__(self):
                 "StageName": "Prod",
 
                 "DefinitionBody": swagger,
+                # "Auth": None,
```
brettstack: Remove

samtranslator/swagger/swagger.py
```diff
@@ -285,6 +286,107 @@ def _make_cors_allowed_methods_for_path(self, path):
...
+        for security in existing_security:
+            if authorizer_names.isdisjoint(security.keys()):
+                existing_non_authorizer_security.append(security)
+            else:
+                existing_authorizer_security.append(security)
+
+        none_idx = -1
+        authorizer_security = []
+
+        # If this is the Api-level DefaultAuthorizer we need to check for an
```
brettstack: This implementation isn't ideal. The existing order of operations is:

1. Apply all Function Event configuration
1. Apply API configuration

This makes it hard to apply a DefaultAuthorizer, especially when 'NONE' is defined on the Function Event. Current logic is to add a 'NONE' entry to the Swagger and then remove it when DefaultAuthorizer comes through. Similarly, DefaultAuthorizer won't be added if it detects an existing entry in `security` which matches a key defined in `Api.Auth.Authorizers`.

Ideally we would add DefaultAuthorizer first and then remove/override as needed, but I'd rather not complicate the order of operations (unless there's already precedent of that which I'm not aware of).

samtranslator/model/eventsources/push.py
```diff
@@ -470,6 +471,28 @@ def _add_swagger_integration(self, api, function):
...
+                        self.relative_id,
+                        'Unable to set Authorizer "{authorizer}" on API method "{method}" for path "{path}" because the related API does not define any Authorizers.'.format(
+                            authorizer=method_authorizer, method=self.Method, path=self.Path))
+
+                if method_authorizer != 'NONE' and not api_authorizers.get(method_authorizer):
+                    raise InvalidEventException(
+                        self.relative_id,
+                        'Unable to set Authorizer "{authorizer}" on API method "{method}" for path "{path}" because it was not defined in the API\'s Authorizers.'.format(
+                            authorizer=method_authorizer, method=self.Method, path=self.Path))
+
```
brettstack: Check `if method_authorizer == 'NONE' and not api_auth.get('DefaultAuthorizer')` and throw

samtranslator/model/apigateway.py
brettstack: Need it. Results in `"authorizerCredentials": null` otherwise

samtranslator/model/apigateway.py
```diff
@@ -85,3 +86,142 @@ def make_auto_deployable(self, stage, swagger=None):
...
+            if reauthorize_every:
+                swagger[APIGATEWAY_AUTHORIZER_KEY]['authorizerResultTtlInSeconds'] = reauthorize_every
+
+            if function_invoke_role:
+                swagger[APIGATEWAY_AUTHORIZER_KEY]['authorizerCredentials'] = function_invoke_role
+
+            if self._get_function_payload_type() == 'REQUEST':
+                swagger[APIGATEWAY_AUTHORIZER_KEY]['identitySource'] = self._get_identity_source()
+
+        if authorizer_type == 'COGNITO_USER_POOLS' or (authorizer_type == 'LAMBDA' and self._get_function_payload_type() == 'TOKEN'):
```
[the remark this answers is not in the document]
brettstack: Done

samtranslator/model/apigateway.py
brettstack: Done

samtranslator/model/api/api_generator.py
brettstack: Confirmed.

samtranslator/model/apigateway.py
brettstack: Removed. `_get_function_invoke_role` already exists.

samtranslator/model/apigateway.py
brettstack: Done

samtranslator/model/apigateway.py
brettstack: Done

samtranslator/model/eventsources/push.py
brettstack: Done

samtranslator/plugins/api/implicit_api_plugin.py
brettstack: Done

samtranslator/model/api/api_generator.py
```diff
@@ -208,6 +213,73 @@ def _add_cors(self):
...
+
+        # Make sure keys in the dict are recognized
+        if not all(key in AuthProperties._fields for key in self.auth.keys()):
+            raise InvalidResourceException(self.logical_id, INVALID_ERROR)
+
+        if not SwaggerEditor.is_valid(self.definition_body):
+            raise InvalidResourceException(self.logical_id, "Unable to add Auth configuration because "
+                                                            "'DefinitionBody' does not contain a valid Swagger")
+        swagger_editor = SwaggerEditor(self.definition_body)
+        auth_properties = AuthProperties(**self.auth)
```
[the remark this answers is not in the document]
brettstack: If you're familiar with JavaScript it's similar to `...self.auth`. Basically passing in all properties of `self.auth` as parameters to `AuthProperties()`

samtranslator/model/api/api_generator.py
```diff
@@ -208,6 +216,113 @@ def _add_cors(self):
...
+        authorizers = self._get_authorizers(auth_properties.Authorizers)
+
+        if not authorizers:
+            return []
+
+        permissions = []
+
+        for authorizer_name, authorizer in authorizers.items():
+            # Construct permissions for Lambda Authorizers only
+            if not authorizer.function_arn:
```
keetonian: Is this the best way to check if this is a lambda authorizer? Are lambda authorizers required to have an ARN at this point, and is it guaranteed that the other authorizer types won't?

brettstack: Yes they must have a FunctionArn and yes this is the best way to differentiate between Lambda and non-Lambda. If new authorizer types are added in the future which have a FunctionArn it is likely they will also need a permission, and on the off-chance they don't then we can update this logic.

brettstack:
Currently working on the Cognito example, followed directly by tests. Then there's just a couple of documentation changes and we can merge this in.

hoegertn:
So do you have an idea when this will be available in production?

brettstack:
As part of our policy, I can't comment on dates. I can say it will be "soon" after it is merged into develop

samtranslator/translator/arn_generator.py
```diff
@@ -3,12 +3,18 @@
...
     @classmethod
-    def generate_arn(cls, partition, service, resource):
+    def generate_arn(cls, partition, service, resource, include_account_id=True):
         if not service or not resource:
             raise RuntimeError("Could not construct ARN for resource.")
 
-        return 'arn:{0}:{1}:${{AWS::Region}}:${{AWS::AccountId}}:{2}'.format(partition, service, resource)
+        arn = 'arn:{0}:{1}:${{AWS::Region}}:'
 
+        if include_account_id:
```
keetonian: Which resources don't require an account ID? I noticed that this is only used for `authorizer_uri`, do you have a link for that documentation handy? If not I'll look for it and read up on it.

samtranslator/model/eventsources/push.py
```diff
@@ -471,6 +472,37 @@ def _add_swagger_integration(self, api, function):
...
+                if method_authorizer != 'NONE' and not api_authorizers.get(method_authorizer):
+                    raise InvalidEventException(
+                        self.relative_id,
+                        'Unable to set Authorizer [{authorizer}] on API method [{method}] for path [{path}] because it '
+                        'wasn\'t defined in the API\'s Authorizers.'.format(
+                            authorizer=method_authorizer, method=self.Method, path=self.Path))
+
+                if method_authorizer == 'NONE' and not api_auth.get('DefaultAuthorizer'):
```
keetonian: Wouldn't this case mean that they do not want an authorizer for this api, and therefore it shouldn't matter if it is set to `NONE`?

samtranslator/model/api/api_generator.py
keetonian: Can you either leave a note here or in the corresponding issue making note of this?

samtranslator/model/eventsources/push.py
```diff
@@ -471,6 +472,37 @@ def _add_swagger_integration(self, api, function):
                     method=self.Method, path=self.Path))
 
         editor.add_lambda_integration(self.Path, self.Method, uri)
+
+        if self.Auth:
+            method_authorizer = self.Auth.get('Authorizer')
+
+            if method_authorizer:
```
keetonian: Looking at the test coverage, it shows that lines 479-504 (this section) were missed in the tests. Could you make sure that tests are hitting this section?

samtranslator/swagger/swagger.py
```diff
@@ -285,6 +286,108 @@ def _make_cors_allowed_methods_for_path(self, path):
...
+        Resource Policy, etc.
+
+        :param string path: Path name
+        :param string method_name: Method name
+        :param dict auth: Auth configuration such as Authorizers, ApiKey, ResourcePolicy (only Authorizers supported
+                          currently)
+        :param dict api: Reference to the related Api's properties as defined in the template.
+        """
+        method_authorizer = auth and auth.get('Authorizer')
+        if method_authorizer:
```
keetonian: It also seems like tests may not be hitting this section of code. Can you verify?

keetonian: (Test readout says that [325->exit] and [360->exit] are not being tested)

examples/2016-10-31/api_cognito_auth/template.yaml
```diff
@@ -0,0 +1,153 @@
...
+      Timeout: 3
+
+  LambdaCognitoUserPoolExecutionPermission:
+    Type: AWS::Lambda::Permission
+    Properties: 
+      Action: lambda:InvokeFunction
+      FunctionName: !GetAtt PreSignupLambdaFunction.Arn
+      Principal: cognito-idp.amazonaws.com
+      SourceArn: !Sub 'arn:aws:cognito-idp:${AWS::Region}:${AWS::AccountId}:userpool/${MyCognitoUserPool}'
+      # TODO: Add a CognitoUserPool Event Source to SAM to create this permission for you.
```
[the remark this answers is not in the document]
brettstack: New feature. I just wanted to document it here for now as we will update this example once it's ready.

versions/2016-10-31.md
```diff
@@ -628,3 +639,53 @@ Cors:
...
+#### API Auth Object
+
+Configure Auth on APIs. Define Lambda and Cognito `Authorizers` and specify a `DefaultAuthorizer`.
+
+```yaml
+Auth:
+  DefaultAuthorizer: MyCognitoAuth # OPTIONAL
+  Authorizers:
+    MyCognitoAuth:
+      Type: COGNITO_USER_POOLS # OPTIONAL; Defaults to 'COGNITO_USER_POOLS' when `UserPoolArn` is specified
```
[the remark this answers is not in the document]
brettstack: LAMBDA_REQUEST and LAMDBA_TOKEN are also valid values.

versions/2016-10-31.md
```diff
@@ -628,3 +639,53 @@ Cors:
 ```
 
 > NOTE: HTTP spec requires the value of Allow properties to be a quoted string. So don't forget the additional quotes in the value. ie. "'www.example.com'" is correct whereas "www.example.com" is wrong
+
+#### API Auth Object
+
+Configure Auth on APIs. Define Lambda and Cognito `Authorizers` and specify a `DefaultAuthorizer`.
```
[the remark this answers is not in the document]
brettstack: Good idea

samtranslator/swagger/swagger.py
brettstack: The terminal coverage report isn't clear. If you look at the html report it is a "partial" miss. In this case it's saying that `method_authorizer` was never false.

samtranslator/model/eventsources/push.py
brettstack: The terminal coverage report isn't clear. If you look at the html report it is a "partial" miss. In this case it's saying that `method_authorizer` was never false.

samtranslator/translator/arn_generator.py
brettstack: Note sure. This was added for reusability.

samtranslator/model/eventsources/push.py
keetonian: My mistake. Thank you

versions/2016-10-31.md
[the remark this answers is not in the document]
keetonian: Could we also include that in our documentation? I know it is an optional field, but this is a very useful if somewhat confusing new feature, and I'd like to include all the details that people would need in order to start using it once it is available.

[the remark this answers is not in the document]
brettstack: Actually this is incorrect. There is no `Type` property as it is inferred. I'll fix.

examples/2016-10-31/api_cognito_auth/template.yaml
```diff
@@ -0,0 +1,153 @@
...
+  CognitoUserPoolClientName:
+    Type: String
+    Default: MyUserPoolClient
+
+Resources:
+  MyApi:
+    Type: AWS::Serverless::Api
+    Properties:
+      StageName: Prod
+      Cors: "'*'"
```
ombozek: This seems wrong to me. It is a SAM thing that we want it to be a string?

examples/2016-10-31/api_lambda_token_auth/template.yaml
```diff
@@ -0,0 +1,54 @@
...
+  MyApi:
+    Type: AWS::Serverless::Api
+    Properties:
+      StageName: Prod
+      Auth:
+        DefaultAuthorizer: MyLambdaTokenAuthorizer
+        Authorizers:
+          MyLambdaTokenAuthorizer:
+            FunctionArn: !GetAtt MyAuthFunction.Arn
+            # FunctionInvokeRole: !Ref MyRole
```
ombozek: Why are these commented out?

examples/2016-10-31/api_lambda_token_auth/template.yaml
```diff
@@ -0,0 +1,54 @@
...
+            Method: get
+            Auth:
+              Authorizer: NONE
+        GetUsers:
+          Type: Api
+          Properties:
+            RestApiId: !Ref MyApi
+            Path: /users
+            Method: get
+            # Auth:
```
ombozek: Same here

samtranslator/model/api/api_generator.py
```diff
@@ -208,6 +216,116 @@ def _add_cors(self):
...
 
+    def _add_auth(self):
+        """
+        Add Auth configuration to the Swagger file, if necessary
+        """
+
+        if not self.auth:
+            return
+
+        INVALID_ERROR = "Invalid value for 'Auth' property"
```
ombozek: Why constify this if you're only going to use it once? It seems like the rest of the error messages are just string literals

samtranslator/model/apigateway.py
```diff
@@ -85,3 +87,170 @@ def make_auto_deployable(self, stage, swagger=None):
...
+        context = identity.get('Context')
+
+        if not headers and not query_strings and not stage_variables and not context:
+            return True
+
+        return False
+
+    def generate_swagger(self):
+        authorizer_type = self._get_type()
+        APIGATEWAY_AUTHORIZER_KEY = 'x-amazon-apigateway-authorizer'
```
ombozek: Nit: Instead of using this key all over in this map, why not assign the submap to a var
`apigw_auth_key = swagger[x-amazon-apigateway-authorizer]`

samtranslator/translator/arn_generator.py
```diff
@@ -3,12 +3,18 @@
...
+            arn += '${{AWS::AccountId}}:'
+
+        arn += '{2}'
```
ombozek: If we don't add an accountID we'd still have to add the corresponding colon and asterix or we'd evaluate the resource as the account id

examples/2016-10-31/api_cognito_auth/template.yaml
brettstack: SAM CORS uses API Gateway response mapping, and this is an implementation detail from API Gateway.

examples/2016-10-31/api_lambda_token_auth/template.yaml
brettstack: This is to show additional configuration options. I think it's useful to users.

examples/2016-10-31/api_lambda_token_auth/template.yaml
brettstack: Same there 👆

samtranslator/model/api/api_generator.py
brettstack: Yeah I considered changing this but it's how it is in other parts of the code so I was being consistent

ajhool:
Very excited for this feature. I'm not entirely where translation from SAM to CloudFormation occurs (whether it is sam-cli, client-side, or behind AWS's walls). Once this PR is merged into master, will it be immediately ready to use?

This is my package and deploy script, which is called from CodePipeline/CodeBuild:
```
aws cloudformation package --template-file ./template.yaml --output-template-file ./build/sam-output.yaml --s3-bucket some-bucket`
sam deploy --template-file ./sam-output.yaml --stack-name some-stack --region us-east-1 --capabilities CAPABILITY_IAM
```

keetonian:
This feature has now been released and is available to use

OndeVai:
@brettstack @keetonian The example template at https://github.com/awslabs/serverless-application-model/blob/master/examples/2016-10-31/api_cognito_auth/template.yaml fails SAM validation via "sam validate -t [filename]". 

> Error: [InvalidResourceException('MyApi', "Cors works only with inline Swagger specified in 'DefinitionBody' property")] ('MyApi', "Cors works only with inline Swagger specified in 'DefinitionBody' property")

OndeVai:
@keetonian I'm on version 0.7.0

keetonian:
Ok I looked into it- `sam validate` doesn't yet support the new auth feature, and errors when neither DefinitionBody nor DefinitionUri is set because it expects one of them to be there (as found [here](https://github.com/awslabs/aws-sam-cli/blob/develop/samcli/commands/validate/lib/sam_template_validator.py#L108)).

OndeVai:
Good deal. Thanks. Good news is that the sam deploy works like a champ :)

keetonian:
@OndeVai Could you create a new issue for this use case, detailing how you would like it to work so that we (and others) can more easily track the design and development of this feature?

OndeVai:
Will do. Thanks.

Issue #535, opened 2018-08-01
1.5.4 vs 1.6.1 behavioral change - JSON_VALIDATION_WARNING  


bedge:
If I upgrade from 1.5.4 to 1.6.1, a template which I was able to transform previously, now generates warnings.

The use case is here: https://github.com/Dunedan/serverless-applications-with-aws-sam/blob/gh-pages/local_samtranslator.py

❯ cat test.yaml | ./tools/local_samtranslator.py
WARNING:root:JSON_VALIDATION_WARNING: OrderedDict([('Properties', OrderedDict([('CodeUri', 's3://infor-devops-dev-appdata-us-east-1/deng/lambda/ig1xm/master/1.0.0-bedge/ig1xm.zip'), ('Environment', OrderedDict([('Variables', OrderedDict([
('GRID_ENV', OrderedDict([('Ref', 'GridEnv')])), ('LOG_LEVEL', 'DEBUG'), ('NODE_CPU_MAX', 80), ('NODE_ERRORS_MAX', 1000), ('NODE_RAM_MAX', 95), ('NODE_WARNINGS_MAX', 1000), ('PD_SERVICE', 'test')]))])), ('Events', OrderedDict([('Schedule'
, OrderedDict([('Properties', OrderedDict([('Schedule', 'rate(4 hours)')])), ('Type', 'Schedule')]))])), ('Handler', 'igm-lambda.poll'), ('Runtime', 'python3.6'), ('Timeout', 60), ('VpcConfig', OrderedDict([('Role', OrderedDict([('Ref', '
LambdaExecRole')])), ('SecurityGroupIds', OrderedDict([('Ref', 'SecurityGroupIds')])), ('SubnetIds', [OrderedDict([('Ref', 'SubnetIds')])])]))])), ('Type', 'AWS::Serverless::Function')]) is not valid under any of the given schemas

Failed validating u'anyOf' in schema[u'properties'][u'Resources'][u'patternProperties'][u'^[a-zA-Z0-9]+$']:
    {u'anyOf': [{u'$ref': u'#/definitions/AWS::Serverless::Api'},
                {u'$ref': u'#/definitions/AWS::Serverless::Function'},
                {u'$ref': u'#/definitions/AWS::Serverless::SimpleTable'},
                {u'$ref': u'#/definitions/CloudFormationResource'}]}

The resulting output is unchanged, but I felt the message was worth reporting.

brettstack:
[...]

PR #549, opened 2018-08-15
fix: remove JSON Schema validation warning

brettstack:
The schema is currently lacking and therefore is just noise. We may add this back in the future or remove it entirely.

*Issue #535, if available:*

*Description of changes:*

### DIFF
--- a/samtranslator/parser/parser.py
+++ b/samtranslator/parser/parser.py
@@ -27,10 +27,10 @@ class Parser:
                 [InvalidTemplateException("'Resources' section is required")])
 
         validation_errors = SamTemplateValidator.validate(sam_template)
-        has_errors = len(validation_errors)
+        # has_errors = len(validation_errors)
 
-        if has_errors:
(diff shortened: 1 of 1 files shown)


samtranslator/parser/parser.py
```diff
@@ -27,10 +27,10 @@ def _validate(self, sam_template, parameter_values):
                 [InvalidTemplateException("'Resources' section is required")])
 
         validation_errors = SamTemplateValidator.validate(sam_template)
```
keetonian: Why leave this line uncommented?

brettstack: The problem is so widespread that it's not feasible for us to fix this due to other priorities. Additionally, the original intent of the schema was not so much for validation, but for linters, IDE plugins etc. and we now have https://github.com/awslabs/cfn-python-lint for that.

Still need the call to `validate()` as it does additional validation, not just schema.

PR #554, opened 2018-08-28
docs: update example apps to use policy templates and parameters

brettstack:
*Issue #, if available:*

*Description of changes:*


jlhood:
Could you add a testing section to your PR description? Just wondering if you actually tried to deploy these updates.

brettstack:
@jlhood @keetonian I'm going to test several of these, but will rely on end-to-end testing when it comes time to redeploy these to the app repo.

examples/apps/cloudfront-ab-test/template.yaml
```diff
@@ -1,6 +1,11 @@
 AWSTemplateFormatVersion: '2010-09-09'
 Transform: 'AWS::Serverless-2016-10-31'
 Description: 'Blueprint for CloudFront ab testing, implemented in NodeJS.'
+Metadata:
+  Name: cloudfrontabtest
```
jlhood: Remove this metadata. It'll actually cause publish to fail because the published name (cloudfront-ab-test) does not match this field. Should just remove the metadata completely.

examples/apps/update-apps.js
```diff
@@ -0,0 +1,32 @@
...
+  readFileSync,
+  writeFileSync
+} = require('fs')
+
+const path = require('path')
+const bluePrintsDirectory = '_Blueprints-apps'
+const bluePrintAppDirectories = readdirSync(bluePrintsDirectory)
+const ignore = ['.DS_Store', 'mass-pub']
+const appIgnore = ['node_modules']
+bluePrintAppDirectories.forEach(blueprintAppDirectoryName => {
```
jlhood: This script doesn't seem like it belongs in this PR. Remove?

examples/apps/lex-make-appointment/lambda_function.py
```diff
@@ -1,214 +0,0 @@
-'''
```
jlhood: Why is this file getting removed? The template still seems to reference it. Same question for some other apps here.

examples/apps/lex-make-appointment/requirements.txt
```diff
@@ -1 +0,0 @@
-requests==2.13.0
```
jlhood: Why remove?

examples/apps/lex-make-appointment/lambda_function.py
keetonian: The template in this folder seems to only references a different node.js file and never uses this python file. Could you specify which template you're talking about?

keetonian: Look at the directory structure. This app is found at `examples/apps/lex-make-appointment/` and is the js version.

jlhood: Oh haha I see now. I was looking at the lex-make-appointment-python template, not lex-make-appointment. Thanks for clarifying!

examples/apps/lex-make-appointment/requirements.txt
jlhood: Disregard.

PR #556, opened 2018-08-30
add EKSDescribePolicy - Gives permission to describe or list Amazon EKS clusters

pahud:
*Issue #, if available:*

*Description of changes:*
Description of changes: [pahud/eks-lambda-drainer](https://github.com/pahud/eks-lambda-drainer) will require this policy template so that Lambda function can describe or list Amazon EKS clusters.


docs/policy_templates_data/policy_templates.json
```diff
@@ -1207,6 +1207,23 @@
...
+    "EKSDescribePolicy": {
+      "Description": "Gives permission to describe or list Amazon EKS clusters",
+      "Parameters": {
+      },
+      "Definition": {
+        "Statement": [
+          {
+            "Effect": "Allow",
+            "Action": [
+              "eks:DescribeCluster",
```
jlhood: Policy templates are meant to be scoped down very tightly for strong security. I understand ListClusters needs Resource: *, but ideally DescribeCluster should be scoped down to a specific cluster, e.g., `arn:${AWS::Partition}:eks:${AWS::Region}:${AWS::AccountId}:cluster/${ClusterName}`.

I'm not clear on what your exact use case is, but is there a way to make this policy template scoped down further? You could break it up into two policy templates, one for listing all clusters, then one for describing a specific cluster.

pahud: Hi,

The use case is when we have multiple Amazon EKS clusters in the same region and each cluster has its nodegroup with some spot instances in it. Whenever any single spot instance is about to terminate, the cloudWatch event rule will receive a signal for "Spot instance interruption warning" and this rule has a Lambda target to handle the notification from all spot instances from multiple Amazon EKS nodegroup.

The Lambda function will:
1) get the instance ID(i.e. i-xxxxxxxxx) from the notification from CloudWatch Event
2) look up the tags of the instance to get the Amazon EKS cluster name it belongs to (see [#1](https://github.com/pahud/eks-lambda-drainer/issues/1))
3) execute [eks:DescribeCluster](https://docs.aws.amazon.com/cli/latest/reference/eks/describe-cluster.html) to get its endpoint
4) use [client-go](https://github.com/kubernetes/client-go) to visit this Kubernetes API endpoint and perform the pod draining or taint-based eviction via kubernetes API

As this scenario may  have multiple Amazon EKS clusters yet only single Lambda function handling all the spot instances termination, we can't scope it down to a specific single Amazon EKS cluster.

I have made this in [pahud/eks-lambda-drainer](https://github.com/pahud/eks-lambda-drainer) and is about to publish to Serverless App Repo and hopefully everyone can deploy this feature with single button click to deploy it from SAR.

brettstack: I think this is okay since it's only List+Describe.

PR #577, opened 2018-09-12
ci: patch dependency on boto3 region_name param

keetonian:
*Issue #, if available:*

*Description of changes:*
If aws-cli is not configured with a region name, many tests used to fail because boto3 uses this configuration to resolve the region name. Additionally, some tests were overriding the environment variable `AWS_DEFAULT_REGION` (set by aws-cli and picked up by boto3) to test for different regions. This PR removes the dependency on having a correct aws-cli configuration for boto3 as well as these overrides in the tests.

### DIFF
--- a/tests/translator/test_translator.py
+++ b/tests/translator/test_translator.py
@@ -23,8 +23,9 @@ from samtranslator.translator.transform import transform
 from mock import Mock, MagicMock, patch
 
 
-input_folder = 'tests/translator/input'
-output_folder = 'tests/translator/output'
+BASE_PATH = os.path.dirname(__file__)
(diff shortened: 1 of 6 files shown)


tests/translator/test_translator.py
```diff
@@ -175,10 +175,7 @@ def test_transform_success(self, testcase, partition_with_region):
         partition_folder = partition if partition != "aws" else ""
         expected = json.load(open(os.path.join(OUTPUT_FOLDER,partition_folder, testcase + '.json'), 'r'))
 
-        old_region = os.environ.get("AWS_DEFAULT_REGION", "")
```
jlhood: lol, glad you're removing this!

Issue #603, opened 2018-09-29
Support storing parameters in SSM Parameter Store


RobRoseKnows:
Currently there's no policy templates that support the `ssm:GetParameter` permission required to do stuff like [store secret parameters](https://aws.amazon.com/blogs/mt/use-parameter-store-to-securely-access-secrets-and-config-data-in-aws-codedeploy/). There's examples of storing secrets in [some of the examples](https://github.com/awslabs/serverless-application-model/blob/master/examples/apps/cloudwatch-alarm-to-slack/index.js), but it just uses KMS. This is mentioned in #48 after it was closed, but it uses inline policies not supported by the SAR.

I'm not sure if this is an extremely desirable feature but since it's unlikely it could be released before the end of the Devpost hackathon, I wouldn't mind doing this as a Hacktoberfest PR.

brettstack:
This sounds like a useful Policy Template and we'd appreciate the PR.

keetonian:
Released!

PR #625, opened 2018-10-13
Added two new policy templates for SSM Parameter Store.

RobRoseKnows:
*Issue #:* #603 

*Description of changes:* Added two new policy templates: `DecodeSSMParameterPolicy` and `DecodeSSMParameterCustomKeyPolicy`. Also added parameters for the SQS queue and Splunk environment variables in two of the example apps. Kinda forgot I did that. 😅 


samtranslator/policy_templates_data/policy_templates.json
```diff
@@ -1430,6 +1430,106 @@
...
+            "Effect": "Allow",
+            "Action": [
+              "ssm:DescribeParameters"
+            ],
+            "Resource": "*"
+          },
+          {
+            "Effect": "Allow",
+            "Action": [
+              "ssm:GetParameters"
```
keetonian: I'm reviewing [SSM actions](https://docs.aws.amazon.com/systems-manager/latest/userguide/auth-and-access-control-permissions-reference.html) and it looks like there is a `ssm:GetParameter` option that only allows accessing one parameter instead of several. Is there a reason why you chose to use `ssm:GetParameters` over `ssm:GetParameter`?

samtranslator/policy_templates_data/policy_templates.json
```diff
@@ -1430,6 +1430,106 @@
...
+        }
+      },
+      "Definition": {
+        "Statement": [
+          {
+            "Effect": "Allow",
+            "Action": [
+              "ssm:DescribeParameters"
+            ],
+            "Resource": "*"
```
keetonian: Is it not possible to scope `ssm:DescribeParameters`?

examples/apps/sqs-poller/template.yaml
```diff
@@ -1,9 +1,13 @@
...
 Description: Periodically polls an SQS queue and asynchronously consumes each message.
-Parameters: 
-  QueueNameParameter: 
+Parameters:
+  QueueNameParameter:
     Type: String
+    Description:
+  QueueUrlParameter:
+    Type: String
+    AllowedPattern: '^https:\/\/sqs\.[a-z\-0-9]+\.amazonaws\.com\/[0-9]{12}\/.{1,80}$'
```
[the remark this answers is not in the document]
RobRoseKnows: Alright, I'll fix that.

samtranslator/policy_templates_data/policy_templates.json
RobRoseKnows: I was basing this off of a tutorial example that used `ssm:GetParameters` instead of `ssm:GetParameter`. I'll look into seeing what the difference is, because it didn't specify.

samtranslator/policy_templates_data/policy_templates.json
RobRoseKnows: Not sure, as I mentioned above that's a legacy of the tutorial I was following, I'll look into whether it's necessary or if it can be scoped.

keetonian: Thanks! It's possible that this API action requires a `"*"` resource, but we would like to scope things where possible and applicable.

RobRoseKnows: As far as I can tell, ssm:DescribeParameters can't be scoped. I don't know if it's necessary though, since [some examples show it in](https://docs.aws.amazon.com/systems-manager/latest/userguide/sysman-paramstore-access.html), but [other examples show it out](https://docs.aws.amazon.com/kms/latest/developerguide/services-parameter-store.html). I'm going to leave it in since it isn't a terribly destructive permission.

samtranslator/policy_templates_data/policy_templates.json
RobRoseKnows: I *think* `ssm:GetParameters` allows the use of wildcards in the parameter names, but I don't know if how/if it differs from `ssm:GetParameter` besides that. I've resolved it by having both of them in the template.

samtranslator/policy_templates_data/policy_templates.json
```diff
@@ -1430,6 +1430,108 @@
...
+            }
+          },
+          {
+            "Effect": "Allow",
+            "Action": [
+              "kms:Decrypt"
+            ],
+            "Resource": {
+              "Fn::Sub": [
+                "arn:${AWS::Partition}:ssm:${AWS::Region}:${AWS::AccountId}:alias/aws/ssm"
```
brettstack: From these docs https://docs.aws.amazon.com/systems-manager/latest/userguide/sysman-paramstore-access.html

|  Each AWS account is assigned a default AWS KMS key. You can view your key by executing the following command from the AWS CLI:

I'm hesitant to grant permission to decrypt this key.

RobRoseKnows: Yes using this key is probably non-ideal, but as far as I can tell, it's what's used when you create a SecureString parameter without a defined key. I did write an alternate template for using a custom key however, as I agree this is non-ideal. 

Do you think changing the name to `DecodeSSMParameterRootKeyPolicy` and the description to indicate it grants access to the root key would be appropriate to alleviate security concerns?

samtranslator/policy_templates_data/policy_templates.json
```diff
@@ -1430,6 +1430,108 @@
...
+              ]
+            }
+          }
+        ]
+      }
+    },
+    "DecodeSSMParameterCustomKeyPolicy": {
```
jlhood: We don't need this policy template. Let's just use this PR to define the `SSMParameterReadPolicy` for SSM API permissions. Then, if users use a custom KMS key to encrypt a parameter, they can add the existing `KMSDecryptPolicy` to give permissions to use that custom key for decrypting the SSM parameter value.

samtranslator/policy_templates_data/policy_templates.json
```diff
@@ -1430,6 +1430,108 @@
           "Resource": "*"
         }]
       }
+    },
+    "DecodeSSMParameterPolicy": {
```
jlhood: Let's call this `SSMParameterReadPolicy` to match naming convention of other policy templates.

RobRoseKnows:
@jlhood Merged in develop in e89a714. I removed the second policy and the second statement in 8eedad6. Looks like the Travis check failed on something related to Flake8. I'm not sure why that'd fail though, all the tests passed on my machine.

PR #637, opened 2018-10-24
Use correct path for SES notifications Lambda

duboff:
SES Notifications coming via SNS do not actually have the `notificationType` param on the `message` object. Instead there is an `eventType` one. At least on the events we've tested.

*Issue #, if available:*

*Description of changes:*
This PR changes the name of the path parameter to the correct one actually supplied by SNS.

### DIFF
--- a/examples/apps/ses-notification-nodejs/index.js
+++ b/examples/apps/ses-notification-nodejs/index.js
@@ -5,7 +5,7 @@ exports.handler = (event, context, callback) => {
     //console.log('Received event:', JSON.stringify(event, null, 2));
     const message = JSON.parse(event.Records[0].Sns.Message);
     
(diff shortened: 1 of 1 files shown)


jlhood:
Path forward:

1. @keetonian will verify message has `eventType` field instead of `notificationType`. If this is the case, our [docs](https://docs.aws.amazon.com/ses/latest/DeveloperGuide/receiving-email-notifications-contents.html) need to be updated as well.
1. @duboff please address @keetonian's callout to change `notificationType` to `eventType` in one other place in the file.

jfuss:
This example (along with all under examples/apps) where moved to https://github.com/aws-samples/serverless-app-examples.

Closing.

Issue #261, opened 2018-01-20
Filter policy support for SNS events


joshm91:
Hi guys,

It'd be great to be able to support the new(-ish) [Filter Policy](https://docs.aws.amazon.com/sns/latest/dg/message-filtering.html) feature in SNS when specifying an SNS event on a Lambda function.  

Something like this would be perfect:

```yaml
  myLambda:
    Type: AWS::Serverless::Function
    Properties:
      Handler: lambda_handlers.handler
      ...
      Events:
        SNS:
          Type: SNS
          Properties:
            Topic: !GetAtt someSnsTopic.Arn
            FilterPolicy:
              AttributeName1:
                - value1
                - value2
              AttributeName2:
                - value3
```

jfuss:
@joshm91 Filter Policy is not available for CloudFormation just yet and therefore not able to be apart of SAM. Marking this as a feature request.

Nr18:
Just yet 🤔so it’s coming 🤪

joshm91:
Faur enough, thanks!

oharaandrew314:
Yes, this would be great.  I have several lambdas listening to one SNS topic, and I would vastly prefer being able to filter outside of the lambda invocation, which skews my invocation counts.

kevohagan:
@jfuss is this available in CloudFormation now? this would be great in sam! thx

vrivellino:
It looks like it's available now: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/ReleaseHistory.html

bottemav:
@joshm91 In the CloudFormation template one can also filter on a prefix or a numeric value matching. Do you need this too or does whitelisting (exact string matching) satisfy your use case?

brettstack:
Happy to take PRs on this.

brettstack:
Thanks!

otaviofff:
Hey guys. How is it coming together? Thanks a lot

brettstack:
We've just merged this into develop branch and will go out in our next release. Thanks @bottemav 🙌

otaviofff:
Fantastic! Thanks for adding the suppport for SNS message filtering.

brettstack:
Closing this since it's been merged into develop and will be released in the next release.

davidtwamley:
[...]

PR #648, opened 2018-10-28
Feature/sns filter policy

bottemav:
*Issue #261 :*

*Description of changes:*
- document new property FilterPolicy of SNS event source
- extend the SNS event source with an optional property filterPolicy
- if specified in the SAM template, the value of the new property is copied to the property filterPolicy of the SNS subscription
- new unit test for SNS event source
- new SAM template for the test translator


brettstack:
This is great! We'll take a look; in the meantime do you think you could add an example? You can check out this example as a good example of an example https://github.com/awslabs/serverless-application-model/tree/master/examples/2016-10-31/api_lambda_request_auth :)

ryanwmarsh:
Very useful! Would love to see this one get merged.

otaviofff:
Ready to merge? Thanks

PR #663, opened 2018-11-14
Update Schema by including a validation rule for the Api Gateway Path

matteofigus:
*Issue #, if available:*
Follows a previous discussion in a Pull Request for SAM CLI: https://github.com/awslabs/aws-sam-cli/pull/767

*Description of changes:*
When using SAM CLI, if using a not valid Api Gateway path (for instance `/~hello`) , SAM CLI succeeds during dev and fails during deploy. By adding this rule in SAM, the cli should automatically pick up the rule, and we should be able to alert the user earlier in the process and improve the developer experience. 

Note: I am not sure about how to test this change. I tried adding a file with a broken rule in the `tests/translator/input` for TDD but the test failed because no output was present (which I was assuming the test would generate for me). Would you be able to provide some help on how I can do that?

Thanks a lot for your help 🙏 

### DIFF
--- a/samtranslator/validator/sam_schema/schema.json
+++ b/samtranslator/validator/sam_schema/schema.json
@@ -311,6 +311,11 @@
           "type": "string"
         },
         "Path": {
+          "patternProperties": {
+            "^[a-zA-Z0-9\\-\\_\\.\/\\{\\}\\+]+$": {
+              "type": "string"
+            }
+          },
           "type": "string"
         },
         "RestApiId": {


jlhood:
Added a comment on test updates to get this PR merged. However, I also want to make you aware that this PR alone won't solve your larger issue of wanting SAM CLI's validate command to catch this error for you. Due to a number of issues with the JSON schema validation (false negative/positive issues), we decided to temporarily disable it in #549. So even with this change merged, it is effectively not run as part of `sam validate`.

We have an open issue (#675) to track fixing the false positive/negative issues so we can re-enable this validation. I still suggest you follow through and complete this PR so once we re-enable validation, your checks are included.

jlhood:
@matteofigus That sounds great. Thanks again!

keetonian:
Actually, @matteofigus the schema was never intended to provide the full validation experience that customers are looking for, and was returning some false errors and warnings before it was turned off. We will be taking a different route for better validation as we are now tracking in https://github.com/awslabs/serverless-application-model/issues/675. We should probably remove the schema, as we aren't using it and it causes some confusion for customers.

PR #682, opened 2018-11-24
Allow Api.Auth in Globals section

sjawhar:
https://github.com/awslabs/serverless-application-model/issues/512 added the ability to set API auth, and the original issue outlined including in the Globals section. However, it looks like the Globals checker actually is excluding Auth as one of the supported properties. Is there a reason that is so? I figured opening a PR was a quick and easy way to start a discussion as well as move toward a fix. Looking forward to your feedback!


sjawhar:
No problem, it was an expected error message that needed updating. This doesn't necessarily address the need for new tests, though. If you think they are needed, please point me to where I can get started.

sjawhar:
I've added the Auth section to the globals_for_api test. Now it looks like the test is failing because unicode is causing the match assertion to fail. Here's a snippet:
image

Any tips?

keetonian:
I'll take a look, I haven't seen this before.

cadetstar:
Is there an ETA for this getting published and available for use? Alternately, can I override the branch aws-sam-cli uses locally to validate the template?

Issue #669, opened 2018-11-18
Allow AWS::Serverless::Function event sources to be created disabled


jlhood:
`AWS::Lambda::EventSourceMapping` has an Enabled property that defaults to true, but can be set to false in CloudFormation. Opening this issue asking to add this property to the Events definitions for supported event types in `AWS::Serverless::Function`, e.g., SQS, so an event source can be initially disabled on create.

Use case: I'm writing an app where I want an SQS queue storing messages to be replayed to another SQS queue in the event of a disaster recovery scenario. I want a Lambda attached and ready to read messages from the replay buffer queue and put them onto a destination queue. However, I want that event source mapping to be disabled by default. Then an operator can manually enable it as part of a disaster recovery runbook to replay recent messages to the destination queue.

Right now I'm working around this by not using the SQS integration in `AWS::Serverless::Function`, but it means I have to create a lot more resources in my template. 😭

brettstack:
I agree. It would be simpler to simply enable a mapping rather than creating the even source mapping during that scenario. This should be a relatively simple addition; here are some places to get started if anyone is interested:

https://github.com/awslabs/serverless-application-model/blob/master/samtranslator/model/eventsources/pull.py#L18.

https://github.com/awslabs/serverless-application-model/blob/master/samtranslator/model/eventsources/pull.py#L63

Docs ands tests too are greatly appreciated.

kurtmaile:
Agreed - I opened the original issue months ago on this.  You get the issue if you disable the ESM (e.g support issue and need to stop it quickly, you actually then get a new ESM ID when you re-enable and creates all sorts of issues with SAM/CF as you get orphaned ESM instances)

Any idea on a release date for this?

Cheers!

kurtmaile:
....my issue was more a one of just being able to toggle it, not necessarily just on create.  I tried this the other day again and it didnt work, also docs dont reference it, am I being a total dofus and missing it somewhere??  Sorry if so!

keetonian:
Hi @kurtmaile- we just had a PR submitted by @tryantwit and this will go out with the next release.

keetonian:
Released!

PR #690, opened 2018-11-29
Event source disabled

tryantwit:
Issue #669 

*Description of changes:*
Added Enabled boolean option for SQS, DynamoDB and Kinesis events.
Updated documentation to show support for this property addition.
Modified translator tests to verify that cloudformation templates generate with Enabled as part of AWS::Lambda::EventSourceMapping.


Inesa09:
Isn't it useful to replicate this change for other types of events? Just found it useful for API in my case.

PR #699, opened 2018-12-03
docs: add example application using Layers

keetonian:
*Description of changes:*
Add an example using Lambda Layers


examples/2016-10-31/image_resize_python/README.md
```diff
@@ -0,0 +1,44 @@
+# AWS::Serverless::LayerVersion Library Code Example
+
+This example shows you how to download and publish an application that bundles its Function dependencies into an AWS::Serverless::LayerVersion resource.
```
paoptu023: Can also add explanation like "The function resizes an image in the source bucket and saves the output to the target bucket" to give a bit more context. I could only find out what the example does after clicking the link at the bottom.

examples/2016-10-31/image_resize_python/src/index.py
```diff
@@ -0,0 +1,33 @@
...
+    for record in event['Records']:
+        bucket = record['s3']['bucket']['name']
+        key = record['s3']['object']['key'] 
+        download_path = '/tmp/{}{}'.format(uuid.uuid4(), key)
+        upload_path = '/tmp/resized-{}'.format(key)
+        
+        s3_client.download_file(bucket, key, download_path)
+        try:
+            resize_image(download_path, upload_path)
+        except Exception as e:
```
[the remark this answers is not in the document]
keetonian: No, this can be removed.

examples/2016-10-31/image_resize_python/README.md
keetonian: Good call, updated

jlhood:
Is there a reason not to merge this directly to the master branch?

PR #700, opened 2018-12-03
fix: add missing parameter to s3 bucket resource

keetonian:
This is not an ideal situation- needing to update this resource with every single new feature. We should consider a way to update our underlying design so we don't need to continually update these resources, but let them pass through parameters to CFN.

*Issue #, if available:*
https://github.com/awslabs/serverless-application-model/issues/679
*Description of changes:*
Add missing `PublicAccessBlockConfiguration` parameter to S3 buckets


carlnordenfelt:
Any idea when this will be released?

Issue #142, opened 2017-05-25
Transforms do not support Conditions


brandond:
If I put a Condition on a resource of type "AWS::Serverless::Function", this condition is neither honored by the transform, nor is it passed through to the resources created by the transform. This makes it impossible to conditionally create transformed resources.

### Tracking Conditions support:
- [x] Serverless::Function
    - [x] Function resource (v1.10.10)
    - [x] Events (v1.10.0) (see https://github.com/awslabs/serverless-application-model/issues/214 for tracking conditions support inside of events, which is not implemented in v1.10.0)
    - [x] Implicit API: Issue https://github.com/awslabs/serverless-application-model/issues/758
    - [x] Alias (v1.10.0)
- [x] Serverless::Api (v1.10.0)
- [x] Serverless::LayerVersion (v1.10.0)
- [x] Serverless::Application (v1.10.0)
- [x] Serverless::SimpleTable (v1.10.0)

sanathkr:
Yup, feature request. We already have being implemented now. Will update once its ready.

mikecbrant:
I would add that my experience has not been having the transform silently ignore conditions, but rather to have the conditions cause the generic "internal transform failure" error message at the point of trying to create a change set (template validation does not catch the issue).

In my case, I was trying to conditionally set the "Path" property on an API event in AWS::Serverless:Function like the following:

```
      Events:
        GetByKey:
          Type: Api
          Properties:
            Path: !If [DynamoHasRangeKey, '/{hashKey}/{rangeKey}', '/{hashKey}']
            Method: get
```

The condition works in other areas of the template (like the AWS::DynamoDB::Table resource that is being defined).

brandond:
@mikecbrant that's not a condition, that's a function that evaluates a conditional test (If). Conditions look like this:

```yaml
AWSTemplateFormatVersion: "2010-09-09"
Conditions: 
  IsTargetRegion: !Equals [ !Ref "AWS::Region", "us-west-2" ]
Resources: 
  NewVolume: 
    Type: "AWS::EC2::Volume"
    Properties: 
      Size: 100
    Condition: IsTargetRegion
```

The resource gets skipped entirely if the condition evaluates to false.

mikecbrant:
[...]

PR #707, opened 2018-12-06
Passing Condition to new LayerVersion and Application for #142

Jacco:
*Issue #142 *

*Description of changes:*

Pass passtrhough_resource_attributes when processing LayerVersion and Application

### DIFF
--- a/samtranslator/model/sam_resources.py
+++ b/samtranslator/model/sam_resources.py
@@ -568,7 +568,7 @@ class SamApplication(SamResourceMacro):
     def _construct_nested_stack(self):
         """Constructs a AWS::CloudFormation::Stack resource
         """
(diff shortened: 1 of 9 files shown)


samtranslator/model/sam_resources.py
```diff
@@ -633,9 +633,10 @@ def _construct_lambda_layer(self, intrinsics_resolver):
         """
         retention_policy_value = self._get_retention_policy_value(intrinsics_resolver)
 
-        retention_policy = {
-            'DeletionPolicy': retention_policy_value
-        }
+        retention_policy = self.get_passthrough_resource_attributes()
```
keetonian: Could you rename `retention_policy` here since it could contain both `DeletionPolicy` and `Condition` attributes?

keetonian:
This is awesome, by the way. Thank you so much for doing this!

PR #710, opened 2018-12-08
WIP: fix: support intrinsics in Location property of Serverless Application

keetonian:
*Issue #, if available:*
https://github.com/awslabs/serverless-application-model/issues/694
*Description of changes:*
Support intrinsics such as Ref and Sub in the Location parameter for s3 strings.

This does not add support for intrinsics in ApplicationId or SemanticVersion.

### DIFF
--- a/samtranslator/plugins/application/serverless_app_plugin.py
+++ b/samtranslator/plugins/application/serverless_app_plugin.py
@@ -3,6 +3,7 @@ from botocore.exceptions import ClientError, EndpointConnectionError
 import logging
 from time import sleep, time
 
+from samtranslator.intrinsics import resolver
 from samtranslator.model.exceptions import InvalidResourceException, InvalidDocumentException
 from samtranslator.plugins import BasePlugin
 from samtranslator.plugins.exceptions import InvalidPluginException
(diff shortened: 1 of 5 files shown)


samtranslator/plugins/application/serverless_app_plugin.py
```diff
@@ -101,10 +102,12 @@ def _can_process_application(self, app):
 
         :param dict app: the application and its properties
         """
+        location = app.properties.get(self.LOCATION_KEY)
         return (self.LOCATION_KEY in app.properties
-                and isinstance(app.properties[self.LOCATION_KEY], dict)
-                and self.APPLICATION_ID_KEY in app.properties[self.LOCATION_KEY]
-                and self.SEMANTIC_VERSION_KEY in app.properties[self.LOCATION_KEY])
+                and isinstance(location, dict)
+                and list(location.keys())[0] not in resolver.DEFAULT_SUPPORTED_INTRINSICS
```
paoptu023: why it only checks for the first key?

samtranslator/plugins/application/serverless_app_plugin.py
```diff
@@ -187,7 +190,8 @@ def on_before_transform_resource(self, logical_id, resource_type, resource_prope
         self._check_for_dictionary_key(logical_id, resource_properties, [self.LOCATION_KEY])
 
         # If location isn't a dictionary, don't modify the resource.
-        if not isinstance(resource_properties[self.LOCATION_KEY], dict):
+        if (not isinstance(resource_properties[self.LOCATION_KEY], dict)
```
paoptu023: Can you make it a util function to check these conditions? You have the same check in the function above.

keetonian: Good idea

samtranslator/plugins/application/serverless_app_plugin.py
keetonian: If it's an S3 location, it will only have one dictionary key at this level:
```
Location:
  Ref: parameter
# or
Location:
  Sub: s3://{something}
```
If it has something else, then it's probably not valid.

This brings up a good point though: this doesn't check for all possible intrinsics. I need to fix that.

hawkins:
Why was this closed? Don't we still need this functionality? I'm unable to use intrinsic functions in Location today in version 0.11.0

Ranger296:
Still running into this issue as well while using an intrinsic function (Fn::Join and I've tried Fn::Sub) with the location property in Serverless::Application

""StatusReason": "Transform AWS::Serverless-2016-10-31 failed with: Invalid Serverless Application Specification document. Number of errors found: 11. Resource with id [someStack] is invalid. Resource is missing the required [ApplicationId] property"

PR #711, opened 2018-12-08
WIP: chore: update schema

keetonian:
*Issue #, if available:*
https://github.com/awslabs/serverless-application-model/issues/675
*Description of changes:*
The schema was returning many false warnings, so it was turned off in #549. This PR turns the schema back on, but only to broadly check the syntax of the SAM template. This way, we can have very basic validation on while we discuss in https://github.com/awslabs/serverless-application-model/issues/675 about how we want to validate templates going forward.


keetonian:
Test failure is an invalid globals section. I hesitate to update the tests yet because the error messages returned by the schema really aren't that great...

PR #715, opened 2018-12-12
adding SES Bulk Templated Policy Template

simalexan:
*Description of changes:*
Add SESBulkTemplatedCrudPolicy which grants the user access to:
- `ses:SendTemplatedEmail`
- `ses:SendBulkTemplatedEmail`

*Use case:*
As a developer I'd like to be able to instantly deploy a serverless component from SAR that can send templated emails and bulk templated emails.
I have published three SAR components for sending emails, two of which are now labeled as having Custom IAM roles, as there isn't a policy template that enables them to use `ses:SendTemplatedEmail` and `ses:SendBulkTemplatedEmail`. 

Here are the links to SAR components:
- [api-lambda-send-templated-email-ses](https://serverlessrepo.aws.amazon.com/applications/arn:aws:serverlessrepo:us-east-1:375983427419:applications~api-lambda-send-templated-email-ses)
- [api-lambda-send-bulk-templated-email-ses](https://serverlessrepo.aws.amazon.com/applications/arn:aws:serverlessrepo:us-east-1:375983427419:applications~api-lambda-send-bulk-templated-email-ses)

Thank you for taking this into consideration. I would be interested in adding more in the future.

### DIFF
--- a/samtranslator/policy_templates_data/policy_templates.json
+++ b/samtranslator/policy_templates_data/policy_templates.json
@@ -697,6 +697,39 @@
         ]
       }
     },
+    "SESBulkTemplatedCrudPolicy": {
+      "Description": "Gives permission to send email, templated email, templated bulk emails and verify identity",
+      "Parameters": {
+        "IdentityName": {
+          "Description": "Identity to give permissions to"
+        }
+      },
+      "Definition": {
+        "Statement": [
+          {
+            "Effect": "Allow",
+            "Action": [
+              "ses:GetIdentityVerificationAttributes",
+              "ses:SendEmail",
+              "ses:SendRawEmail",
+              "ses:SendTemplatedEmail",
+              "ses:SendBulkTemplatedEmail",
+              "ses:VerifyEmailIdentity"
+            ],
+            "Resource": {
+              "Fn::Sub": [
+                "arn:${AWS::Partition}:ses:${AWS::Region}:${AWS::AccountId}:identity/${identityName}",
+                {
+                  "identityName": {
+                    "Ref": "IdentityName"
+                  }
+                }
+              ]
+            }
+          }
+        ]
+      }
+    },
     "SNSCrudPolicy": {
       "Description": "Gives permissions to create, publish and subscribe to SNS topics",
       "Parameters": {
--- a/examples/2016-10-31/policy_templates/all_policy_templates.yaml
+++ b/examples/2016-10-31/policy_templates/all_policy_templates.yaml
@@ -71,6 +71,9 @@ Resources:
         - SESCrudPolicy:
             IdentityName: name
 
+        - SESBulkTemplatedCrudPolicy:
+            IdentityName: name
+
         - SNSCrudPolicy:
             TopicName: name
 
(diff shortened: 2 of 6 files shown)


samtranslator/policy_templates_data/policy_templates.json
```diff
@@ -697,6 +697,39 @@
...
+              "ses:GetIdentityVerificationAttributes",
+              "ses:SendEmail",
+              "ses:SendRawEmail",
+              "ses:SendTemplatedEmail",
+              "ses:SendBulkTemplatedEmail",
+              "ses:VerifyEmailIdentity"
+            ],
+            "Resource": {
+              "Fn::Sub": [
+                "arn:${AWS::Partition}:ses:${AWS::Region}:${AWS::AccountId}:identity/${identityName}",
```
jlhood: Are you sure this works? I read this in the [SES IAM docs](https://docs.aws.amazon.com/IAM/latest/UserGuide/list_amazonses.html):

> Amazon SES has no service-defined resources that can be used as the Resource element of an IAM policy statement.

simalexan: Yes, that is correct, you don't need any resource, but there are 2 things that made me include this `Resource` section:
1. The previous `SESCrudPolicy`, which is just for sending regular SES emails, requires an identity as well. It's the previous policy template just above this one. It is required for this Policy Template. I didn't intend to add it, but when I saw it there, as it was one the first ones added here, I just added it for consistency.
2. This identity represents the verified identity (email or domain) from which you want to be allowed to send emails from. 

I have no problem from removing this section if necessary. The `SESCrudPolicy` is working, so I think this one will too.

simalexan:
Last but not least, this is not urgent or so for me. I've also chosen this template policy numbering in order of the reviewed existing PRs with new Policy Templates.  `KitchenSinkFunctionRolePolicy47`

jlhood:
Sorry, I clearly looked at this too quickly this morning and missed a lot of the details you'd provided. Sorry if I created confusion. Let me clarify:

> I guess that you are classifying which policy templates should be Custom based on the level of security implications possible.

No, we're still encouraging people to contribute SAM policy templates and are not drawing any lines regarding which are more or less secure. It's unfortunately common for users to use `Resource: '*'` out of convenience so we enforce that policy templates should specify a specific resource unless a specific resource is actually unsupported for the given action. We've received positive customer feedback on policy templates as a more convenient way to help developers to do the right thing from a security standpoint so keep them coming!

From a SAR standpoint, we also still encourage using policy templates as they help increase discoverability of your apps since they'll show up to consumers by default in search if they don't require any IAM or resource policy capabilities.

Again, sorry for the confusion and really appreciate you submitting these policy templates. Look forward to more from you in the future!

simalexan:
@keetonian Thank you very much! I was going to get it after finishing work. But am very very grateful! Much appreciated!

simalexan:
@jlhood your colleague finished it up for me, kudos to @keetonian ! :)

simalexan:
Also, didn't know that the tests were actually checking up on the order of policy templates. What is the goal of such a specified order?

PR #730, opened 2018-12-19
Example of a Hello World using Layers

dgomesbr:
**Issue #, if available:** none

**Description of changes:** This adds an example of Hello World using layers in SAM.


jfuss:
Closing since there was no response to previous feedback. If this is still an example worth wild, I would suggest re-submitting or submitting the example to https://github.com/awslabs/aws-sam-cli-app-templates/ which would make the example consumable through `sam init`

PR #737, opened 2018-12-27
Add example on how to use S3 events for lambda functions written in Go using sam

shuza:
**Description of changes**: This PR shows how to use aws sam to develop application that will trigger lambda function written in Go on S3 events like image upload.

### DIFF
--- a/examples/2016-10-31/image_resize_golang/README.md
+++ b/examples/2016-10-31/image_resize_golang/README.md
@@ -1,0 +1,27 @@
+# AWS::Serverless::S3 Event Code Example
+This example shows you how to get events of S3 bucket. When you upload an image in the source bucket it will resize that image and save in the destination bucket.
+###### Note
+Don't forget to chnage S3 bucket name, source bucket prefix and `DestBucket` name in template file.
+
+## Package and deploy template
+#### Build
+Run the following command to build binary of your code
+```
+GOOS=linux go build -o main
+```
+#### Package
+To package your application run the following command
+```
+sam package --template-file template.yaml --output-template-file serverless-output.yaml --s3-bucket <YOUR_S3_BUCKET>
+```
+It will validate the template, zip you applicaiton, upload to your S3 bucket and generates the output template. Replace the `<YOUR_S3_BUCKET>` with your bucket name.
+#### Deploy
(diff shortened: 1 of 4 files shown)


examples/2016-10-31/image_resize_golang/template.yaml
```diff
@@ -0,0 +1,54 @@
...
+      Runtime: go1.x
+      Tracing: Active
+      Policies:
+      - S3ReadPolicy:
+          BucketName: !Sub "${BucketNamePrefix}-source-bucket"
+      - S3CrudPolicy:
+          BucketName: !Sub "${BucketNamePrefix}-dest-bucket"
+      Environment:
+        Variables:
+          DestBucket: "sam-exmaple-dest-bucket"   # BucketNamePrefix-dest-bucket
```
keetonian: Needs to be `DestBucket: !Sub "${BucketNamePrefix}-dest-bucket"` as well in case a value other than the default is used.

bs-shuza: Thanks for the advise. I have updated the template file according to the instruction.

shuza: Thanks for the advice. I have updated the template file according to the instruction.

examples/2016-10-31/image_resize_golang/README.md
```diff
@@ -0,0 +1,27 @@
...
+It will validate the template, zip you applicaiton, upload to your S3 bucket and generates the output template. Replace the `<YOUR_S3_BUCKET>` with your bucket name.
+#### Deploy
+Run the following command replacing  `<BUCKET_PREFIX>` with your desire prefix name to deploye your application
+```
+sam deploy --template-file serverless-output.yaml --stack-name image-resizer --capabilities CAPABILITY_IAM --parameter-overrides BucketNamePrefix=<BUCKET_PREFIX>
+```
+It will create necessary resources and link them according to the template using cloudformation.
+
+## Test
+Uload an image in `JPEG` format to the `SourceBucket` defined in the template and verify the same image with smaller size in `DestBucket`.
```
paoptu023: Typo Uload -> Upload, please fix before merging.

PR #742, opened 2019-01-02
feat: add support for Conditions in Serverless API

keetonian:
*Issue #, if available:*
#142 
*Description of changes:*
Add support for Conditions in Serverless API

### DIFF
--- a/samtranslator/model/api/api_generator.py
+++ b/samtranslator/model/api/api_generator.py
@@ -27,7 +27,8 @@ class ApiGenerator(object):
     def __init__(self, logical_id, cache_cluster_enabled, cache_cluster_size, variables, depends_on,
                  definition_body, definition_uri, name, stage_name, endpoint_configuration=None,
                  method_settings=None, binary_media=None, cors=None, auth=None, access_log_setting=None,
-                 canary_setting=None, tracing_enabled=None):
+                 canary_setting=None, tracing_enabled=None, resource_attributes=None,
+                 passthrough_resource_attributes=None):
         """Constructs an API Generator class that generates API Gateway resources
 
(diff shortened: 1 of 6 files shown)


mlaff:
Could you add a nontrivial condition test case and confirm it works as expected ? For all the cases I see here, both options are `true`

paoptu023:
Just curious, what's the use of these output files?

samtranslator/model/api/api_generator.py
```diff
@@ -27,7 +27,8 @@ class ApiGenerator(object):
     def __init__(self, logical_id, cache_cluster_enabled, cache_cluster_size, variables, depends_on,
                  definition_body, definition_uri, name, stage_name, endpoint_configuration=None,
                  method_settings=None, binary_media=None, cors=None, auth=None, access_log_setting=None,
-                 canary_setting=None, tracing_enabled=None):
+                 canary_setting=None, tracing_enabled=None, resource_attributes=None,
```
paoptu023: https://github.com/awslabs/serverless-application-model/blob/9206b05f049ba7465d11737c42be9ccb23c94239/samtranslator/model/api/api_generator.py#L30-L31

Can you add definitions for the newly added params in function docstring?

keetonian:
1. Conditions: SAM does not evaluate the result of any conditions; this is just to add the framework to output the same conditions on the CFN resources that are given to the SAM (Serverless) resources.
1. Output files: These output files are what the translator output is compared against. This makes it easier to see any changes and catch unwanted regressions or differences.

samtranslator/model/api/api_generator.py
keetonian: Good catch. Will add

PR #753, opened 2019-01-09
fix: handle LayerName intrinsic function resolution

keetonian:
*Description of changes:*
Issue reported on slack as changes were being rolled out. This change fixes this issue.

The Fn::Sub intrinsic function was not correctly being handled. This PR fixes how it and other intrinsics are handled for parameters in Serverless::LayerVersions.

### DIFF
--- a/samtranslator/model/__init__.py
+++ b/samtranslator/model/__init__.py
@@ -434,7 +434,8 @@ class SamResourceMacro(ResourceMacro):
         if not parameter_value:
             return parameter_value
         value = intrinsics_resolver.resolve_parameter_refs(parameter_value)
-        if not isinstance(value, string_types):
+
+        if not isinstance(value, string_types) and not isinstance(value, dict):
             raise InvalidResourceException(self.logical_id,
                                            "Could not resolve parameter for '{}' or parameter is not a String."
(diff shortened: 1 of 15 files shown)


samtranslator/model/__init__.py
```diff
@@ -434,7 +434,8 @@ def _resolve_string_parameter(self, intrinsics_resolver, parameter_value, parame
         if not parameter_value:
             return parameter_value
         value = intrinsics_resolver.resolve_parameter_refs(parameter_value)
-        if not isinstance(value, string_types):
+
+        if not isinstance(value, string_types) and not isinstance(value, dict):
```
keetonian: When resolving intrinsics, if SAM can't or won't resolve all of them, it still returns an intrinsic dictionary. The previous version of this function did not allow for a dictionary to be returned from the intrinsic resolver.

paoptu023: So are string and dictionary the only 2 cases that will be allowed?

keetonian: Yes. Correct values here are either a string or an intrinsic function, which is interpreted as a dictionary.

PR #755, opened 2019-01-10
fix: add conditions to events and add tests

keetonian:
*Issue #, if available:*
#142 
*Description of changes:*
Adds another test for Conditions, as well as makes sure they are added to all Event types (except for API). Will follow up with a fix (or detailed error for the meantime) for Conditions for API event types.

### DIFF
--- a/samtranslator/model/eventsources/cloudwatchlogs.py
+++ b/samtranslator/model/eventsources/cloudwatchlogs.py
@@ -47,5 +47,7 @@ class CloudWatchLogs(PushEventSource):
         subscription_filter.LogGroupName = self.LogGroupName
         subscription_filter.FilterPattern = self.FilterPattern
         subscription_filter.DestinationArn = function.get_runtime_attr("arn")
+        if 'Condition' in function.resource_attributes:
+            subscription_filter.set_resource_attribute('Condition', function.resource_attributes['Condition'])
 
         return subscription_filter
--- a/samtranslator/model/eventsources/pull.py
+++ b/samtranslator/model/eventsources/pull.py
@@ -64,6 +64,8 @@ class PullEventSource(ResourceMacro):
         lambda_eventsourcemapping.StartingPosition = self.StartingPosition
         lambda_eventsourcemapping.BatchSize = self.BatchSize
         lambda_eventsourcemapping.Enabled = self.Enabled
+        if 'Condition' in function.resource_attributes:
+            lambda_eventsourcemapping.set_resource_attribute('Condition', function.resource_attributes['Condition'])
 
         if 'role' in kwargs:
             self._link_policy(kwargs['role'])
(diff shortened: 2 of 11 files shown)


samtranslator/model/eventsources/cloudwatchlogs.py
```diff
@@ -47,5 +47,7 @@ def get_subscription_filter(self, function, permission):
         subscription_filter.LogGroupName = self.LogGroupName
         subscription_filter.FilterPattern = self.FilterPattern
         subscription_filter.DestinationArn = function.get_runtime_attr("arn")
+        if 'Condition' in function.resource_attributes:
```
jlhood: Is there a way we can not have to check this in every event source (and all new event sources moving forward)? Is there something we can do that's more DRY? I understand for certain event sources, we may need to override and do something different, e.g., S3. If there's no obvious way to do this, then I'm ok leaving it this way.

samtranslator/model/eventsources/push.py
```diff
@@ -281,9 +283,9 @@ def _depend_on_lambda_permissions_using_tag(self, bucket, permission):
             properties['Tags'] = tags
         dep_tag = {
             'sam:ConditionalDependsOn:' + permission.logical_id: {
-                'Fn:If': [
+                'Fn::If': [
```
jlhood: Ugh, really don't like seeing that we missed stuff like this the first time around. We need a way for contributors to be able to run some sort of integration tests and launch real CloudFormation stacks so we can catch these kinds of errors much earlier in the process.

Also, why are we adding this tag at all? What purpose does it serve? Looks like it was added as part of the original [Conditional PR](https://github.com/awslabs/serverless-application-model/commit/16f512f98bef15659cf7da67fca8ad734fc11554#diff-471a2c4ace1cdbf649830e4272a9bbc8R281). Was there discussion on the slack channel related to this?

keetonian: It's a conditional depends_on from the original PR- see [this Stack Overflow post](https://stackoverflow.com/questions/34607476/cloudformation-apply-condition-on-dependson).

I totally agree. We need a way to actually deploy these changes to the develop branch; it would make our turnaround time to release to master much faster as well as catch these bugs better.

jlhood: > It's a conditional depends_on from the original PR

Ok I see the comment now. I didn't realize DependsOn doesn't support conditionals. 😞

samtranslator/model/eventsources/cloudwatchlogs.py
keetonian: I think so, I'll look into it when I get a chance.

samtranslator/model/eventsources/pull.py
```diff
@@ -64,6 +64,8 @@ def to_cloudformation(self, **kwargs):
         lambda_eventsourcemapping.StartingPosition = self.StartingPosition
         lambda_eventsourcemapping.BatchSize = self.BatchSize
         lambda_eventsourcemapping.Enabled = self.Enabled
+        if 'Condition' in function.resource_attributes:
```
jfuss: If would be awesome to have a way to do this as a function and not need to worry about it in every resource. As we add more Events, resources, etc, this can be very easily forgotten.

Issue #665, opened 2018-11-15
Bug: Trailing slashes in the paths in the Events type cause errors


sthulb:
**Description:**

Given the following template:
```yaml
  Func: 
    Type: AWS::Serverless::Function
    Properties:
      Handler: bar.handler
      Events:
        HTTP:
          Type: Api
          Properties:
            Path: /bar/
            Method: GET
```

When calling the `/bar/` endpoint in the API Gateway, it will always error with a message of: 
```
Thu Nov 15 15:22:53 UTC 2018 : Execution failed due to configuration error: Invalid permissions on Lambda function
```

I've deployed the template with and WITHOUT the trailing slash on the path, the template without works.

To add further info, using `aws lambda get-policy --function-name ...` to get the policies of both versions, they're identical.

**WITH**:
```json
{
    "Policy": "{\"Version\":\"2012-10-17\",\"Id\":\"default\",\"Statement\":[{\"Sid\":\"test-FuncBHTTPPermissionTest-QA18Q84KEKFR\",\"Effect\":\"Allow\",\"Principal\":{\"Service\":\"apigateway.amazonaws.com\"},\"Action\":\"lambda:invokeFunction\",\"Resource\":\"arn:aws:lambda:eu-west-1:1234567890:function:test-FuncB-1X4GNWC7MCTBL\",\"Condition\":{\"ArnLike\":{\"AWS:SourceArn\":\"arn:aws:execute-api:eu-west-1:1234567890:5v9ewolbwc/*/*/bar\"}}},{\"Sid\":\"test-FuncBHTTPPermissionProd-1I13SM8SHZ1I\",\"Effect\":\"Allow\",\"Principal\":{\"Service\":\"apigateway.amazonaws.com\"},\"Action\":\"lambda:invokeFunction\",\"Resource\":\"arn:aws:lambda:eu-west-1:1234567890:function:test-FuncB-1X4GNWC7MCTBL\",\"Condition\":{\"ArnLike\":{\"AWS:SourceArn\":\"arn:aws:execute-api:eu-west-1:777977323905:5v9ewolbwc/Prod/*/bar\"}}}]}",
    "RevisionId": "018fad31-a77e-4bf2-97cd-2066f7313216"
}
```

**WITHOUT**:
```json
{
[...]

PR #763, opened 2019-01-14
fix: Removes trailing slashes from APIGW permissions

sthulb:
*Issue #, if available:* #665

*Description of changes:*
Adds a regex sub to remove the trailing slash from the path of the Api event source

### DIFF
--- a/tests/model/eventsources/test_api_event_source.py
+++ b/tests/model/eventsources/test_api_event_source.py
@@ -62,18 +62,3 @@ class ApiEventSource(TestCase):
             raise AttributeError("Arn not found")
 
         return arn
-
-# {
-#     'Type': 'AWS::Lambda::Permission',
-#     'Properties': {
-#         'Action': 'lambda:invokeFunction',
-#         'FunctionName': {'Ref': 'func'},
-#         'Principal': 'apigateway.amazonaws.com',
-#         'SourceArn': {
-#             'Fn::Sub': [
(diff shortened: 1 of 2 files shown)


tests/model/eventsources/test_api_event_source.py
```diff
@@ -0,0 +1,64 @@
...
+
+        try:
+            arn = self._extract_path_from_arn("{}PermissionTest".format(self.logical_id), perm)
+        except AttributeError:
+            self.fail("Permission class isn't valid")
+
+        self.assertEqual(arn, "arn:aws:execute-api:${AWS::Region}:${AWS::AccountId}:${__ApiId__}/${__Stage__}/GET/foo")
+
+    @patch("boto3.session.Session.region_name", "eu-west-2")
+    def test_get_permission_with_trailing_slash(self):
```
jlhood: Please add a test where the path is just '/' to ensure we don't accidentally remove the slash in that case.

jlhood:
Oh also, please make your changes against the develop branch, not master and resolve conflicts.

brettstack:
Is there any chance that this would break existing templates? At the very least this will likely force a new deployment in cases where a user hasn't actually modified their API.

sthulb:
@brettstack no. Anyone who fell into this trap and added a / at the end of their path has a broken API.

tests/model/eventsources/test_api_event_source.py
sthulb: Done :)

PR #769, opened 2019-01-17
s3CrudPolicy allows using canned ACL for s3.PutObject

gojko:
*Description of changes:* the old S3CrudPolicy would not allow s3.PutObject operations if they supplied a canned ACL (such as `public-read`), because S3 also requires a PutObjectAcl permission for that. This pull request extends the S3CrudPolicy to allow canned ACLs in API requests to write objects.


gojko:
thanks!

PR #799, opened 2019-02-08
Add Resource Groups readonly policy to SAM

fh:
*Description of changes:*
Adding read-only permissions for AWS Resource Groups to allow examples including Resource Groups.

### DIFF
--- a/samtranslator/policy_templates_data/policy_templates.json
+++ b/samtranslator/policy_templates_data/policy_templates.json
@@ -1560,6 +1560,24 @@
           }
         ]
       }
+    },
+    "ResourceGroupsReadOnlyAccessPolicy": {
+      "Description": "Gives readonly permissions to AWS Resource Groups and list group contents",
+      "Parameters": {},
+      "Definition": {
+        "Statement": [{
+          "Effect": "Allow",
+          "Action": [
+            "resource-groups:Get*",
+            "resource-groups:List*",
+            "resource-groups:Search*",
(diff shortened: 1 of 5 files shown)


samtranslator/policy_templates_data/policy_templates.json
```diff
@@ -1560,6 +1560,24 @@
...
+      "Definition": {
+        "Statement": [{
+          "Effect": "Allow",
+          "Action": [
+            "resource-groups:Get*",
+            "resource-groups:List*",
+            "resource-groups:Search*",
+            "tag:Get*",
+            "cloudformation:DescribeStacks",
+            "cloudformation:ListStackResources"
```
brettstack: Could you update this existing Policy Template https://github.com/awslabs/serverless-application-model/blob/d97341b61a0c18387753fc1b42038d30e391f02d/samtranslator/policy_templates_data/policy_templates.json#L397 with the ListStackResources Action? Then update your new Policy Template:

1. Remove the cloudformation: actions
1. Be explicit with the other Actions (remove '*')
1. If applicable, be explicit on the Resource instead of '*' (if '*' is the only suitable Resource value, please link to the documentation). It looks like it can be scoped down though https://docs.aws.amazon.com/IAM/latest/UserGuide/list_awsresourcegroups.html#awsresourcegroups-resources-for-iam-policies. You may need to split the tag action into a separate policy template; could you share documentation on this action?

PR #804, opened 2019-02-09
feat(Conditions): add support for Conditions on API event resources

keetonian:
*Issue #, if available:*
https://github.com/awslabs/serverless-application-model/issues/758

*Description of changes:*
Add ability to honor conditions placed on functions with API events.

*Testing:*
Templates were generated using SAM locally and deployed via CFN to test parts of this implementation. I did this with the normal and CORS templates; I have not tested the Auth example (should work the same).

*Note to reviewers:*
This PR is very long, mostly because I am testing several things related to APIs. The actual code difference is about ~170 lines, all the rest are for tests. The output files for these templates with APIs are very large and hard to read; I have run the translator on similar templates and verified that they deploy to CFN.

*Callout:*
I have not tested to see if CFN will allow SAM to create extra Conditions. I assume that this is something that is allowed by CFN.


tests/translator/output/api_with_cors_and_conditions_no_definitionbody.json
```diff
@@ -0,0 +1,381 @@
...
+                  "type": "mock",
+                  "requestTemplates": {
+                    "application/json": "{\n  \"statusCode\" : 200\n}\n"
+                  },
+                  "responses": {
+                    "default": {
+                      "statusCode": "200",
+                      "responseParameters": {
+                        "method.response.header.Access-Control-Allow-Origin": "'www.example.com'",
+                        "method.response.header.Access-Control-Allow-Methods": "'DELETE,GET,OPTIONS,POST'"
```
keetonian: DELETE method doesn't have a condition on it, while the other 2 do. I verified that this can still be deployed to CFN if the other methods (POST, GET) are not included in the path (condition = false) but included here in CORS. Is it fine if only the DELETE method exists, but CORS allows GET and POST as well (which are removed by CFN by resolving conditions)?

tests/translator/output/aws-cn/api_with_auth_and_conditions_all_max.json
```diff
@@ -0,0 +1,695 @@
...
+                  }
+                },
+                "responses": {},
+                "security": [
+                  {
+                    "MyLambdaTokenAuth": []
+                  }
+                ]
+              },
+              "patch": {
```
keetonian: Here's an example of a condition where an authorizer is also applied to the method.

Note: Conditions have not been added to authorizers nor have authorizers been modified in any way.

Code lifted out of the file for convenience:
```json
              "patch": {
                "Fn::If": [
                  "PathCondition",
                  {
                    "x-amazon-apigateway-integration": {
                      "type": "aws_proxy",
                      "httpMethod": "POST",
                      "uri": {
                        "Fn::If": [
                          "PathCondition",
                          {
                            "Fn::Sub": "arn:aws-cn:apigateway:${AWS::Region}:lambda:path/2015-03-31/functions/${MyFunctionWithConditional.Arn}/invocations"
                          },
                          {
                            "Ref": "AWS::NoValue"
                          }
                        ]
                      }
                    },
                    "responses": {},
                    "security": [
                      {
                        "MyLambdaTokenAuthNoneFunctionInvokeRole": []
                      }
                    ]
                  },
                  {
                    "Ref": "AWS::NoValue"
                  }
                ]
              },
```

tests/translator/output/aws-us-gov/implicit_api_with_many_conditions.json
```diff
@@ -0,0 +1,1507 @@
...
+        },
+        {
+          "Condition": "Cond8"
+        },
+        {
+          "Condition": "Cond9"
+        }
+      ]
+    },
+    "SamImplicitApiCondition": {
```
keetonian: Here's an example of the SamImplicitApiCondition, which is added when there are 2+ conditions on an implicit API resource. See https://github.com/awslabs/serverless-application-model/issues/758 under "Combining Conditions" for the explanation on why this is necessary.

samtranslator/plugins/api/implicit_api_plugin.py
```diff
@@ -39,6 +39,8 @@ def __init__(self):
 
         self.implicit_api_logical_id = GeneratedLogicalId.implicit_api()
         self.existing_implicit_api_resource = None
+        self.conditions = set()
+        self.implicit_api_condition = 'SamImplicitApiCondition'
```
keetonian: This is the name of the condition that SAM adds to the template. Is this fine, or should the name be something else?

samtranslator/swagger/swagger.py
```diff
@@ -92,12 +116,15 @@ def add_lambda_integration(self, path, method, integration_uri):
...
         method = self._normalize_method_name(method)
-
         if self.has_integration(path, method):
             raise ValueError("Lambda integration already exists on Path={}, Method={}".format(path, method))
 
         self.add_path(path, method)
 
+        # Wrap the integration_uri in a Condition if one exists on that function:
+        if condition:
+            integration_uri = make_conditional(condition, integration_uri)
```
keetonian: This is necessary, or else CFN complains that it can't resolve a resource reference.

samtranslator/plugins/api/implicit_api_plugin.py
```diff
@@ -195,6 +204,32 @@ def _add_api_to_swagger(self, event_id, event_properties, template):
...
+                implicit_api_resource['Condition'] = self.implicit_api_condition
+                self._add_condition_to_template(template_dict)
+
+    def _add_condition_to_template(self, template_dict):
+        """
+        Adds necessary conditions to the template for the implicit api
+        :param dict template_dict: SAM template dictionary
+        """
+        if not template_dict.get('Conditions'):
+            template_dict['Conditions'] = {}
```
keetonian: By default, if Conditions exist on these resources, they should also exist in the Conditions section of the template. Should I remove this check?

samtranslator/plugins/api/implicit_api_plugin.py
```diff
@@ -195,6 +204,32 @@ def _add_api_to_swagger(self, event_id, event_properties, template):
...
+        template_conditions = template_dict['Conditions']
+        implicit_api_conditions = make_combined_condition(sorted(list(self.conditions)), self.implicit_api_condition)
```
keetonian: Sorting the conditions assures:
1. The conditions are applied the same each time SAM is run (same conditions get wrapped up in the same SAM condition)
1. Tests can run and pass (need the same conditions in each SAM condition as in the test output files)

samtranslator/swagger/swagger.py
keetonian: Sample output:

The contents of the method (in this case, **post** are wrapped in a `Fn::If` conditional. CFN doesn't recognize that the `uri` field is wrapped in this conditional (creates an "unresolved resource reference" error), so another conditional is needed around that particular section. This is why we need two conditionals for one template block.

```json
"paths": {
            "/": {
              "post": {
                "Fn::If": [
                  "MyCondition", 
                  {
                    "x-amazon-apigateway-integration": {
                      "httpMethod": "POST", 
                      "type": "aws_proxy", 
                      "uri": {
                        "Fn::If": [
                          "MyCondition", 
                          {
                            "Fn::Sub": "arn:aws:apigateway:${AWS::Region}:lambda:path/2015-03-31/functions/${ImplicitApiFunction.Arn}/invocations"
                          }, 
                          {
                            "Ref": "AWS::NoValue"
                          }
                        ]
                      }
                    }, 
                    "responses": {}
                  }, 
                  {
                    "Ref": "AWS::NoValue"
                  }
                ]
              }, 
```

samtranslator/plugins/api/implicit_api_plugin.py
keetonian: Yes, it is called "ServerlessRestApi". Will change.

samtranslator/swagger/swagger.py
```diff
@@ -50,6 +52,28 @@ def has_path(self, path, method=None):
...
+        """
+        Returns the swagger contents of the given method. This checks to see if a conditional block
+        has been used inside of the method, and, if so, returns the method contents that are
+        inside of the conditional.
+
+        :param dict method: method dictionary
+        :return: swagger components of the method
+        """
+        if self._CONDITIONAL_IF in method:
+            return method[self._CONDITIONAL_IF][1]
```
brettstack: This assumes the method is in this position, but this is not always true. For example, I could have a condition that when evaluated to true returns NoValue and when false returns the method.

samtranslator/model/intrinsics.py
```diff
@@ -26,6 +30,75 @@ def make_conditional(condition, data):
...
+    Makes a combined condition using Fn::Or. Since Fn::Or only accepts up to 10 conditions,
+    this method optionally creates multiple conditions. These conditions are named based on
+    the condition_name parameter that is passed into the method.
+
+    :param list conditions_list: list of conditions
+    :param string condition_name: base name desired for new condition
+    :return: dictionary of condition_name: condition_value
+    """
+    if len(conditions_list) < 2:
+        # Can't make a condition if <2 conditions provided.
```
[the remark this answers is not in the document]
keetonian: Yes, that single condition is then added to the implicit API. This is for combining multiple conditions, which requires >1 given to this function.

samtranslator/swagger/swagger.py
keetonian: Good point. SAM doesn't set something like that up, but someone writing Swagger could.

tests/translator/input/api_with_cors_and_conditions_no_definitionbody.yaml
```diff
@@ -0,0 +1,55 @@
...
+      Runtime: nodejs8.10
+      Events:
+        DeleteHtml:
+          Type: Api
+          Properties:
+            RestApiId: !Ref ExplicitApi
+            Path: /
+            Method: delete
+
+  ImplicitApiFunction:
```
brettstack: This says "ImplicitApiFunction" but it's referencing an ExplicitApi

samtranslator/model/intrinsics.py
```diff
@@ -26,6 +30,75 @@ def make_conditional(condition, data):
...
+    max_conditions = 10
+
+    conditions = {}
+    conditions_length = len(conditions_list)
+    # Get number of conditions needed, then minus one to use them as 0-based indices
+    num_conditions = calculate_number_of_conditions(conditions_length, max_conditions) - 1
+
+    while len(conditions_list) > 1:
+        new_condition_name = condition_name
+        # If more than 1 new condition is needed, add a # to the end of the name
```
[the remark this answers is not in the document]
keetonian: Ok. Will fix.

samtranslator/model/intrinsics.py
```diff
@@ -26,6 +30,75 @@ def make_conditional(condition, data):
...
+
+    :param int conditions_length: total # of conditions to handle
+    :param int max_conditions: maximum number of conditions that can be put in an Fn::Or statement
+    :return: the number (int) of necessary additional conditions.
+    """
+    num_conditions = 1 + (conditions_length - 2) // (max_conditions - 1)
+    return num_conditions
+
+
+def make_combined_condition(conditions_list, condition_name):
```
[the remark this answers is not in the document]
keetonian: Sure. It's outlined in the linked issue, I'll copy some of that into the code as an explanation.

samtranslator/plugins/api/implicit_api_plugin.py
```diff
@@ -195,6 +204,32 @@ def _add_api_to_swagger(self, event_id, event_properties, template):
         resource.properties["DefinitionBody"] = editor.swagger
         template.set(api_id, resource)
 
+    def _maybe_add_condition_to_implicit_api(self, template_dict):
+        """
+        Decides whether to add a condition to the implicit api resource.
+        :param dict template_dict: SAM template dictionary
+        """
+        if None not in self.conditions and len(self.conditions) > 0:
```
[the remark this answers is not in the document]
keetonian: That's for functions that don't have conditions on them. None still gets added to this conditions list/set.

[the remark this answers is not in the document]
keetonian: If there is a function that references the implicit API that doesn't have a condition on it, then SAM doesn't need to create an aggregate condition for that API (it will always need to create the API).

samtranslator/plugins/api/implicit_api_plugin.py
```diff
@@ -195,6 +204,32 @@ def _add_api_to_swagger(self, event_id, event_properties, template):
...
+            implicit_api_resource = template_dict.get('Resources').get(self.implicit_api_logical_id)
+            if len(self.conditions) == 1:
+                condition = self.conditions.pop()
```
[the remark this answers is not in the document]
keetonian: Because it's a python set and that's the only way to get the condition out of it (we could convert to a list and get it normally that way)

samtranslator/swagger/swagger.py
keetonian: This is a very good callout. Will have to think about this.

tests/translator/input/api_with_cors_and_conditions_no_definitionbody.yaml
keetonian: Copy error 🤦‍♂️

samtranslator/model/intrinsics.py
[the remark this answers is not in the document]
keetonian: Commented more about this in the implicit api plugin in the docstring for `_add_condition_to_template`.

samtranslator/model/intrinsics.py
```diff
@@ -26,6 +30,75 @@ def make_conditional(condition, data):
...
+        return None
+
+    # Total number of conditions allows in an Fn::Or statement. See docs:
+    # https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-conditions.html#intrinsic-function-reference-conditions-or
+    max_conditions = 10
+
+    conditions = {}
+    conditions_length = len(conditions_list)
+    # Get number of conditions needed, then minus one to use them as 0-based indices
+    num_conditions = calculate_number_of_conditions(conditions_length, max_conditions) - 1
```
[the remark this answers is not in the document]
keetonian: Done

samtranslator/swagger/swagger.py
keetonian: Ok. Couple of thoughts here.

This has never worked up to this point, which means that no existing SAM templates will have a conditional inside of a method configuration. So, whatever we do here won't *break* any existing customers, but could potentially affect *future* uses of SAM.

SAM uses this code, I think, only for two things:
1. Creating Swagger for the user (generating APIGW paths and methods)
1. Adding an authorizer to a method.

This means that the only time (at least as SAM is now) a customer would have SAM read their swagger is when they are using custom authorizers. We should still assume that other use cases could arise.

Possible solutions:

1. Leave as is:
    1. Document that SAM only modifies the first element of a Condition that is nested under a method.
    1. This means that, with authorizers, SAM would only add an authorizer to the first item in the conditional and ignore the second entirely. 
    1. SAM will also throw an error if the first item is invalid.
1. Check the first item to see if it is a valid path, and, if not, return the second item.
    1. This is an either/or situation. It still wouldn't handle the case where both paths would need to be modified.
1. Just do whatever we're doing to both items in the list, and throw an error only if both of them are invalid paths.
    1. Might be the safest approach.  
    1. Would a customer ever want SAM to only modify the first one and not touch the second? 
    1. Or should we return both parts of this conditional list and make sure that it is a valid path before modifying it (for authorizers or other similar things we do in the future)?

keetonian: The third option might be the best option here, should we go ahead and modify all valid method configurations that are in the conditional list? The change is pretty small I think

brettstack: 3rd option sounds the best

tests/translator/output/api_with_cors_and_conditions_no_definitionbody.json
brettstack: This is fine.

tests/translator/output/aws-cn/api_with_auth_and_conditions_all_max.json
brettstack: So you're saying that even though the "MyLambdaTokenAuthNoneFunctionInvokeRole" isn't actually used, it will still get created (added to securityDefinitions later)? I think this is fine.

keetonian: That role is created elsewhere and is included in the auth section. I kept authorizers the same, since they exist separate from these other resources.

keetonian:
Ok. Several things need to happen that come out of a discussion about paths.

**Problem Description:**
In the current implementation, APIGW methods are added based on the conditions on their underlying functions. However, this can leave orphaned paths with no attached methods to them, which doesn't seem like what we want to do.

Example: 
image

**Solution:**
We want to add a condition to the paths much like we add a condition to the implicit API, which condition is either:
* the condition on the method(s) if only one condition is present
* an aggregate of all conditions on the methods inside of that condition
* no condition if any method doesn't have a condition attached to it.

**Example JSON:**

The "<GeneratedPathConditionName>" should be replaced with the name given to the condition that is created that contains both "Cond" and "MyCondition" conditions that are found on the methods inside of the "/sub" path.

```json
"paths": {
  "Fn::If": [
    "<GeneratedPathConditionName>",
    {
      "/sub": {
        "post": {
          "Fn::If": [
            "Cond",
            {
              "x-amazon-apigateway-integration": {
                "httpMethod": "POST",
                "type": "aws_proxy",
                "uri": {
                  "Fn::If": [
                    "Cond",
                    {
                      "Fn::Sub": "arn:aws:apigateway:${AWS::Region}:lambda:path/2015-03-31/functions/${helloworld.Arn}/invocations"
                    },
                    {
                      "Ref": "AWS::NoValue"
                    }
                  ]
                }
              },
              "responses": {}
            },
            {
              "Ref": "AWS::NoValue"
            }
          ]
        },
        "get": {
          "Fn::If": [
            "MyCondition",
            {
              "x-amazon-apigateway-integration": {
                "httpMethod": "POST",
                "type": "aws_proxy",
                "uri": {
                  "Fn::If": [
                    "MyCondition",
                    {
                      "Fn::Sub": "arn:aws:apigateway:${AWS::Region}:lambda:path/2015-03-31/functions/${hello.Arn}/invocations"
                    },
                    {
                      "Ref": "AWS::NoValue"
                    }
                  ]
                }
              },
              "responses": {}
            },
            {
              "Ref": "AWS::NoValue"
            }
          ]
        }
      }
    },
    {
      "Ref": "AWS::NoValue"
    },
  ]
}
```

**Necessary Updates:**

Swagger tests: 
* https://github.com/keetonian/serverless-application-model/blob/release/v1.10.0-api-event/tests/swagger/test_swagger.py#L95
* https://github.com/keetonian/serverless-application-model/blob/release/v1.10.0-api-event/tests/swagger/test_swagger.py#L260

All new API integration tests (included inside of this PR)

samtranslator/swagger/swagger.py
```diff
@@ -347,62 +409,69 @@ def add_auth_to_method(self, path, method_name, auth, api):
...
-                del existing_non_authorizer_security[none_idx]
-
-            # Existing Authorizer found (defined at Function/Path/Method); use that instead of default
-            elif existing_authorizer_security:
-                authorizer_security = existing_authorizer_security
-
-            # No existing Authorizer found; use default
+
+        # It is possible that the method could have two definitions in a Fn::If block.
+        for method_definition in self.get_method_contents(self.get_path(path)[normalized_method_name]):
```
keetonian: @brettstack here's a section of iterating over both parts of an IF conditional.

samtranslator/plugins/api/implicit_api_plugin.py
```diff
@@ -204,37 +208,103 @@ def _add_api_to_swagger(self, event_id, event_properties, template):
...
         template.set(api_id, resource)
 
+    def _get_api_id(self, event_properties):
+        """
+        Get API logical id from API event properties.
+
+        Handles case where API id is not specified or is a reference to a logical id.
+        """
+        api_id = event_properties.get("RestApiId")
+        if isinstance(api_id, dict) and "Ref" in api_id:
```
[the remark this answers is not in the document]
keetonian: No, it shouldn't be (and shouldn't ever be for this case)

samtranslator/plugins/api/implicit_api_plugin.py
```diff
@@ -130,8 +133,14 @@ def _process_api_events(self, function, api_events, template, condition=None):
             if not event_properties:
                 continue
 
-            if self._add_implicit_api_id_if_necessary(event_properties):
-                self.conditions.add(condition)
+            self._add_implicit_api_id_if_necessary(event_properties)
+
+            api_id = self._get_api_id(event_properties)
+            path = event_properties["Path"]
+            method = event_properties["Method"]
```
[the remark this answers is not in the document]
keetonian: No guaranteed validation at this point, this is one of the earliest pieces of code to run in the translator.

samtranslator/plugins/api/implicit_api_plugin.py
```diff
@@ -204,37 +208,103 @@ def _add_api_to_swagger(self, event_id, event_properties, template):
...
+                continue
+
+            swagger = api.properties.get("DefinitionBody")
+            editor = SwaggerEditor(swagger)
+
+            for path in editor.iter_on_path():
+                all_method_conditions = set(
+                    [condition for method, condition in self.api_conditions[api_id][path].items()]
+                )
+                if len(all_method_conditions) > 0 and None not in all_method_conditions:
```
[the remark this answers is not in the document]
keetonian: This is similar to what I did on the APIGW resources. If None is in these conditions, that means that there is a referencing function *without* a condition and there will *always* be a method under this path, which means the path doesn't need a condition at all.

samtranslator/plugins/api/implicit_api_plugin.py
```diff
@@ -204,37 +208,103 @@ def _add_api_to_swagger(self, event_id, event_properties, template):
...
-        those conditions in order to conditionally create the Implicit Api. See RFC:
-        https://github.com/awslabs/serverless-application-model/issues/758
+        Add top-level template condition that combines the given list of conditions.
 
         :param dict template_dict: SAM template dictionary
+        :param string condition_name: Name of top-level template condition
+        :param list conditions_to_combine: List of conditions that should be combined (via OR operator) to form
+                                           top-level condition.
+        """
+        template_conditions = template_dict.setdefault('Conditions', {})
```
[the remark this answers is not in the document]
keetonian: We want to make sure it exists since we'll be mutating it below (adding conditions).

tests/translator/output/implicit_api_with_auth_and_conditions_max.json
```diff
@@ -1,232 +1,251 @@
...
     "FunctionCondition2": {
       "Fn::Equals": [
-        true, 
+        true,
         false
       ]
-    }, 
-    "ServerlessRestApiCondition": {
+    },
+    "SamApiPathConditionServerlessRestApiSLASHusers": {
```
keetonian: The `SLASH` is a little distracting. Any ideas for something better?

samtranslator/plugins/api/implicit_api_plugin.py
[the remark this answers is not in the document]
brettstack: api_id["Ref"] is insufficient then, right?

samtranslator/plugins/api/implicit_api_plugin.py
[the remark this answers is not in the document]
brettstack: Okay that makes sense. I think a comment to this effect in both places would be helpful

samtranslator/plugins/api/implicit_api_plugin.py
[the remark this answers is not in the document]
jlhood: The `_add_api_to_swagger` method is already assuming Path and Method exist so this isn't inherently less safe. It is a bug though that we should fix in a separate PR.

samtranslator/plugins/api/implicit_api_plugin.py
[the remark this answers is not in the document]
jlhood: Any `!Ref` has been converted to a dict with `Ref` as the key at this point. This was a refactor of existing code so I could reuse this functionality, not changing assumptions.

samtranslator/plugins/api/implicit_api_plugin.py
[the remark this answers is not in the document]
jlhood: TODO: will add a precondition check that `conditions_to_combine` contains at least 2 entries.

samtranslator/plugins/api/implicit_api_plugin.py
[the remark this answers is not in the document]
jlhood: TODO: Will add better named boolean variables to make this more readable.

samtranslator/plugins/api/implicit_api_plugin.py
```diff
@@ -204,37 +208,103 @@ def _add_api_to_swagger(self, event_id, event_properties, template):
...
-            template_dict['Conditions'] = {}
-        template_conditions = template_dict['Conditions']
-        implicit_api_conditions = make_combined_condition(sorted(list(self.conditions)), self.implicit_api_condition)
-        for key, value in implicit_api_conditions.items():
-            template_conditions[key] = value
+        # only valid characters for CloudFormation logical id are [A-Za-z0-9], but swagger paths can contain
+        # slashes and curly braces for templated params, e.g., /foo/{customerId}. So we'll replace
+        # non-alphanumeric characters.
+        path_logical_id = path.replace('/', 'SLASH').replace('{', 'OB').replace('}', 'CB')
+        return 'SamApiPathCondition{}{}'.format(api_id, path_logical_id)
```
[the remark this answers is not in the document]
jlhood: TODO: will change

tests/translator/output/implicit_api_with_auth_and_conditions_max.json
jlhood: Discussed as a group. Can't think of anything substantially better so will leave it.

samtranslator/plugins/api/implicit_api_plugin.py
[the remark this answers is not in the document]
jlhood: Addressed in latest commit

samtranslator/plugins/api/implicit_api_plugin.py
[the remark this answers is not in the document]
jlhood: Addressed in latest commit

samtranslator/plugins/api/implicit_api_plugin.py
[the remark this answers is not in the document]
jlhood: Addressed in latest commit

PR #826, opened 2019-02-26
fix(LayerVersion): add UpdateReplacePolicy to LayerVersion resource

keetonian:
*Issue #, if available:*
https://github.com/awslabs/serverless-application-model/issues/703

*Description of changes:*
Add the `UpdateReplacePolicy` to LayerVersion resources so that old version aren't deleted.

Deployed template with a layer to CFN, then updated the template to verify the old version doesn't get deleted.

*Checklist:*

- [x] Write/update tests
- [x] `make pr` passes
- [x] Update documentation

### DIFF
--- a/tests/translator/test_translator.py
+++ b/tests/translator/test_translator.py
@@ -42,6 +42,8 @@ LINT_IGNORE_WARNINGS = [
     'E3001', # Check for resource availability in a region.
     'W7001', # Check if mappings are used. Serverless::Application uses mappings, the output CFN doesn't use them anymore.
     'W1020', # Sub isn't needed if it doesn't have a variable defined. SAM leaves `!Sub` in even if it tries to resolve variables.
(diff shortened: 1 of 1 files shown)


keetonian:
With the new `UpdateReplacePolicy`, changing the LayerVersion logical id is no longer necessary. However, if we were to remove this, it could trigger a new deployment of all current SAM layer version resources.

brettstack:
This is changing existing functionality/tests. Any concerns with breaking existing users using this?

Will this force an update because we changed UpdateReplacePolicy?

keetonian:
In testing, adding the `UpdateReplacePolicy` to the layer triggered a stack update but did not trigger an update on the LayerVersion resource.

keetonian:
Update documentation: should use Lambda::LayerVersion resource and add the resource properties to it.

PR #828, opened 2019-02-28
Bugfix: CORS Security

### LINKED ISSUE #717: Cannot set DefaultAuthorizer and have CORS enabled
**Description:**

I want to deploy an API Gateway that both has a custom lambda authorizer and uses CORS.  However, the configuration always ends up in a non-working state.  Because of #650 , the only authorizer you can specify is the DefaultAuthorizer (if you are referencing a swagger at all).  This means the authorizer is applied to every single endpoint in the API - including OPTIONS endpoints generated by setting CORS.  However, CORS methods are not meant to be authorized - particularly if you are using Headers authorization, since the `Authorization` header is stripped out from CORS pre-flight checks.  Which leaves us with the following options:

1. Do not authorize our API. (Obviously cannot do this)
2. Do not enable CORS (Also cannot do this, as we must allow or web application to talk with our API)
3. Manually, in the AWS console, remove the IdentitySource for the authorizer in the API Gateway after every single automated deployment (not sustainable or practical)
4. Manually, in the AWS Console, remove the authorizer from every single OPTIONS endpoint (also not sustainable or practical)

I understand why IdentitySource is required in the serverless template and in aws cli (which begs the question why it can be removed in AWS Console at all), but because #650 is not fixed, we cannot manually associate the authorizer with our explicit endpoints - thereby leaving the generated CORS endpoints unauthorized.  Therefore, one cannot have custom authorizers and CORS enabled on an API Gateway

**Steps to reproduce the issue:**
1. Create a serverless template that enables CORS, creates a custom authorizer (with Header `IdentitySources`), sets that authorizer as the `DefaultAuthorizer`, and uses an inline `DefinitionBody`

Example template:

```yaml
Description: Users lambda functions

Parameters:
  EnvironmentName:
    Type: String
    Description: 'Enter: dev, stage, or prod'
  LambdaRole:
    Type: String
    Default: 'arn:aws:iam::[PII]:role/api-lambda-role'

Globals:
  Function:
    CodeUri: ''
    Runtime: dotnetcore2.1
    MemorySize: 256
    Timeout: 30	

Resources:
  ApiGatewayApi:
    Type: AWS::Serverless::Api
    Properties:
      StageName: !Ref EnvironmentName
      Cors:
        AllowOrigin: "'*'"
        AllowHeaders: "'*'"
      Auth:
        DefaultAuthorizer: ApiAuthorizer
        Authorizers:
          ApiAuthorizer:
            FunctionPayloadType: REQUEST
            FunctionArn: !GetAtt Authorize.Arn
            Identity:
              Headers:
                - Authorization
              ReauthorizeEvery: 0
      DefinitionBody:
        swagger: 2.0
        info:
          title: 
            Fn::Sub: "api-${EnvironmentName}"
          version: "1.0"
        servers:
        - url: "https://api-${EnvironmentName}.example.com/{basePath}"
          variables:
            basePath:
              default:
                Fn::Sub: "/${EnvironmentName}/v1"
        
        paths:
          /user/{id}:
            get:
              x-amazon-apigateway-integration:
                uri:
                  Fn::Sub: "arn:aws:apigateway:${AWS::Region}:lambda:path/2015-03-31/functions/${GetUser.Arn}/invocations"
                passthroughBehavior: "when_no_match"
                httpMethod: "POST"
                type: "aws_proxy"
  
  GatewayResponseDefault4XX:
    Type: 'AWS::ApiGateway::GatewayResponse'
    Properties:
      ResponseParameters:
        gatewayresponse.header.Access-Control-Allow-Origin: "'*'"
        gatewayresponse.header.Access-Control-Allow-Headers: "'*'"
      ResponseType: DEFAULT_4XX
      RestApiId: !Ref ApiGatewayApi

  Authorize:
    Properties:
      Handler: GlobalEditAPI.Users::GlobalEditAPI.Users.Functions.AuthorizerFunctions::Authorize
      Role: !Ref LambdaRole

  GetUser:
    Properties:
      Handler: GlobalEditAPI.Users::GlobalEditAPI.Users.Functions.UserFunctions::GetUserAsync
      Role: !Ref LambdaRole
      Events:
        ProxyApi:
          Type: Api
          Properties:
            RestApiId: !Ref ApiGatewayApi
            Path: /user/{id}
            Method: GET

Outputs:
  ApiUrl:
    Description: URL of Users API endpoint
    Value: !Join
      - ''
      - - https://
        - !Ref ApiGatewayApi
        - '.execute-api.'
        - !Ref 'AWS::Region'
        - '.amazonaws.com/'
        - !Ref EnvironmentName
```

2. Deploy the template to AWS

(We are using dotnet, but underneath it is basic AWS CLI)
```
dotnet lambda deploy-serverless --s3-bucket "api-example" --stack-name "api-example" --template-parameters EnvironmentName="example"
```

**Observed result:**

The authorizer has been associated with the `OPTIONS` endpoints, which will cause them to fail.

**Expected result:**

I should be able to either:

* Not have authorizers associated with "generated" endpoints from CORS, even if the authorizer is set a Default
* Have #650 fixed so I can manually associate authorizers with endpoints, thereby forcing the authorizer to not associate with the generated CORS endpoints
* Have Authorizers in API Gateway automatically give OPTIONS requests a pass and do not try to authorize them

jbutz:
*Issue #:* #717

*Description of changes:*

I have updated the logic that creates the Swagger OPTIONS path operations objects for CORS and added a security requirements object with an object indicating no security. This resolves #717 and is similar to how a CloudFormation Macro was used to fix the issue caused by having both a default authorizer object and CORS object configured.

I have used the CorsFixer, described [here](https://github.com/awslabs/serverless-application-model/issues/717#issuecomment-454835323), to resolve this issue separately and things look to be matching what I got from it fairly closely. Even with the tests failing locally with mocking errors the output is matching the expected out. I have not been able to run the transform from the CLI, I get an error when I run `bin/sam-translate.py` locally. The unit tests pass for Python 3 on Travis but not for Python 2.7. The errors for 2.7 also occur on my local machine and I have no idea what the issue it and could use help in resolving it. I'm not a python developer and I welcome any advice or help.

*Checklist:*

- [X] Write/update tests
- [ ] `make pr` passes
- [X] Update documentation

### DIFF
--- a/samtranslator/model/api/api_generator.py
+++ b/samtranslator/model/api/api_generator.py
@@ -18,8 +18,9 @@ CorsProperties = namedtuple("_CorsProperties", ["AllowMethods", "AllowHeaders", 
 # Default the Cors Properties to '*' wildcard and False AllowCredentials. Other properties are actually Optional
 CorsProperties.__new__.__defaults__ = (None, None, _CORS_WILDCARD, None, False)
 
-AuthProperties = namedtuple("_AuthProperties", ["Authorizers", "DefaultAuthorizer"])
-AuthProperties.__new__.__defaults__ = (None, None)
+AuthProperties = namedtuple("_AuthProperties", ["Authorizers", "DefaultAuthorizer",
+                                                "AddDefaultAuthorizerToCorsPreflight"])
+AuthProperties.__new__.__defaults__ = (None, None, True)
 
 
 class ApiGenerator(object):
@@ -264,7 +265,8 @@ class ApiGenerator(object):
 
         if authorizers:
             swagger_editor.add_authorizers(authorizers)
-            self._set_default_authorizer(swagger_editor, authorizers, auth_properties.DefaultAuthorizer)
+            self._set_default_authorizer(swagger_editor, authorizers, auth_properties.DefaultAuthorizer,
+                 auth_properties.AddDefaultAuthorizerToCorsPreflight)
 
         # Assign the Swagger back to template
         self.definition_body = swagger_editor.swagger
@@ -335,7 +337,8 @@ class ApiGenerator(object):
 
         return permissions
 
-    def _set_default_authorizer(self, swagger_editor, authorizers, default_authorizer):
+    def _set_default_authorizer(self, swagger_editor, authorizers, default_authorizer,
+                                add_default_authorizer_to_cors_preflight):
         if not default_authorizer:
             return
 
@@ -344,7 +347,8 @@ class ApiGenerator(object):
                                            default_authorizer + "' was not defined in 'Authorizers'")
 
         for path in swagger_editor.iter_on_path():
-            swagger_editor.set_path_default_authorizer(path, default_authorizer, authorizers=authorizers)
+            swagger_editor.set_path_default_authorizer(path, default_authorizer, authorizers=authorizers,
+                           addDefaultAuthorizerToCorsPreflight=add_default_authorizer_to_cors_preflight)
 
     def _set_endpoint_configuration(self, rest_api, value):
         """
--- a/samtranslator/swagger/swagger.py
+++ b/samtranslator/swagger/swagger.py
@@ -310,7 +310,8 @@ class SwaggerEditor(object):
         for authorizerName, authorizer in authorizers.items():
             self.security_definitions[authorizerName] = authorizer.generate_swagger()
 
-    def set_path_default_authorizer(self, path, default_authorizer, authorizers):
+    def set_path_default_authorizer(self, path, default_authorizer, authorizers,
+                                    add_default_authorizer_to_cors_preflight = True):
         """
         Sets the DefaultAuthorizer for each method on this path. The DefaultAuthorizer won't be set if an Authorizer
         was defined at the Function/Path/Method level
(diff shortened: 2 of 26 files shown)


brettstack:
Thanks for the PR @jbutz. I don't think we should be adding this in every scenario. Please add an additional property to the `Auth` property called `AddDefaultAuthorizerToCorsPreflight`. This property should have a default value of `True` so that we don't break anyone relying on this functionality.

jbutz:
Going back to redo this, I think have the code to do it mostly there right now but the unit tests are giving me troubles.

Closing this PR

keetonian:
@jbutz reach out if you need help with the unit tests. It's usually the hash on the end of the logical id for APIGW deployment resources. If you run them locally, run it in Python 2.7 if possible as that's what the tests check for those hashes.

xrendan:
@brettstack @keetonian I'm looking to pick this up and I've worked passed it now, but with the current implementation of the test suite, when the test encounters an exception, it throws an UnboundLocalError instead of giving a stacktrace with reference to this bug: https://github.com/wolever/parameterized/issues/66 in parameterized. 

Right now I'm using the workaround specified in the thread but it could be worth committing it so that contributing is easier for other people in the future.

Issue #815, opened 2019-02-15
Create AWS::Serverless::GatewayResponse resource


sthulb:
There's a number of issues raised by customers regarding the mix of `AWS::Serverless::Api` and `AWS::ApiGateway::GatewayResponse`

We should do better and create a resource to ensure this works. From my understanding this is a result of merge behaviour and quirks in CFN.

#623

carlnordenfelt:
Would it make sense, to extend the `AWS::Severless:Api` resource with gateway responses to avoid adding another Serverless resource?

Example:
    ExampleAPI:
      Type: 'AWS::Serverless::Api'
      Properties:
        GatewayResponses:
          - ResponseParameters:
               gatewayresponse.header.Access-Control-Allow-Origin: "'*'"
             ResponseType: 'DEFAULT_4XX'

I think the above would be very useful and a 'simple' solution to the problem.

brettstack:
@carlnordenfelt is correct, we would need to extend the `AWS::Serverless::Api` with a `GatewayResponses` parameter instead of creating a new resource. In this case we would probably just pass through anything defined on `GatewayResponses` and apply it as `x-amazon-apigateway-gateway-responses` in the Swagger body.

Note we are talking more about the "Merge" feature I alluded to in #623 which will effectively do this pass through.

brettstack:
For anyone who wants to have a go at implementing this, here are some relevant places in code:

1. https://github.com/awslabs/serverless-application-model/blob/master/samtranslator/model/apigateway.py#L11
1. https://github.com/awslabs/serverless-application-model/blob/master/samtranslator/model/api/api_generator.py#L52
1. Apply the value to swagger https://github.com/awslabs/serverless-application-model/blob/master/samtranslator/swagger/swagger.py#L33

carlnordenfelt:
If only it wasn't python :)

brettstack:
@carlnordenfelt curious which language you would prefer.

carlnordenfelt:
@brettstack feels like we are hijacking the issue a bit but my go to is nodejs or even Java. Not too familiar with the ins and outs of python.

Buffer0x7cd:
@brettstack I would love to work on the above issue.

keetonian:
@Buffer0x7cd Absolutely! Go ahead and give it a try, and reach out if you have any questions!

chrisoverzero:
[...]

PR #841, opened 2019-03-05
feat(gatewayresponses): add support for API Gateway Responses

chrisoverzero:
~This is a WIP. Remaining:~

1. ~Documentation~
2. ~Example~
3. Edge Cases Which Will Be Pointed Out to Me, _I'm Sure???_

*Issue #, if available:*

#815

*Description of changes:*

Added new property `GatewayResponses` to `AWS::Serverless::API`. This translates to `x-amazon-apigateway-gateway-responses` in the generated Swagger with some convenience transformations applied for SAMier syntax.

In addition to the present test suite, ran a company project through the translator. It deployed and worked correctly. (Then I reluctantly took it back down.)

*Checklist:*

- [x] Write/update tests
- [x] `make pr` passes
- [x] Update documentation
- [x] Create example


samtranslator/model/api/api_generator.py
```diff
@@ -269,6 +288,69 @@ def _add_auth(self):
...
+                self.logical_id, "GatewayResponses works only with inline Swagger specified in "
+                                 "'DefinitionBody' property")
+
+        # Make sure keys in the dict are recognized
+        for responses_key, responses_value in self.gateway_responses.items():
+            if responses_key not in GatewayResponsesProperties._fields:
+                raise InvalidResourceException(
+                    self.logical_id, "Invalid key '{}' for 'GatewayResponses' property".format(responses_key))
+
+            if 'StatusCode' not in responses_value.keys():
```
carlnordenfelt: If I'm not looking at this wrong, StatusCode should be optional.
It should only be provided if you want to change the original status code:
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-apigateway-gatewayresponse.html#cfn-apigateway-gatewayresponse-statuscode

chrisoverzero: You are right. I was developing against [this documentation][], which doesn’t mention that. When I get back to my desk, I'll compare what I wrote to the CF docs to see if there's anything else like this. 

[this documentation]: 
https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-swagger-extensions-gateway-responses.gatewayResponse.html

chrisoverzero: Done!

chrisoverzero:
Documentation is added. I tried to match house style, but lemme know if it's n-not?

I _intend_ to have an example committed today, but Day Job may interfere.

chrisoverzero:
OK, I think we in it.

I am the farthest thing from a professional Python developer, so please do let me know where I've fallen down.

samtranslator/model/apigateway.py
```diff
@@ -93,6 +93,38 @@ def make_auto_deployable(self, stage, swagger=None):
...
+        self.status_code = status_code
+
+    def generate_swagger(self):
+        swagger = {
+            "responseParameters": self._add_prefixes(self.response_parameters),
+            "responseTemplates": self.response_templates
+        }
+
+        # Prevent "null" being written.
+        if self.status_code:
```
brettstack: Could you add some regex validation here?

samtranslator/swagger/swagger.py
```diff
@@ -31,6 +31,7 @@ def __init__(self, doc):
         self._doc = copy.deepcopy(doc)
         self.paths = self._doc["paths"]
         self.security_definitions = self._doc.get("securityDefinitions", {})
+        self.gateway_responses = self._doc.get('x-amazon-apigateway-gateway-responses', {})
```
brettstack: Define `_X_APIGW_GATEWAY_RESPONSES` above (line 18)

chrisoverzero:
Awesome. Thanks so much. I'm past EoD here in my time zone, but I'll jump on these in the morning.

samtranslator/swagger/swagger.py
chrisoverzero: Ugh, y'know, I had this originally, but I made it rhyme with the use of `"securityDefinitions"`.

examples/2016-10-31/api_gateway_responses/template.yaml
```diff
@@ -0,0 +1,45 @@
...
+          FunctionArn: !GetAtt Authorizer.Arn
+          Identity:
+            ValidationExpression: "^Bearer +[-0-9a-zA-Z\\._]*$"
+            ReauthorizeEvery: 300
+    GatewayResponses:
+      UNAUTHORIZED:
+        ResponseParameters:
+          Headers:
+            Access-Control-Expose-Headers: "'WWW-Authenticate'"
+            Access-Control-Allow-Origin: "'*'"
```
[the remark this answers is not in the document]
chrisoverzero: My CORS cheatsheet says that the only headers for a non-preflight response are:

- `Access-Control-Expose-Headers`
- `Access-Control-Allow-Origin`
- `Access-Control-Allow-Credentials`

The first two are covered, and the third secretly means "cookies". So I think this is set. (Also, I tested it in a web browser.)

examples/2016-10-31/api_gateway_responses/template.yaml
```diff
@@ -0,0 +1,45 @@
+AWSTemplateFormatVersion: "2010-09-09"
+Transform: AWS::Serverless-2016-10-31
+Description: Simple webservice deomnstrating gateway responses.
+Globals:
+  Api:
```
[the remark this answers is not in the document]
chrisoverzero: Personally, I like the implicit API, but I can see how -- for an example -- being as explicit as possible is important. I've made this change.

samtranslator/model/apigateway.py
chrisoverzero: I've validated via:

```python
re.fullmatch(r'[0-9]{1,3}', str(status_code))
```

I hope that's what you meant.

samtranslator/model/api/api_generator.py
```diff
@@ -269,6 +276,61 @@ def _add_auth(self):
...
+        if not self.gateway_responses:
+            return
+
+        if self.gateway_responses and not self.definition_body:
+            raise InvalidResourceException(
+                self.logical_id, "GatewayResponses works only with inline Swagger specified in "
+                                 "'DefinitionBody' property")
+
+        # Make sure keys in the dict are recognized
+        for responses_key, responses_value in self.gateway_responses.items():
```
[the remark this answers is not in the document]
chrisoverzero: I've moved most of the validation there. I left behind the validation for the presence of properties other than `ResponseParameters`, `ResponseTemplates` and `StatusCode` because they're accessed in order to _create_ an `ApiGatewayResponse`.

But if ignored, overflow properties are actually OK, I can scratch it entirely.

chrisoverzero:
Apologies for the force pushes -- trying to get the merge conflict resolved.

This is now caught up with current feedback.

samtranslator/model/apigateway.py
```diff
@@ -93,6 +95,51 @@ def make_auto_deployable(self, stage, swagger=None):
...
+    def __init__(self, api_logical_id=None, response_parameters=None, response_templates=None, status_code=None):
+        if response_parameters:
+            for response_parameter_key in response_parameters.keys():
+                if response_parameter_key not in ApiGatewayResponse.ResponseParameterProperties:
+                    raise InvalidResourceException(
+                        api_logical_id,
+                        "Invalid gateway response parameter '{}'".format(response_parameter_key))
+
+        # status_code must look like a status code, if present. Let's not be judgmental; just check 0-999.
+        if status_code and not match(r'^[0-9]{1,3}$', str(status_code)):
```
brettstack: Yes, this regex is sufficient, however, I don't think we should forbid strings here. The docs say StatusCode should be a string https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-apigateway-gatewayresponse.html#cfn-apigateway-gatewayresponse-statuscode but it may accept both String and Number (with the latter getting parsed to the former)

chrisoverzero: Sure thing. That suggests the following tests to me:

- `StatusCode: 401` => PASS
- `StatusCode: '401'` => PASS
- `StatusCode: four-oh-one` => FAIL

> [...] I don't think we should forbid strings here.

I don't think I am anymore, am I? I removed the numeric type test. I'm inexperienced at Python, but my understanding is that `str` will turn both `401` and `'401'` into `'401'`.

chrisoverzero: Oh, I think I see now. You must mean "Property 'StatusCode' must be numeric", right?

I guess I was thinking of "numeric" as "_looks like_ a number". Do you have a suggestion of better wording that would explain allowing `200` and `'200'`?

brettstack: Ignore me, I missed the part where you were casting it to a string. I think the error message as-is is fine.

chrisoverzero: >Ignore me [...]

Never!

>I think the error message as-is is fine.

Cool. I think I'm caught up, then.

carlnordenfelt:
Really glad to see this merged! Thanks @chrisoverzero for picking this up!

@brettstack @keetonian do you guys have any idea when this will be released? I'm in dire need of this to avoid having to define the entire API myself at the moment.

keetonian:
We're working on it now. We just cut a release branch: https://github.com/awslabs/serverless-application-model/tree/release/v1.11.0

We should be able to release this soon, barring any unforeseen issues with the new features.

Issue #836, opened 2019-03-04
Lambda Custom Deployment configurations in CodeDeploy cant be referenced in SAM template under DeploymentPreference


dan-schw:
**Description:**

When creating a Lambda function with a deployment configuration (DeploymentPreference) you need to specify a Type. There are some defaults as defined here: https://docs.aws.amazon.com/codedeploy/latest/userguide/deployment-configurations.html#deployment-configurations-predefined-lambda

However we are unable to use a custom deployment configuration from CodeDeploy. When we make the Type of the DeploymentPreference the name of the custom deployment configuration (eg TestDeploymentConfig) then Cloudformation looks for the deployment configuration CodeDeployDefault.LambdaTestDeploymentConfig. 

**Steps to reproduce the issue:**
1. Create a custom deployment configuration in CodeDeploy called TestDeploymentConfiguration
2. Try to reference it in DeploymentPreference - Type
3. The CF stack made from SAM shows the following error: No deployment configuration found for name: CodeDeployDefault.LambdaTestDeploymentConfiguration

**Code**

This naming seems hardcoded here:

serverless-application-model/samtranslator/model/preferences/deployment_preference_collection.py
Line: 113

```
deployment_group.DeploymentConfigName = fnSub("CodeDeployDefault.Lambda${ConfigName}",
                                                      {"ConfigName":  deployment_preference.deployment_type})
```

With this naming we wont be able to point our deployment preferences to a custom deployment configuration because it always places "CodeDeployDefault.Lambda" in front of the name.

Here (https://docs.aws.amazon.com/codedeploy/latest/userguide/deployment-configurations.html#deployment-configuration-lambda)
 it says "You can also create your own custom canary or linear deployment configuration."

How do we reference a custom deployment configuration in SAM?

dan-schw:
SAM Template

```
Transform: 'AWS::Serverless-2016-10-31'
Resources:
  MyLambdaFunction:
    Type: AWS::Serverless::Function
    Properties:
      Handler: index.handler
      Runtime: nodejs6.10
      CodeUri: s3://bucket/code.zip
      AutoPublishAlias: live
      MemorySize: 1024
      Timeout: 15 
      Policies:
        - AWSLambdaExecute # Managed Policy
      DeploymentPreference:
        Enabled: true
        Type: TestDeployment

```

CF Processed template:

```
[...]

PR #848, opened 2019-03-09
Added Support for referencing custom configuration in CodeDeploy

Buffer0x7cd:
*Issue #, if available:*
[#836](https://github.com/awslabs/serverless-application-model/issues/836)
*Description of changes:*
Added a list of predefined configurations provided by CodeDeploy. If the user provided configuration is one of the predefined configurations then the generate Cloudformation will have `"CodeDeployDefault.Lambda` as a prefix, but if it's a custom CodeDeploy Configuration then the Prefix will not be added 
Added relevant unit tests and the end to end tests.
*Checklist:*

- [ x] Write/update tests
- [ x] `make pr` passes
- [x] Update documentation


Buffer0x7cd:
@jlhood There were few tests that are using custom deployments type in deployment preference and expecting the generated cloudformation to have `CodeDeployDefault.Lambda` prefix(https://github.com/awslabs/serverless-application-model/blob/805cc85bc26cfda16e9611c047e38f78560ad6e6/tests/translator/model/preferences/test_deployment_preference_collection.py#L66) While few of them are having a typo(https://github.com/awslabs/serverless-application-model/blob/805cc85bc26cfda16e9611c047e38f78560ad6e6/tests/translator/input/function_with_deployment_preference.yaml#L10). Due to this, I needed to update those test cases so that they can work with the changes introduced in this PR.

samtranslator/model/preferences/deployment_preference_collection.py
```diff
@@ -109,8 +120,13 @@ def deployment_group(self, function_logical_id):
                                                       'Events': ['DEPLOYMENT_FAILURE',
                                                                  'DEPLOYMENT_STOP_ON_ALARM',
                                                                  'DEPLOYMENT_STOP_ON_REQUEST']}
-        deployment_group.DeploymentConfigName = fnSub("CodeDeployDefault.Lambda${ConfigName}",
-                                                      {"ConfigName": deployment_preference.deployment_type})
+
+        if deployment_preference.deployment_type in CODEDEPLOY_PREDEFINED_CONFIGURATIONS_LIST:
```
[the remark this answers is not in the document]
Buffer0x7cd: Thanks for the Feedback, will be working on the Changes as requested :+1:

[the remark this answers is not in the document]
Buffer0x7cd: @jlhood Thanks for the feedback. Due to some health Issues, I was not able to work on this issue earlier, but now I have integrated your suggestion about supporting dict types values in the deployment_type property. please feel free to review and add any additional suggestions.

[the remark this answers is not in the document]
jlhood: @Buffer0x7cd Sorry to hear about your health issues. Hope you're feeling better! Thanks for the updates. We'll review them soon!

samtranslator/model/preferences/deployment_preference_collection.py
```diff
@@ -109,8 +120,20 @@ def deployment_group(self, function_logical_id):
                                                       'Events': ['DEPLOYMENT_FAILURE',
                                                                  'DEPLOYMENT_STOP_ON_ALARM',
                                                                  'DEPLOYMENT_STOP_ON_REQUEST']}
-        deployment_group.DeploymentConfigName = fnSub("CodeDeployDefault.Lambda${ConfigName}",
-                                                      {"ConfigName": deployment_preference.deployment_type})
+        if isinstance(deployment_preference.deployment_type, dict):
```
jlhood: The way you did the recursive function is very specific to `Fn::If` rather than working for any intrinsic function. I think we can do this in a more concise way that's also more general to work with any intrinsic function. How about something like this instead:

```python
        deployment_group.DeploymentConfigName = self._replace_deployment_types(copy.deepcopy(deployment_preference.deployment_type))
...

    def _replace_deployment_types(self, value):
        if isinstance(value, list):
            for i in range(len(value)):
                value[i] = self._replace_deployment_types(value[i])
            return value
        elif isinstance(value, dict):
            for (k, v) in value.items():
                value[k] = self._replace_deployment_types(v)
            return value
        else:
            if value in CODEDEPLOY_PREDEFINED_CONFIGURATIONS_LIST:
                return fnSub("CodeDeployDefault.Lambda${ConfigName}", {"ConfigName": value})
            return value
```

tests/translator/input/function_with_custom_conditional_codedeploy_deployment_preference.yml
```diff
@@ -0,0 +1,17 @@
...
+  IsDevEnv2: !Equals [!Ref EnvType, prod]
+Resources:
+  HelloWorldFunction:
+    Type: AWS::Serverless::Function
+    Properties:
+      Handler: index.handler
+      Runtime: nodejs8.10
+      AutoPublishAlias: live
+      DeploymentPreference:
+        Type: !If [IsDevEnv, !If [IsDevEnv2, AllAtOnce, TestCustomDeploymentConfig], Canary10Percent15Minutes]
```
jlhood: Thanks for adding this test! Can you also add an example using `!Sub` or `!Join` to test intrinsic functions that don't follow the same format as `!If`?

samtranslator/model/preferences/deployment_preference_collection.py
Buffer0x7cd: Hi @jlhood  Thanks for your feedback.  while working on this I noticed that the deployment_group function is called twice (I am still trying to figure that section out). this caused the final dictionary containing `Fn::Sub` to get processed once more and resulted in an invalid output, I added that check for `Fn::If` to fix that issue ( I am still learning Cloudformation so I missed the case with other intrinsic functions ). After seeing this feedback  I tried implenting your solution above but due to the `deployment_group` function get called twice the final output gets processed twice. I am attaching the output section which i got after applying above solution.

```json
        "DeploymentConfigName": {
          "Fn::If": [
            "IsDevEnv", 
            "TestingConfig", 
            {
              "Fn::Sub": [
                "CodeDeployDefault.Lambda${ConfigName}", 
                {
                  "ConfigName": {
                    "Fn::Sub": [
                      "CodeDeployDefault.Lambda${ConfigName}", 
                      {
                        "ConfigName": "Canary10Percent15Minutes"
                      }
                    ]
                  }
                }
              ]
            }
          ]
        }
```
As we can see in the above output, when the internal `ConfigName` block get processed during the second call to `deployment_group` it replace the config name with the `Fn::Sub` Behavior, which it should not in this case.

Buffer0x7cd: To solve this issue, I have extended the check which I was doing earlier for `Fn::If`. The modified version of _replace_deployment_types is below.

```python
    def _replace_deployment_types(self, value):
        if isinstance(value, list):
            for i in range(len(value)):
                value[i] = self._replace_deployment_types(value[i])
            return value
        elif is_instrinsic(value):
            for (k, v) in value.items():
                value[k] = self._replace_deployment_types(v)
            return value
        else:
            if value in CODEDEPLOY_PREDEFINED_CONFIGURATIONS_LIST:
                return fnSub("CodeDeployDefault.Lambda${ConfigName}", {"ConfigName": value})
            return value
```
This way we can avoid processing the dictionaries that are not instrinsic dictionaries. Pleas let me know your feedback about the change.

Buffer0x7cd: Also can you please provide an example of the uses of `Join` or `!Sub` like you did earlier fof `!If`, because i am not able see a clear usages of these functions in user provided input. the only use case i can think of is that `!Join` can be used to create the Name of a custom configuration option from an array.

jlhood: Thanks for the updates!

> the deployment_group function is called twice (I am still trying to figure that section out). this caused the final dictionary containing Fn::Sub to get processed once more and resulted in an invalid output

Did you do the deep copy on `deployment_preference.deployment_type` I had suggested above?

```python
        deployment_group.DeploymentConfigName = self._replace_deployment_types(copy.deepcopy(deployment_preference.deployment_type))
```

Calling the `deployment_group()` method twice should be ok as long as it's always only modifying a copy of `deployment_preference.deployment_type`. If you forgot to add the deepcopy call, that would explain why it got double-modified.

> Also can you please provide an example of the uses of Join or !Sub like you did earlier fof !If

I'm honestly having a difficult time coming up with a concrete example that doesn't seem completely contrived. Looking at the other [intrinsic functions](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference.html) I don't know that we should recursively replace all of the strings that match the supported deployment pref types. Maybe `Fn::If` is the only one that makes sense to use. However, note that `Fn::If`s can be nested within each other. So maybe an example to test (although still a bit contrived) would be:

```yaml
DeploymentPreference:
  Type: !If
    - IsDevEnv
    - !If
      - IsGammaEnv
      - Canary10Percent5Minutes
      - AllAtOnce
    - Canary10Percent15Minutes
```

Apologies for the churn on this one!

Buffer0x7cd: @jlhood I have updated my implementation with your feedbacks also I have added tests for nested if conditions. please let me know if some changes are required.

tests/translator/input/function_with_custom_conditional_codedeploy_deployment_preference.yml
jlhood: Nice!

tests/translator/model/preferences/test_deployment_preference_collection.py
```diff
@@ -77,12 +76,70 @@ def test_deployment_group_with_minimal_parameters(self):
...
+                                      'Canary10Percent15Minutes']}
+
+        expected_deployment_config_name = {'Fn::If':
+                                           ['IsDevEnv', {'Fn::If':
+                                                         ['IsDevEnv1', {'Fn::Sub': [
+                                                                'CodeDeployDefault.Lambda${ConfigName}',
+                                                                {
+                                                                   'ConfigName': 'AllAtOnce'
+                                                                }
+                                                                ]
```
jlhood: Whitespace looks off here. Can you format it?

tests/translator/output/aws-cn/function_with_custom_codedeploy_deployment_preference.json
```diff
@@ -0,0 +1,142 @@
...
+            "DeploymentType": "BLUE_GREEN", 
+            "DeploymentOption": "WITH_TRAFFIC_CONTROL"
+          }
+        }
+      }, 
+      "MinimalFunctionRole": {
+        "Type": "AWS::IAM::Role", 
+        "Properties": {
+          "ManagedPolicyArns": [
+            "arn:aws:iam::aws:policy/service-role/AWSLambdaBasicExecutionRole"
```
jlhood: Did `make pr` pass? ARNs in the CN partitions should be of the form `arn:aws-cn:iam:...`. Similarly in other partitions, e.g., GovCloud, there's ARNs have a different partition value. `make pr` will show you these errors so you can correct them. The Travis build also runs these tests and that has to pass in order for the PR to be merged.

jlhood: I see what's happening now. The new test input templates you've added are not getting run as part of the unit tests. You need to add them [here](https://github.com/Buffer0x7cd/serverless-application-model/blob/develop/tests/translator/test_translator.py#L210) in order to actually run them during the tests. Then run `make pr` and fix any errors.

Buffer0x7cd: @jlhood Is there are any other steps involved for adding the new end to end test templates ?. I added my test templates in the location mentioned above and also fixed the ARN schema for China and Us Gov Cloud output templates, but I am receiving a strange error regarding when running make pr. 
Screenshot from 2019-04-09 01-43-05
I have attached the relevant section's output in the above image. as the test_translatior.py codebase is new for me, I am thinking that I might be missing some necessary steps. please let me know your thoughts about this?

brettstack: `patching` isn't in scope inside the `except` block. You can move `try/except` to be inside the iterator.

keetonian: I think I just fixed the issue with these tests.

jlhood: Nice! Thanks for fixing those, @keetonian. Really excited to get this one merged. 🙏

Buffer0x7cd: @keetonian  @jlhood apologies for not able to dedicate time for this issue through the end. can you please let me know, what was the issue with the test?

jlhood: @Buffer0x7cd These were the changes @keetonian made to enable the two tests and address the failures:

https://github.com/awslabs/serverless-application-model/pull/848/commits/0b67e7aca16a0bd074f82d4c4dcff64a1129849f

keetonian: It was mostly the partition names: China uses aws-cn and GovCloud uses aws-us-gov. Otherwise, things were pretty well correct.

PR #856, opened 2019-03-15
feat: Add support for FindInMap in serverless app plugin

honglu:
*Description of changes:*
- Add FindInMap action
- Support intrinsics in serverless app plugin to resolve ApplicationId and SemanticVersion under Location

Added unit tests and end to end tests

*Checklist:*

- [x] Write/update tests
- [x] `make pr` passes

### DIFF
--- a/samtranslator/intrinsics/actions.py
+++ b/samtranslator/intrinsics/actions.py
@@ -533,9 +533,9 @@ class FindInMapAction(Action):
         top_level_key = self.resolve_parameter_refs(value[1], parameters)
         second_level_key = self.resolve_parameter_refs(value[2], parameters)
 
-        if not isinstance(map_name, basestring) or \
-                not isinstance(top_level_key, basestring) or \
-                not isinstance(second_level_key, basestring):
+        if not isinstance(map_name, string_types) or \
+                not isinstance(top_level_key, string_types) or \
+                not isinstance(second_level_key, string_types):
             return input_dict
 
         if map_name not in parameters or \
--- a/samtranslator/intrinsics/actions.py
+++ b/samtranslator/intrinsics/actions.py
@@ -503,3 +503,50 @@ class GetAttAction(Action):
             input_dict[key] = [resolved_value] + remaining
 
         return input_dict
+
+
+class FindInMapAction(Action):
+    """
+    This action can't be used along with other actions.
+    """
+    intrinsic_name = "Fn::FindInMap"
+
+    def resolve_parameter_refs(self, input_dict, parameters):
+        """
+        Recursively resolves "Fn::FindInMap"references that are present in the mappings and returns the value.
+        If it is not in mappings, this method simply returns the input unchanged.
+
+        :param input_dict: Dictionary representing the FindInMap function. Must contain only one key and it
+                           should be "Fn::FindInMap".
+
+        :param parameters: Dictionary of mappings from the SAM template
+        """
+        if not self.can_handle(input_dict):
+            return input_dict
+
+        value = input_dict[self.intrinsic_name]
+
+        if not isinstance(value, list) or len(value) != 3:
+            return input_dict
+
+        map_name = self.resolve_parameter_refs(value[0], parameters)
+        top_level_key = self.resolve_parameter_refs(value[1], parameters)
+        second_level_key = self.resolve_parameter_refs(value[2], parameters)
+
+        if not isinstance(map_name, basestring) or \
+                not isinstance(top_level_key, basestring) or \
+                not isinstance(second_level_key, basestring):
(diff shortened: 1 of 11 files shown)


brettstack:
Please fix the Travis error

samtranslator/intrinsics/actions.py
```diff
@@ -503,3 +503,50 @@ def _get_resolved_dictionary(self, input_dict, key, resolved_value, remaining):
...
+                           should be "Fn::FindInMap".
+
+        :param parameters: Dictionary of mappings from the SAM template
+        """
+        if not self.can_handle(input_dict):
+            return input_dict
+
+        value = input_dict[self.intrinsic_name]
+
+        if not isinstance(value, list) or len(value) != 3:
```
brettstack: Can you add a comment, like "FindInMap expects an array with 3 values". Wondering if we should add some validation here and throw?

samtranslator/intrinsics/actions.py
```diff
@@ -503,3 +503,50 @@ def _get_resolved_dictionary(self, input_dict, key, resolved_value, remaining):
...
+            return input_dict
+
+        if map_name not in parameters or \
+                top_level_key not in parameters[map_name] or \
+                second_level_key not in parameters[map_name][top_level_key]:
+            return input_dict
+
+        return parameters[map_name][top_level_key][second_level_key]
+
+    def resolve_resource_refs(self, input_dict, supported_resource_refs):
```
brettstack: Why do we need this and `resolve_resource_id_refs`?

tests/intrinsics/test_actions.py
```diff
@@ -932,3 +932,203 @@ def test_return_value_if_cannot_handle(self, can_handle_mock):
...
+                "TopKey1": {
+                   "SecondKey2": "value3"
+               },
+                "TopKey2": {
+                    "SecondKey1": "TopKey1"
+                }
+            }
+        }
+        input = {
+            "Fn::FindInMap": ["MapA", {"Fn::FindInMap": ["MapA", "TopKey2", "SecondKey1"]}, "SecondKey2"]
```
brettstack: Can you add a similar test, but one which looks up an entry in `MapA` based on an entry in `MapB`?

tests/plugins/application/test_serverless_app_plugin.py
```diff
@@ -192,7 +198,24 @@ def test_sar_service_calls(self):
         app_id = 'app_id'
         semver = '1.0.0'
         response = self.plugin._sar_service_call(service_call_lambda, logical_id, app_id, semver)
-        self.assertEqual(app_id, response['ApplicationId'])
+        self.assertEqual(app_id, response['ApplicationId'])    \
```
brettstack: What is that trailing \\?

samtranslator/intrinsics/actions.py
honglu: Do you mean we should throw `InvalidTemplateException` here instead of silently return the input?

brettstack: Right. I'm wondering in what circumstances would the user want it to silently return instead of having some validation which will warn them the template is invalid before it creates/executes the change set.

samtranslator/intrinsics/actions.py
```diff
@@ -503,3 +504,46 @@ def _get_resolved_dictionary(self, input_dict, key, resolved_value, remaining):
...
+
+        :param parameters: Dictionary of mappings from the SAM template
+        """
+        if not self.can_handle(input_dict):
+            return input_dict
+
+        value = input_dict[self.intrinsic_name]
+
+        # FindInMap expects an array with 3 values
+        if not isinstance(value, list) or len(value) != 3:
```
brettstack: Should we also throw errors in the other scenarios where we `return input_dict`? Or are there cases where we can't statically resolve them but they may get successfully resolved at runtime during CFN change set execution?

honglu: You are right. In other scenarios, it is possible that some intrinsic functions are still left unresolved.

tests/translator/input/application_with_intrinsics.yaml
```diff
@@ -0,0 +1,40 @@
...
+    us-gov-west-1:
+      ApplicationId: arn:aws-gov:serverlessrepo:us-gov-west-1:123456789012:applications/hello-world
+      Version: 1.0.3
+
+Resources:
+  ApplicationRefParameter:
+    Type: 'AWS::Serverless::Application'
+    Properties:
+      Location:
+        ApplicationId: !Ref ApplicationIdParam
```
keetonian: Good, thank you for adding support for this.

PR #862, opened 2019-03-21
refactor: move methods that modify parameter values to a separate class

honglu:
*Description of changes:*
Move methods that modify parameter values to a separate class

*Checklist:*

- [x] Write/update tests
- [x] `make pr` passes


samtranslator/translator/translator.py
```diff
@@ -46,8 +46,10 @@ def translate(self, sam_template, parameter_values):
         :returns: a copy of the template with SAM resources replaced with the corresponding CloudFormation, which may \
                 be dumped into a valid CloudFormation JSON or YAML template
         """
-        parameter_values = self._add_default_parameter_values(sam_template, parameter_values)
-        parameter_values = self._add_pseudo_parameter_values(parameter_values)
+        sam_parameter_values = SamParameterValues(parameter_values)
```
brettstack: Could you instead add 2 boolean parameters to the constructor (add_default_parameters, and add_psuedo_parameters). When these are true, call `self._ add_default_parameter_values` and `self._ add_pseudo_parameter_values` respectively. SAM already has a lot of mutating calls, but I'd prefer to mitigate where we can.

tests/sdk/test_parameter.py
```diff
@@ -0,0 +1,136 @@
...
+
+        expected = {
+            "Param1": "value1"
+        }
+
+        sam_parameter_values = SamParameterValues(parameter_values)
+        sam_parameter_values.add_default_parameter_values(sam_template)
+        self.assertEqual(expected, sam_parameter_values.parameter_values)
+
+    def test_add_default_parameter_values_must_skip_params_without_defaults(self):
```
brettstack: This would fail CFN execution, right? All Params require a value. Can (should) we fail early? I like the idea of SAM failing faster than CFN when possible. Then again, it could be possible that a Plugin could fill in Params, but I can't think of a use-case. I'd like to get James and Keeton's feedback too.

tests/sdk/test_parameter.py
```diff
@@ -0,0 +1,136 @@
...
+
+        # String
+        param("something"),
+
+        # Some other non-parameter looking dictionary
+        param({"Param1": {"Foo": "Bar"}}),
+
+        param(None)
+    ])
+    def test_add_default_parameter_values_must_ignore_invalid_template_parameters(self, template_parameters):
```
brettstack: Again, if they're invalid can we fail early?

samtranslator/translator/translator.py
honglu: I used `copy.deepcopy` so it won't modify the original parameter_values. I actually prefer to leave those two methods separated instead of bundling them together to give the caller more flexibility to decide whether to just use one of them or both of them

PR #873, opened 2019-03-26
feat: update bin/sam-translate.py

brettstack:
* Add additional commands
* Update parameters to be consistent with SAM CLI
* Add additional parameters
* Improve logging
* Fixes


bin/sam-translate.py
```diff
@@ -5,15 +5,25 @@
 Known limitations: cannot transform CodeUri pointing at local directory.
 
 Usage:
-  sam-translate.py --input-file=sam-template.yaml [--output-file=<o>]
+  sam-translate.py --template-file=sam-template.yaml [--verbose] [--output-template=<o>]
+  sam-translate.py package --template-file=sam-template.yaml --s3-bucket=my-bucket [--verbose] [--output-template=<o>]
```
jlhood: Can you confirm you've manually tested these new commands with this latest version of the fix before merging?

tests/translator/input/error_application_properties.yaml
```diff
@@ -33,4 +33,13 @@ Resources:
...
-        SemanticVersion:
\ No newline at end of file
+        SemanticVersion:
+
+  IntrinsicProperties:
+    Type: 'AWS::Serverless::Application'
+    Properties:
+      Location:
+        ApplicationId:
+          Sub: foobar
```
keetonian: Could make one of these be `Fn::Sub` to test for unsupported intrinsics as well as random dict keys.

samtranslator/plugins/application/serverless_app_plugin.py
```diff
@@ -211,11 +219,23 @@ def on_before_transform_resource(self, logical_id, resource_type, resource_prope
                                        [self.APPLICATION_ID_KEY, self.SEMANTIC_VERSION_KEY])
 
         app_id = resource_properties[self.LOCATION_KEY].get(self.APPLICATION_ID_KEY)
+
         if not app_id:
             raise InvalidResourceException(logical_id, "Property 'ApplicationId' cannot be blank.")
+
+        if isinstance(app_id, dict):
+            raise InvalidResourceException(logical_id, "Property 'ApplicationId' cannot be resolved. Only FindInMap "
```
keetonian: The error messages for both of these instances are practically the same, any way we could re-use the verbage and insert the property name?

samtranslator/plugins/application/serverless_app_plugin.py
```diff
@@ -86,10 +87,17 @@ def on_before_transform_template(self, template_dict):
 
             app_id = self._replace_value(app.properties[self.LOCATION_KEY],
                                          self.APPLICATION_ID_KEY, intrinsic_resolvers)
+
             semver = self._replace_value(app.properties[self.LOCATION_KEY],
                                          self.SEMANTIC_VERSION_KEY, intrinsic_resolvers)
 
+            if isinstance(app_id, dict) or isinstance(semver, dict):
+                key = (json.dumps(app_id), json.dumps(semver))
+                self._applications[key] = False
```
keetonian: Why did you choose to set this to `False`?

tests/translator/input/error_resource_properties_not_dict.yaml
```diff
@@ -0,0 +1,4 @@
+Resources:
+  Function:
+    Type: AWS::Serverless::Function
+    Properties:
```
keetonian: What if `Properties` doesn't exist? Does this cover this case as well? Or is that already caught?

tests/translator/input/error_resource_properties_not_dict.yaml
```diff
@@ -0,0 +1,4 @@
+Resources:
+  Function:
+    Type: AWS::Serverless::Function
```
keetonian: Do we have a test where `Type` doesn't exist? I saw that in a log dive as well.

tests/translator/input/error_resource_properties_not_dict.yaml
brettstack: I think that is already covered and it was just this case that wasn't.

tests/translator/input/error_resource_properties_not_dict.yaml
brettstack: Yes I believe this is already covered.

tests/translator/input/error_function_invalid_autopublishalias.yaml
```diff
@@ -0,0 +1,115 @@
...
+      AutoPublishAlias:
+        Ref: MyAlias
+      Events:
+        Get:
+          Type: Api
+          Properties:
+            Path: "/path"
+            Method: GET
+# ---
+# Parameters:
```
keetonian: Why do we have all of this if it is all commented out?

brettstack: Removed

samtranslator/plugins/application/serverless_app_plugin.py
brettstack: Just needed a value here. Wanted to get any other recommendations from you.

keetonian: Ok. We're getting a value here because we don't resolve intrinsics until later, right? Could we resolve them earlier so that the flow in this area remains the same?

tests/translator/test_translator.py
```diff
@@ -353,19 +353,24 @@ def _generate_new_deployment_hash(self, logical_id, dict_to_hash, rest_api_to_sw
     'error_api_gateway_responses_unknown_responseparameter',
     'error_api_gateway_responses_unknown_responseparameter_property',
     'error_api_invalid_auth',
+    'error_api_invalid_',
```
jlhood: Can you finish this file name?

samtranslator/plugins/application/serverless_app_plugin.py
keetonian: I think you can actually stick an error instead of `False` into `self._applications[key]`. If the `app_id` or `semver` is still a dict, we won't resolve it any further so we'll just have to return an error at some point.

PR #883, opened 2019-04-10
chore: Update example apps from nodejs6.10 to nodejs8.10 and add samp…

carvantes:
*Description of changes:*
Upgrade several example apps from nodejs6.10 to nodejs8.10. Also, added a `testEvent.json` with a valid test event on each of them to make them easier to test with sam local invoke.
Most test events are taken from `sam local generate-event`

`sam local invoke -e testEvent.json` ran successfully on all examples

The exceptions that cannot be tested that way are:

- `s3-get-object`, deployed with sam package/deploy and manually tested.
- `microservice-http-endpoint`, deployed with sam package/deploy and created a simple DDB table to manually test it

*Checklist:*

- [ ] Write/update tests. N/A, There are no actual tests for example apps, the added `testEvent.json` is useful for testing though
- [x] `make pr` passes
- [ ] Update documentation. N/A, there are no docs to update


examples/apps/cloudfront-http-redirect/testEvent.json
```diff
@@ -0,0 +1 @@
+{}
```
paoptu023: Seems meaningless to include an event for this app. You can simply run [sam local invoke --no-event](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-cli-command-reference-sam-local-invoke.html) without an event.

Applies to all apps that empty event files were added.

carvantes: At first it was hard to tell which examples could run with `--no-event` so I added the empty event to make it explicit that the lambda function does not interact with the incoming event in any way.

Maybe the intent of that empty test event is not as intuitive as I thought. I can remove it

carvantes: I agree those empty events are weird, I've removed them

PR #904, opened 2019-04-29
feat(policy-templates): add new policy for allowing step functions ex…

ljacobsson:
…ecutions

*Description of changes:*
Added policy template allowing `states:StartExecution` on a Step Functions state machine

I have used the policy statement in my own stacks. 

*Checklist:*

- [X] Write/update tests
- [X] `make pr` passes
- [X] Update documentation

### DIFF
--- a/samtranslator/policy_templates_data/policy_templates.json
+++ b/samtranslator/policy_templates_data/policy_templates.json
@@ -1579,9 +1579,9 @@
             ],
             "Resource": {
               "Fn::Sub": [
-                "arn:aws:states:${AWS::Region}:${AWS::AccountId}:stateMachine:${parameterName}",
+                "arn:aws:states:${AWS::Region}:${AWS::AccountId}:stateMachine:${stateMachineName}",
                 {
-                  "parameterName": {
+                  "stateMachineName": {
                     "Ref": "StateMachineName"
                   }
                 }
--- a/samtranslator/policy_templates_data/policy_templates.json
+++ b/samtranslator/policy_templates_data/policy_templates.json
@@ -1562,6 +1562,34 @@
           }
         ]
       }
+    },
+    "StepFunctionsExecutionPolicy": {
+      "Description": "Gives permission to start a Step Functions state machine execution",
+      "Parameters": {
+        "StateMachineName": {
+          "Description":"The name of the state machine to execute."
+        }
+      },
+      "Definition": {
+        "Statement": [
+          {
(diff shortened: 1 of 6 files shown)


samtranslator/policy_templates_data/policy_templates.json
```diff
@@ -1562,6 +1562,34 @@
...
+      "Definition": {
+        "Statement": [
+          {
+            "Effect": "Allow",
+            "Action": [
+              "states:StartExecution"
+            ],
+            "Resource": {
+              "Fn::Sub": [
+                "arn:aws:states:${AWS::Region}:${AWS::AccountId}:stateMachine:${parameterName}",
```
jlhood: Change `parameterName` to `stateMachineName` for better readability. Don't forget to also change it on line 1584.

samtranslator/policy_templates_data/policy_templates.json
```diff
@@ -1562,6 +1562,34 @@
...
+        "StateMachineName": {
+          "Description":"The name of the state machine to execute."
+        }
+      },
+      "Definition": {
+        "Statement": [
+          {
+            "Effect": "Allow",
+            "Action": [
+              "states:StartExecution"
```
jlhood: Are there other permissions we should include here as well? Are you sure this is the only logical action to give to a function to interact with a state machine? `states:StopExecution` seems to make sense, but looking at [the docs](https://docs.aws.amazon.com/IAM/latest/UserGuide/list_awsstepfunctions.html), it doesn't look like it's scoped to a state machine (which seems weird).

I'm not going to block on this since we could add additional permissions later, but I'll also check around with others to see if any other permissions are logical to add to this policy.

ljacobsson: I see what you mean and yes it seems weird that it's not scoped to a state machine. The reason why I only include `StartExecution` and not `StopExecution` is that I think in most cases one Lambda won't be responsible for both starting and stopping a state machine. 

I'll leave this as it is and as you said more permissions can be added later.

ericallam:
I think it'd be nice to include the `states:DescribeExecution` action in this policy template. The use case being the lambda that starts an execution will receive an `executionArn` in the response and should be able to call `describeExecution` on that arn. The `Resource` could be structured using the State Machine name like so:

`arn:${AWS::Partition}:states:${AWS::Region}:${AWS::AccountId}:execution:${StateMachineName}:*`

Issue #892, opened 2019-04-17
Internal transform failure on bad Condition syntax


jlhood:
**Description:**

Customer pinged me that deployment was failing with "FAILED - Transform AWS::Serverless-2016-10-31 failed with: Internal transform failure."

The issue was a syntax error in their template, but SAM should handle that case more gracefully and provide a helpful error message.

The syntax error was in the use of Condition support on SAM resources. Sample template that reproduces the issue:

```yaml
"Conditions":
  "MyCondition":
    "Fn::Equals":
      - true
      - true

"Resources":
 "MyFunction":
    "Type": "AWS::Serverless::Function"
    "Condition":
      "Ref": "MyCondition"  # bad template syntax: don't need to Ref the condition name
    "Properties":
      ...
      "Events":
        "PostApi":
          "Type": "Api"
          "Properties":
            "Method": "POST"
            "Path": "/foo"
```

**Steps to reproduce the issue:**
1. Try to deploy the sample template using SAM.

**Observed result:**
ChangeSet fails with “FAILED - Transform AWS::Serverless-2016-10-31 failed with: Internal transform failure.”

**Expected result:**
ChangeSet fails with more helpful error message from SAM translator.

keetonian:
Fix for this issue:

1. Validate that the Condition (if it exists) is a string: https://github.com/awslabs/serverless-application-model/blob/develop/samtranslator/sdk/resource.py#L27
  * Note: If something other than a string is used as a Condition on a resource, CFN throws an error: `An error occurred (ValidationError) when calling the CreateChangeSet operation: Template format error: Every Condition member must be a string.`
1. Add error case test: https://github.com/awslabs/serverless-application-model/blob/develop/tests/translator/test_translator.py#L444

PR #908, opened 2019-05-03
 Added fix for Internal transform failure on bad Condition syntax

jadhavmanoj:
*Issue #892 :*

*Internal transform failure on bad Condition syntax:*

*Checklist:*

- [x]  Write/update tests
- [x] `make pr` passes
- [x] Update documentation

### DIFF
--- a/samtranslator/sdk/resource.py
+++ b/samtranslator/sdk/resource.py
@@ -36,19 +36,23 @@ class SamResource(object):
         # As long as the type is valid and type string.
         # validate the condition should be string
         validate_type_str = is_type(str)
-        encoded_condition_name = self.condition
-        if sys.version_info.major == 2:
-            # In Py2, only unicode needs to be encoded.
-            if isinstance(self.condition, unicode):
-                encoded_condition_name = self.condition.encode('utf-8')
-        else:
-            encoded_condition_name = self.condition.encode('utf-8')
 
-        if encoded_condition_name and not validate_type_str(encoded_condition_name, should_raise=False):
-            raise InvalidDocumentException([
-                InvalidTemplateException("An error occurred (ValidationError) when calling the CreateChangeSet "
-                                         "operation: Template format error: Every Condition member must be a"
-                                         " string")])
+        if self.condition:
+
+            if sys.version_info.major == 2:
+                # In Py2, only unicode needs to be encoded.
+                if isinstance(self.condition, unicode):
+                    encoded_condition_name = self.condition.encode('utf-8')
+                else:
+                    encoded_condition_name = self.condition
+            else:
+                encoded_condition_name = self.condition.encode('utf-8')
+
+            if not validate_type_str(encoded_condition_name, should_raise=False):
+                raise InvalidDocumentException([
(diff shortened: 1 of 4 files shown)


samtranslator/sdk/resource.py
```diff
@@ -30,7 +33,24 @@ def valid(self):
...
+
+            if sys.version_info.major == 2 and isinstance(self.condition, unicode):
+                # In Py2, only unicode needs to be encoded.
+                encoded_condition_name = self.condition.encode('utf-8')
+            else:
+                encoded_condition_name = self.condition
+
+            if not validate_type_str(encoded_condition_name, should_raise=False):
+                raise InvalidDocumentException([
+                    InvalidTemplateException("An error occurred (ValidationError) when calling the CreateChangeSet "
```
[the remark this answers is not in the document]
keetonian: I think we don't even need the `"Error"` part- if you look in the error case test that @ jadhavmanoj wrote, the error message already shows that the template is invalid. I think we can just leave this as `"Every Condition member must be a string."`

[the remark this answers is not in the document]
jlhood: Agreed, that's better.

samtranslator/sdk/resource.py
```diff
@@ -30,7 +32,15 @@ def valid(self):
 
         :return: True, if the resource is valid
         """
-        # As long as the type is valid.
+        # As long as the type is valid and type string.
+        # validate the condition should be string
+
+        if self.condition:
+
+            if not isinstance(self.condition, string_types):
```
jlhood: Can we use `is_str()` instead? I'd just like to be consistent in how we do type checking across the codebase.

jadhavmanoj:
> Really appreciate you taking the time to fix this bug! We reviewed and have some minor comments, but in verifying our comments, we ended up making the changes ourselves so we'll go ahead and push those commits and get this merged in. Thanks again!

Thanks for the review & comments @jlhood and @keetonian

samtranslator/sdk/resource.py
```diff
@@ -30,7 +32,17 @@ def valid(self):
 
         :return: True, if the resource is valid
         """
-        # As long as the type is valid.
+        # As long as the type is valid and type string.
+        # validate the condition should be string
+
+        validate_type_str = is_str()
```
jlhood: Do we need this extra var? Can't we just use `is_str()` directly below?

tests/translator/input/error_function_with_invalid_condition_name.yaml
```diff
@@ -0,0 +1,17 @@
...
+  MyCondition:
+    Fn::Equals:
+      - true
+      - false
+
+Resources:
+  ConditionFunction:
+    Type: 'AWS::Serverless::Function'
+    Condition:
+      ref: MyCondition
```
jlhood: `ref` -> `Ref` so it's valid CloudFormation syntax.

keetonian:
@jadhavmanoj no problem! Thank you for fixing this!

PR #909, opened 2019-05-03
Add S3 filter example

charliejllewellyn:
It would be useful to have an example of a filter

*Issue #, if available:*

*Description of changes:*

*Checklist:*

- [ ] Write/update tests
- [ ] `make pr` passes
- [ ] Update documentation

### DIFF
--- a/versions/2016-10-31.md
+++ b/versions/2016-10-31.md
@@ -409,6 +409,11 @@ Type: S3
 Properties:
   Bucket: my-photo-bucket
   Events: s3:ObjectCreated:*
+  Filter:
+  S3Key:
+  Rules:
+    - Name: prefix|suffix
+      Value: my-prefix|my-suffix
 ```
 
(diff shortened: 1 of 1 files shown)


versions/2016-10-31.md
```diff
@@ -409,6 +409,11 @@ Type: S3
 Properties:
   Bucket: my-photo-bucket
   Events: s3:ObjectCreated:*
+  Filter:
+  S3Key:
```
jlhood: Thanks for adding this. However, you need to indent the fields. It should look like this:

```yaml
    Events: s3:ObjectCreated:*
    Filter:
      S3Key:
        Rules:
          - Name: prefix|suffix
            Value: my-prefix|my-suffix
```

charliejllewellyn: Sorry wasn't concentrating :/

Issue #652, opened 2018-11-06
Authorizers: Support AuthorizationScopes


keetonian:
**Description:**

This was brought up in Slack:

> Starting to work on a PoC using the new authorizer features in SAM and already notice one huge shortcoming I didn't think about.
> `AuthorizationScopes` - this would have been really powerful to include so I could define required scopes for authenticating tokens at the API level instead of inside the code.

This is also necessary for using `access_tokens`.

Docs: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-apigateway-method.html#cfn-apigateway-method-authorizationscopes

SAM should support `AuthorizationScopes`.

brysontyrrell:
This was my thread in the channel. My proposal would look something like this:

```
  MyApi:
    Type: AWS::Serverless::Api
    Properties:
      StageName: Prod
      Auth:
        DefaultAuthorizer: MyCognitoAuthorizer
        Authorizers:
          MyCognitoAuthorizer:
            UserPoolArn: !Ref CognitoUserPoolArn
              OAuthScopes:
                - scope1
                - scope2

  MyFunction:
    Type: AWS::Serverless::Function
    Properties:
      CodeUri: ./src
      Handler: lambda.handler
      MemorySize: 1024
      Runtime: nodejs8.10
      Events:
        Root:
          Type: Api
          Properties:
            RestApiId: !Ref MyApi
            Path: /
            Method: GET
            Auth:
              Authorizer: MyCognitoAuthorizer
              OAuthScopes:
                - scope3
```

This is a swagger doc from one of my templates to show what it looks like with scopes defined:

```
ApiGateway:
    Type: AWS::Serverless::Api
    Properties:
      StageName: Prod
      EndpointConfiguration: EDGE

      DefinitionBody:
        swagger: 2.0
        info:
          title: !Ref AWS::StackName
        securityDefinitions:
          CognitoAuth:
            type: apiKey
            name: Authorization
            in: header
            x-amazon-apigateway-authtype: cognito_user_pools
            x-amazon-apigateway-authorizer:
              type: cognito_user_pools
              providerARNs:
                - !Ref CognitoUserPoolArn

        paths:
          "/messages":
            post:
              x-amazon-apigateway-integration:
                httpMethod: post
                type: aws_proxy
                uri: !Sub arn:aws:apigateway:${AWS::Region}:lambda:path/2015-03-31/functions/${SendMessage.Arn}/invocations
[...]

PR #917, opened 2019-05-14
Add authorizationscopes

klmz:
*Issue #, if available:* 
#652 

 *Description of changes:* 
This adds support for authorisation scopes to the SAM spec. It is a small hanges to the swagger translator that add authorization scopes per method.

I have added an e2e test in the from of an example yaml file (`api_with_auth_with_scopes.yaml`) with all usage options. I also added swagger specific tests to `test_swagger.py`.

*Checklist:*

- [x] Write/update tests
- [x] `make pr` passes
- [x] Update documentation
- [x] Add swagger test

### DIFF
--- a/tests/translator/input/api_with_auth_with_default_scopes.yaml
+++ b/tests/translator/input/api_with_auth_with_default_scopes.yaml
@@ -1,0 +1,88 @@
+Resources:
+  MyApiWithCognitoAuth:
+    Type: "AWS::Serverless::Api"
+    Properties:
+      StageName: Prod
+      Auth:
+        DefaultAuthorizer: MyDefaultCognitoAuth
+        Authorizers:
+          MyDefaultCognitoAuth:
+            UserPoolArn: arn:aws:1
+            AuthorizationScopes:
+              - default.write
+              - default.read
+          MyCognitoAuthWithDefaultScopes:
+            UserPoolArn: arn:aws:2
+            AuthorizationScopes:
+              - default.delete
+              - default.update
+
+  MyFn:
+    Type: AWS::Serverless::Function
+    Properties:
+      CodeUri: s3://bucket/key
+      Handler: index.handler
+      Runtime: nodejs8.10
+      Events:
+        CognitoAuthorizerWithDefaultScopes:
+          Type: Api
+          Properties:
+            RestApiId: !Ref MyApiWithCognitoAuth
+            Method: get
+            Path: /cognitoauthorizerwithdefaultscopes
+            Auth:
+              Authorizer: MyCognitoAuthWithDefaultScopes
+        CognitoDefaultScopesDefaultAuthorizer:
+          Type: Api
+          Properties:
+            RestApiId: !Ref MyApiWithCognitoAuth
+            Method: get
+            Path: /cognitodefaultscopesdefaultauthorizer
+        CognitoWithAuthNone:
+          Type: Api
+          Properties:
+            RestApiId: !Ref MyApiWithCognitoAuth
+            Method: get
+            Path: /cognitowithauthnone
+            Auth:
+              Authorizer: NONE
+        CognitoDefaultScopesWithOverwritten:
+          Type: Api
+          Properties:
+            RestApiId: !Ref MyApiWithCognitoAuth
+            Method: get
+            Path: /cognitodefaultscopesoverwritten
+            Auth:
+              Authorizer: MyDefaultCognitoAuth
+              AuthorizationScopes: 
+                - overwritten.read
+                - overwritten.write
+        CognitoAuthorizerScopesOverwritten:
+          Type: Api
+          Properties:
+            RestApiId: !Ref MyApiWithCognitoAuth
+            Method: get
+            Path: /cognitoauthorizercopesoverwritten
+            Auth:
+              Authorizer: MyCognitoAuthWithDefaultScopes
+              AuthorizationScopes: 
+                - overwritten.read
+                - overwritten.write
+        CognitoDefaultScopesNone:
+          Type: Api
+          Properties:
+            RestApiId: !Ref MyApiWithCognitoAuth
+            Method: get
+            Path: /cognitodefaultscopesnone
+            Auth:
+              Authorizer: MyDefaultCognitoAuth
+              AuthorizationScopes: []
+        CognitoDefaultAuythDefaultScopesNone:
+          Type: Api
+          Properties:
+            RestApiId: !Ref MyApiWithCognitoAuth
+            Method: get
+            Path: /cognitodefaultauthdefaultscopesnone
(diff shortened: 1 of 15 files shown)


klmz:
Ok. I must admit that i dont know how. But I'm guessing it had something to do with linting. Once I made sure ```make pr``` passed the test worked on python 2 as well. Sorry for the noise, still have to right some test though. I will do this somewhere this week!

fyayc-ale:
@keetonian, sorry to ping you, would be possible to merge this? We are using a lambda to achieve the same and this will reduce errors on our side by a lot!

klmz:
> @keetonian, sorry to ping you, would be possible to merge this? We are using a lambda to achieve the same and this will reduce errors on our side by a lot!
Depends on how much you need this but you can pull this branch and run the transformation manually as described here: [Dev guide](https://github.com/awslabs/serverless-application-model/blob/master/DEVELOPMENT_GUIDE.rst#verifying-transforms)

If you ever worked with python and virtualenv you can have it done in minutes.

keetonian:
@fyayc-ale @klmz Will review today. Thank you for pinging me.

tests/translator/input/api_with_auth_with_scopes.yaml
```diff
@@ -0,0 +1,52 @@
+Resources:
+  MyApiWithCognitoAuth:
+    Type: "AWS::Serverless::Api"
+    Properties:
+      StageName: Prod
+      Auth:
+        DefaultAuthorizer: MyCognitoAuth
+        Authorizers:
+          MyCognitoAuth:
```
brettstack: Could you add `AuthorizationScopes` here also.

brettstack:
@klmz thanks! Looks like it's missing `AuthorizationScopes` at the Api level? This should set scopes on all methods which use this authorizer (and can be overridden at the method level)

klmz:
@brettstack I didn't do that because I tried to stay as close to the cloudformation spec. A default scope doesn't make a whole lot of sense to me. If it is default you can just as well leave the scope out I guess? Although I guess there are use cases where it could make sense. If you want most endpoints to be with a scope and just a couple without scopes. Or if you use the same token with different api's or something like that? Would like to hear you thoughts on this.
Actually, would also like your thoughts on my assumption of staying close to the CF spec. Is that something you strive for? Maybe this is more of general discussion for somewhere else 😅.

That being said, I guess there is some use for it so I expect to have some time to add the api level scopes this week.

brettstack:
Yes it's the scenario where you want most endpoints to have a certain configuration and you override the outliers.

eterjoost:
Lol. That was much more complex than I anticipated. I think I got the behavior down, but I need to write test still. I expect to have some time to finish this on Thursday!

brettstack:
Thanks @eterjoost we'll review when it's all ready. 👍

klmz:
I need some help with fixing the tests. Some tests are now failing, but it is very hard to see why in the output. Am I doing something wrong? Or is this how it works? Maybe there is a better way of using the tests instead of just running them all?

klmz:
It is an issue with the china and us-gov region. The output doesn't match the expected output, but I can't see what is different.

keetonian:
@klmz ok, we can look at it, thanks for bringing it up!

Michael1d:
Hi All,

This is a hugely important pull request. Having the ability to use scopes within a SAM template is the only way to implement access tokens in an enterprise scenario.

Can someone have a look, help @klmz to get to the end goal or provide some input to how we could fix this?

Thanks in advance,

Mike

leantorres73:
+1, This is a very important feature for implementing access_token and client_credentials grant type.

louism517:
+1 - the client_credentials flow actually _requires_ a scope for authentication.

MohammedAlSafwan:
Any progress on this issue?!

keetonian:
@leantorres73 @louism517 I'm not familiar with those features, could you provide some more information about them?

@Michael1d I resolved the merge conflicts locally, but had a couple of questions regarding this feature and the overall scope. Could you give some more information about access tokens and how you use/would like to use them?

Michael1d:
Hi @keetonian,

Apologies for not get back to you sooner. In my scenario, I have API's that have many potential "clients" - think an SPA, an analytics ETL process, a B2B customer etc. I have created Cognito Application Clients that represent those use cases, assigning a single scope to each one. Therefore, I have the requirement to add one or more scopes selectively to serverless functions deployed by SAM. I hope that makes sense.

On the standard CF AWS::ApiGateway::Method, there is a property for AuthorizationScopes and the same functionality is required here with a AWS::Serverless::Function in SAM. I also like the idea of being able to set default scopes on a AWS::Serverless::Api as discussed in here as well.

Please let me know if you have any other questions or clarifications.

keetonian:
Just pushed up a few changes to fully support this feature.

tests/translator/input/api_with_auth_with_default_scopes.yaml
```diff
@@ -0,0 +1,88 @@
...
+        CognitoDefaultScopesNone:
+          Type: Api
+          Properties:
+            RestApiId: !Ref MyApiWithCognitoAuth
+            Method: get
+            Path: /cognitodefaultscopesnone
+            Auth:
+              Authorizer: MyDefaultCognitoAuth
+              AuthorizationScopes: []
+        CognitoDefaultAuythDefaultScopesNone:
```
praneetap: typo

keetonian:
Scopes match the Swagger/OpenApi spec: https://swagger.io/docs/specification/authentication/

keetonian:
No APIGW extensions for scopes, they use the same implementation: https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-swagger-extensions-authorizer.html

keetonian:
Implementation/deployment verified for scopes changes.

faxad:
For some reason I couldn't get it to work.
I'm using SAM CLI, version 0.6.0

SAM deploy doesn't seem to give any error.
However Authorization and OAuth Scopes doesn't get set when I check on the console.

What am I missing?

```
  FooAPI:
    Type: AWS::Serverless::Api
    Properties:
      StageName: Prod
      Auth:
        DefaultAuthorizer: FooAPICognitoAuthorizer
        Authorizers:
          FooAPICognitoAuthorizer:
            Identity:
              Header: Authorization
            UserPoolArn: !Ref UserPoolArn
```

```
  FooFunction:
    Properties:
      CodeUri: package/deployment.zip
      Handler: app.foo_func
      Events:
        ApiEvent:
          Type: Api
          Properties:
            RestApiId: !Ref FooAPI
            Method: get
            Path: /
            Auth:
              Authorizer: FooAPICognitoAuthorizer
              AuthorizationScopes:
                - scope_x
```

Issue #191, opened 2017-08-31
SAM Creates stage named "Stage" by default?


sdole:
I have a template like this:

```
AWSTemplateFormatVersion: '2010-09-09'
Transform: 'AWS::Serverless-2016-10-31'
Resources:
  Api:
    Properties:
      StageName: dv01
      DefinitionUri: api_spec_SAM.json
    Type: AWS::Serverless::Api
```

After package and deploy with the cli, it creates an API with two stages named "Stage" and "dv01". In reality, I do not want to create either stages. The StageName property is "required". Not sure why. My intention is to create a separate set of templates to create the stages, deployments, lambda function versions etc. 

The reason I need separation is because we have multiple code pipelines that are owned by different people (dev, build teams). I would like these templates to live in separate git repos.

Is there a way to create an API without any stages? 

Else, the work around is to simply ignore this one default stage created via SAM. Thanks!

pauljbrice:
I am also seeing this issue, however I do require a stage set in the AWS::Serverless::Api resource but I am seeing a second stage created as "Stage". I validated in the "processed" cfn template the resource AWS::ApiGateway:Stage is referencing a AWS:::Api::Deployment but the stage name in the deployment is "Stage" not "qic" as set in the stage resource.

**SAM Template:**

ARSAPIGateway:
    Type: AWS::Serverless::Api
    Properties:
      DefinitionUri: apigateway/swagger.yaml
      StageName: qic
      Variables:
        LambdaFunctionName: !Ref LambdaFunction17

**Processed CFN Template:**

"APIGatewayDeploymenta727bb8729": {
      "Type": "AWS::ApiGateway::Deployment",
      "Properties": {
        "RestApiId": {
          "Ref": "APIGateway"
        },
        "Description": "RestApi deployment id: a727bb872967fd6e7bd621271f597376226329ee",
        **"StageName": "Stage"**
      }
    },
"APIGatewayStage": {
      "Type": "AWS::ApiGateway::Stage",
      "Properties": {
        "Variables": {
          "LambdaFunctionName": {
            "Ref": "LambdaFunction17"
          }
        },
        "RestApiId": {
          "Ref": "APIGateway"
        },
        "DeploymentId": {
          "Ref": "APIGatewayDeploymenta727bb8729"
        },
        **"StageName": "qic"**
      }
    },

hagoss01:
I am having a related problem.

[...]

Issue #627, opened 2018-10-16
CORS issue when using OAS3 spec for Serverless::Api


xscottx:
**Description:**

Tried to convert my API over tonight from swagger 2.0 to openapi specification 3.0, and got an error at the Cloudformation Changeset creation and SAM transformation step.  This might be due to the use of CORS in our template to modify the openapi file.

**Steps to reproduce the issue:**
1. Take a working API, deployed to a stage
2. goto Stages > Export > Select 'Export as OpenAPI 3'
3. Delete the extraneous 'options' methods in the file
4. Validate swagger syntax and format using swagger-cli v2.1.3
5. In the SAM template.yml, use CORS and DefinitionBody (ex below)
`
MyApi:
    Type: AWS::Serverless::Api
    Properties:
      ...
      Cors:
        AllowOrigin: "'*'"
        AllowMethods: "'*'"
        AllowHeaders: "'*"
      DefinitionBody:
          'Fn::Transform':
            Name: 'AWS::Include'
            Parameters:
              Location: !Sub 's3://path-to-oas/oas3.yml'`
6. Run command, aws cloudformation package etc
7. Should fail with the following

**Observed result:**
Failed to create the changeset: Waiter ChangeSetCreateComplete failed: Waiter encountered a terminal failure state Status: FAILED. Reason: Transform AWS::Serverless-2016-10-31 failed with: Invalid Serverless Application Specification document. Number of errors found: 1. Resource with id [MyApi] is invalid. Unable to add Cors configuration because 'DefinitionBody' does not contain a valid Swagger

**Expected result:**
Changeset to be created, ready for aws cloudformation deployment.

mikebroberts:
Yup, us too. Looks like https://github.com/awslabs/serverless-application-model/blob/master/samtranslator/swagger/swagger.py#L301 needs to be updated (thanks to @johnchapin for finding that)

scarlier:
any update on this issue?

hoegertn:
Anything new?

keetonian:
Adding a priority label to help us get this scheduled soon. Sorry for the wait.

I'm not personally familiar with OAS3; could someone direct me to something that explains the differences between it and Swagger so I can better understand what needs to change to support it? Can I still use swagger syntax and call it OAS3 instead of Swagger 2.0?

txase:
@keetonian I'm aware of only two impacts for SAM:

[...]

PR #932, opened 2019-05-23
feat: openapi version 3 support non-auth changes

praneetap:
*Issue #, if available:*
#191 #627 
*Description of changes:*
Adding OpenApi 3.0 support for non-auth use cases. This adds a global/api flag called OpenApiVersion, which when specified - 
1. with value 2.0 or 3.0, fixes the issue #191 
2. With value 3.0, changes the swagger to openapi 3.0 format. 

Added tests, and end to end tests.
*Checklist:*

- [x] Write/update tests
- [x] `make pr` passes
- [x] Update documentation


praneetap:
#191

praneetap:
#627

samtranslator/model/api/api_generator.py
```diff
@@ -93,6 +96,10 @@ def _construct_rest_api(self):
             raise InvalidResourceException(self.logical_id,
                                            "Specify either 'DefinitionUri' or 'DefinitionBody' property and not both")
 
+        if self.open_api_version and self.open_api_version not in OpenApiVersionsSupported:
```
keetonian: So I think people often use the patch version for OpenApi as well (ie '3.0.0' or '3.0.1'); I think we just want to verify that they're using some version of OpenApi 3 and not care about the minor/patch versions.

samtranslator/model/api/api_generator.py
```diff
@@ -147,7 +154,8 @@ def _construct_deployment(self, rest_api):
         deployment = ApiGatewayDeployment(self.logical_id + 'Deployment',
                                           attributes=self.passthrough_resource_attributes)
         deployment.RestApiId = rest_api.get_runtime_attr('rest_api_id')
-        deployment.StageName = 'Stage'
+        if not self.open_api_version:
+            deployment.StageName = 'Stage'
```
keetonian: I am so glad we are finally getting rid of this.

I proposed also getting rid of it if we detect that they're using OpenApi 3.x in their swagger (but don't have the OpenApi flag set), are we going ahead with that idea or not?

samtranslator/model/apigateway.py
```diff
@@ -20,7 +20,8 @@ class ApiGatewayRestApi(Resource):
             'Parameters': PropertyType(False, is_type(dict)),
             'EndpointConfiguration': PropertyType(False, is_type(dict)),
             "BinaryMediaTypes": PropertyType(False, is_type(list)),
-            "MinimumCompressionSize": PropertyType(False, is_type(int))
+            "MinimumCompressionSize": PropertyType(False, is_type(int)),
+            'OpenApiVersion': PropertyType(False, is_str())
```
keetonian: Remove this line. This is the ApiGateway resource (maps directly to the CFN equivalent). The OpenApiVersion flag only needs to be added to the `Serverless::Api` resource found in the `sam_resources.py` file.

samtranslator/model/apigateway.py
```diff
@@ -39,7 +40,7 @@ class ApiGatewayStage(Resource):
             'DeploymentId': PropertyType(True, is_str()),
             'Description': PropertyType(False, is_str()),
             'RestApiId': PropertyType(True, is_str()),
-            'StageName': PropertyType(True, one_of(is_str(), is_type(dict))),
+            'StageName': PropertyType(False, one_of(is_str(), is_type(dict))),
```
keetonian: Why did you change this to `False`?

samtranslator/swagger/swagger.py
```diff
@@ -554,10 +554,13 @@ def is_valid(data):
...
         """
-        return bool(data) and \
-            isinstance(data, dict) and \
-            bool(data.get("swagger")) and \
-            isinstance(data.get('paths'), dict)
+        if bool(data) and isinstance(data, dict) and isinstance(data.get('paths'), dict):
+            if bool(data.get("swagger")):
+                return True
+            elif bool(data.get("openapi")):
+                return "3.0" in data["openapi"]
```
keetonian: Same callout about OpenAPI versions (3.0.0 should be supported, I think)

praneetap: 3.0.0 is supported, I have a test for it..

samtranslator/model/apigateway.py
praneetap: I followed Sanath's previous change for stage stage fix. If this is required I will have to set it, which is not what we want.

samtranslator/model/api/api_generator.py
praneetap: Do you recommend adding a regex here?

samtranslator/model/api/api_generator.py
praneetap: Unfortunately there are cases where we don't read through the swagger document (definitionUri) so I don't think that is a consistent experience.

brettstack: Could you elaborate? We should be able to get rid of StageStage if they specify **any** value for `OpenApi` (i.e only if they don't specify a value do we keep it).

samtranslator/swagger/swagger.py
keetonian: Oh yeah, this will do string matching. `"3.0"` is inside of `"3.0.0"`. 👍

samtranslator/model/apigateway.py
keetonian: I think (but am not sure) that we still want this StageName property set, but not the other one that is on the `Deployment` resource. Either way, this will still work (we're still setting the StageName property on the Stage resource and conditionally removing it from the Deployment resource), so this is more of a technicality than a requirement.

samtranslator/model/api/api_generator.py
keetonian: That would work

samtranslator/model/apigateway.py
brettstack: I'd love to make this optional. Is there any reason not to?

samtranslator/model/apigateway.py
praneetap: will do thanks!

samtranslator/model/apigateway.py
brettstack: If we decide to make this optional, we need to decide on a default value ("prod"?) and update docs.

praneetap: I think that should be a part of another cr. I don't want to stray too far from Sanath's changes for this bug previously.

praneetap: I removed this change thanks Keeton!

samtranslator/model/api/api_generator.py
praneetap: Had an offline conversation about this, we will only fix stage stage if this flag is specified. If they don't have it specified and use oas 3 swagger in their api definitionbody it will still create an additional stage. This is because we want to remain backwards compatible and dont want to fix the stage stage bug without the customer opting in (with use of the OpenApiVersion flag)

samtranslator/swagger/swagger.py
```diff
@@ -554,10 +555,15 @@ def is_valid(data):
         :param dict data: Data to be validated
         :return: True, if data is a Swagger
         """
-        return bool(data) and \
-            isinstance(data, dict) and \
-            bool(data.get("swagger")) and \
-            isinstance(data.get('paths'), dict)
+        openapi_version_supported_regex = r"\A3(\.\d)(\.\d)?$"
```
keetonian: Could you consolidate these regexes? It would be easier to maintain them in case we need to update the version in the future. I've seen two different ones, but one of them is used many more times than the other.

samtranslator/model/api/api_generator.py
```diff
@@ -93,6 +97,11 @@ def _construct_rest_api(self):
             raise InvalidResourceException(self.logical_id,
                                            "Specify either 'DefinitionUri' or 'DefinitionBody' property and not both")
 
+        if self.open_api_version:
+            if re.match(OpenApiVersionsSupportedRegex, self.open_api_version) is None:
+                raise InvalidResourceException(
+                    self.logical_id, "The OpenApiVersion value must be of the format 2.x.x or 3.x.x")
```
brettstack: "2.x.x" isn't valid in that regex, right? This is confusing because "2.x.x" is a commonly accepted value for semantic versions.

praneetap: 2.x.x is a valid value for the flag. This is the regex - r"\A[2-3](\.\d)(\.\d)?$"

samtranslator/swagger/swagger.py
praneetap: Where would the right place to consolidate be?

docs/cloudformation_compatibility.rst
```diff
@@ -171,6 +171,7 @@ BinaryMediaTypes                    All
 MinimumCompressionSize              All
 Cors                                All
 TracingEnabled                      All
+OpenApiVersion                      All
```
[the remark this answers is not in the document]
praneetap: Not sure if that is necessary? Can Brett comment here?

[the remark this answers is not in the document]
brettstack: Preference is to support; otherwise we need to change this to "None"

samtranslator/model/api/api_generator.py
brettstack: 'x' isn't a digit/number though. Maybe there's some Python thing I'm missing here.

docs/cloudformation_compatibility.rst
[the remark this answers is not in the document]
praneetap: I am not sure yet whether this is necessary right now, so I am going to change this to None and we can tackle it in a separate task?

samtranslator/model/api/api_generator.py
praneetap: Got it, resolved. Thanks!

samtranslator/model/api/api_generator.py
```diff
@@ -98,9 +97,9 @@ def _construct_rest_api(self):
                                            "Specify either 'DefinitionUri' or 'DefinitionBody' property and not both")
 
         if self.open_api_version:
-            if re.match(OpenApiVersionsSupportedRegex, self.open_api_version) is None:
+            if re.match(SwaggerEditor.get_openapi_versions_supported_regex(), self.open_api_version) is None:
                 raise InvalidResourceException(
-                    self.logical_id, "The OpenApiVersion value must be of the format 2.x.x or 3.x.x")
+                    self.logical_id, "The OpenApiVersion value must be 3.0.0")
```
brettstack: "must be of the format 3.0.0" is fine; it's just that "2.x.x" was ambiguous.

praneetap: done. thanks!

samtranslator/model/api/api_generator.py
```diff
@@ -93,6 +96,11 @@ def _construct_rest_api(self):
             raise InvalidResourceException(self.logical_id,
                                            "Specify either 'DefinitionUri' or 'DefinitionBody' property and not both")
 
+        if self.open_api_version:
+            if re.match(SwaggerEditor.get_openapi_versions_supported_regex(), self.open_api_version) is None:
+                raise InvalidResourceException(
+                    self.logical_id, "The OpenApiVersion value must be of the format 3.0.0")
```
keetonian: nit: `3.0.0` doesn't seem like a format, but I think this is still understandable.

praneetap:
> Looks good. Have you deployed any of these OpenApi 3.0 templates to verify that they work in CFN?
@keetonian 
I could deploy this successfully
```
Globals:
  Api:
    OpenApiVersion: '3.0.1'

Resources:
  ImplicitApiFunction:
    Properties:
      CodeUri: s3://praneep-test-2/thumbnails.zip
      Handler: index.gethtml
      Runtime: nodejs8.10
      Events:
        GetHtml:
          Type: Api
          Properties:
            Path: /
            Method: get
```

Issue #547, opened 2018-08-10
Add API Key capabilities to AWS::Serverless::Api


mariano-calandra-xp:
Hi, I was thinking about API key capabilities to `AWS::Serverless::Api` resource. 

Something like:

```
  MyRestApi:
    Type: AWS::Serverless::Api
    Properties:
      ...
      ApiKeys: 
        - Enabled: true 
          Name: myFirstApiKey
          UsagePlanId: !Ref MyUsagePlan

  MyUsagePlan: 
    Type: AWS::Serverless::ApiUsagePlan
    Properties:
      Name: String
      Description: String
      Quota: QuotaSettings
      Throttle: ThrottleSettings
```

That will be translated into:

```
  MyRestApi:
    Type: AWS::Serverless::Api
    ...

  MyFirstApiKey: 
    Type: AWS::ApiGateway::ApiKey
    DependsOn: 
      - MyUsagePlan
    Properties: 
      Name: MyFirstApiKey
      Enabled: "true"
      StageKeys: 
        - RestApiId: !Ref MyRestApi
          StageName: !Ref MyRestApi.Stage

  MyUsagePlan:
    Type: AWS::ApiGateway::UsagePlan
    DependsOn:
      - MyRestApi
    Properties:
      ApiStages:
      - ApiId: !Ref MyRestApi
        Stage: !Ref MyRestApi.Stage
      UsagePlanName: MyUsagePlan
      Quota: QuotaSettings
      Throttle: ThrottleSettings

  MyUsagePlanKey:
    Type: AWS::ApiGateway::UsagePlanKey
    Properties : 
      KeyId: !Ref MyFirstApiKey
      UsagePlanId: !Ref MyUsagePlan
      KeyType: API_KEY
```

`AWS::Serverless::ApiUsagePlan` could be optional with a default. 

The current way do the job but it's a little bit more verbose

Does it sound?

brettstack:
Absolutely! We've taken our first steps into Auth with https://github.com/awslabs/serverless-application-model/pull/546 and plan on following up with API Keys, AWS_IAM, and Resource Policies. Thanks for the suntan proposal; I'll tag you when I have an RFC up

brettstack:
Note I added some instructions in #802 for anyone who wants to submit a PR the instructions should be similar.

mkamioner:
Any update on this?

jlhood:
[...]

PR #943, opened 2019-05-31
Add ApiKey Auth support

### LINKED ISSUE #867: Rebase and refactor ApiKeyRequired feature
This feature (https://github.com/awslabs/serverless-application-model/pull/444/files) was merged into a feature branch (https://github.com/awslabs/serverless-application-model/tree/feat/auth) while we worked on the design for Auth as a whole. Now that we have that, we should rebase this change and refactor it to use the `Auth` property so that we can get this merged into develop.

cfbarbero:
*Issue #, if available:*
Fixes #867 

*Description of changes:*
Update the `Auth` property to include a new sub-property `ApiKeyRequired: <boolean>`. This will be supported on both the [Api Auth Method](https://github.com/cfbarbero/serverless-application-model/blob/master/versions/2016-10-31.md#api-auth-object) and the [Function Auth Method](https://github.com/cfbarbero/serverless-application-model/blob/master/versions/2016-10-31.md#function-auth-object). The function level specification will override the API level.

For example:
```yaml
Description: API Gateway with ApiKey Auth
Resources:
  MyApi:
    Type: AWS::Serverless::Api
    Properties:
      StageName: Prod
      Auth:
        ApiKeyRequired: true

  MyFunction:
    Properties:
      CodeUri: .
      Handler: index.handler
      Runtime: nodejs8.10
      Events:
        ApiKeyFalse:
          Type: Api
          Properties:
            RestApiId: !Ref MyApi
            Path: /
            Method: get
            # This is a public endpoint
            Auth:
              ApiKeyRequired: false
        ApiKeyTrue:
          Type: Api
          Properties:
            RestApiId: !Ref MyApi
            Path: /
            Method: get
            Auth:
              ApiKeyRequired: true
        ApiKeyDefault:
          Type: Api
          Properties:
            RestApiId: !Ref MyApi
            Path: /other
            Method: get
```

This is roughly based upon #444.  It implements part of #547. 

In implementing this change, I refactored the support for the Authorizers in the [swagger.py](samtranslator/swagger/swagger.py) file. ~~By doing this, it changed the order in which some of the generated swagger properties are translated and there by required updates to the following translator test output files:~~
- ~~tests/translator/output/api_with_aws_iam_auth_overrides.json~~
- ~~tests/translator/output/api_with_default_aws_iam_auth.json~~

Added additional translator tests:
- tests/translator/input/api_with_apikey_default_override.yaml
- tests/translator/input/api_with_apikey_required.yaml
Updated existing translator tests:
- tests/translator/input/globals_for_api.yaml
- tests/translator/input/implicit_api_with_auth_and_conditions_max.yaml

*Checklist:*

- [x] Write/update tests
- [x] `make pr` passes
- [x] Update documentation

### DIFF
--- a/samtranslator/model/api/api_generator.py
+++ b/samtranslator/model/api/api_generator.py
@@ -21,7 +21,8 @@ CorsProperties = namedtuple("_CorsProperties", ["AllowMethods", "AllowHeaders", 
 CorsProperties.__new__.__defaults__ = (None, None, _CORS_WILDCARD, None, False)
 
 AuthProperties = namedtuple("_AuthProperties",
-                            ["Authorizers", "DefaultAuthorizer", "InvokeRole", "AddDefaultAuthorizerToCorsPreflight", "ApiKeyRequired"])
+                            ["Authorizers", "DefaultAuthorizer", "InvokeRole", "AddDefaultAuthorizerToCorsPreflight",
+                             "ApiKeyRequired"])
 AuthProperties.__new__.__defaults__ = (None, None, None, True, None)
 
 GatewayResponseProperties = ["ResponseParameters", "ResponseTemplates", "StatusCode"]
@@ -310,7 +311,7 @@ class ApiGenerator(object):
         if authorizers:
             swagger_editor.add_authorizers_security_definitions(authorizers)
             self._set_default_authorizer(swagger_editor, authorizers, auth_properties.DefaultAuthorizer,
-                                auth_properties.AddDefaultAuthorizerToCorsPreflight)
+                                         auth_properties.AddDefaultAuthorizerToCorsPreflight)
 
         if auth_properties.ApiKeyRequired:
             swagger_editor.add_apikey_security_definition()
--- a/samtranslator/model/api/api_generator.py
+++ b/samtranslator/model/api/api_generator.py
@@ -116,6 +116,8 @@ class ApiGenerator(object):
         if self.definition_uri:
             rest_api.BodyS3Location = self._construct_body_s3_dict()
         elif self.definition_body:
+            # # Post Process OpenApi Auth Settings
+            self.definition_body = self._openapi_auth_postprocess(self.definition_body)
             rest_api.Body = self.definition_body
 
         if self.name:
@@ -318,8 +320,8 @@ class ApiGenerator(object):
             self._set_default_apikey_required(swagger_editor)
 
         # Assign the Swagger back to template
-
-        self.definition_body = self._openapi_auth_postprocess(swagger_editor.swagger)
+        # self.definition_body = self._openapi_auth_postprocess(swagger_editor.swagger)
+        self.definition_body = swagger_editor.swagger
 
     def _openapi_auth_postprocess(self, definition_body):
         """
--- a/samtranslator/model/api/api_generator.py
+++ b/samtranslator/model/api/api_generator.py
@@ -21,8 +21,8 @@ CorsProperties = namedtuple("_CorsProperties", ["AllowMethods", "AllowHeaders", 
 CorsProperties.__new__.__defaults__ = (None, None, _CORS_WILDCARD, None, False)
 
 AuthProperties = namedtuple("_AuthProperties",
-                            ["Authorizers", "DefaultAuthorizer", "InvokeRole", "AddDefaultAuthorizerToCorsPreflight"])
-AuthProperties.__new__.__defaults__ = (None, None, None, True)
+                            ["Authorizers", "DefaultAuthorizer", "InvokeRole", "AddDefaultAuthorizerToCorsPreflight", "ApiKeyRequired"])
+AuthProperties.__new__.__defaults__ = (None, None, None, True, None)
 
 GatewayResponseProperties = ["ResponseParameters", "ResponseTemplates", "StatusCode"]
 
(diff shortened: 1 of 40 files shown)


brettstack:
>  it changed the order in which some of the generated swagger properties

Please try to maintain the Swagger integrity as SAM will create an APIGW Deployment when it detects Swagger has changed and we strongly try not to do an APIGW Deployment unless the user actually made a change to their API.

cfbarbero:
@brettstack 
 I have fixed the bug that caused the AWS_IAM security definition to get reordered.  I have reverted the test output files for the 2 tests (across all 3 regions) that I had changed to accomodate the updated ordering and they are passing.

Let me know if there's anything else you'd like me to address.

samtranslator/model/eventsources/push.py
```diff
@@ -562,6 +562,16 @@ def _add_swagger_integration(self, api, function):
                             'is only a valid value when a DefaultAuthorizer on the API is specified.'.format(
                                 method=self.Method, path=self.Path))
 
+            apikey_required_setting = self.Auth.get('ApiKeyRequired')
+            apikey_required_setting_is_false = apikey_required_setting is not None and not apikey_required_setting
+            if apikey_required_setting_is_false and not api_auth.get('ApiKeyRequired'):
```
praneetap: Throwing an exception here is not required, as this would be a no-op on api gw.

samtranslator/model/api/api_generator.py
```diff
@@ -274,9 +274,13 @@ def _add_auth(self):
...
-            swagger_editor.add_authorizers(authorizers)
+            swagger_editor.add_authorizers_security_definitions(authorizers)
             self._set_default_authorizer(swagger_editor, authorizers, auth_properties.DefaultAuthorizer)
 
+        if auth_properties.ApiKeyRequired:
+            swagger_editor.add_apikey_security_definition()
+            self._set_default_apikey_required(swagger_editor)
+
         # Assign the Swagger back to template
         self.definition_body = swagger_editor.swagger
```
praneetap: Could you please rebase and add/update a test to ensure this works with openapi 3 as well.

samtranslator/model/eventsources/push.py
cfbarbero: Like the Authorizers implementation, I am storing the explicit ApiKeyRequired: False for a function in the security block for that function as `api_key_false`. 
```
"security": { "api_key_false": [] }
```
This security option gets cleaned up by the swagger.set_path_default_apikey_required method, however, this is only called when there is a default ApiKeyRequired setting on the Serverless API.  If there is no default setting (or no explicit Serverless API) then we never loop back and cleanup the `api_key_false` security setting.

I assumed we didn't want to spit out this security setting, however, I'll defer to your judgement.  I'm also open to other ideas on this.

samtranslator/model/api/api_generator.py
cfbarbero: @praneetap , I have rebased.  I am happy to add a test for openapi 3.  Can you provide a little direction on that?  Are there other tests I can refer to as an example?

cfbarbero: @praneetap , I've added a couple of openapi3 translator tests.  Please let me know if there's anything else you'd like to see on this.

New Test:
- api_with_apikey_required_openapi_3

Updated Tests:
- api_with_auth_all_maximum_openapi_3
- api_with_auth_all_maximum

praneetap: Yes, will be happy to! OpenApi tests are run [here](https://github.com/awslabs/serverless-application-model/blob/develop/tests/translator/test_translator.py#L345).
You can find more examples of input templates [here](https://github.com/awslabs/serverless-application-model/tree/develop/tests/translator/input) if you search for "openapi".

samtranslator/model/api/api_generator.py
```diff
@@ -108,6 +109,8 @@ def _construct_rest_api(self):
 
         self._add_cors()
         self._add_auth()
+        # Post Process OpenApi Auth Settings
+        self.definition_body = self._openapi_auth_postprocess(self.definition_body)
```
cfbarbero: @praneetap In order to support the ApiKeyRequired setting at the Function level (without requiring an explicit setting on the API), I had to move the `_openapi_auth_postprocess` method call up a level to ensure it was called on every template reqardless of if there's an explicit API Auth setting.  If you have another suggestion I'm happy to change.

cfbarbero:
@praneetap I have addressed your PR comments, this is ready for review again.

samtranslator/model/api/api_generator.py
```diff
@@ -115,6 +116,8 @@ def _construct_rest_api(self):
         if self.definition_uri:
             rest_api.BodyS3Location = self._construct_body_s3_dict()
         elif self.definition_body:
+            # # Post Process OpenApi Auth Settings
+            self.definition_body = self._openapi_auth_postprocess(self.definition_body)
```
cfbarbero: @praneetap In order to support the ApiKeyRequired setting at the Function level (without requiring an explicit setting on the API), I had to move the _openapi_auth_postprocess method call up a level to ensure it was called on every template reqardless of if there's an explicit API Auth setting. If you have another suggestion I'm happy to change.

jlhood:
@emonhaider We are working on the v1.14.0 release now. We can't give exact dates, but it's coming soon.

PR #949, opened 2019-06-03
feat: openapi support for auth usecases

praneetap:
*Issue #, if available:*
#191 #627
*Description of changes:*
Adding OpenApi 3.0 support for auth use cases. This adds a global/api flag called OpenApiVersion, which when specified as 3.. adds securitySchemes as per OpenApiSpec 3.
End to end tests,
unit tests
deployed locally transformed template on Cfn
*Checklist:*

- [x] Write/update tests
- [x] `make pr` passes
- [x] Update documentation


samtranslator/model/api/api_generator.py
```diff
@@ -287,7 +287,34 @@ def _add_auth(self):
...
+        If the definition body contains "openapi" and not "swagger", we enforce that the OpenApiVersion
+        flag is specified for cases with auth.
+        """
+        if definition_body.get('swagger') is not None:
+            return definition_body
+
+        if definition_body.get('openapi') is not None:
+            if self.open_api_version is None:
+                raise InvalidResourceException(
+                    self.logical_id, "OpenApiVersion must be specified")
```
brettstack: Minor error message improvement: ", e.g. OpenApiVersion: {definition_body.get('openapi')}"

samtranslator/plugins/globals/globals.py
```diff
@@ -129,7 +129,7 @@ def fix_openapi_definitions(cls, template):
         :param dict template: SAM template
         :return: Modified SAM template with corrected swagger doc matching the OpenApiVersion.
         """
-        resources = template["Resources"]
+        resources = template.get("Resources") or {}
```
brettstack: I think you can just pass `{}` as the second parameter to `.get`

samtranslator/model/api/api_generator.py
```diff
@@ -287,7 +287,34 @@ def _add_auth(self):
...
+        and openapi defined in the definition body, we return if there is "swagger" defined in the
+        definition body.
+
+        If the definition body contains "openapi" and not "swagger", we enforce that the OpenApiVersion
+        flag is specified for cases with auth.
+        """
+        if definition_body.get('swagger') is not None:
+            return definition_body
+
+        if definition_body.get('openapi') is not None:
```
keetonian: Some thoughts about this section:

I know we discussed that some people might be adding in `Swagger: '2.0'` as well as `openapi: 3.0.1`  to their templates right now to get around SAM's swagger check. If it turns out that enough people are using this workaround, we want to ship this section of the code.

However, I don't agree with that workaround (and I am also not sure if it actually works). I'll try to verify if it works, and, if it does and it appears as if customers are using that workaround, we can ship this code. If not, I'd rather make the validation in this section tighter so that SAM is a little more explicit about what is supported and what isn't.

praneetap: @keetonian @brettstack If we can ship with this validation in place and later remove it if unnecessary, I would much rather do that. Not sure what you mean by "little more explicit about what is supported and what isn't." @keetonian

keetonian: That's fair.

tests/translator/input/error_api_with_openapi_definition_body_no_flag.yaml
```diff
@@ -0,0 +1,58 @@
...
+          Properties:
+            Path: /
+            Method: get
+
+  ExplicitApi:
+    Type: AWS::Serverless::Api
+    Properties:
+      StageName: SomeStage
+      DefinitionBody:
+        openapi: 3.1.1
```
keetonian: I would rather that this just worked instead of being an error. This is absolutely valid syntax.

@brettstack had an earlier comment:
> Currently we fail if we don't detect "swagger," however, we know that some developers add both "openapi" and "swagger" to get around this. Do we think we will potentially break customers who do this workaround? If so, we should fail with error message saying they need to specify OpenApiVersion 3 (also add code comment with this reason)

This goes along with my other comment. This doesn't change the existing behavior, so I guess we can change this later if it seems like we can. As a developer though, if I see that "SAM Supports OpenApi 3", my immediate thought is that I can just use OpenApi 3 in the normal way.

praneetap: I agree. I am onboard with fixing this so we treat having "openapi" in the def-body without the OpenApiVersion flag specified as valid syntax

praneetap: Fixed!

samtranslator/model/api/api_generator.py
```diff
@@ -287,7 +287,29 @@ def _add_auth(self):
...
         # Assign the Swagger back to template
-        self.definition_body = swagger_editor.swagger
+
+        self.definition_body = self._openapi_auth_postprocess(swagger_editor.swagger)
+
+    def _openapi_auth_postprocess(self, definition_body):
+        """
+        Convert auth components to openapi 3 in definition body if OpenApiVersion flag is specified.
+
+        If there is swagger defined in the definition body, we treat it as a swagger spec and do not
```
keetonian: nit: `If "swagger" is defined... we use the swagger spec and do not...`

samtranslator/model/api/api_generator.py
```diff
@@ -287,7 +287,29 @@ def _add_auth(self):
...
+        if definition_body.get('openapi') is not None:
+            if self.open_api_version is None:
+                self.open_api_version = definition_body.get('openapi')
+
+        if self.open_api_version and re.match(SwaggerEditor.get_openapi_version_3_regex(), self.open_api_version):
+            if definition_body.get('securityDefinitions'):
+                definition_body['components'] = {}
+                definition_body['components']['securitySchemes'] = definition_body['securityDefinitions']
+                del definition_body['securityDefinitions']
+        return definition_body
```
keetonian: Next question: should we error on an `else` statement here? (using openapi and not swagger, and the openapi version doesn't match a version we support).

Or is this already caught elsewhere?

tests/translator/input/api_with_swagger_and_openapi_with_auth.yaml
```diff
@@ -0,0 +1,59 @@
...
+            Path: /
+            Method: get
+
+  ExplicitApi:
+    Type: AWS::Serverless::Api
+    Properties:
+      StageName: SomeStage
+      DefinitionBody:
+        openapi: 3.1.1
+        swagger: 2.0
```
keetonian: If a user wants to do this, we can let them, and allow them to deal with whatever ApiGateway thinks about it. I think this will be fine (but we definitely don't recommend doing anything of the sort).

Issue #931, opened 2019-05-22
RFC: Allow customizations on AWS::Serverless::Function API event types


beck3905:
Swagger seems to be an all-or-nothing approach to AWS::Serverless::Api resources.  If the swagger definition is provided in the API then the entire API needs to defined in swagger.  I often have a single AWS::Serverless::Function that needs some customization in how it should be defined in the API, but in order to accomplish that I would need to provide boilderplate swagger for the entire API plus my customization.

It would be much more convenient and simple if I could just add my customizations to the Lambda API event and the translator would apply these to the API definition for that specific function.

keetonian:
@beck3905 I like this idea. Like you posted, there are several features in https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-apigateway-method.html that could be useful in API Events.

Some of these issues are:
* Authorization Scopes (#652)
* Request Models (#895)
* API Key Required (and usage plans) (#248)

Could you provide any sample syntax of how you might like to use some of these features? Or any other features that you think would be good to add to the API Event?

beck3905:
@keetonian I recently had a use case where I had an API with caching enabled that requires an Authorization token.  Because the caching only looks at the URL and not the headers the cache would return valid values for users with invalid tokens and would return data for other users.  Obviously, this was a problem.  I solved it by adding the Authorization header parameter to the Method Request and enabling caching on that parameter.  Since Serverless::API is pretty much all or nothing with swagger, the only way to accomplish this was with swagger.  However, if the https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-apigateway-method.html#cfn-apigateway-method-requestparameters were available in the API event then I would've only had to add a small bit of configuration there to accomplish the same thing.

I suggest exposing the following properties in the API event source type:

- ApiKeyRequired
- AuthorizationScopes
- RequestModels (linked to Models defined in the Serverless::API resource)
- RequestParameters
- RequestValidatorId
- ResponseModels (mapping of status code to model and linked to Models defined in the Serverless::API resource)
- ResponseParameters (mapping of status code to parameters)

For example:

```
Resource:
  Api:
    Type: AWS::Serverless::Api
    Properties:
      Models: # use swagger format to define models. This gets passed through to the `definitions` section of swagger.
        User:   
          type: object
          required:
            - userName
          properties:
            userName:
              type: string
            firstName:
              type: string
            lastName:
              type: string
     ...

  Function:
[...]

PR #953, opened 2019-06-05
Add RequestParameters Support

beck3905:
*Issue #, if available:*
#931

*Description of changes:*
Added support for `RequestParameters` in the `Api` event source properties

unit tests
end to end tests
ran transform locally and ran created a stack from the transformed template

*Checklist:*

- [x] Write/update tests
- [x] `make pr` passes
- [x] Update documentation


beck3905:
I'm not really sure what needs to be done to update documentation.  I couldn't see in the repo how the documentation gets generated in order to update it.

beck3905:
@keetonian My use case for request parameters is to include an Authorization header in my cache key.  In order to accomplish this with API Gateway, I have to defined a parameter in the Method Request of the method and then add it to the `cacheKeyParameters` of the Method Integration.

I listed Response Parameters in the initial issue, however I don't personally have a use case for this yet so I haven't looked at an implementation yet.

beck3905:
@keetonian In regard to Lambda Proxy integration, I think you can still use request parameters.  For example, using SAM if I specify my path in the Api event source as `path/{id}` the sam translator will automatically add a path request parameter and that will be available in the event that is passed in via Lambda Proxy integration to my lambda function.  There is no way to provide a mapping template to transform parameters, but they will be available to the lambda function.

beck3905:
@praneetap @keetonian I have updated this PR with examples and documentation.  Would you please take a look?

beck3905:
@keetonian @praneetap Just bumping this again to get your attention.This is a feature my team and I are waiting for so we don't have to use standard CloudFormation templates and can continue to use SAM for our use cases.  Thanks.

praneetap:
Sorry for the delays on our end, the changes look good. Just want to make sure you have verified that  transformed template deploys and application functions as expected? Please try a template with OpenApiVersion 3 as well, since you are modifying swagger.

beck3905:
@praneetap Thanks for taking a look.  I did deploy a transformed template and everything worked as expected.  I did both swagger and OpenApiVersion 3.

beck3905:
@praneetap @jlhood @keetonian It appears that the latest version of `cfn-lint==0.22.0` is breaking 4 tests in my branch.  These tests are also broken in the `develop` branch.  I'm not sure if I should pin `requirements/dev.txt` to `cfn-lint==0.21.6`, which does pass, and I'm not sure how best to fix the failures as they are related to `AWS::Cognito::UserPools` present in tests templates that have nothing to do with the feature I added.  Any suggestions?

beck3905:
The build is failing due to this issue with cfn-lint v0.22.0 which is also impacting the `develop` branch: https://github.com/awslabs/serverless-application-model/issues/1004.  I've noticed several commits in the develop branch that are also failing tests due to this issue.

beck3905:
@keetonian @praneetap @jlhood Looks like the build issue has been fixed.  Any chance of getting the last review?

samtranslator/model/eventsources/push.py
```diff
@@ -600,6 +603,56 @@ def _add_swagger_integration(self, api, function):
...
+            for parameter in self.RequestParameters:
+
+                if isinstance(parameter, dict):
+
+                    parameter_name, parameter_value = next(iter(parameter.items()))
+
+                    if not re.match('method\.request\.(query|path|header)', parameter_name):
+                        raise InvalidEventException(
+                            self.relative_id,
+                            "Invalid value for 'RequestParameters' property")
```
brettstack: Add ". Keys must be in the format 'method.request.[query|path|header].{value}', e.g 'method.request.header.Authorization'."

samtranslator/model/eventsources/push.py
```diff
@@ -600,6 +603,56 @@ def _add_swagger_integration(self, api, function):
...
+
+                    if not isinstance(parameter_value, dict) or not all(key in RequestParameterProperties
+                                                                        for key in parameter_value.keys()):
+                        raise InvalidEventException(
+                            self.relative_id,
+                            "Invalid value for 'RequestParameters' property")
```
brettstack: Add ". Values must be an object, e.g { Required: true, Caching: false }"

samtranslator/model/eventsources/push.py
```diff
@@ -600,6 +603,56 @@ def _add_swagger_integration(self, api, function):
...
+                    settings = default_value.copy()
+                    settings.update(parameter_value)
+                    settings.update({'Name': parameter_name})
+
+                    parameters.append(settings)
+
+                elif not isinstance(parameter, string_types):
+                    raise InvalidEventException(
+                        self.relative_id,
+                        "Invalid value for 'RequestParameters' property")
```
brettstack: Add ". Property must be either a string or an object"

samtranslator/model/eventsources/push.py
```diff
@@ -600,6 +603,56 @@ def _add_swagger_integration(self, api, function):
...
+
+                else:
```
brettstack: Can we swap this and the previous condition around, so check if it's a string, and then the final else (no elif) will be the case that it's not an object or a string.

samtranslator/model/eventsources/push.py
```diff
@@ -600,6 +603,56 @@ def _add_swagger_integration(self, api, function):
...
+                    if not re.match('method\.request\.(query|path|header)', parameter):
+                        raise InvalidEventException(
+                            self.relative_id,
+                            "Invalid value for 'RequestParameters' property")
```
brettstack: Update similar to the message above with the expected format.

beck3905:
@brettstack I believe I've addressed your comments.  Would you please have a look?

samtranslator/swagger/swagger.py
```diff
@@ -788,6 +788,50 @@ def add_models(self, models):
...
+            existing_parameters = method_definition.get('parameters', [])
+
+            for request_parameter in request_parameters:
+
+                parameter_name = request_parameter['Name']
+                location_name = parameter_name.replace('method.request.', '')
+                location, name = location_name.split('.')
+
+                parameter = {
+                    'in': location,
```
jlhood: Ok here in the swagger doc, you do need to use `query` instead of `querystring`, but what I'm not clear on is if you need to use `querystring` in the API Gateway `cacheKeyParameters` section. Have you tested and verified this works correctly?

samtranslator/swagger/swagger.py
```diff
@@ -788,6 +788,50 @@ def add_models(self, models):
...
+                    'required': request_parameter['Required'],
+                    'type': 'string'
+                }
+
+                existing_parameters.append(parameter)
+
+                if request_parameter['Caching']:
+
+                    integration = method_definition[self._X_APIGW_INTEGRATION]
+                    cache_parameters = integration.get('cacheKeyParameters', [])
```
jlhood: Can you make `cacheKeyParameters` a constant, similar to `self._X_APIGW_INTEGRATION`?

examples/2016-10-31/function_request_parameters/template.yaml
```diff
@@ -0,0 +1,28 @@
...
+      Events:
+        GetApi:
+          Type: Api
+          Properties:
+            Path: /post
+            Method: POST
+            RequestParameters:
+              - method.request.header.Authorization:
+                  Required: True
+                  Caching: True
```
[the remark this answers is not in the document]
beck3905: You're right, caching the parameter means nothing if the cache cluster on the stage is not enabled.  I've updated the template with the Globals section you suggested.

I've also changed those booleans to lowercase as suggested.

examples/2016-10-31/function_request_parameters/template.yaml
```diff
@@ -0,0 +1,28 @@
...
+              - method.request.query.type
```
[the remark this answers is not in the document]
beck3905: I did a test template with both and found that only the `query` parameter was added to the API successfully.  This is because the doc you referenced is for the Amazon extensions to OpenAPI/Swagger, but it turns out that the swagger docs say to use `query` instead of `querystring`.  The Amazon extension must make this conversion behind the scenes.  https://swagger.io/docs/specification/describing-parameters/#query-parameters

samtranslator/model/eventsources/push.py
```diff
@@ -17,6 +17,8 @@
 
 CONDITION = 'Condition'
 
+RequestParameterProperties = ["Required", "Caching"]
```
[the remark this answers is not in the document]
beck3905: Updated to all caps with underscores.  I left this outside the class because I followed a similar pattern from https://github.com/awslabs/serverless-application-model/blob/master/samtranslator/model/api/api_generator.py#L27, which btw does not follow that same convention of all caps with underscores.

samtranslator/model/eventsources/push.py
```diff
@@ -607,6 +610,59 @@ def _add_swagger_integration(self, api, function):
...
+            }
+
+            parameters = []
+            for parameter in self.RequestParameters:
+
+                if isinstance(parameter, dict):
+
+                    parameter_name, parameter_value = next(iter(parameter.items()))
+
+                    if not re.match('method\.request\.(query|path|header)', parameter_name):
```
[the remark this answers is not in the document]
beck3905: see comments above about the swagger spec.  It should be `query` and not `querystring` even though the Amazon extension uses `querystring` instead of `query`.  Testing also verified this.

examples/2016-10-31/function_request_parameters/template.yaml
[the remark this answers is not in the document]
beck3905: On second thought, I decided for consistency sake to update this to use `querystring` even though that is not what the swagger docs say.  Users will more likely be familiar with the AWS usage that swagger usage.

samtranslator/model/eventsources/push.py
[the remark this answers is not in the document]
beck3905: I've updated this to use `querystring`.

samtranslator/swagger/swagger.py
beck3905: It turns out the cacheKeyParameters requires `querystring` and not `query`.  Therefore I've updated the code and docs to require `querystring` and convert `querystring` to `query` for the swagger.

samtranslator/swagger/swagger.py
beck3905: done.

beck3905:
@jlhood @praneetap I've made updates based on more review comments.  Please have look.

samtranslator/model/eventsources/push.py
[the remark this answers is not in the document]
jlhood: Ah apologies for the inconsistency there, but thanks for fixing this one!

examples/2016-10-31/function_request_parameters/template.yaml
```diff
@@ -0,0 +1,33 @@
...
+    CacheClusterSize: '0.5'
+
+Resources:
+
+  MyLambdaFunction:
+    Type: AWS::Serverless::Function
+    Properties:
+      Handler: index.handler
+      Runtime: nodejs8.10
+      CodeUri: s3://gbdx-build-artifacts/bbeck-test/423f9056eb70d25480852fc472b43dd5 #src/
```
jlhood: Guessing you meant to revert this back to `src/`

```suggestion
      CodeUri: src/
```

examples/2016-10-31/function_request_parameters/template.yaml
```diff
@@ -0,0 +1,33 @@
...
+              - method.request.header.Authorization:
+                  Required: true
+                  Caching: true
+              - method.request.querystring.type
+
+Outputs:
+
+  ApiURL:
+    Description: "API endpoint URL for Prod environment"
+    Value: !Sub "https://${ServerlessRestApi}.execute-api.${AWS::Region}.amazonaws.com/prod/get"
```
jlhood:
```suggestion
    Value: !Sub "https://${ServerlessRestApi}.execute-api.${AWS::Region}.amazonaws.com/Prod/post"
```

samtranslator/model/eventsources/push.py
```diff
@@ -607,6 +610,61 @@ def _add_swagger_integration(self, api, function):
...
+            }
+
+            parameters = []
+            for parameter in self.RequestParameters:
+
+                if isinstance(parameter, dict):
+
+                    parameter_name, parameter_value = next(iter(parameter.items()))
+
+                    if not re.match('method\.request\.(querystring|path|header)', parameter_name):
```
jlhood: Should add a trailing period so strings like `method.request.pathmoreletters` isn't considered valid.

```suggestion
                    if not re.match('method\.request\.(querystring|path|header)\.', parameter_name):
```

samtranslator/model/eventsources/push.py
```diff
@@ -607,6 +610,61 @@ def _add_swagger_integration(self, api, function):
...
+                            "e.g { Required: true, Caching: false }")
+
+                    settings = default_value.copy()
+                    settings.update(parameter_value)
+                    settings.update({'Name': parameter_name})
+
+                    parameters.append(settings)
+
+                elif isinstance(parameter, string_types):
+                    if not re.match('method\.request\.(querystring|path|header)', parameter):
```
jlhood:
```suggestion
                    if not re.match('method\.request\.(querystring|path|header)\.', parameter):
```

examples/2016-10-31/function_request_parameters/template.yaml
beck3905: Yes, thanks for catching that.  I'm surprised with all the back and forth and testing I haven't committed that already.

beck3905:
@keetonian @praneetap @jlhood Thank you for approving my PR and merging.  I noticed that it is staged for release in v1.15.0.  We are currently on v1.13.2, which suggests that it will be a while until this is released since we still have to go through v1.14.x.  Is there any way to get this released faster?  I know this is considered a new feature, however in my case the lack of the ability to cache parameters is more like a bug.  That is why I submitted this PR.

For example, I have a function with the following path: `/my-api/thing/{thing_id}`.  SAM automatically will add a parameter to the API swagger definition for `thing_id` as a path parameter, however it does not add it as a cache key parameter.  This means that I get the following unexpected behavior from my API that has caching enabled on the stage:

GET request to `/my-api/thing/thing1` returns `thing1`
GET request to `/my-api/thing/thing2` returns `thing1` when I expect `thing2`.  `thing1` is returned because it is in the cache and the `thing_id` path parameter is not part of the cache key.

This seems like basic functionality that is required if caching is enabled, otherwise the API will return incorrect resources from the cache.  I had a similar example with the Authorization header where regardless of the token in the Authorization header if a resource was in the cache, a user could access it.

Thanks again and please advise.

jlhood:
@beck3905 This change missed the cutoff for the v1.14.0 release so it got tagged for v1.15.0. Unfortunately, we are not able to share specific dates for releases, but v1.15.0 will be the next release we cut after v1.14.0, which is currently in progress. Patch releases like v1.13.1 and v1.13.2 were due to bugs found during the release where we needed to cherry-pick a specific bugfix related to the v1.13.0 release.

We're working on making our release process more transparent to external customers, but we're not quite there yet. For now, you can see when we cut [release branches](https://github.com/awslabs/serverless-application-model/branches/active), which can give you some indication of when we begin working on a specific release.

azarboon:
Adding `RequestParameters` to `AWS::Serverless::Function`, doesn't enforce it. Even though the query string is defined as required, but it's not enforced. 

```
            RequestParameters: 
              - method.request.querystring.myString: 
                  Required: true
```

And in API GW console, under URL Query String Parameters, it shows this warning: 

> You have marked some query string parameter as required but thee request validator assigned to this method is not configured to validate parameters. To ensure that incoming HTTP requests include the required query string parameters, select an appropriate request validator.

This makes us to need to updated RequestParameters in template (in order to have updated Swagger specs) and to update Lambda business logic (to perform validation). So it's extra work and cna lead to inconsistency.

Issue #566, opened 2018-09-06
Best way to pass binary content through ApiGateway unmodified


0xdevalias:
I searched the existing issues for [`ContentHandling `](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-apitgateway-method-integration.html#cfn-apigateway-method-integration-contenthandling) and most that contain it (as part of their example code) seem to be setting at as part of a fully custom OpenAPI/Swagger spec.

I was wondering if there is a canonical way for setting this on the generated 'implicit ApiGateway', or if I should be resorting to manually creating my entire spec for this kind of behaviour?

0xdevalias:
Reading a little deeper.. the base64 encoding may actually be occurring at the `awslabs/aws-lambda-go-api` proxy level:
* https://github.com/awslabs/aws-lambda-go-api-proxy/blob/master/core/response.go#L81-L86
* https://github.com/awslabs/aws-lambda-go-api-proxy/blob/master/gorillamux/adapter.go#L31-L36

So I'm not sure whether this is even something that would be configurable/fixable in SAM.

0xdevalias:
Potentially related issues:
* https://github.com/awslabs/aws-sam-cli/issues/653
* https://github.com/awslabs/serverless-application-model/issues/553
* https://github.com/awslabs/serverless-application-model/issues/248

0xdevalias:
Decided to try implementing this myself using the more manual version of ``. By inspecting the end CloudFormation json, and converting/working backwards into SAM yaml, I came up with this:

```yaml
  FooTestApi:
    Type: AWS::Serverless::Api
    Properties:
      Name: FooTestApi
      StageName: Prod
      DefinitionBody:
        swagger: '2.0'
        info:
          version: '1.0'
          title: !Ref AWS::StackName
        paths:
          "/{proxy+}":
            x-amazon-apigateway-any-method:
              x-amazon-apigateway-integration:
                type: aws_proxy
                uri: !Sub arn:aws:apigateway:${AWS::Region}:lambda:path/2015-03-31/functions/${AgentProxy.Arn}/invocations
                httpMethod: POST
                contentHandling: CONVERT_TO_BINARY
              responses: {}
```

I also created this permission manually (though I expect if I set the `RestApiId` property on my function it may have been generated for me):

```
  AgentProxyCatchAllPermissionProdFooTestApi:
    Type: AWS::Lambda::Permission
    Properties:
[...]

Issue #561, opened 2018-09-05
Globals/Api/BinaryMediaTypes is not honored when SAM template is updated and re-deployed


abhilash1in:
**Description:**
`BinaryMediaTypes` is removed from APIGW settings when SAM template is updated and re-deployed

**Steps to reproduce the issue:**
1. Create a simple SAM application with `Globals` as follows:
```
Globals:
    Api:
        BinaryMediaTypes:
            - "*~1*"
```
2. Package and deploy. Observe `Settings` section of the new APIGW API. `Binary Media Types` will have `*/*`

3. Make some changes to the SAM template. Re-package and re-deploy. Observe `Settings` section of the APIGW API. `Binary Media Types` config is empty/missing

**Observed result:**
`Binary Media Types` config is empty/missing after SAM template is updated and redeployed

**Expected result:**
`Binary Media Types` does not change after stack update.

**Output of `sam --version`**: `SAM CLI, version 0.6.0`

**Optional Debug logs**:

`Add --debug flag to command you are running`

jfuss:
Closing as this does not relate to this project. I responded on the issue that was made to AWS SAM CLI.

mrjomp:
This needs to be reopened, it has not been fixed and it's part of the SAM templates

keetonian:
@mrjomp could you provide more details of what your error is and how you encountered it? Is it the exact same as the original post in this issue?

emanuelhjertzen:
Ran into this issue today also. Exactly as described in original description by @abhilash1in.
I first set to '*~1*' in template.yaml. Then changed to 'image~1png' and packages/deployed. After that the BinaryMediaTypes have been removed. Changing them further does not have any effect.

emanuelhjertzen:
Did further tests:
1. In template set 
BinaryMediaTypes:
  - image~1gif
  - image~1png
Package/deploy got the expected result.

2. Change template to
BinaryMediaTypes:
  - image~1gif
  - image~1png
  - application~1vnd.openxmlformats-officedocument.spreadsheetml.sheet
Package/deploy resulted in only application/1vnd.openxmlformats-officedocument.spreadsheetml.sheet set in console.

3. Change template to
BinaryMediaTypes:
  - image~1gif
  - image~1png
  - application~1vnd.openxmlformats-officedocument.spreadsheetml.sheet
  - image~1tiff
Package/deploy resulted in only image/tiff set in console

4. Change template to
BinaryMediaTypes:
[...]

Issue #553, opened 2018-08-24
Unable to set "contentHandling"


tayler-kemsley:
I've found that setting the 'x-amazon-apigateway-binary-media-types' to '*/*' causes 500 errors.

Using the API to set the content handling to "CONVERT_TO_TEXT" resolves the issue, but there doesn't seem to be a way to define that in the SAM CORS configuration, so the requests will always fail. (See below for error)

The error from API gateway is as follows:

```
a5bfc9 14:52:31  (53c17c18-a7ad-11e8-a099-4b78be23626d) Method request body before transformations: [Binary Data]
a5bfc9 14:52:31  (53c17c18-a7ad-11e8-a099-4b78be23626d) Execution failed due to configuration error: Unable to transform request
a5bfc9 14:52:31  (53c17c18-a7ad-11e8-a099-4b78be23626d) Gateway response type: API_CONFIGURATION_ERROR with status code: 500
a5bfc9 14:52:31  (53c17c18-a7ad-11e8-a099-4b78be23626d) Gateway response body:
{
    "detail":  "Internal server error",
    "title": "API_CONFIGURATION_ERROR",
    "status": 500
}
```

This is the output of the API integration in a _working_ case:

```
{
    "integrationResponses": {
        "200": {
            "responseTemplates": {
                "application/json": null
            },
            "selectionPattern": "",
            "statusCode": "200"
        }
    },
    "contentHandling": "CONVERT_TO_TEXT",
    "timeoutInMillis": 29000,
    "passthroughBehavior": "NEVER",
    "requestTemplates": {
        "application/json": "{\"statusCode\": 200}"
    },
    "cacheNamespace": "qfgrnb",
    "type": "MOCK",
    "cacheKeyParameters": []
}
```

keetonian:
Hi @tayler-kemsley! Thank you for bringing this up. If I understand this right, it looks like we would need to surface another CORS field, `contentHandling`, that would be added to the `x-amazon-apigateway-integration` resource in the output CloudFormation. Would SAM need to make any other template changes to match the working case that you posted?
I read up a little on [resource conversions in apigw](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-payload-encodings-workflow.html), but I'm still not very familiar with them.

tayler-kemsley:
[...]

PR #954, opened 2019-06-05
fix: fix API Binary Media Types update issue

brettstack:
Related Issues: #566 #553 #561


tests/translator/input/api_with_binary_media_types_definition_body.yaml
```diff
@@ -0,0 +1,26 @@
+Globals:
```
praneetap: Could you also add an example of this with a template containing inline swagger?

brettstack: Added

Issue #923, opened 2019-05-16
AWS_IAM auth does not allow InvokeRole override


theburningmonk:
**Description:**

When using the new `AWS_IAM` auth type, the `InvokeRole` is always set to `CALLER_CREDENTIALS` even when I specify an override. The problem here is that, it forces the caller to have both API Gateway's invoke permission as well as `lambda:InvokeFunction` permission. This breaks the API abstraction and leaks implementation details (that there's a Lambda behind API Gateway, and the name of the function).

**Steps to reproduce the issue:**
1. create API with auth type set to `AWS_IAM` and set `InvokeRole` to `null`

![image](https://user-images.githubusercontent.com/546969/57856560-3dd38800-77e5-11e9-908d-816c97d2f6b4.png)

**Observed result:**

API endpoints still uses `CALLER_CREDENTIALS`

![image](https://user-images.githubusercontent.com/546969/57856663-6a879f80-77e5-11e9-8ec5-653a717afcda.png)

**Expected result:**

* `Execution role` to be `null
* `Invoke with caller credentials` to be `disabled`

theburningmonk:
For now, our workaround is to have a custom macro that is triggered after the initial SAM macro and strips the `credentials` field.

![image](https://user-images.githubusercontent.com/546969/57862751-c99ee180-77f0-11e9-854e-1d93f713fd77.png)

tavolate:
We have the same issue

keetonian:
It looks like we're forcing `InvokeRole` to be non-null here: https://github.com/awslabs/serverless-application-model/blob/develop/samtranslator/swagger/swagger.py#L192

A simple fix for this would be to allow null or blank values and add a test that enforces this behavior.

jadhavmanoj:
@keetonian if InvokeRole in None. should we remove credentials key from template?

keetonian:
That's a good question. @brettstack might know, otherwise I'll investigate more and see what SAM should do in this case.

brettstack:
@jadhavmanoj that's correct.

benkehoe:
Is the plan to change the default to not use caller credentials? I am in favor of such a plan.

brettstack:
Unfortunately that ship has sailed. Changing default behavior would be a breaking change. The plan is to add `InvokeRole: null` (or 'NONE') as originally intended.

benkehoe:
[...]

PR #986, opened 2019-06-19
fix: allow setting InvokeRole to NONE or null

keetonian:
*Issue #, if available:*
#923 

*Description of changes:*
Allows invoke role override (removes the "credentials" field as discussed in #923)

Verified that the transformed template removes the "credentials" section when InvokeRole is set to 'None' or null. Also deployed template to CFN to verify that "Invoke with caller credentials" and "execution role" were not set on the api method.

I also attempted to make it more explicit that `CALLER_CREDENTIALS` is the default value, due to a conversation in the above issue.

*Checklist:*

- [x] Write/update tests
- [x] `make pr` passes
- [x] Update documentation

### DIFF
--- a/examples/2016-10-31/api_aws_iam_auth/template.yaml
+++ b/examples/2016-10-31/api_aws_iam_auth/template.yaml
@@ -8,7 +8,7 @@ Resources:
       StageName: Prod
       Auth:
         DefaultAuthorizer: AWS_IAM
-        InvokeRole: CALLER_CREDENTIALS
+        InvokeRole: CALLER_CREDENTIALS # default, set to NONE or null to not require credentials
 
   MyFunction:
     Type: AWS::Serverless::Function
--- a/examples/2016-10-31/api_aws_iam_auth/template.yaml
+++ b/examples/2016-10-31/api_aws_iam_auth/template.yaml
@@ -8,7 +8,7 @@ Resources:
       StageName: Prod
       Auth:
         DefaultAuthorizer: AWS_IAM
-        InvokeRole: CALLER_CREDENTIALS # default, set to NONE or null to not require credentials
+        InvokeRole: CALLER_CREDENTIALS # default, can specify other role or NONE
 
   MyFunction:
     Type: AWS::Serverless::Function
--- a/samtranslator/swagger/swagger.py
+++ b/samtranslator/swagger/swagger.py
@@ -169,10 +169,18 @@ class SwaggerEditor(object):
         api_auth_config = api_auth_config or {}
         if method_auth_config.get('Authorizer') == 'AWS_IAM' \
            or api_auth_config.get('DefaultAuthorizer') == 'AWS_IAM' and not method_auth_config:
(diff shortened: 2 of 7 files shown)


samtranslator/swagger/swagger.py
```diff
@@ -169,10 +169,18 @@ def add_lambda_integration(self, path, method, integration_uri,
         api_auth_config = api_auth_config or {}
         if method_auth_config.get('Authorizer') == 'AWS_IAM' \
            or api_auth_config.get('DefaultAuthorizer') == 'AWS_IAM' and not method_auth_config:
-            self.paths[path][method][self._X_APIGW_INTEGRATION]['credentials'] = self._generate_integration_credentials(
-                method_invoke_role=method_auth_config.get('InvokeRole'),
-                api_invoke_role=api_auth_config.get('InvokeRole')
+            method_invoke_role = method_auth_config.get('InvokeRole')
+            if not method_invoke_role and 'InvokeRole' in method_auth_config:
+                method_invoke_role = 'NONE'
+            api_invoke_role = api_auth_config.get('InvokeRole')
```
keetonian: This section looks for two things:
* If there is an `InvokeRole` set, it gets the value
* If `InvokeRole` is set to null, it sets it to 'NONE'
* If `InvokeRole` is not set, it is passed as usual and is set to `CallerCredentials`.

tests/translator/input/api_with_aws_iam_auth_overrides.yaml
```diff
@@ -84,3 +84,35 @@ Resources:
...
+      Events:
+        API3:
+          Type: Api
+          Properties:
+            RestApiId: !Ref MyApiWithAwsIamAuth
+            Method: get
+            Path: /MyFunctionNONEInvokeRole
+            Auth:
+              Authorizer: AWS_IAM
+              InvokeRole: NONE
```
praneetap: Please add a global level test

examples/2016-10-31/api_aws_iam_auth/template.yaml
```diff
@@ -8,7 +8,7 @@ Resources:
       StageName: Prod
       Auth:
         DefaultAuthorizer: AWS_IAM
-        InvokeRole: CALLER_CREDENTIALS
+        InvokeRole: CALLER_CREDENTIALS # default, set to NONE or null to not require credentials
```
brettstack: This isn't accurate. IAM credentials are still required, they just aren't used as the invocation role for the Lambda Function. Setting to NONE will use the Lambda Function's configured role; setting to an different IAM role will invoke using that role.

kosmakoff:
Sorry for posting to inappropriate thread, but I have a question. Since this branch is already merged, and I see that it has a release 1.14.0 tag on it, does it mean it should already work on AWS, or does this change only apply to `sam cli`? Sorry for confusing terms, but I have the same exact problem with `InvokeRole`, and I am using `dotnet lambda` for development. Please advise.

Issue #1014, opened 2019-07-04
Unable to use NONE authorization method with AWS_IAM default authorizer


TDaglis:
**Description:**

It is not possible to create an API with just AWS_IAM authorizer and add a public route to it. Adding any other authorizer to the API, even an unused one, fixes the issue, but it's hacky. The source of the issue is probably not taking into account that the AWS_IAM authorizer is declared only as a default authorizer and not listed in the `Authorizers` dictionary.

**Steps to reproduce the issue:**
1. Create a sam template with an API with AWS_IAM default authorizer.
2. Create 2 endpoints
3. In one endpoint, add `Authorizer: "NONE"` as the method authorizer.
4. Try to deploy the template

Example sam template:
```yaml
AWSTemplateFormatVersion: '2010-09-09'
Transform: AWS::Serverless-2016-10-31
Description: Test API

Globals:
  Function:
    Runtime: python3.6
    Timeout: 600
    MemorySize: 1024

Resources:
  myapi:
    Type: AWS::Serverless::Api
    Properties:
      StageName: v1
      Auth:
        DefaultAuthorizer: AWS_IAM

  private:
    Type: AWS::Serverless::Function
    Properties:
      CodeUri: .
      Handler: test.lambda_handler
      Events:
        GetEvent:
          Type: Api
          Properties:
            Path: /private
            Method: get
            RestApiId: !Ref myapi

  public:
    Type: AWS::Serverless::Function
    Properties:
      CodeUri: .
      Handler: test.lambda_handler
      Events:
        GetEvent:
          Type: Api
          Properties:
            Path: /public
            Method: get
            RestApiId: !Ref myapi
            Auth:
                Authorizer: "NONE"
```
**Observed result:**

Deployment fails with the following message:
```Failed to create the changeset: Waiter ChangeSetCreateComplete failed: Waiter encountered a terminal failure state Status: FAILED. Reason: Transform AWS::Serverless-2016-10-31 failed with: Invalid Serverless Application Specification document. Number of errors found:1. Resource with id [public] is invalid. Event with id [GetEvent] is invalid. Unable to set Authorizer [NONE] on API method [get] forpath [/public] because the related API does not define any Authorizers.```

**Expected result:**

Deployment succeeds with the `/private` resource having AWS_IAM authorizer and the `/public` route having no authorizer.

ShreyaGangishetty:
Closing this issue as v1.15.0 is released

PR #1015, opened 2019-07-04
Fix Authorizer: NONE support with AWS_IAM authorizer

TDaglis:
*Issue #1014*

*Description of changes:*
Change the validation of `method_authorizer` to support the combination of `AWS_IAM` default authorizer and `NONE` method authorizer.

When using `AWS_IAM` permission, it is only specified in the `DefaultAuthorizer`.  [Reference](https://github.com/awslabs/serverless-application-model/blob/master/versions/2016-10-31.md#api-auth-object). This is a special case which leads to the `api_authorizers` being empty and results to not being able to add
 ```
Auth:
  Authorizer: 'NONE'
```
to a specific function, in order to override the default authorizer and make the function public.

*Checklist:*

- [x] Write/update tests
- [x] `make pr` passes
- [ ] Update documentation

### DIFF
--- a/samtranslator/model/eventsources/push.py
+++ b/samtranslator/model/eventsources/push.py
@@ -552,7 +552,7 @@ class Api(PushEventSource):
                 api_authorizers = api_auth and api_auth.get('Authorizers')
 
                 if method_authorizer != 'AWS_IAM':
-                    if not api_authorizers:
+                    if method_authorizer != 'NONE' and not api_authorizers:
                         raise InvalidEventException(
                             self.relative_id,
                             'Unable to set Authorizer [{authorizer}] on API method [{method}] for path [{path}] '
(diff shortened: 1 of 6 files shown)


praneetap:
Thanks for the PR! The Travis build is failing. Once those errors are fixed we will give this a deeper review.

keetonian:
Ran tests locally and they all passed. Re-running travis.

jhaenchen:
I need this fix folks, just checking in!

praneetap:
Thank you for the contribution! Could you please add end to end tests (Input/Output) that verifies the fix? You can take a look at this [pr](https://github.com/awslabs/serverless-application-model/pull/1043/files) for guidance. Please make sure you follow the pr checklist.

TDaglis:
I attempted to add an end to end test, but it appears that it has issues in Python2.7. I don't understand what the issue is exactly. Any direction on this would be helpful. I'll get back to it.

TDaglis:
It appears my output files had issues with Python2.7. I fixed them and now the tests pass.

Issue #1002, opened 2019-06-28
Transform fails when Api StageName contains '_'


hikerninja:
If an AWS:Serverless::Api resource has a StageName that contains `_` the transform fails with the following error:

```
Failed to create CloudFormation change set: Transform AWS::Serverless-2016-10-31 failed with: Invalid Serverless Application Specification document. Number of errors found: 2. Resource with id [Apivendor_apiStage] is invalid. Logical ids must be alphanumeric. Resource with id [ControllerFunctionPostHRResultsPermissionvendor_api] is invalid. Logical ids must be alphanumeric.
```

Below is a snippet of the template that was used:

```
Api:
    Type: AWS::Serverless::Api
    Properties:
      StageName: vendor_api
```

The template successfully deploys with a StageName that does not contain `_`.  The documentation for API Gateway specifies that a stage name can contain `_`, and we are currently getting around this issue by using a different stage name, then deploying to the `vendor_api` stage using an additional script.

I believe this is a result of the transform using a Cloudformation logical resource id derived from the AWS::Serverless::Api logical resource id and the stage name, and logical resource ids cannot include non-alphanumerical characters.

keetonian:
Correct, the CloudFormation logical ID of the ApiGateway::Stage resource that SAM creates is derived from the `StageName` property and the name of the `Serverless::Api`.

praneetap:
If anyone is interested in contributing this fix, [here](https://github.com/awslabs/serverless-application-model/blob/develop/samtranslator/model/api/api_generator.py#L188) is where the fix would need to be made.

eduardovra:
So the solution should be to not concatenate the StageName string in the ApiGateway::Stage logical ID anymore. Is this correct ?

ShreyaGangishetty:
Closing this issue as v1.15.0 is released

PR #1035, opened 2019-07-23
fix: randomize logical IDs of API stage and Lambda permission

yuimam:
*Issue #1002*

*Description of changes:*

("-") is not accepted in CFn logical ID, but the stage name of API Gateway and Lambda permission may include ("-").  I changed the way to generate logical ID by logical_id_generator.

I tried the following template, and it worked.
And I also added E2E tests.

```yaml
Description: HelloWorld
Resources:
  MyAPI:
    Type: "AWS::Serverless::Api"
    Properties:
      StageName: test_test
  HelloWorld:
    Properties:
      Handler: index.handler
      Runtime: nodejs10.x
      CodeUri: s3://xxx/xxx
      Description: HelloWorld
      MemorySize: 128
      Timeout: 3
      Events:
        GetResource:
          Type: Api
          Properties:
            Path: /hello
            Method: get
            RestApiId: !Ref MyAPI
```

*Checklist:*

- [x] Write/update tests
- [x] `make pr` passes
- [ ] Update documentation

### DIFF
--- a/samtranslator/model/api/api_generator.py
+++ b/samtranslator/model/api/api_generator.py
@@ -10,6 +10,7 @@ from samtranslator.region_configuration import RegionConfiguration
 from samtranslator.swagger.swagger import SwaggerEditor
 from samtranslator.model.intrinsics import is_instrinsic, fnSub
 from samtranslator.model.lambda_ import LambdaPermission
+from samtranslator.translator import logical_id_generator
 from samtranslator.translator.arn_generator import ArnGenerator
 from samtranslator.model.tags.resource_tagging import get_tag_list
 
@@ -184,8 +185,12 @@ class ApiGenerator(object):
         # If StageName is some intrinsic function, then don't prefix the Stage's logical ID
         # This will NOT create duplicates because we allow only ONE stage per API resource
         stage_name_prefix = self.stage_name if isinstance(self.stage_name, string_types) else ""
-
-        stage = ApiGatewayStage(self.logical_id + stage_name_prefix + 'Stage',
+        if '-' in stage_name_prefix or '_' in stage_name_prefix:
+            generator = logical_id_generator.LogicalIdGenerator(self.logical_id + 'Stage', stage_name_prefix)
+            stage_logical_id = generator.gen()
+        else:
+            stage_logical_id = self.logical_id + stage_name_prefix + 'Stage'
+        stage = ApiGatewayStage(stage_logical_id,
                                 attributes=self.passthrough_resource_attributes)
         stage.RestApiId = ref(self.logical_id)
         stage.update_deployment_ref(deployment.logical_id)
--- a/samtranslator/model/api/api_generator.py
+++ b/samtranslator/model/api/api_generator.py
@@ -185,11 +185,11 @@ class ApiGenerator(object):
         # If StageName is some intrinsic function, then don't prefix the Stage's logical ID
         # This will NOT create duplicates because we allow only ONE stage per API resource
         stage_name_prefix = self.stage_name if isinstance(self.stage_name, string_types) else ""
-        if '-' in stage_name_prefix or '_' in stage_name_prefix:
+        if stage_name_prefix.isalnum():
+            stage_logical_id = self.logical_id + stage_name_prefix + 'Stage'
+        else:
             generator = logical_id_generator.LogicalIdGenerator(self.logical_id + 'Stage', stage_name_prefix)
             stage_logical_id = generator.gen()
-        else:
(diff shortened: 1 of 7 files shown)


samtranslator/model/api/api_generator.py
```diff
@@ -184,8 +185,12 @@ def _construct_stage(self, deployment, swagger):
         # If StageName is some intrinsic function, then don't prefix the Stage's logical ID
         # This will NOT create duplicates because we allow only ONE stage per API resource
         stage_name_prefix = self.stage_name if isinstance(self.stage_name, string_types) else ""
-
-        stage = ApiGatewayStage(self.logical_id + stage_name_prefix + 'Stage',
+        if '-' in stage_name_prefix or '_' in stage_name_prefix:
```
jlhood: According to the [CloudFormation docs](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/resources-section-structure.html#resources-section-structure-logicalid), the only valid characters are alphanumeric (A-Za-z0-9). Why don't you change this check to be for any non-alphanumeric characters rather than just `-` or `_`.

samtranslator/model/api/api_generator.py
```diff
@@ -184,8 +185,12 @@ def _construct_stage(self, deployment, swagger):
...
+            generator = logical_id_generator.LogicalIdGenerator(self.logical_id + 'Stage', stage_name_prefix)
+            stage_logical_id = generator.gen()
+        else:
```
jlhood: Really appreciate you thinking about backwards-compatibility!

samtranslator/model/eventsources/push.py
```diff
@@ -47,9 +48,13 @@ def _construct_permission(self, function, source_arn=None, source_account=None,
         :returns: the permission resource
         :rtype: model.lambda_.LambdaPermission
         """
-        lambda_permission = LambdaPermission(self.logical_id + 'Permission' + suffix,
+        if '-' in suffix or '_' in suffix:
```
jlhood: Same here

samtranslator/model/api/api_generator.py
yuimam: I just changed the condition in the latest commit.

samtranslator/model/eventsources/push.py
yuimam: I just changed the condition in the latest commit.

samtranslator/model/api/api_generator.py
```diff
@@ -184,8 +185,12 @@ def _construct_stage(self, deployment, swagger):
         # If StageName is some intrinsic function, then don't prefix the Stage's logical ID
         # This will NOT create duplicates because we allow only ONE stage per API resource
         stage_name_prefix = self.stage_name if isinstance(self.stage_name, string_types) else ""
-
-        stage = ApiGatewayStage(self.logical_id + stage_name_prefix + 'Stage',
+        if stage_name_prefix.isalnum():
+            stage_logical_id = self.logical_id + stage_name_prefix + 'Stage'
+        else:
+            generator = logical_id_generator.LogicalIdGenerator(self.logical_id + 'Stage', stage_name_prefix)
+            stage_logical_id = generator.gen()
```
keetonian: I think I like your approach more than something like `stage_logical_id = re.sub('[^A-Za-z0-9]+', '', stage_logical_id)` to remove all non-alphanumeric characters, as your approach differentiates between the stage names that you provided in your tests.

samtranslator/model/eventsources/push.py
```diff
@@ -47,9 +48,13 @@ def _construct_permission(self, function, source_arn=None, source_account=None,
         :returns: the permission resource
         :rtype: model.lambda_.LambdaPermission
         """
-        lambda_permission = LambdaPermission(self.logical_id + 'Permission' + suffix,
+        if suffix.isalnum():
+            permission_logical_id = self.logical_id + 'Permission' + suffix
+        else:
+            generator = logical_id_generator.LogicalIdGenerator(self.logical_id + 'Permission', suffix)
+            permission_logical_id = generator.gen()
```
keetonian: The difficulty with changing the Lambda permission naming here is that it will be difficult to reference the function permissions that are created. I think that this will be fine, as nothing should depend on those resources.

Could you make a note on https://github.com/awslabs/serverless-application-model/blob/master/docs/internals/generated_resources.rst#api that if the `StageName` parameter has a value that contains non alpha-numeric characters, the generated names will follow this new convention?

jlhood:
@yuimam Any updates on this? I think @keetonian's only suggestion is adding a note to the documentation. Then this should be good to approve/merge. Really appreciate your contribution!

Issue #514, opened 2018-07-18
API Gateway Resource Policy


brettstack:
Resources:

https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-resource-policies.html

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-apigateway-restapi.html#cfn-apigateway-restapi-policy

Amazon API Gateway resource policies are JSON policy documents that you attach to an API to control whether a specified principal (typically an IAM user or role) can invoke the API. You can use API Gateway resource policies to allow your API to be securely invoked by:

- users from a specified AWS account

- specified source IP address ranges or CIDR blocks

- specified virtual private clouds (VPCs) or VPC endpoints (in any account)

```yaml
Resources:
  MyApi:
    Type: AWS::Serverless::Api
    Properties:
      ...
      Auth:
        ResourcePolicy:
          # IAM-based policies
          IAMAllowlist: [
          '123456789012:role/myRole',
          '${AWS::AccountId}:user/myUser',
          '123456789012:root',
          '123456789012:*'
          ],
          IAMDenylist: [],

          # IP-based policies
          IpAllowlist: ['12.123.234.213'],
          IpDenylist: [],

          # VPC-based policies
          SourceVpcAllowlist: ['vpc-ab1234cd'],
          SourceVpcDenylist: []

          # Custom statements. These must be actual Resource Policy Statements.
          CustomStatements: [{
            Action: 'execute-api:Invoke', # Optional; Default: execute-api:Invoke
            Resource: ['execute-api:/*/*/*'], # Optional; constructed based on Stagename, Path, and Method.
            ... # Additional properties get passed through to the resulting statement
          }]

  MyFn:
    Type: AWS::Serverless::Function
    Properties:
      Events:
        GetRoot:
          Type: Api
          Properties:
            Auth:
              ResourcePolicy:
                ... # Same as above; the Statement Resource will be created differently when defined here (i.e. it will use the Method and Path)
```

When `Auth.ResourcePolicy` is set on an API `Event`, the `Path` and `Method` of the `Event` will be used to construct the `Resource`. When `Auth.ResourcePolicy` is set on an API resource, the `Path` and `Method` parts of `Resource` will be *; that is, the policy will apply to the entire API. For the Stage part of `Resource`, we can inject the `StageName`, however, we do need to consider how we will make it work when we implement multi-stage support.

[...]

PR #1045, opened 2019-07-26
feat: sam support to add resource policies to api properties

praneetap:
*Issue #, if available:*
#514 
*Description of changes:*
Added support for resource policies on api definition in SAM. 
Transformed this template locally and deployed it on CFN successfully. I could invoke the API successfully.

```yaml
Parameters:
  StageName:
    Default: 'prod'
    Type: String
Resources:
  MyFunction:
    Properties:
      InlineCode: |
        exports.handler = async () => "Hello Worlds!"
      Handler: index.handler
      Runtime: nodejs8.10
      Events:
        Api:
          Type: Api
          Properties:
            RestApiId: !Ref LambdaAPIDefinition
            Method: Put
            Path: /get
```
*Checklist:*

- [x] Write/update tests
- [x] `make pr` passes
- [x] Update documentation


tests/translator/input/api_with_resource_policy.yaml
```diff
@@ -0,0 +1,12 @@
+Resources:
+ ExplicitApi:
+    Type: AWS::Serverless::Api
+    Properties:
+      StageName: Prod
+      Auth:
+        ResourcePolicy:
+          CustomStatements: [{
+              Action: 'execute-api:Invoke', # Optional; Default: execute-api:Invoke
```
keetonian: This isn't currently optional, correct? Making this optional would happen in a future PR.

praneetap: will remove. thanks!

examples/2016-10-31/api_resource_policy/template.yaml
```diff
@@ -0,0 +1,38 @@
...
+        ResourcePolicy:
+          CustomStatements: [{
+            "Effect": "Allow",
+            "Principal": {
+              "Service": "lambda.amazonaws.com"
+            },
+            "Action": "sts:AssumeRole"
+          }]
+      StageName: !Ref StageName
+      BinaryMediaTypes:
```
jlhood: Is custom stage name and binary media types necessary or is it copy/paste from some other example? Would prefer the example focus on the resource policy feature and not include extra, unrelated features.

samtranslator/swagger/swagger.py
```diff
@@ -788,6 +789,32 @@ def add_models(self, models):
...
+
+        custom_statements = resource_policy.get('CustomStatements')
+        if custom_statements is not None:
+            self.resource_policy['Version'] = '2012-10-17'
+            if self.resource_policy.get('Statement') is None:
+                self.resource_policy['Statement'] = custom_statements
+            else:
+                statement = self.resource_policy['Statement']
+                if type(statement) is list:
+                    statement.extend(custom_statements)
```
jlhood: Do we specifically validate that what's in `CustomStatements` ahead of this code? If not, this could blow up with a 5xx. Should we support them just adding a single statement that's not a list (is a dict)?

samtranslator/swagger/swagger.py
```diff
@@ -788,6 +789,32 @@ def add_models(self, models):
...
+                else:
+                    statement = [statement, custom_statements]
```
jlhood: If `custom_statements` is a list, doesn't this mean it'll be a nested list inside the outer list? Is that what we want?

tests/translator/input/api_with_resource_policy_global.yaml
```diff
@@ -0,0 +1,36 @@
...
+      DefinitionBody: {
+        "info": {
+          "version": "1.0",
+          "title": {
+            "Ref": "AWS::StackName"
+          }
+        },
+        "paths": {},
+        "swagger": "2.0",
+        "x-amazon-apigateway-policy": {
```
jlhood: I'm having trouble finding the API GW docs on this OpenApi extension. Could you point me to them? Just want to make sure they're in the right place in the overall swagger doc.

tests/translator/input/api_with_resource_policy_global.yaml
```diff
@@ -0,0 +1,36 @@
...
+  Api:
+    Auth:
+      ResourcePolicy:
+        CustomStatements: [{
+            Action: 'execute-api:Invoke', 
+            Resource: ['execute-api:/*/*/*']
+          }]
+
+Resources:
+ ExplicitApi:
```
jlhood: So custom statements will override an explicitly defined OpenApi doc? It's not strictly limited to implicit APIs? I guess we're doing that elsewhere with auth specifically, but just wanted to confirm. Also, should we have an example with implicit API only (lambda function implies an API, globals adds custom statements to resource policy)?

examples/2016-10-31/api_resource_policy/template.yaml
praneetap: Sure, I can make it more concise

tests/translator/input/api_with_resource_policy_global.yaml
keetonian: I don't see it in the OpenApi extensions: https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-swagger-extensions.html

praneetap: I got this from API gateway, by manually updating the API with resource policies on the apigw console and exporting the swagger with the apigw extensions.

tests/translator/input/api_with_resource_policy_global.yaml
praneetap: Yes, I think this applies to all the auth properties in general. 

Added implicit-only example.

examples/2016-10-31/api_resource_policy/template.yaml
```diff
@@ -0,0 +1,28 @@
+AWSTemplateFormatVersion: '2010-09-09'
+Transform: AWS::Serverless-2016-10-31
+Globals:
+  Api:
+    Auth:
+      ResourcePolicy:
+        CustomStatements: [{
```
jlhood: What is the use case for adding this custom statement? For tests, the validity of the custom statement isn't as important, just that we transform it correctly. However, for examples, I think we should show a valid use case for custom statements. I'm not sure this is one.

samtranslator/model/api/api_generator.py
```diff
@@ -325,7 +325,11 @@ def _add_auth(self):
             swagger_editor.add_apikey_security_definition()
             self._set_default_apikey_required(swagger_editor)
 
-        self.definition_body = self._openapi_postprocess(swagger_editor.swagger)
+        if auth_properties.ResourcePolicy:
+            swagger_editor.add_resource_policy(auth_properties.ResourcePolicy)
+        # Assign the Swagger back to template
+        # self.definition_body = self._openapi_auth_postprocess(swagger_editor.swagger)
```
jlhood: Why is this commented out? Remove?

samtranslator/swagger/swagger.py
jlhood: I don't think my question has been addressed in the latest version of this PR. Are we validating custom_statements is a list? If not, this can blow up if custom_statements is not a list, e.g., dictionary, string.

tests/translator/input/api_with_resource_policy_global.yaml
jlhood: You should contact them to update their docs to add this extension.

tests/translator/input/api_with_resource_policy_global_intrinsic.yaml
```diff
@@ -0,0 +1,26 @@
+Globals:
```
jlhood: This file should be renamed: `api_with_resource_policy_global_intrinsic` -> `api_with_resource_policy_global_implicit`

samtranslator/swagger/swagger.py
keetonian: True. This only works if `custom_statements` is an iterable. If it's anything else, this code will fail.

examples/2016-10-31/api_resource_policy/template.yaml
praneetap: I changed it to allowing access based on the IP address.

samtranslator/model/api/api_generator.py
praneetap: looks like a merge issue, will fix

samtranslator/swagger/swagger.py
```diff
@@ -788,6 +789,37 @@ def add_models(self, models):
...
+        :return:
+        """
+        if resource_policy is None:
+            return
+
+        custom_statements = resource_policy.get('CustomStatements')
+
+        if custom_statements is not None:
+            if not isinstance(custom_statements, list):
+                raise ValueError("Invalid input. CustomStatements must be a list.")
```
keetonian: Did we want to enforce this? I think the swagger property requires a list (can't tell for sure though, since they don't document it), but it might be nice if we were to wrap this into a list instead of fail.

If we are going to fail here, could you add an error case test to make sure that we catch the ValueError and don't just throw a 5xx error that gets returned as an "Internal server error" to the user?

praneetap: I changed it to wrap into a list instead of throwing an error.

Issue #1058, opened 2019-08-01
iot_backend Event properties are incorrect


singledigit:
**Description:**

iot_backend example does not function properly.

**Steps to reproduce the issue:**
1. cd examples/2016-10-31/iot_backend
2. run `sam validate -t template.yaml`
3.

**Observed result:**

Error: [InvalidResourceException('IoTLambdaFunctionIoT', "Type of property 'AwsIotSqlVersion' is invalid.")] ('IoTLambdaFunctionIoT', "Type of property 'AwsIotSqlVersion' is invalid.")

**Expected result:**

template.yaml is a valid SAM Template

jlhood:
@singledigit Hi Eric! Thanks for opening this issue and submitting a PR to fix it! 😊

ShreyaGangishetty:
Closing this issue as v1.15.0 is released

PR #1059, opened 2019-08-01
Fixed iot_backend example

singledigit:
*Issue #, if available:*
#1058

*Description of changes:*
Updated function event properties to remove 'AwsIotSqlVersion' property. Also removed key: id from Table as it will be auto created. Updated to nodejs10.x

sam validate -t template.yaml

*Checklist:*

- [na ] Write/update tests
- [ ] `make pr` passes
- [X] Update documentation
- [X] Add/update example to `examples/2016-10-31`

### DIFF
--- a/examples/2016-10-31/iot_backend/template.yaml
+++ b/examples/2016-10-31/iot_backend/template.yaml
@@ -19,6 +19,7 @@ Resources:
         IoT:
           Type: IoTRule
           Properties:
+            AwsIotSqlVersion: '2016-03-23'
             Sql: "SELECT * FROM 'iot2ddb'"
   Table:
     Type: AWS::Serverless::SimpleTable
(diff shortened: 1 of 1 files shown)


examples/2016-10-31/iot_backend/template.yaml
```diff
@@ -19,11 +19,6 @@ Resources:
         IoT:
           Type: IoTRule
           Properties:
-            AwsIotSqlVersion: 2016-03-23
```
jlhood: To fix the specific error you call out in the issue, you don't have to remove this property, you can use quotes to make it explicit that it's a string:

```yaml
            AwsIotSqlVersion: '2016-03-23'
```

singledigit: Making changes based on feedback and will submit

singledigit:
Updated and pushed for review

PR #1060, opened 2019-08-01
Adding delimiter between the hash data components.

praneetap:
*Issue #, if available:*
#191 
*Description of changes:*
Adding a delimiter between the hash components as best practice.
make pr 
*Checklist:*

- [x] Write/update tests
- [x] `make pr` passes
- [ ] Update documentation

### DIFF
--- a/samtranslator/model/apigateway.py
+++ b/samtranslator/model/apigateway.py
@@ -62,6 +62,8 @@ class ApiGatewayAccount(Resource):
 
 
 class ApiGatewayDeployment(Resource):
+    _X_HASH_DELIMITER = '||'
+
     resource_type = 'AWS::ApiGateway::Deployment'
     property_types = {
             'Description': PropertyType(False, is_str()),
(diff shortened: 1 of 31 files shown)


samtranslator/model/apigateway.py
```diff
@@ -90,9 +92,10 @@ def make_auto_deployable(self, stage, openapi_version=None, swagger=None):
         # to prevent redeployment when API has not changed
 
         # NOTE: `str(swagger)` is for backwards compatibility. Changing it to a JSON or something will break compat
+        # NOTE: If extending the hash data, make sure to add delimiter between the previous data and the new data.
         hash_input = str(swagger)
         if openapi_version:
-            hash_input = hash_input + str(openapi_version)
+            hash_input = hash_input + self._X_HASH_DELIMITER + str(openapi_version)
```
jlhood: Put the hash inputs into an array (only adding openapi_version in the if), then use [join()](https://www.programiz.com/python-programming/methods/string/join).

praneetap: done

PR #1077, opened 2019-08-09
feat: Resource policy Iam, Vpc and Ip whitelist/blacklist support

praneetap:
*Issue #, if available:*
#514 
*Description of changes:*
Adding support in SAM to create resource policies for IAM/IP/VPC whitelist/blacklist.
still WIP
*Checklist:*

- [x] Write/update tests
- [x] `make pr` passes
- [ ] Update documentation


praneetap:
Some things that are still needed to be done - 
1. Add method/path to the Resource section of the resource policy
1. Validate that if there is a Vpc allow/deny list then we have to make sure the endpoint is PRIVATE (Its a private API)
1. Verify what different policies can be set in what combinations (iam+ip, ip+vpc, iam+vpc, iam+ip+vpc) by deploying on CFN

samtranslator/model/api/api_generator.py
```diff
@@ -326,7 +326,9 @@ def _add_auth(self):
             self._set_default_apikey_required(swagger_editor)
 
         if auth_properties.ResourcePolicy:
-            swagger_editor.add_resource_policy(auth_properties.ResourcePolicy)
+            for path in swagger_editor.iter_on_path():
+                swagger_editor.add_resource_policy(auth_properties.ResourcePolicy, path,
```
keetonian: This is for the case where a resource policy is added to a function event, correct? This particular path needs to be able to have a policy on a particular path and not apply to the entire API.

samtranslator/swagger/swagger.py
```diff
@@ -799,26 +802,193 @@ def add_resource_policy(self, resource_policy):
...
-            if self.resource_policy.get('Statement') is None:
-                self.resource_policy['Statement'] = custom_statements
-            else:
-                statement = self.resource_policy['Statement']
-                if isinstance(statement, list):
-                    statement.extend(custom_statements)
-                else:
-                    statement = [statement]
-                    statement.extend(custom_statements)
+        if iam_deny_list is not None:
```
keetonian: For this section (allow and deny lists): does order matter? Do we need to allow then deny or deny then allow? I've seen some policies that deny everything then only allow a few things, and I'm not quite sure how that works.

samtranslator/swagger/swagger.py
```diff
@@ -799,26 +802,193 @@ def add_resource_policy(self, resource_policy):
...
+            self.resource_policy['Statement'] = policy_statement
+        else:
+            statement = self.resource_policy['Statement']
+            if not isinstance(statement, list):
+                statement = [statement]
+            statement.extend(policy_statement)
+            self.resource_policy['Statement'] = statement
+
+    def _get_method_path_uri_list(self, path, api_id, stage):
+        # It turns out that APIGW doesn't like trailing slashes in paths (#665)
```
keetonian: Could you turn this into a docstring? `"""`

samtranslator/swagger/swagger.py
```diff
@@ -799,26 +802,193 @@ def add_resource_policy(self, resource_policy):
...
+        # and removes as a part of their behaviour, but this isn't documented.
+        # The regex removes the tailing slash to ensure the permission works as intended
+        # At this point, value of Swagger path should be a dictionary with method names being the keys
```
keetonian: Will need to look more into this- what if there's a CFN condition on one of the paths? Try adding a test that has a condition on a function with an api event and another function without a condition that also has an api event.

samtranslator/swagger/swagger.py
```diff
@@ -799,26 +802,193 @@ def add_resource_policy(self, resource_policy):
...
+        methods = list(self.get_path(path).keys())
+        uri_list = []
+        path = re.sub(r'{([a-zA-Z0-9._-]+|proxy\+)}', '*', path)
```
keetonian: It looks like this regex is copied from elsewhere, could we keep the regex in one place as a variable and use that variable in multiple places?

tests/translator/input/api_with_resource_policy.yaml
```diff
@@ -9,4 +9,19 @@ Resources:
               Action: 'execute-api:Invoke', 
               Resource: ['execute-api:/*/*/*']
             }
+  ImplicitApiFunction:
```
keetonian: Thank you for adding functions to these tests, it helps to have an API definition generated in the tests. We missed that last open api review.

tests/translator/input/api_with_resource_policy_global.yaml
```diff
@@ -12,27 +12,22 @@ Globals:
...
+            RestApiId:
+              Ref: ExplicitApi
+            Path: /
+            Method: get
+
+  ExplicitApi:
     Type: AWS::Serverless::Api
     Properties:
       StageName: Prod
-      DefinitionBody: {
```
keetonian: You removed the explicit definition body, but it might be nice to have one test that has an explicit definition body that we can test with.

samtranslator/swagger/swagger.py
praneetap: I would need a sample template for this.

samtranslator/swagger/swagger.py
praneetap: I don't believe order matters for policy documents.

keetonian: Yeah looked into it- it appears order does not matter.

tests/translator/output/api_with_ip_deny_list.json
```diff
@@ -0,0 +1,206 @@
...
+                        "__Stage__": "Prod", 
+                        "__ApiId__": "ServerlessRestApi"
+                      }
+                    ]
+                  }
+                ], 
+                "Effect": "Allow", 
+                "Principal": {
+                  "AWS": [
+                    "12345"
```
jlhood: Ok now I see why you included IAM whitelist along with your IP deny... This is pretty interesting. The way API GW resource policies work is pretty confusing when you're hand-editing the aspen policy, but the SAM shorthand syntax used in certain combinations could lead customers to think they've secured their API when they've really opened it up to any IAM user. For example, any resource policy that adds Allow Principal: "*", but mixes that with IAM auth of specific accounts could be problematic, right (I'm actually not sure what the API GW behavior is in this case)? We'll need to discuss this with security. SAM may need to do some checking and disallow certain combinations of the shorthand resource policy types.

praneetap: Thanks for the valuable feedback @jlhood! Working my way through these...should update soon!

praneetap: > Ok now I see why you included IAM whitelist along with your IP deny... This is pretty interesting. The way API GW resource policies work is pretty confusing when you're hand-editing the aspen policy, but the SAM shorthand syntax used in certain combinations could lead customers to think they've secured their API when they've really opened it up to any IAM user. For example, any resource policy that adds Allow Principal: "*", but mixes that with IAM auth of specific accounts could be problematic, right (I'm actually not sure what the API GW behavior is in this case)? We'll need to discuss this with security. SAM may need to do some checking and disallow certain combinations of the shorthand resource policy types.

I just checked the behavior on the console - It throws an error "Invalid policy document. Please check the policy syntax and ensure that Principals are valid." When my policy statement looks like this - 
```json
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Principal": {
                "AWS": [
                    "arn:aws:iam::{{otherAWSAccountID}}:root",
                    "arn:aws:iam::{{otherAWSAccountID}}:user/{{otherAWSUserName}}",
                    "arn:aws:iam::{{otherAWSAccountID}}:role/{{otherAWSRoleName}}"
                ]
            },
            "Action": "execute-api:Invoke",
            "Resource": [
                "execute-api:/{{stageNameOrWildcard*}}/{{httpVerbOrWildcard*}}/{{resourcePathOrWildcard*}}"
            ]
        },
        {
            "Effect": "Deny",
            "Principal": "*",
            "Action": "execute-api:Invoke",
            "Resource": "execute-api:/{{stageNameOrWildcard}}/{{httpVerbOrWildcard}}/{{resourcePathOrWildcard}}",
            "Condition" : {
                "IpAddress": {
                    "aws:SourceIp": [ "{{sourceIpOrCIDRBlock}}", "{{sourceIpOrCIDRBlock}}" ]
                }
            }
        },
        {
            "Effect": "Allow",
            "Principal": "*",
            "Action": "execute-api:Invoke",
            "Resource": "execute-api:/{{stageNameOrWildcard}}/{{httpVerbOrWildcard}}/{{resourcePathOrWildcard}}"
        }

    ]
}
```

Given this, do we validate on SAM or do you think that this error is sufficient?

praneetap:
@keetonian @jlhood this is ready for your review again. Tried to address most of the comments the best I could!

samtranslator/swagger/swagger.py
keetonian: Add to the top of the template:
```yaml
Conditions:
  C1:
    Fn::Equals:
      - true
      - true
```

Then on any function, add under the logical ID:
```yaml
Condition: C1
```

samtranslator/model/eventsources/push.py
```diff
@@ -637,6 +639,15 @@ def _add_swagger_integration(self, api, function):
             if method_authorizer or apikey_required_setting is not None:
                 editor.add_auth_to_method(api=api, path=self.Path, method_name=self.Method, auth=self.Auth)
 
+            if self.Auth.get('ResourcePolicy'):
+                resource_policy = self.Auth.get('ResourcePolicy')
+                """
+                At this point the stage name resolution hasnt occured yet so we are going to apply
+                the resource policy on all stages of the API
```
jlhood: This comment no longer seems accurate.

tests/translator/output/api_with_ip_deny_list.json
keetonian: Which principal did it say was invalid?

samtranslator/model/eventsources/push.py
```diff
@@ -449,6 +449,7 @@ class Api(PushEventSource):
 
             # Api Event sources must "always" be paired with a Serverless::Api
             'RestApiId': PropertyType(True, is_str()),
+            'Stage': PropertyType(False, is_str()),
```
keetonian: Is this the only way to get the Stage here? This surfaces the Stage parameter to the template author, they can then override this value to be whatever they want.

tests/translator/output/api_with_ip_deny_list.json
jlhood: Ok great. If API GW's control plane is doing the error checking, I don't think we need to add extra error checking in the SAM layer. Thanks for verifying that!

samtranslator/model/eventsources/push.py
praneetap: ah, of course. will take care of it

samtranslator/swagger/swagger.py
```diff
@@ -800,26 +803,206 @@ def add_resource_policy(self, resource_policy):
...
+        uri_list = []
+        path = SwaggerEditor.get_path_without_trailing_slash(path)
 
-            self._doc[self._X_APIGW_POLICY] = self.resource_policy
+        for m in methods:
+            method = '*' if (m.lower() == self._X_ANY_METHOD or m.lower() == 'any') else m.upper()
+
+            # RestApiId can be a simple string or intrinsic function like !Ref. Using Fn::Sub will handle both cases
+            resource = '${__ApiId__}/' + '${__Stage__}/' + method + path
+            partition = ArnGenerator.get_partition_name(None)
```
keetonian: Do you think we should keep doing this (I know we do it elsewhere), or should we use ${AWS::Partition} pseudo parameter?

samtranslator/model/eventsources/push.py
```diff
@@ -637,6 +639,15 @@ def _add_swagger_integration(self, api, function):
             if method_authorizer or apikey_required_setting is not None:
                 editor.add_auth_to_method(api=api, path=self.Path, method_name=self.Method, auth=self.Auth)
 
+            if self.Auth.get('ResourcePolicy'):
+                resource_policy = self.Auth.get('ResourcePolicy')
+                """
+                At this point the stage name resolution hasnt occured yet so we are going to apply
+                the resource policy on all stages of the API
+                """
+                editor.add_resource_policy(resource_policy=resource_policy,
```
keetonian: `condition` is in scope (see line 595), you should be able to pass the `condition` in to the path and set it on the resource policy inside of an Fn::If statement.

You'll have to test if `AWS::NoValue` works inside of a resource policy, I assume that it will.

samtranslator/swagger/swagger.py
praneetap: I think we should change that in one big sweeping change in all tests.

samtranslator/model/eventsources/push.py
praneetap: We don't modify the stage property here. Is there a better way?

PR #1090, opened 2019-08-16
docs: Clarify `AWS::Serverless::SimpleTable.PrimaryKey` Default Value

chrisoverzero:
In the prose of the description for `PrimaryKey`, the default key name is said to be "ID". In the examples and in the translator tests, the default value is "id".

May no one else be bitten by this.

*Description of changes:*

In short, I made the normative and descriptive texts match.

- Changed "ID" to "`id`" in the prose of the specification. This is a non-normative change, as the specified default is in the "Primary Key Object" section -- and it matches the normative value anyway.
- Added text specifying the default `Type` value. (`String`) This one _is_ a normative change, but it's adding something to the specification text which was already specified.

I checked the values in the normative part of the specification and the output of the translator tests.

*Checklist:*

- [x] Update documentation

### DIFF
--- a/versions/2016-10-31.md
+++ b/versions/2016-10-31.md
@@ -313,7 +313,7 @@ The `AWS::Serverless::SimpleTable` resource creates a DynamoDB table with a sing
 
 Property Name | Type | Description
 ---|:---:|---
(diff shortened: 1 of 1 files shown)


chrisoverzero:
Thanks so much!

PR #1113, opened 2019-08-31
Add codecommit policy unit tests

ArendAMZN:
*Issue #, if available:*

*Description of changes:*

This adds unit tests that should have been included in the previous PR
https://github.com/awslabs/serverless-application-model/pull/1101

I also updated the unit test error messaging to be more readable. Previously it would output the entire output json blob and it was difficult to find the error. Now it shows a diff. See below for example of issue I was running into

```
E       AssertionError: assert {'Resources':...:IAM::Role'}}} == {'Resources': ...:IAM::Role'}}}
E         Differing items:
E         {'Resources': {'KitchenSinkFunction': {'Properties': {'Code': {'S3Bucket': 'sam-demo-bucket', 'S3Key': 'hello.zip'}, '...icy50'}, {'PolicyDocument': {...}, 'PolicyName': 'KitchenSinkFunctionRolePolicy51'}, ...]}, 'Type': 'AWS::IAM::Role'}}} != {'Resources': {'KitchenSinkFu
nction': {'Properties': {'Code': {'S3Bucket': 'sam-demo-bucket', 'S3Key': 'hello.zip'}, '...icy50'}, {'PolicyDocument': {...}, 'PolicyName': 'KitchenSinkFunctionRolePolicy51'}, ...]}, 'Type': 'AWS::IAM::Role'}}}
E         Use -v to get the full diff

tests/translator/test_translator.py:197: AssertionError
---------------------------------------------------------------------------------------------------------------------------------- Captured stdout call -----------------------------------------------------------------------------------------------------------------------------------
--- expected
+++ actual
@@ -17,8 +17,8 @@
         "Runtime": "python2.7", 
         "Tags": [
           {
-            "Value": "SAM", 
-            "Key": "lambda:createdBy"
+            "Value": "SAM"
           }
         ]
       }
@@ -221,9 +221,7 @@
                   "Resource": {
                     "Fn::Sub": [
                       {
-                        "repositoryName": {
-                          "Ref": "RepositoryName"
-                        }
+                        "repositoryName": "name"
                       }, 
                       "arn:${AWS::Partition}:codecommit:${AWS::Region}:${AWS::AccountId}:${repositoryName}"
                     ]

```

*Checklist:*

- [Y] Write/update tests
- [Y] `make pr` passes
- [n/a] Update documentation
- [n/a] Verify transformed template deploys and application functions as expected
- [n/a] Add/update example to `examples/2016-10-31`


tests/translator/test_translator.py
```diff
@@ -349,14 +350,26 @@ def test_transform_success_openapi3(self, testcase, partition_with_region):
             output_fragment = transform(
                 manifest, parameter_values, mock_policy_loader)
 
-        print(json.dumps(output_fragment, indent=2))
```
jlhood: Thanks for adding the diff output, but this print is convenient when developing new features. As long as the output template is correct, we can copy/paste it as the output file for our unit tests.

jlhood: I tried adding back just the print, but we have this print logic in multiple functions in this test file, and I don't have time to do the refactor to include diff output everywhere. Since this change is to `test_transform_success_openapi3()`, which isn't even relevant to this PR (policy templates are tested by `test_transform_success()`), I went ahead and reverted the change to `test_templates.py` so we can get these policy template unit tests merged.

Issue #1127, opened 2019-09-07
[Custom Domains][M1] Implement multiple basepath mappings for single domain single api.


praneetap:
## Related Issues
#783 
## Description 
This is a proposal to add Basepath mappings for Custom Domains usecase.

```yaml
  Type: AWS::Serverless::Api
  Properties:
    Domain:
      DomainName: example.com
      Certificate: !Ref CertificateARN 
      BasePath: [pathone, pathtwo] # '/' is no longer a valid basepath if you have multiple. They all map to Api.Stage.
```

SAM must create a `AWS::ApiGateway::BasePathMapping` resource per base path.
SAM must also add the basepath to the swagger.

Issue #1126, opened 2019-09-07
[Custom Domains][M1] Implement the Edge endpoint for CustomDomains


praneetap:
## Related Issues
#783 
## Description 
This is a proposal to implement the handling EDGE endpoint for Custom Domains (with no Route53 support)
```yaml
  Type: AWS::Serverless::Api
  Properties:
    Domain:
      DomainName: example.com
      CertificateArn: !Ref CertificateARN
      EndpointConfiguration: EDGE 
```

The SAM generated resources that should change are -
```yaml
  APIDomainName:
    Type: 'AWS::ApiGateway::DomainName'
    Properties:
      EndpointConfiguration:
        Types: [EDGE]
      CertificateArn: !Ref CertificateARN
      DomainName: !Ref DomainName
```

Issue #1125, opened 2019-09-07
[Custom Domains][M1]Implement the base case


praneetap:
## Related Issues
#783 
## Description 
This is a proposal to implement the base case of Custom Domains feature.
SAM input -
```yaml
  Type: AWS::Serverless::Api
  Properties:
    Domain:
      DomainName: example.com
      CertificateArn: !Ref CertificateARN 
```
This takes in a valid certificate ARN and creates a Regional endpoint mapping of the custom domain to the API basepath `/`. The CFN resources that are created are - 
```yaml
  APIDomainName:
    Type: 'AWS::ApiGateway::DomainName'
    Properties:
      EndpointConfiguration:
        Types: [REGIONAL]
      RegionalCertificateArn: !Ref CertificateARN
      DomainName: !Ref DomainName

  APIBasePathMapping:
    Type: AWS::ApiGateway::BasePathMapping
    Properties:
      DomainName: !Ref APIDomainName
      RestApiId: !Ref MyApi
      BasePath: /
      Stage: Prod
```
Apart from template creation, couple more in Acceptance Criteria : 
1. This must add base path and server fields to swagger/openapi doc. 
1. Also add this to the hash so that API GW redeploys if there is a change in base path or domain resources. (Since changing the base path does not update the swagger every time)

deleugpn:
Is this still waiting to be released? Does it work with API Gateway Http Api?

keetonian:
@deleugpn it is going out with v1.20.0. 

Http Api: custom domains work a little differently with Http Apis, that hasn't been implemented yet.

PR #1144, opened 2019-09-17
feat: custom domains Api Gateway support

praneetap:
*Issue #, if available:*
#1125, #1126, #1127 
*Description of changes:*
Adding basic custom domains handling to SAM. This only creates ApiGateway resources with a single basepath mapping to '/' 
I found that I don't need to edit the swagger, APIGateway does this for me automatically when I create the BasePath Mappings.
Tested deploying on CFN

*Checklist:*

- [x] Write/update tests
- [x] `make pr` passes
- [x] Update documentation


samtranslator/model/apigateway.py
```diff
@@ -153,6 +156,26 @@ def _status_code_string(self, status_code):
         return None if status_code is None else str(status_code)
 
 
+class ApiGatewayDomainName(Resource):
+    resource_type = 'AWS::ApiGateway::DomainName'
+    property_types = {
+            'RegionalCertificateArn': PropertyType(False, is_str()),
+            'DomainName': PropertyType(True, is_str()),
```
jlhood: Does `is_str()` on a lot of these properties mean intrinsics won't be supported?

docs/cloudformation_compatibility.rst
```diff
@@ -184,6 +184,7 @@ MinimumCompressionSize              All
 Cors                                All
 TracingEnabled                      All
 OpenApiVersion                      None
+Domain                              All
```
jlhood: Are you sure this is true?

samtranslator/model/api/api_generator.py
```diff
@@ -204,20 +206,81 @@ def _construct_stage(self, deployment, swagger):
...
         return stage
 
+    def _construct_api_domain(self, rest_api):
+        """
+        Constructs and returns the ApiGateway Domain and BasepathMapping
+        """
+        if self.domain is None or \
+           self.domain.get('DomainName') is None or \
+           self.domain.get('CertificateArn') is None:
+            return None, None
```
jlhood: I'm wary of silent failures like this. So if I specify `DomainName` but not `CertificateArn` it'll deploy successfully, but not have my custom domain? That seems very frustrating from a customer standpoint. If certain properties are required, we should throw a 400 back to the customer so they can understand what's wrong with their template, right?

samtranslator/model/api/api_generator.py
```diff
@@ -204,20 +206,81 @@ def _construct_stage(self, deployment, swagger):
...
+
+        logical_id = logical_id_generator.LogicalIdGenerator("", self.domain).gen()
+
+        domain = ApiGatewayDomainName('ApiGatewayDomainName' + logical_id,
+                                      attributes=self.passthrough_resource_attributes)
+        domain.DomainName = self.domain.get('DomainName')
+        endpoint = self.domain.get('EndpointConfiguration')
+
+        if endpoint is None or endpoint not in ['EDGE', 'REGIONAL']:
```
jlhood: Again, should we be silently falling back to default values? What if they use an intrinsic function in EndpointConfiguration? I don't think it's fair to say we support all intrinsics for the Domain property. What if I wanted to do something like this?

```yaml
    EndpointConfiguration: !If
      - SomeCondition
      - REGIONAL
      - EDGE
```

samtranslator/model/api/api_generator.py
```diff
@@ -204,20 +206,81 @@ def _construct_stage(self, deployment, swagger):
...
+            basepaths = None
+
+        basepath_resource_list = []
+
+        if basepaths is None:
+            basepath_mapping = ApiGatewayBasePathMapping(self.logical_id + 'BasePathMapping',
+                                                         attributes=self.passthrough_resource_attributes)
+            basepath_mapping.DomainName = self.domain.get('DomainName')
+            basepath_mapping.RestApiId = ref(rest_api.logical_id)
+            basepath_mapping.Stage = ref(rest_api.logical_id + '.Stage')
```
jlhood: Does this work? I thought <api logical id>.Stage is a special SAM parameter. CloudFormation won't understand that. Will this be correctly translated to the right actual logical name in the output template?

samtranslator/model/api/api_generator.py
```diff
@@ -204,20 +206,81 @@ def _construct_stage(self, deployment, swagger):
...
+            basepath_resource_list.extend([basepath_mapping])
+        else:
+            for path in basepaths:
+                path = path.strip('/')
```
jlhood: Can a base path have multiple slashes in it? Like, basepath = /foo/bar/? If so, then this won't be a valid logical id

samtranslator/model/api/api_generator.py
```diff
@@ -204,20 +206,81 @@ def _construct_stage(self, deployment, swagger):
...
+        domain.EndpointConfiguration = {"Types": [endpoint]}
+
+        # Create BasepathMappings
+        if self.domain.get('BasePath') and isinstance(self.domain.get('BasePath'), string_types):
+            basepaths = [self.domain.get('BasePath')]
+        elif self.domain.get('BasePath') and isinstance(self.domain.get('BasePath'), list):
+            basepaths = self.domain.get('BasePath')
+            # Cloudformation restricts you from adding '/' with other paths. '/' is also the default basepath value
+            if '/' in basepaths:
+                basepaths = None
```
jlhood: Again, it seems like we're making decisions on behalf of the customer here. I'm not sure I understand the implications of this well enough to say that we're making the right decision.

samtranslator/model/apigateway.py
```diff
@@ -95,6 +95,9 @@ def make_auto_deployable(self, stage, openapi_version=None, swagger=None):
         hash_input = [str(swagger)]
         if openapi_version:
             hash_input.append(str(openapi_version))
+        if domain:
+            hash_input.append(self._X_HASH_DELIMITER)
```
jlhood: Really happy to see you're ensuring we trigger a new deployment on custom domain changes! But why are you adding an extra delimiter here? The join below will add the delimiter between all entries in the list.

Also, we need to be careful here with doing simple if checks for appending to the hash. What if we had 2 different parameters, x and y that are part of the hash with logic like this:

```python
    if x:
        hash_input.append(x)
    if y:
        hash_input.append(y)
```

Then I have a template where I set x to 'foo' and don't set y. Then later, I remove the x property and set y to 'foo'. The deployment hash will be the same even though I've made an actual change to my template. We should be building this array such that we avoid this problem. For example, we could avoid the above scenario by doing this:

```python
    if x:
        hash_input.append('x:{}'.format(x))
    if y:
        hash_input.append('y:{}'.format(y))
```

Whatever we do, we need to make it backwards-compatible with the current behavior, but moving forward, we should do something like this with new values that we decide to add to the hash array.

samtranslator/model/sam_resources.py
```diff
@@ -526,13 +527,17 @@ def to_cloudformation(self, **kwargs):
...
-        rest_api, deployment, stage, permissions = api_generator.to_cloudformation()
+        rest_api, deployment, stage, permissions, domain, basepath_mapping = api_generator.to_cloudformation()
 
         resources.extend([rest_api, deployment, stage])
         resources.extend(permissions)
-
+        if domain:
+            resources.extend([domain])
+        if basepath_mapping:
+            resources.extend(basepath_mapping)
```
jlhood: Doesn't it need to be this?

```suggestion
            resources.extend([basepath_mapping])
```

versions/2016-10-31.md
```diff
@@ -233,7 +237,7 @@ AccessLogSetting | [CloudFormation AccessLogSetting property](https://docs.aws.a
 CanarySetting | [CloudFormation CanarySetting property](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-apigateway-stage-canarysetting.html) | Configure a Canary Setting to a Stage of a regular deployment. This value is passed through to Cloudformation, so any value supported by `CanarySetting` is supported here.
 TracingEnabled | `boolean` | Indicates whether active tracing with X-Ray is enabled for the stage.
 Models | `List of JSON or YAML objects` | JSON schemas that describes the models to be used by API methods.
-
+Domain | `Object` | Configuration settings for custom domains on API. Must contain `DomainName` and `CertificateArn`
```
jlhood: We usually don't just say `Object` for the type when it's an object we're defining. Can you do something similar to the `Cors` property and define a `Domain Configuration` object so you can specify the parameters on that object in the spec?

samtranslator/model/api/api_generator.py
praneetap: Yes, Cloudformation also waits for the stage to be deployed in this case, which ensures correct dependency among the resources.

samtranslator/model/sam_resources.py
praneetap: It always returns a list

samtranslator/model/api/api_generator.py
praneetap: Agreed, totally missed this. Will add the error

samtranslator/model/api/api_generator.py
praneetap: Will add a test

samtranslator/model/api/api_generator.py
praneetap: If you don't specify a basepath, cfn defaults to '/'. If you specify basepaths like so - ['/', 'get', 'foo'] api gw errors out with the message - "Only one base path mapping is allowed if the base path is empty.". It only allows one domain to one api if there is an empty basepath. We can either throw an error here, or ignore other paths of '/' is specified. What is your recommendation?

samtranslator/model/apigateway.py
praneetap: > Really excited to see movement on custom domain support! Lots of questions around edge case behavior. I don't think a single happycase test is sufficient here.
> 
> Also, are you planning to add an example as part of this PR?

I probably should add an example. Let me add that in the next rev thanks!

jlhood: Great, thanks! We've received feedback from customers that they really value the examples so we should definitely be adding them for new features.

samtranslator/model/api/api_generator.py
jlhood: I see, that makes sense. I've definitely been bitten by SAM lacking support for `!If` on certain parameters so I'd like to see that supported here, ideally. I'm fine with having EndpointConfiguration be optional defaulting to REGIONAL. However, I'm not ok with us defaulting to REGIONAL if someone puts `EndpointConfiguration: INVALID` in their template. I'd rather we throw an error in that case.

samtranslator/model/api/api_generator.py
jlhood: Again, I don't like us trying to auto-correct customer template errors. We should either throw an error or pass it through to API GW to throw an error, whichever makes more sense.

samtranslator/model/apigateway.py
jlhood: Ah ok, then this case is probably satisfied and we should only apply my solution to strings or other properties that could be ambiguous.

samtranslator/model/api/api_generator.py
```diff
@@ -204,20 +206,81 @@ def _construct_stage(self, deployment, swagger):
...
+        else:
+            basepaths = None
```
ShreyaGangishetty: Here, it looks like `basepaths = None` in both if and else section. instead of if-else we can just have `basepaths = None`. I am not sure if i am understanding this correct

samtranslator/model/api/api_generator.py
praneetap: Argh you are right, I wonder how I should name this logical id.

samtranslator/model/api/api_generator.py
praneetap: I think I will change it to passing it to API GW to error out.

docs/cloudformation_compatibility.rst
praneetap: I added the intrinsic support now and added a test for it.

Jacco:
Just a note: I have a template with two AWS::Serverless::Api that share the same AWS::ApiGateway::DomainName each with a different AWS::ApiGateway::BasePathMapping.

I guess this solution does not support that?

Also found that using custom domains the AWS::ApiGateway::BasePathMapping needs a DepensOn on de Stage generated by AWS::Serverless::Api for succesful deploying

    DependsOn:
      - BackofficeAPIStage

praneetap:
> Just a note: I have a template with two AWS::Serverless::Api that share the same AWS::ApiGateway::DomainName each with a different AWS::ApiGateway::BasePathMapping.
> 
> I guess this solution does not support that?
> 
> Also found that using custom domains the AWS::ApiGateway::BasePathMapping needs a DepensOn on de Stage generated by AWS::Serverless::Api for succesful deploying
> 
> ```
> DependsOn:
>   - BackofficeAPIStage
> ```

No unfortunately this doesn't support referencing an existing DomainName.

BasePathMapping references Api.Stage (this is a referable property exposed by SAM), which creates the necessary dependency chain. The BasePathMapping resource looks like this - 
```yaml
MyApiBasePathMapping:
    Type: AWS::ApiGateway::BasePathMapping
    Properties:
      DomainName: !Ref APIDomainName
      RestApiId: !Ref MyApi
      BasePath: one
      Stage: !Ref MyApi.Stage ## this creates the dependency
```

samtranslator/model/apigateway.py
```diff
@@ -76,7 +76,7 @@ class ApiGatewayDeployment(Resource):
         "deployment_id": lambda self: ref(self.logical_id),
     }
 
-    def make_auto_deployable(self, stage, openapi_version=None, swagger=None):
+    def make_auto_deployable(self, stage, openapi_version=None, swagger=None, domain=None):
         """
         Sets up the resource such that it will trigger a re-deployment when Swagger changes
         or the openapi version changes.
```
keetonian: nit: add "domain" to this comment.

praneetap:
> Overall, looks good. This might be a silly question, but can you have multiple AWS::ApiGateway::DomainName resources that all have the same domain name + certificate but different logical Id's?

No, you can only create one ApiGateway DomainName per domain "www.example.com". CloudFormation deploy fails if you try to create another ApiGateway DomainName resource.

Issue #1129, opened 2019-09-07
[Custom Domains][M2] Support creating Route53 IpV6 records


praneetap:
## Related Issues
#783 
## Description 
Milestone 2 targets adding Route53 support to Custom Domains feature. This task is to track adding Route 53 type AAAA records to a provided HostedZone by adding a IpV6 flag to SAM template.
SAM input
```yaml
  Type: AWS::Serverless::Api
  Properties:
    Domain:
      DomainName: example.com
      Certificate: !Ref MyApiSimpleDomainCertificate 
      EndpointConfiguration: REGIONAL # this is default, so same as not having this here
      Route53:
        EvaluateTargetHealth: true #false by default
        HostedZoneId: xyz # required, must be string
        IpV6: true # false by default
Distribution:
  Type: AWS::CloudFront::Distribution # Must have a cloudfront distribution set up
  ....
```
New Resource created should look like this -
```yaml
  DomainNameIpV6Route53RecordSet:
    Type: AWS::Route53::RecordSetGroup
    Properties:
      HostedZoneId: !Ref HostedZoneId
      RecordSets:
      - Name: !Ref DomainName
        Type: A
        AliasTarget:
          HostedZoneId: Z2FDTNDATAQYW2 # (const value for cloudfront distributions)
          DNSName: !GetAtt Distribution.DomainName
      - Name: !Ref DomainName
        Type: AAAA
        AliasTarget:
          HostedZoneId: Z2FDTNDATAQYW2 # (const value for cloudfront distributions)
          DNSName: !GetAtt Distribution.DomainName
```

Issue #1128, opened 2019-09-07
[Custom Domains][M2] Support creating Route53 IpV4 records


praneetap:
## Related Issues
#783 
## Description 
Milestone 2 targets adding Route53 support to Custom Domains feature. This task is to track adding Route 53 type A records to a provided HostedZone. If using EDGE, the `AWS::CloudFront::Distribution` resource must be defined in the same template.
SAM input -
```yaml
  Type: AWS::Serverless::Api
  Properties:
    Domain:
      DomainName: example.com
      Certificate: !Ref MyCertificate 
      EndpointConfiguration: REGIONAL | EDGE 
      Route53:
        EvaluateTargetHealth: true #false by default, must be false for EDGE
        HostedZoneId: xyz # required
```

Resource changes - Must create `AWS::Route53::RecordSetGroup` 
REGIONAL - 
```yaml
  APIDomainName:
    Type: 'AWS::ApiGateway::DomainName'
    Properties:
      EndpointConfiguration:
        Types: [REGIONAL] 
      RegionalCertificateArn: !Ref Certificate
      DomainName: !Ref DomainName

  DomainNameTypeARoute53RecordSet:
    Type: AWS::Route53::RecordSetGroup
    Properties:
      HostedZoneId: !Ref HostedZoneId
      RecordSets:
        - Name: !Ref DomainName
          Type: A
          AliasTarget:
            DNSName: !GetAtt APIDomainName.RegionalDomainName
            HostedZoneId: !GetAtt APIDomainName.RegionalHostedZoneId
```
EDGE -
```yaml
  APIDomainName:
    Type: 'AWS::ApiGateway::DomainName'
    Properties:
      EndpointConfiguration:
        Types: [EDGE]
      CertificateArn: !Ref CertificateARN
      DomainName: !Ref DomainName
  DomainNameTypeARoute53RecordSet:
    Type: AWS::Route53::RecordSetGroup
    Properties:
      HostedZoneId: !Ref HostedZoneId
      RecordSets:
      - Name: !Ref DomainName
        Type: A
        AliasTarget:
          HostedZoneId: Z2FDTNDATAQYW2 # (const value for cloudfront distributions)
          DNSName: !GetAtt Distribution.DomainName

PR #1165, opened 2019-09-27
feat: Custom domains route53 support

praneetap:
*Issue #, if available:*
#1128  #1129 
*Description of changes:*
Adding Route53 resource generation for custom domains feature.
unit tests only so far.
*Checklist:*

- [x] Write/update tests
- [x] `make pr` passes
- [x] Update documentation


praneetap:
This is not ready for review yet, until the following- 
1. First pr of custom domains is through
1. I have tested end to end
1. I have added docs
1. I have added examples

samtranslator/model/api/api_generator.py
```diff
@@ -276,7 +280,58 @@ def _construct_api_domain(self, rest_api):
...
+
+    def _construct_record_sets_for_domain(self, domain):
+        recordset_list = []
+        recordset = {}
+        route53 = domain.get('Route53')
+
+        recordset['Name'] = domain.get('DomainName')
+        recordset['Type'] = 'A'
+        recordset['AliasTarget'] = self._construct_alias_target(self.domain)
+        recordset_list.extend([recordset])
```
keetonian: Why extend here? Are we going to make recordset into a list later? I don't see any existing values for recordset_list.

samtranslator/model/api/api_generator.py
```diff
@@ -276,7 +280,58 @@ def _construct_api_domain(self, rest_api):
...
+
+        recordset_ipv6 = {}
+        if route53.get('IpV6') is not None and route53.get('IpV6') is True:
+            recordset_ipv6['Name'] = domain.get('DomainName')
+            recordset_ipv6['Type'] = 'AAAA'
+            recordset_ipv6['AliasTarget'] = self._construct_alias_target(self.domain)
+            recordset_list.extend([recordset_ipv6])
```
keetonian: Again, why extend instead of append?

samtranslator/model/api/api_generator.py
```diff
@@ -218,23 +219,26 @@ def _construct_api_domain(self, rest_api):
...
+            return None, None, None
 
         if self.domain.get('DomainName') is None or \
            self.domain.get('CertificateArn') is None:
             raise InvalidResourceException(self.logical_id,
                                            "Custom Domains only works if both DomainName and CertificateArn"
                                            " are provided")
 
-        logical_id = logical_id_generator.LogicalIdGenerator("", self.domain).gen()
+        self.domain['ApiDomainName'] = "{}{}".format('ApiGatewayDomainName',
```
keetonian: I don't remember from the last PR- is it possible to reference the domain resource by using <ApiName>.Domain? If not, we might want to add support for that in a future PR for this feature.

Also, I'm still not sure if I like this name but can't think of a better one so it will probably stick 👍

samtranslator/model/api/api_generator.py
praneetap: Just to keep it consistent and future proof (in case we add more types of recordsets in the future)

samtranslator/model/sam_resources.py
```diff
@@ -489,6 +489,7 @@ class SamApi(SamResourceMacro):
     referable_properties = {
         "Stage": ApiGatewayStage.resource_type,
         "Deployment": ApiGatewayDeployment.resource_type,
+        "DomainName": ApiGatewayDomainName.resource_type
     }
```
praneetap: @keetonian You can refer the domainname created with `Api.DomainName` because of this line here.

samtranslator/model/api/api_generator.py
praneetap: Yes, that change is in this PR.

Yeah, naming is a hard one. Let me know if you come up with something better.

PR #1170, opened 2019-09-30
Updated MethodSettings from API Gateway

alexfrosa:
Make it clear that the property MethodSettings is a list.

### DIFF
--- a/versions/2016-10-31.md
+++ b/versions/2016-10-31.md
@@ -8,7 +8,7 @@ The AWS Serverless Application Model (SAM) is licensed under [The Apache License
 
 ## Table of contents
 * [Introduction](#introduction)
- * [Specification](#specification)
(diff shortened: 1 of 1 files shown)


versions/2016-10-31.md
```diff
@@ -8,7 +8,7 @@ The AWS Serverless Application Model (SAM) is licensed under [The Apache License
 
 ## Table of contents
 * [Introduction](#introduction)
-* [Specification](#specification)
+ * [Specification](#specification)
```
praneetap: please get rid of the indent here.

praneetap:
@alexfrosa Thanks for fixing the docs! I just have one very minor comment.

versions/2016-10-31.md
alexfrosa: done

Issue #1141, opened 2019-09-13
RFC: Usage Plans for Auth


ShreyaGangishetty:
**Resources**
[Api keys](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-setup-api-key-with-console.html), [create and use Api keys with Usage Plans](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-api-usage-plans.html), [Auth](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/serverless-controlling-access-to-apis.html)

**Associated Issues**
[#248](https://github.com/awslabs/serverless-application-model/issues/248), [#547](https://github.com/awslabs/serverless-application-model/issues/547)

**Description**
Api keys can be generated by creating a Usage Plan or by using existing Usage Plan.

The template for Usage plans in SAM would look like:
```yaml
  ...
      UsagePlan:  
        CreateUsagePlan: SHARED  # Accepted values: SHARED | NONE | SINGLE # Required field
        UsagePlanId: !Ref usageplanId # Optional field
  ...
```
* SAM creates three resources, `AWS::ApiGateway::ApiKey` with name `ServerlessApiKey`, `AWS::ApiGateway::UsagePlan` with name `ServerlessUsagePlan`, and `AWS::ApiGateway::UsagePlanKey` with name `ServerlessUsagePlanKey` for a stack when `CreateUsagePlan` is set to `SHARED` in Globals / Api.
* SAM creates three resources, `AWS::ApiGateway::ApiKey` with name `<ApiName>ApiKey`, `AWS::ApiGateway::UsagePlan` with name `<ApiName>UsagePlan`, and `AWS::ApiGateway::UsagePlanKey` with name `<ApiName>UsagePlanKey` for an Api when `CreateUsagePlan` is set to `SINGLE` in Globals / Api.
* SAM creates `AWS::ApiGateway::ApiKey` and `AWS::ApiGateway::UsagePlanKey` when `UsagePlanId` is given. SAM will not create/update `AWS::ApiGateway::UsagePlan` resource and expects `ApiStages` to be defined by the user seperately.

**Input to SAM**
```yaml
Globals:
  Api:
    Auth:
      ApiKeyRequired: true
      UsagePlan:
        CreateUsagePlan: SHARED

Resources:
  ...
  MyApi:
    Type: AWS::Serverless::Api
    Properties:
      StageName: Prod

  MyDifferentApi:
[...]

PR #1179, opened 2019-10-07
feat: usage plans support for Api Auth

ShreyaGangishetty:
*Issue #, if available:*
[#1141](https://github.com/awslabs/serverless-application-model/issues/1141), [#248](https://github.com/awslabs/serverless-application-model/issues/248), [#547](https://github.com/awslabs/serverless-application-model/issues/547)

*Description of changes:*
added support for usage plans

added test input templates and also deployed the generated templates in CloudFormation and validated
*Checklist:*

- [x] Write/update tests
- [x] `make pr` passes
- [x] Update documentation


examples/2016-10-31/usage_plan/template.yaml
```diff
@@ -0,0 +1,98 @@
...
+     Type: AWS::Serverless::Api
+     Properties:
+       StageName: Prod
+
+   MyApiTwo:
+     Type: AWS::Serverless::Api
+     Properties:
+       StageName: Prod
+       Auth:
+         ApiKeyRequired: true
```
praneetap: you shouldnt need this to be defined here again if its in globals

samtranslator/model/api/api_generator.py
```diff
@@ -459,6 +469,129 @@ def _add_auth(self):
...
+        if auth_properties.UsagePlan is None:
+            return []
+        usage_plan_properties = auth_properties.UsagePlan
+
+        # throws error if the property invalid/ unsupported for UsagePlan
+        if not all(key in UsagePlanProperties._fields for key in usage_plan_properties.keys()):
+            raise InvalidResourceException(
+                self.logical_id, "Invalid property for 'UsagePlan'")
+
+        usage_plan_type = usage_plan_properties.get('CreateUsagePlan')
```
praneetap: please call this create_usage_plan for readability

samtranslator/model/api/api_generator.py
```diff
@@ -459,6 +469,129 @@ def _add_auth(self):
...
+        usage_plan = None
+
+        # doesn't allow giving UsagePlanId when CreateUsagePlan is SINGLE or SHARED
+        if usage_plan_type is None:
+            raise InvalidResourceException(self.logical_id, 'Parameter "CreateUsagePlan" is a required parameter '
+                                                            'for "UsagePlan" but was not provided.')
+
```
praneetap: Add a check 
```
if create_usage_plan not in ['SHARED', 'SINGLE', 'NONE]:
  <raise error>
```

tests/translator/input/api_with_usageplans.yaml
```diff
@@ -0,0 +1,152 @@
...
+
+   MyApiTwo:
+     Type: AWS::Serverless::Api
+     Properties:
+       StageName: Prod
+       Auth:
+         ApiKeyRequired: true
+         UsagePlan:
+          CreateUsagePlan: NONE
+          UsagePlanId: !Ref MyApiTwoUsagePlan
```
keetonian: My thought with this-

I thought that, if `CreateUsagePlan: None` is set, we won't create the `UsagePlanKey` or `ApiKey`. Or, did we decide to create those if the `UsagePlanId` is set, and if that property isn't set, we don't do anything?

I want to make sure for the case where, if someone has set up their UsagePlan, ApiKey, and UsagePlanKey, we don't mess things up and they can turn off the auto-generated SAM portions.

ShreyaGangishetty: I have documented this [here](https://github.com/awslabs/serverless-application-model/pull/1179/files#diff-ebb9eb355c6f4c24d565c7e8517549a5R920). 

When `CreateUsagePlan:NONE` it works similar to `CreateusagePlan:SINGLE` where we create `ApiKey and UsagePlanApiKey` and expect the `UsagePlan` to be defined by the user. If the user has already set up the three resources (they don't want to create a usage plan) they will not add `UsagePlan` section for that `Api`. One problem here would be if `CreateUsagePlan` is added in the globals section we cannot turn it off for a single Api at the `Resources` level. The user has to either give a `UsagePlanId` or allow us to create the `UsagePlan`. The workaround for this would be to add `CreateUsagePlan` for the required Apis at the `Resource` level instead of adding in the `Globals` section.
Currently, this code throws error when `CreateUsagePlan:None` is given and `usagePlanId` is not given. Instead of throwing error I could do:
1) create the `ApiKey`, `UsagePlanApiKey` when `UsagePlanId` is given
2) do not create any resource when `UsagePlanId` is not given

samtranslator/model/api/api_generator.py
```diff
@@ -459,6 +469,129 @@ def _add_auth(self):
...
+        """
+        if not self.auth:
+            return []
+        auth_properties = AuthProperties(**self.auth)
+        if auth_properties.UsagePlan is None:
+            return []
+        usage_plan_properties = auth_properties.UsagePlan
+
+        # throws error if the property invalid/ unsupported for UsagePlan
+        if not all(key in UsagePlanProperties._fields for key in usage_plan_properties.keys()):
```
praneetap: we don't error on unsupported properties today, so we shouldnt be adding this check here. We only validate that we have the required properties set.

examples/2016-10-31/usage_plan/template.yaml
ShreyaGangishetty: Will update the template

samtranslator/model/api/api_generator.py
ShreyaGangishetty: We are throwing errors for [invalid Auth properties](https://github.com/awslabs/serverless-application-model/blob/master/samtranslator/model/api/api_generator.py#L312). So, I have thrown errors for UsagePlan properties as well

examples/2016-10-31/usage_plan/template.yaml
```diff
@@ -0,0 +1,96 @@
+Conditions:
+  C1:
+    Fn::Equals:
+      - true
+      - true
```
keetonian: I don't think we need this condition in the examples, but it's great to include in the tests.

examples/2016-10-31/usage_plan/template.yaml
```diff
@@ -0,0 +1,96 @@
...
+     Properties:
+       StageName: Prod
+
+   MyApiTwo:
+     Type: AWS::Serverless::Api
+     Properties:
+       StageName: Prod
+       Auth:
+         UsagePlan:
+          CreateUsagePlan: SHARED
```
keetonian: Does this not create a UsagePlan now?

examples/2016-10-31/usage_plan/template.yaml
```diff
@@ -0,0 +1,96 @@
...
+
+   MyApiTwoUsagePlan:
+      Type: AWS::ApiGateway::UsagePlan
+      DependsOn:
+        - MyApiTwo
+      Properties:
+        ApiStages:
+          - ApiId: !Ref MyApiTwo
+            Stage: !Ref MyApiTwo.Stage
```
keetonian: This doesn't get linked, is there a reason why we're explicitly creating this?

examples/2016-10-31/usage_plan/template.yaml
```diff
@@ -0,0 +1,96 @@
...
+        ApiKeyRequired: true
+        UsagePlan:
+          CreateUsagePlan: !Ref UsagePlanType
+
+Resources:
+   MyApiOne:
+     Condition: C1
+     Type: AWS::Serverless::Api
+     Properties:
+       StageName: Prod
```
keetonian: Might be out of scope for this example, since this doesn't use a usage plan at all.

samtranslator/model/api/api_generator.py
keetonian: I like this here 👍

samtranslator/model/api/api_generator.py
```diff
@@ -518,6 +532,134 @@ def _add_auth(self):
...
+            usage_plan_key = self._construct_usage_plan_key(usage_plan_logical_id, create_usage_plan, api_key)
+
+        if usage_plan_properties.get("UsagePlanName"):
+            usage_plan.UsagePlanName = usage_plan_properties.get("UsagePlanName")
+        if usage_plan_properties.get("Description"):
+            usage_plan.Description = usage_plan_properties.get("Description")
+        if usage_plan_properties.get("Quota"):
+            usage_plan.Quota = usage_plan_properties.get("Quota")
+        if usage_plan_properties.get("Tags"):
+            usage_plan.Tags = usage_plan_properties.get("Tags")
```
keetonian: We might want to add any tags added to the API to this Usage Plan as well, what do you think?

examples/2016-10-31/usage_plan/template.yaml
ShreyaGangishetty: Thanks for catching this. You are right this doesn't get used anywhere, I will remove

examples/2016-10-31/usage_plan/template.yaml
ShreyaGangishetty: it creates a usage plan with logical id  `ServerlessUsagePlan`

samtranslator/model/api/api_generator.py
ShreyaGangishetty: a usage plan can contain multiple apis so i think its better not to propagate api tags to usage plan

samtranslator/model/api/api_generator.py
```diff
@@ -518,6 +532,134 @@ def _add_auth(self):
...
+        create_usage_plan = usage_plan_properties.get("CreateUsagePlan")
+        usage_plan = None
+        api_key = None
+        usage_plan_key = None
+
+        if create_usage_plan is None:
+            raise InvalidResourceException(self.logical_id, "'CreateUsagePlan' is a required field for UsagePlan")
+        if create_usage_plan not in ["SHARED", "PER_API", "NONE"]:
+            raise InvalidResourceException(
+                self.logical_id, "'CreateUsagePlan' accepts only NONE, PER_API and SHARED values"
```
praneetap: Make it extensible by putting it in a list and printing the list here.

Issue #1187, opened 2019-10-08
add EventBridgePutEventsPolicy policy template


jlhood:
**Description:**

We should add a policy template for putting Amazon EventBridge events. We can call it EventBridgePutEventsPolicy and it should allow the `events:PutEvents` action. From the [EventBridge IAM documentation](https://docs.aws.amazon.com/IAM/latest/UserGuide/list_amazoneventbridge.html), it looks like you can't scope the permission down to a specific event bus, although that doesn't sound right so I'm doublechecking on that. However, if that turns out to be the case, this policy template would take no parameters and specify `Resource: *`.

Here's an example of a past PR that added a new policy template for reference: #904

Russ-K:
Hi @jlhood ,
I've been looking at this and, while the AWS console complains that:
> The actions in your policy do not support resource-level permissions and require you to choose All resources
I'm absolutely able to create a policy in CloudFormation with a specific EventBus as the resource so I'm unsure of which way to go with this.
I'll see if I can actually use that policy and / or if it restricts writing to other buses but I'm also a little reluctant to have the new policy specify a resource in case that behaviour is removed.
Thoughts?

jlhood:
@Russ-K I appreciate you following up on this! I'm still trying to get more information internally on plans to support a specific event bus resource within an IAM policy. However, that's mainly because I haven't had time to test it myself. I'd love it if you could test and verify if restricting PutEvents permissions to a specific event bus is currently honored by the service. Knowing the current behavior would help us move forward with this change. If restricting PutEvents permissions to a specific event bus is supported, we should be able to rely on that functionality staying. Again, I'll work on confirming from my side, but if you could run some tests to understand the current behavior, that would be terrific!

jlhood:
Thanks @mikedeck. Looks like this issue is blocked until we get this support. I don't want to create a policy template that has no parameters and uses `Resource: *` only to have it become obsolete later if the service adds per-resource permissions.

jmnarloch:
The API should now support the resource level policy in all regions. A parameter can be now added to specify the event bus name in the policy.

chrisoverzero:
@jmnarloch Can I ask how you figured that out? The [IAM page for EventBridge][] still lists no resource type for `events:PutEvents`.

[IAM page for EventBridge]: https://docs.aws.amazon.com/IAM/latest/UserGuide/list_amazoneventbridge.html

jmnarloch:
[...]

PR #1197, opened 2019-10-14
add EventBridgePutEventsPolicy policy template

Russ-K:
*Issue #, if available:*
#1187 

*Description of changes:*
Added EventBridgePutEventsPolicy policy for adding events to an EventBus

*Checklist:*

- [x] Write/update tests
- [x] `make pr` passes
- [ ] Update documentation

### DIFF
--- a/examples/2016-10-31/policy_templates/all_policy_templates.yaml
+++ b/examples/2016-10-31/policy_templates/all_policy_templates.yaml
@@ -96,3 +96,6 @@ Resources:
 
         - CodeCommitReadPolicy:
             RepositoryName: name
+
+        - EventBridgePutEventsPolicy:
+            EventBusName: name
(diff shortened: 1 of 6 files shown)


Russ-K:
From the document [here](https://docs.aws.amazon.com/eventbridge/latest/userguide/eventbridge-cross-account-event-delivery.html), the section "Migrate a Sender-Receiver Relationship to Use AWS Organizations" specifies an EventBus as a resource in a separate account. So it is possible to specify a resource.

However, from investigating further, it is not possible to create a policy with just PutEvents while specifying a specific resource using the AWS console.

jlhood:
@Russ-K Why close this? Even if the policy template doesn't take a specific event bus as a parameter and sets `Resource: *`, it's still helpful to have the policy template to simplify the permissions experience for SAM customers.

Russ-K:
@jlhood There were a number of commits that weren't necessary if using `Resource: *` so I'm going to resubmit it.

PR #1230, opened 2019-10-31
fix: adding support to point to api gateway distribution 

praneetap:
*Issue #, if available:*
Found an issue where we don't support using the distribution supported by apigateway for custom domains.
*Description of changes:*

validated end to end
*Checklist:*

- [x] Write/update tests
- [x] `make pr` passes
- [x] Update documentation


examples/2016-10-31/custom_domains_with_route53/README.md
```diff
@@ -19,3 +18,5 @@ $ sam deploy \
     --stack-name my-new-stack \
     --capabilities CAPABILITY_IAM
 ```
+
+Curl to the endpoint "http://example.com/home/fetch" should hit the Api.
```
keetonian: `home` is the base path and `fetch` is the api path, correct? That's what it looks like looking at the example template, just want to make sure I understand this.

Issue #854, opened 2019-03-14
Feature Request: Add SQS option to SNS event


jlhood:
**Description:**

SAM offers simplified syntax for SNS -> Lambda and with #261 supports adding subscription filter policies. However, a common use case is SNS -> SQS -> Lambda. SAM offers simplified syntax for SQS -> Lambda, but you have to use standard CloudFormation to setup the SNS -> SQS connection and subscription filter policy, which can be verbose.

SAM could add a lot of value by adding an option to the existing SNS event type to create an SQS queue.

Here's an example that uses simple default settings for the queue:

```yaml
Type: SNS
Properties:
  Topic: arn:aws:sns:us-east-1:123456789012:my_topic
  # new property that creates an SQS queue between the topic and Lambda function
  # When only the boolean is specified, SAM uses the following defaults:
  # AWS::SQS::Queue logicalId: <function logicalId><SNS event key>Queue
  # AWS::SQS::QueuePolicy logicalId: <function logicalId><SNS event key>QueuePolicy
  # BatchSize: default behavior - don't add to `AWS::Lambda::EventSourceMapping` resource so default value is used.
  # Enabled: default behavior - don't add to `AWS::Lambda::EventSourceMapping` resource so default value is used.
  SqsSubscription: true
  FilterPolicy:
    store: 
      - example_corp
    price_usd: 
      - numeric: 
          - ">="
          - 100
```

Here's an example that specifies additional options for the queue:

```yaml
Type: SNS
Properties:
  Topic: arn:aws:sns:us-east-1:123456789012:my_topic
  # new property that creates an SQS queue between the topic and Lambda function
  # Default values can be overridden by passing an object instead of a boolean. Supported keys:
  # QueuePolicyLogicalId: overrides default logicalId naming of the AWS::SQS::QueuePolicy resource.
  # QueueArn: allows the user to specify their own queue instead of having SAM create one on their behalf, allowing them to specify non-default properties on the queue.
  # BatchSize: if specified, add BatchSize property with given value to `AWS::Lambda::EventSourceMapping` resource.
  # Enabled: if specified, add BatchSize property with given value to `AWS::Lambda::EventSourceMapping` resource.
  SqsSubscription:
    QueuePolicyLogicalId: CustomQueuePolicyLogicalId
    QueueArn: !GetAtt MyCustomQueue.Arn
    BatchSize: 5
    Enabled: false
  FilterPolicy:
    store: 
      - example_corp
    price_usd: 
[...]

PR #1231, opened 2019-10-31
feat: Add an existing SQS queue option to SNS event

53ningen:
readable diff:
https://github.com/awslabs/serverless-application-model/pull/1231/files?w=1

*Issue #, if available:*
#854 

*Description of changes:*
Add an existing SQS queue option to SNS event

- make pr
- verify transformed template deploys and application functions as expected

*Checklist:*

- [x] Write/update tests
- [x] `make pr` passes
- [x] Update documentation


53ningen:
ready for review

samtranslator/model/eventsources/push.py
```diff
@@ -392,17 +392,42 @@ def to_cloudformation(self, **kwargs):
...
+            resources.append(queue_policy)
+            resources.append(subscription)
+            return resources
+
+        # SNS -> SQS(Existing) -> Lambda
         resources = []
-        queue = self._inject_sqs_queue()
-        queue_policy = self._inject_sqs_queue_policy(self.Topic, queue)
+        queue_arn = self.SqsSubscription.get('QueueArn', None)
+        queue_url = self.SqsSubscription.get('QueueUrl', None)
```
ShreyaGangishetty: If `QueueArn` and `QueueUrl` are not given then we will be adding empty Strings for `Resource` and `Queues` fields in `PolicyDocument`. Instead we need to throw errors if these fields are not given.

timoschilling:
The informations of `QueueArn` and `QueueUrl` seams to be redundant. Isn't it possible to build the `QueueUrl` from the `QueueArn`?

chrisoverzero:
@timoschilling Gominingen described why this was sometimes not possible (and why a separate URL can be a feature) [back in the original issue](https://github.com/awslabs/serverless-application-model/issues/854#issuecomment-548891190). In short, `QueueArn` is rarely going to be a string. It will almost always be an object with a key of `Fn::GetAtt`.

I can imagine that and similar being features of the next round of the, uh, feature:

- If no `QueueUrl` specified:
  - If `QueueArn` is string, construct URL as string.
  - If it is `Fn::GetAtt` object, get logical ID, construct `Ref` object.
  - Bail on other intrinsics. 

…which sound like really nice checks, assuming I haven’t forgotten why they’re impossible.

samtranslator/model/eventsources/push.py
53ningen: Fixed in 40cba76

tests/model/eventsources/test_sns_event_source.py
```diff
@@ -64,3 +64,29 @@ def test_to_cloudformation_passes_the_filter_policy(self):
...
         self.assertRaises(TypeError, self.sns_event_source.to_cloudformation)
+
+    def test_to_cloudformation_throws_when_queue_url_or_queue_arn_not_given(self):
+        sqsSubscription = {
+            'BatchSize': 5
+        }
+        self.sns_event_source.SqsSubscription = sqsSubscription
+        self.assertRaises(TypeError, self.sns_event_source.to_cloudformation)
+
+    def test_to_cloudformation_when_sqs_subscription_disable(self):
```
53ningen: > Could you please add/update any of the existing unit tests for a condition where SqsSubscritption: False.

I've added on this line.

53ningen:
@ShreyaGangishetty 
Thank very much you for reviewing this PR and I'm sorry for my late response.
I have addressed your pull request comments.

Ready for review again.

ljacobsson:
What's the status on this?

It looks like it was included in the v1.20.0 release, but I can't get it working and failing on [this line](https://github.com/aws/serverless-application-model/blob/bcb1aacc079bba29e16f9328105ede0551423625/samtranslator/model/eventsources/push.py#L427). I've even tried to define both `QueueArn` and `QueueUrl`

Also, there's no sign of this feature in the documentation.

chrisoverzero:
>Also, there's no sign of this feature in the documentation.

It is in [the documentation](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-property-function-sns.html) and specification.

Issue #634, opened 2018-10-23
Renaming a Serverless::Function used by an Api Gateway does not deploy the Gateway


jceloi:
**Description:**

Only renaming (no code change) a Serverless::Function using a FunctionName property does not trigger an Api Gateway stage deploy and the API Gateway stage still points to the previous lambda, which in that case does not exist anymore, generating an internal server error.

Manually triggering an API Gateway deploy does solve the problem, but according to me that should be done automatically when the FunctionName changes.

**Steps to reproduce the issue:**
1. Using cloudformation, create an AWS::Serverless::Api pointing at a AWS::Serverless::Function . Deploy it, check that the calls are correctly triggering the lambda.
2. Add a static FunctionName property to the Function, different than the one previously generated during the first deploy. Redeploy the stack.
3. Try to call the lambda through the api gateway again.

**Observed result:**
* Error 500, Internal Server Error code
* No event is routed to the lambda function

**Expected result:**
* Correct invocation of the lambda function

jvanbrunschot-coolblue:
We're running in the same issue. Renaming functions (without a code change) results in permission errors when invoking functions. 
`Execution failed due to configuration error: Invalid permissions on Lambda function`

brettstack:
You should be able to change the logical resource id instead of setting `FunctionName` (which is best practice) and should trigger an API Gateway deploy. We have also discussed adding a new property to *always* deploy your API even if we can't detect a change to the API.

keetonian:
Closed in favor of #660

eliottparis:
> You should be able to change the logical resource id instead of setting `FunctionName` (which is best practice) and should trigger an API Gateway deploy.

I'm facing the same issue here, but I need my `FunctionName` to be dynamically computed depending on an input template parameter.

```yml
Parameters:
  Stage:
    Type: String

  MyFunction:
    Type: AWS::Serverless::Function
    Properties:
      FunctionName: !Join [ "", [ "MyFunction-", !Ref Stage]]
```

#660 could be an option in the future, but is there a way to achieve that for now?

jlhood:
Actually, it looks like a draft PR had been put up to make this change, but wasn't completed. https://github.com/awslabs/serverless-application-model/pull/1134

PR #1235, opened 2019-11-07
fix: redeploy api when function name changes

ShreyaGangishetty:
*Issue #, if available:*
#634 
*Description of changes:*
Updated the code to take function name when redeploying api
This PR resolves only `Ref` and `Fn::Sub` intrinsics for parameter values

added a test which updates the deployment hash when functionname changes

*Checklist:*

- [x] Write/update tests
- [x] `make pr` passes


samtranslator/translator/translator.py
```diff
@@ -31,6 +31,39 @@ def __init__(self, managed_policy_map, sam_parser, plugins=None):
         self.plugins = plugins
         self.sam_parser = sam_parser
 
+    function_names = dict()
+
+    def _get_function_names(self, resource_dict, parameter_values):
```
ShreyaGangishetty: This function appends all the function names of an API as an API can have multiple functions. If atleast one function name changes, the api will be redeployed.

samtranslator/translator/translator.py
```diff
@@ -31,6 +31,39 @@ def __init__(self, managed_policy_map, sam_parser, plugins=None):
...
+        if resource_dict.get("Type", "").strip() == 'AWS::Serverless::Function':
+            if resource_dict.get('Properties', "").get('Events', ""):
+                events = list(resource_dict.get('Properties', "").get('Events', "").values())
+                for item in events:
+                    # If the function event type is `Api`then gets the function name and
+                    # adds to the function_names dict with key as the api_name and value as the function_name
+                    if item.get('Type', "") == 'Api' and item.get('Properties', None) and item.get('Properties', "")\
```
praneetap: You don't need ", None) in 'get', that's the default behavior

samtranslator/translator/translator.py
```diff
@@ -31,6 +31,39 @@ def __init__(self, managed_policy_map, sam_parser, plugins=None):
...
+                        if api_name:
+                            function_name = resource_dict.get('Properties', "").get("FunctionName", None)
+                            if type(function_name) == str:
+                                # All the function_names for associated functions will be concatenated
+                                self.function_names[api_name] = self.function_names.get(api_name, "") +\
+                                                            resource_dict.get('Properties', "").get('FunctionName', "")
+                            # Resolving intrinsics and gets the function name
+                            # This gets the function names only for Ref and Fn::Sub intrinsics
+                            elif type(function_name) == dict:
+                                if function_name.get('Ref'):
```
ShreyaGangishetty: This case works for `!Ref FunctionParameter`

samtranslator/translator/translator.py
```diff
@@ -31,6 +31,39 @@ def __init__(self, managed_policy_map, sam_parser, plugins=None):
...
+                                    self.function_names[api_name] = self.function_names.get(api_name, "") + \
+                                        parameter_values.get(function_name.get('Ref'))
+                                if function_name.get('Fn::Sub'):
```
ShreyaGangishetty: This resolves `Fn::Sub: FunctionName` to a string

samtranslator/model/apigateway.py
```diff
@@ -99,7 +101,12 @@ def make_auto_deployable(self, stage, openapi_version=None, swagger=None, domain
             hash_input.append(str(openapi_version))
         if domain:
             hash_input.append(json.dumps(domain))
-
+        if redeploy_restapi_parameters:
+            function_names = redeploy_restapi_parameters.get('function_names')
+        else:
+            function_names = None
+        if function_names and function_names.get(self.logical_id[:-10], None):
```
keetonian: Can you remind me again why `self.logical_id[:-10]` is needed here? Maybe add a comment about what you're trying to do?

samtranslator/translator/translator.py
```diff
@@ -31,6 +31,36 @@ def __init__(self, managed_policy_map, sam_parser, plugins=None):
         self.plugins = plugins
         self.sam_parser = sam_parser
 
+    def _get_function_names(self, resource_dict, parameter_values):
```
keetonian: Is it not possible to use the intrinsic resolver? This looks like a lot of duplicated effort.

Intrinsic resolver does additional things, like resolving "!Sub ${parameter}-name", which this doesn't appear to handle.

samtranslator/model/apigateway.py
ShreyaGangishetty: Sure. I will add. self.logical_id has the api_name and the word "deployment". I am removing the word "deployment" and just getting the api_name to search for corresponding string of function_name/names

samtranslator/translator/translator.py
ShreyaGangishetty: intrinsics resolver is not being called from this file. I will check and get back if i can call intrinsics resolver here.

samtranslator/translator/translator.py
ShreyaGangishetty: Thank you. I have updated the code without ""/None

patrickgreenwell:
This is still an issue if you reference the RestApiId via a Ref and have a FunctionName with a Sub in it. The OrderDict object seems to be more specific than needs be and templates can still fall into the else statement under the right conditions.  I've made a branch for a change that will resolve the issue.

PR #1258, opened 2019-11-21
Update policy_templates.rst

dalumiller:
The first sentence says that "SAM automatically creates the IAM Role required to run the function." But then the next sentence seems to negate that when it reads, "you need to give your function explicit permissions..." So, shouldn't the first sentence say "SAM does NOT automatically create the IAM Role" because you have to explicitly give it to your function?

*Issue #, if available:*

*Description of changes:* Documentation wording needs updating.

*Description of how you validated changes:* n/a

*Checklist:*

- [ ] Write/update tests
- [ ] `make pr` passes
- [ ] Update documentation

### DIFF
--- a/docs/policy_templates.rst
+++ b/docs/policy_templates.rst
@@ -1,7 +1,7 @@
 Policy Templates
 ================
 
-When you define a Serverless Function, SAM automatically creates the IAM Role required to run the function. Let's say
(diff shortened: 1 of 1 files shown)


keetonian:
SAM does create the IAM role needed to run the function, but doesn't automatically add all of the necessary policies for interacting with other AWS resources and services. This wording seems correct, but thank you for submitting the PR!

PR #1288, opened 2019-12-04
Amend provisioned concurrency examples and templates to work for false condition

me2resh:
*Issue #, if available:*
None

*Description of changes:*
The current examples and templates for lambda provisioned concurrency works fine for a true valid condition, but if the condition in the example is false, it will fail with and error because the parent property ProvisionedConcurrencyConfig will still be present, but will have no child configuration settings.

In the example templates, change the condition AliasProvisionedConcurrencyEnabled to false
Then when you try to deploy the template, cloudformation deploy will fail with the error:
```
Property ProvisionedConcurrentExecutions cannot be empty.
```

I also added a missing condition in the lambda_edge/template.yaml 

*Checklist:*

- [x] Write/update tests
- [x] `make pr` passes
- [x] Update documentation

### DIFF
--- a/examples/2016-10-31/lambda_edge/README.md
+++ b/examples/2016-10-31/lambda_edge/README.md
@@ -30,8 +30,10 @@ LambdaEdgeFunctionSample:
         Timeout: 5
         # More info at https://github.com/awslabs/serverless-application-model/blob/master/docs/safe_lambda_deployments.rst
         AutoPublishAlias: live 
-        ProvisionedConcurrencyConfig:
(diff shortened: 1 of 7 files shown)


praneetap:
Changing base to develop.

examples/2016-10-31/cloudwatch-event-to-msteams/template.yaml
```diff
@@ -45,6 +45,7 @@ Resources:
           Type: CloudWatchEvent
           Description: Detects EC2 Security Group Events to Send to Teams
           Properties:
+            EventBusName:  event-bus-name
```
praneetap: could you undo this change?

me2resh:
Hi @praneetap, the branch currently is rebased against the latest develop and has only my intended changes. please re-review.

PR #1299, opened 2019-12-05
feat: add sqs and sns policies based on destination config

ShreyaGangishetty:
*Issue #, if available:*
SAM expects the users to define `sqs:SendMessage` and `sns:publish` policies when `DestinationConfig` property is set for Kinesis and DynamoDb event types ([documentation](https://github.com/awslabs/serverless-application-model/blob/master/versions/2016-10-31.md#properties-8)).

*Description of changes:*
This PR removes the need to specify these policies explicitly if a property `Type` is specified by the user in `OnFailure` property of `DestinationConfig`.

If `Type` is not given users have to specify the policiles explicitly. Hence, this change is backwards compatible.

Deployed the template and verified if the right policies are added and also verified if the DestinationConfig is displayed without `Type` parameter in the console

*Checklist:*

- [x] Write/update tests
- [x] `make pr` passes
- [x] Update documentation


samtranslator/model/iam.py
```diff
@@ -63,3 +63,31 @@ def dead_letter_queue_policy(cls, action, resource):
                 }]
             }
         }
+
+    @classmethod
+    def sqs_send_message_role_policy(cls, queue_arn, logical_id):
+        document = {
+            'PolicyName': 'SQSPublishPolicy' + logical_id,
```
ShreyaGangishetty: Should I change the naming convention to logical_id + `SQSPublishPolicy`?

keetonian: What is the existing convention?

ShreyaGangishetty: existing convention is FunctionName+'integer number'. But I cannot follow this convention as it might conflict with the explicit `Policies` section.
`logical_id` is `FunctionName+EventName`

tests/translator/output/function_with_event_source_mapping.json
```diff
@@ -61,30 +64,35 @@
         ],
         "Policies": [
           {
-            "PolicyName": "MyFunctionForBatchingExampleRolePolicy0",
+            "PolicyName": "SQSPublishPolicyMyFunctionForBatchingExampleDynamoDBStreamEvent",
```
ShreyaGangishetty: `SQSPublishPolicyMyFunctionForBatchingExampleDynamoDBStreamEvent` will be changed as `MyFunctionForBatchingExampleDynamoDBStreamEventSQSPublishRolePolicy`

samtranslator/model/iam.py
ShreyaGangishetty: updated the policy name to <event-logicalid>+<sqspolicy/snspolicy>

Issue #1305, opened 2019-12-06
AutoPublishAlias is not triggered for environment variable changes


cakepietoast:
**Description:**

Changing or adding/removing environment variables for a function does not cause a new version to be published when using `AutoPublishAlias`. 

**Steps to reproduce the issue:**
1. Deploy a SAM template with an `AWS::Serverless::Function` resource, with `AutoPublishAlias` set, and including an environment variable:  
```
HelloWorldFunction:
    Type: AWS::Serverless::Function
    Properties:
      CodeUri: hello_world/
      Handler: app.lambda_handler
      Runtime: python3.8
      AutoPublishAlias: live
      Environment:
        Variables:
          TestKey: TestVal1
```  
2. Change `TestVal1` to `TestVal2`.  
3. Run sam deploy . 

**Observed result:**
The lambda function is updated with the new environment variable, but no new version is published, and the defined alias still points to the last published version.

**Expected result:**
A new function version is published, and the defined alias is updated to point to it.

keetonian:
If the lambda function is invoked, does it have the old or new value for the environment variable?

keetonian:
It looks like we can add the `Environment` dictionary to the hash that gets computed to determine if a new version needs to be published.

Section of code to update: https://github.com/awslabs/serverless-application-model/blob/master/versions/2016-10-31.md#awsserverlessapi

cakepietoast:
When invoking via api gateway, the "old" version gets invoked.

@szilveszter Did you paste the right url? Was expecting a link to a section of code but your link goes to documentation.

dalumiller:
@cakepietoast  @keetonian This is similar to the #1267 where environment variables stored in Secrets Manager that are changed don't trigger a new version of the lambda.

cakepietoast:
Closing this as it was resolved with #1310.

heyjustcode:
@cakepietoast this still doesn't work if the environment variable change was due to an updated value of a referenced parameter

Issue #413, opened 2018-05-02
AutoPublishAlias: Support updating version when things other than CodeUri Changes


rmmeans:
**Description:**

Per the documentation, AutoPublishAlias only supports publishing another function version when the CodeUri changes. While having AutoPublishAlias in its current state is still beneficial, I have found this has caused much confusion and even bugs that took quite some time to track down. Specifically, when things such as environment variables are changed on the stack.

The "gotcha" that occurs is because an environment variable might be passed as a stack parameter (for instance). When doing a stack update and updating the parameter, everything appears to execute cleanly. The stack change set picks up that the function needs to be updated, etc and the update is marked as successful. This causes the false belief that your update got applied successfully, only on close inspection of the lambda function do you discover that the lambda function was updated, but a new version was not tied to the alias.

I am curious if there is plans to remove this limitation? While the documentation does spell this out clearly, unless someone updating the stack knows that function(s) in the stack are using the `AutoPublishAlias` feature and knows about this caveat - they they might think an update was successful with no other action needed.

digioak:
Appears to be a bug in how we process `AutoPublishAlias`.  We're going to look deeper into a potential solution.

nick-romano-kci:
Just wanted to bump this, it's been an issue for us as well. Our issue is when changing memory allocations.

brettstack:
This is similar to https://github.com/awslabs/serverless-application-model/issues/479 for APIs only deploying when the Swagger changes. We do have a proposal for APIs to always deploy even when we can't detect a change https://github.com/awslabs/serverless-application-model/issues/660. However, I believe that Lambda won't allow you to create a new version of a Function if there have been no changes at all, making this solution not viable in this case.

We could detect changes to the entire `Function` configuration itself, however, we don't currently transform anything based on Parameter values which we would need to do to cover this issue completely.

lanefelker:
Any updates on this? Is there a solution to make a new lambda version created by environment variable change be tied to the alias?

chris-malley-alertme:
Also an issue for us; we use an environment variable to change log level and should to be able to update this without rebuilding the code.

nwb-ilan:
I don't understand the use for `AutoPublishAlias` if it doesn't support configuration changes such as env variable.  
The workaround that comes to mind is to `aws lambda update-alias` on all the functions in the stack to the `'$latest'` version after every `sam deploy`, but that defeats the purpose of having `AutoPublishAlias`.   
Am I missing something?

[...]

PR #1310, opened 2019-12-09
feat: Trigger auto publish alias on env var/memorysize change

cakepietoast:
*Issue #, if available:* #1305, #413 

*Description of changes:* Update how Logical ID is generated for lambda function version so it takes into account environment variables and memorysize configuration. Changes to these values will now cause an update to the alias when AutoPublishAlias is enabled on the function.

*Description of how you validated changes:* Translated (sam-translate.py) and deployed function several times with (and without) changing values for env vars and memorysize configuration to confirm expected functionality.

*Checklist:*

- [x] Write/update tests
- [x] `make pr` passes
- [ ] Update documentation


cakepietoast:
Travis build is running against master instead of develop.

PR #1319, opened 2019-12-12
updated the lambda permission with the correct case

tde908:
*Issue #, if available:*
The default policy action that get's added to the function permissions is in the incorrect case. It is being inserted as lambda:invokeFunction instead of lambda:InvokeFunction (note the i). 
This means the permissions don't work as expected. 

The correct case can be seen in the documentation here:
https://docs.aws.amazon.com/lambda/latest/dg/lambda-api-permissions-ref.html

When triggering the function via API gateway as a GET request the response returns as: 

<AccessDeniedException>
    <Message>Unable to determine service/operation name to be authorized</Message>
</AccessDeniedException>

This occurred when creating a CFN resource with Type: AWS::Serverless::Function

*Description of changes:*
I have updated the lambda action so that it is in the correct case: lambda:InvokeFunction

When creating the lambda the default policy is being added as: 
"Principal": {
  "Service": "apigateway.amazonaws.com"
},
"Action": "lambda:invokeFunction",

I have not updated the tests as of yet as there are a large number of files with the lowercase character. If you are happy with the change, please add a comment and I can go and update the remaining files before the change is merged. 


tde908:
Hi @ShreyaGangishetty ,
I have updated the rest of the tests to include the correct case. There are a few conflicts that need to be resolved, however I don't have the ability to update them myself. 

Is there anything else that you need from my end? I will be happy to help where possible. 

Cheers,
Tom

ShreyaGangishetty:
@tde908 I will check and resolve these merge conflicts. Thank you for submitting this PR

Issue #350, opened 2018-03-29
"Enabled" field in "DeploymentPreference" for "AWS::Serverless::Function" is ignored if it is a Condition


ksskarthik628:
I have a SAM template with Parameters, API, and Lambda, and I have three stacks with "Stage" parameter set as "dev-user1", "alpha", or "beta". Now, I am trying to update my lambda to support traffic shifting based on what stack it is as follows:
```
Conditions:
  IsBetaStage:
    Fn::Equals: [Ref: Stage, beta]
  IsServiceStage:
    Fn::Or: ["Fn::Equals": [Ref: Stage, alpha], Condition: IsBetaStage]
Resources:
  LambdaFunction:
    Type: AWS::Serverless::Function
    Properties:
      FunctionName: "LambdaFunction"
      Handler: 'com.company.lambda.LambdaHandler::handleRequest'
      Runtime: java8
      Timeout: 30
      Tracing: Active
      AutoPublishAlias: live
      DeploymentPreference:
        Enabled: IsServiceStage
        Type:
          Fn::If: [IsBetaStage, Linear10PercentEvery10Minutes, AllAtOnce]
        Alarms:
          - { Ref: IncreasedFaults }
          - { Ref: IncreasedLatency }
```
According to the template, "Enabled" should result to false in stack with "Stage" parameter set to "dev-user1" and true in every other stack, and hence there should be no traffic shifting in this stack. But I am observing the opposite of this, i.e, I am seeing traffic shifting in the stack with "Stage" parameter set to "dev-user1"

ksskarthik628:
I even tried this and it still didn't work:
```
Parameters:
  DevStage:
    Type: String
    Default: dev-user1
Conditions:
  IsBetaStage:
    Fn::Equals: [Ref: Stage, beta]
  IsServiceStage:
    Fn::Or: ["Fn::Equals": [Ref: Stage, alpha], Condition: IsBetaStage]
  IsNotDevStage:
    Fn::Not: ["Fn::Equals": [Ref: DevStage, dev-user1]]
Resources:
  LambdaFunction:
    Type: AWS::Serverless::Function
    Properties:
      FunctionName: "LambdaFunction"
      Handler: 'com.company.lambda.LambdaHandler::handleRequest'
      Runtime: java8
      Timeout: 30
      Tracing: Active
      AutoPublishAlias: live
      DeploymentPreference:
        Enabled: IsNotDevStage
        Type:
[...]

PR #1361, opened 2020-01-03
Add support for string/parameter based Enabled for DeploymentPreference

aketcham0691:
*Issue #, if available:*
#350 
*Description of changes:*
This change allows developers to reference string parameters to enable or disable deployment preferences. Previously, any non-empty string evaluated to true and any empty string evaluated to false.

Created new unit tests and ran make test

*Checklist:*

- [ x] Write/update tests
- [ ] `make pr` passes
`make pr` would not work, as `make black-check` was failing, even before making my changes. I tried using both python2.7 and python3.6. Any guidance here would be appreciated. 
- [ ] Update documentation


aketcham0691:
I see black-check is failing. I think you may want to update your onboarding doc/makefile to download the correct versions of python, black, etc.

I'll try to fix and re-submit.

Issue #1372, opened 2020-01-09
Issue with multiple VPCE's in SourceVpcWhitelist


keanesf:
**Description:**

If multiple VPCE's are specified in the SourceVpcWhitelist traffic from both VPCE's is denied. 

**Steps to reproduce the issue:**
1. Specify multiple VPCEs in the SourceVpcWhitelist and set the EndpointConfiguration to PRIVATE

Example

```yaml
Resources:        
  ApiGateway:
    Type: AWS::Serverless::Api
    Properties:
      StageName: Prod
      EndpointConfiguration: PRIVATE
      Auth:
        ResourcePolicy:
          SourceVpcWhitelist:
            - vpce-1
            - vpce-2
```

2. Send a request in which vpce-1 or vpce-2 is the sourceVpce and the response is an explict deny.  

**Observed result:**

Traffic is not permitted when multiple vpce's are whitelisted. If only once vpce is white listed the issue does not occur.  

**Expected result:**

Traffic is permitted from both whitelisted source vpce's and traffic is denied from a vpce not whitelisted.

duartemendes:
I've run into this issue a few weeks ago actually. The temporary fix is the same as [#1249](https://github.com/awslabs/serverless-application-model/issues/1249), you'll need to create a custom statement.

I think it would help if you added in your comment the API Resource Policy that was generated, it would be easier to understand the problem.

If I remember correctly the problem is that the conditions are being evaluated with an "and" instead of an "or". Meaning that, the incoming request needs to come from all vpc's, which is impossible.

Issue #1249, opened 2019-11-16
SourceVpcWhitelist is unable to resolve intrinsic functions


sanjP10:
**Description:**

Within the Auth object, trying to use SourceVpcWhitelist and finding that intrinsic functions are causing issues with cloudformation creating or updating a stack.

https://github.com/awslabs/serverless-application-model/blob/master/versions/2016-10-31.md#api-auth-object

```
...
Auth:
  DefaultAuthorizer: AWS_IAM
    ResourcePolicy:
      SourceVpcWhitelist:
        - !Ref VPCEndpoint
...
```

**Steps to reproduce the issue:**
1. Create an Serveless::API with a vpc or vpcendpoint
2. Put the above into the Auth section for resource policies

**Observed result:**
Transform AWS::Serverless-2016-10-31 failed with: Internal transform failure.

cfn-lint returns E0001 Error transforming template: expected string or bytes-like object.

**Expected result:**

Cloudformation updates stack and is able to resolve the intrinsic functions

**Note**
If you do not use intrinsic functions like !Sub or !Ref this works fine
I have not tried this on the other options. I have tried with CustomStatements and had no issues

duartemendes:
👋 
Any update on this?
I'm algo experiencing this issue, the workaround I have for now is an ugly custom statement.

`Auth:  
      ResourcePolicy:
        CustomStatements:
          - Effect: Allow
            Principal: '*'
            Action: execute-api:Invoke
            Resource: execute-api:/*
          - Effect: Deny
            Principal: '*'
            Action: execute-api:Invoke
            Resource: execute-api:/*
            Condition:
              StringNotEquals:
                aws:SourceVpce: !Ref VpcEndpoint`

henkesde:
Having the same issue. Going to hardcode the vpc until this is solved. I got lucky that it's not an issue for me yet.

keanesf:
I have ran into this issue as well.

ncanibano:
Facing this issue as well. My understanding is this PR: https://github.com/aws/serverless-application-model/pull/1395 already fixes the !Ref use-case, but using Fn::ImportValue is still not supported within a SourceVpcWhitelist, is that correct?

jfuss:
We created a set of new properties to support the Intrinsic use cases: https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-property-api-resourcepolicystatement.html#sam-api-resourcepolicystatement-IntrinsicVpcWhitelist

[...]

Issue #1218, opened 2019-10-23
Can not use Fn::if within CustomStatements:


eerb-kyruus:
**Description:**

I am attempting to write a custom policy for my Serverless api that depends on deployment type. The serverless transform expects a list inside the CustomStatements: label, preventing use of fn:If to conditionally set policies. The obvious work-around is to specify the entire swagger document myself, but this creates lots of duplication and a huge template.

**Steps to reproduce the issue:**
NOTE: Confidential parts of the template have been redacted
1. Create a serverless API template with this in the conditional section:
Conditions:
  IsProd:
     <redacted>
2. Create a conditional resource policy like the following in the AWS:Serverless:Api Auth: section:

```yaml
        ResourcePolicy:
           CustomStatements:
            Fn::If:
            - IsProd
            - - Principal: '*'
                Effect: Allow
                Action: <redacted>
                Resource: <redacted>
            - - Principal: '*'
                Effect: Allow
                Action: <redacted>
                Resource: <redacted>
                Condition:
                  <redacted>
```

**Observed result:**
The macro creates the following within the swagger doc, reformatted for readability. Note carefully the extra set of square brackets around the policy statement:

```json
"x-amazon-apigateway-policy":
  {
  "Version":"2012-10-17","Statement":
    [
      [
        {
        Condition":
          {
             <redacted>
         },
        "Action": <redacted>
        "Resource": <redacted>
        "Effect":"Allow",
        "Principal":"*"
        }
      ]
   ]
  },
```

The deploy fails due to that extra set of square brackets in the policy statement

**Expected result:**
The swagger gateway policy is created with a single set of square brackets and the deploy succeeds.

jlhood:
@eerb-kyruus Thanks for reporting this. Looks like this code isn't taking intrinsic functions into account:

https://github.com/awslabs/serverless-application-model/blob/440790a8c2278fc16acbe34c83183668c5f40c38/samtranslator/swagger/swagger.py#L987-L988

For now, a possible workaround would be to embed the `Fn::If` inside each statement, e.g.,:

```yaml
        ResourcePolicy:
           CustomStatements:
            - !If
              - IsProd
              - Principal: '*'
                Effect: Allow
                Action: <redacted>
                Resource: <redacted>
              - !Ref AWS::NoValue
            - !If
              - IsProd
              - Principal: '*'
                Effect: Allow
                Action: <redacted>
                Resource: <redacted>
                Condition:
                  <redacted>
[...]

PR #1395, opened 2020-01-15
fix: bug fixes in resource policies

praneetap:
Added support for intrinsics Ref for api auth
Added support for fn::if in CustomStatements
Fixed the policy getting created for vpc whitelist to be 'OR' instead of 'AND'

*Issue #, if available:*
#1249 #1372 #1218 
*Description of changes:*

*Checklist:*

- [ ] Write/update tests
- [ ] `make pr` passes
- [ ] Update documentation

### DIFF
--- a/samtranslator/model/intrinsics.py
+++ b/samtranslator/model/intrinsics.py
@@ -146,23 +146,6 @@ def is_instrinsic(input):
     return False
 
 
-def is_instrinsic_fn(input):
-    """
-    Checks if the given input is an intrinsic function dictionary. Intrinsic function is a dictionary with single
-    key that is the name of the intrinsics.
-
-    :param input: Input value to check if it is an intrinsic (all except Ref)
-    :return: True, if yes
-    """
-
-    if input is not None and isinstance(input, dict) and len(input) == 1:
-
-        key = list(input.keys())[0]
-        return key == "Condition" or key.startswith("Fn::")
-
-    return False
-
-
 def is_intrinsic_if(input):
     """
     Is the given input an intrinsic if? Intrinsic function 'if' is a dictionary with single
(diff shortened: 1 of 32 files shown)


samtranslator/model/intrinsics.py
```diff
@@ -146,6 +146,23 @@ def is_instrinsic(input):
...
+    key that is the name of the intrinsics.
+
+    :param input: Input value to check if it is an intrinsic (all except Ref)
+    :return: True, if yes
+    """
+
+    if input is not None and isinstance(input, dict) and len(input) == 1:
+
+        key = list(input.keys())[0]
+        return key == "Condition" or key.startswith("Fn::")
```
ShreyaGangishetty: If `!Sub` is given in the template does that get resolved to `Fn::Sub` at this point?

praneetap: sorry this change snuck in. Not using this new fn anywhere so will be getting rid of it

samtranslator/swagger/swagger.py
```diff
@@ -824,16 +823,32 @@ def add_resource_policy(self, resource_policy, path, api_id, stage):
...
+                if not is_intrinsic_no_value(if_list[1]):
+                    self._add_custom_statement(if_list[1])
+                    if_true = self.resource_policy
+                self.resource_policy = original
+                if not is_intrinsic_no_value(if_list[2]):
+                    self._add_custom_statement(if_list[2])
+                    if_false = self.resource_policy
+                self.resource_policy = make_conditional(if_list[0], if_true, if_false)
+            else:
+                self._add_custom_statement(custom_statements)
```
keetonian: Why do we need to do this? Does it not work to just pass it straight through and let CFN resolve the if statement?

praneetap: The merge between global custom statements and local with `if` won't be handled.

keetonian: Could you add a test case then that tests this flow? I don't see a test case that has both local and global custom statements.

PR #1396, opened 2020-01-15
fix: unhashable type: 'dict_node' when referencing RestApiId by Ref in Lambda event mapping on redeploy check

patrickgreenwell:
This handles the error of when referencing an API via Ref so that the dictionary check actually occurs. This makes the type checks equivalent and isn't as restrictive as the OrderedDict class. Helps for when things like cfn-lint pass in templates which is where I noticed the bug after updating to the latest version.

*Issue #, if available:*

*Description of changes:*

*Checklist:*

- [ ] Write/update tests
- [ ] `make pr` passes
- [ ] Update documentation

### DIFF
--- a/samtranslator/translator/translator.py
+++ b/samtranslator/translator/translator.py
@@ -1,6 +1,5 @@
 import copy
 from samtranslator.model import ResourceTypeResolver, sam_resources
-from collections import OrderedDict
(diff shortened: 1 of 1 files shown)


ShreyaGangishetty:
@patrickgreenwell  Thank you for fixing this!!

Issue #1402, opened 2020-01-20
Custom Domains - Support for HostedZoneName instead of HostedZoneId


mbarneyjr:
Support using [`HostedZoneName`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-route53-recordset.html#cfn-route53-recordset-hostedzonename) instead of [`HostedZoneId`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-route53-recordset.html#cfn-route53-recordset-hostedzoneid) in the `Domain` section of an `AWS::Serverless::Api`, like this:

```yml
Globals:
  Api:
    Domain:
      DomainName: www.domain.com
      CertificateArn: arn:aws:acm:us-east-1:000000000000:certificate/abcxyz
      EndpointConfiguration: REGIONAL
      Route53:
        HostedZoneName: www.domain.com
        DistributionDomainName: www.domain.com
```

This property would just need to be passed directly to the `HostedZoneName` property of the `AWS::Route53::RecordSet` created resource, since CloudFormation itself handles looking up the hosted zone ID.
> When you create a stack using an AWS::Route53::RecordSet that specifies HostedZoneName, AWS CloudFormation attempts to find a hosted zone whose name matches the HostedZoneName. If AWS CloudFormation cannot find a hosted zone with a matching domain name, or if there is more than one hosted zone with the specified domain name, AWS CloudFormation will not create the stack.

praneetap:
@mbarneyjr this sounds great! We can make `HostedZoneId` not required and always use the domain name.
If you are interested in contributing, the change would go [here](https://github.com/awslabs/serverless-application-model/blob/develop/samtranslator/model/api/api_generator.py#L341-L344)

mbarneyjr:
Submitted a PR #1408

PR #1408, opened 2020-01-22
Adds support for HostedZoneName instead of HostedZoneId in Domain section of the Api

mbarneyjr:
*Issue #, if available:*

Issue #1402 

*Description of changes:*

Adds support for inclusion of `HostedZoneName` instead of `HostedZoneId` in the `Domain` section of an `AWS::Serverless::Api`. This allows users to not need to worry about looking up the hosted zone ID, and letting CloudFormation handle it for them via the `AWS::Route53::RecordSetGroup` resource.

`HostedZoneName` is forwarded to the `RecordSetGroup` resource just as `HostedZoneId` is. 

The following should now be valid SAM:

```yml
Globals:
  Api:
    Domain:
      DomainName: www.my-domain.com
      CertificateArn: my-valid-cert-arn
      EndpointConfiguration: EDGE
      Route53:
        HostedZoneName: my-domain.com.
```

Not 100% sure how to validate the changes, didn't immediately see any documentation around it

*Checklist:*

- [x] Write/update tests
- [x] `make pr` passes (It doesn't have any more failures than what is on `master`)
- [x] Update documentation

### DIFF
--- a/examples/2016-10-31/custom_domains_with_route53_hosted_zone_name/template.yaml
+++ b/examples/2016-10-31/custom_domains_with_route53_hosted_zone_name/template.yaml
@@ -1,0 +1,71 @@
+Parameters:
+  DomainName:
+    Type: String
+    Default: 'example.com'
+  ACMCertificateArn:
+    Type: String
+    Default: 'cert-arn-in-us-east-1'
+Resources:
+  MyFunction:
+    Type: AWS::Serverless::Function
+    Properties:
+      InlineCode: |
+        exports.handler = async (event) => {
+          const response = {
+            statusCode: 200,
+            body: JSON.stringify('Hello from Lambda!'),
+          };
+          return response;
+        };
+      Handler: index.handler
+      Runtime: nodejs8.10
+      Events:
+        Fetch:
+          Type: Api
+          Properties:
+            RestApiId: !Ref MyApi
+            Method: Post
+            Path: /fetch
+
+  MyApi:
+    Type: AWS::Serverless::Api
+    Properties:
+      OpenApiVersion: 3.0.1
+      StageName: Prod
+      Domain:
+        DomainName: !Ref DomainName
+        CertificateArn: !Ref ACMCertificateArn
+        EndpointConfiguration: EDGE
+        BasePath:
+          - /home
+        Route53:
+         HostedZoneName: www.my-domain.com.
+         IpV6: true
+## ====== Everything below here is optional, leave this out if you want to use the internal Api Gateway distribution =======
+         DistributionDomainName: !GetAtt Distribution.DomainName
+
+  Distribution:
+    Type: AWS::CloudFront::Distribution
+    Properties:
+      DistributionConfig:
+        Enabled: true
+        HttpVersion: http2
+        Origins:
+        - DomainName: !Ref DomainName
+          Id: !Ref DomainName
+          CustomOriginConfig:
+            HTTPPort: 80
+            HTTPSPort: 443
+            OriginProtocolPolicy: https-only
+        DefaultCacheBehavior:
+          AllowedMethods: [ HEAD, DELETE, POST, GET, OPTIONS, PUT, PATCH ]
+          ForwardedValues:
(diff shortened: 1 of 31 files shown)


examples/2016-10-31/custom_domains_with_route53_hosted_zone_name/template.yaml
```diff
@@ -0,0 +1,71 @@
...
+      StageName: Prod
+      Domain:
+        DomainName: !Ref DomainName
+        CertificateArn: !Ref ACMCertificateArn
+        EndpointConfiguration: EDGE
+        BasePath:
+          - /home
+        Route53:
+         HostedZoneName: www.my-domain.com.
+         IpV6: true
```
timoschilling: I think `IpV6` is also optional

samtranslator/model/api/api_generator.py
```diff
@@ -338,15 +338,18 @@ def _construct_api_domain(self, rest_api):
         record_set_group = None
         if self.domain.get("Route53") is not None:
             route53 = self.domain.get("Route53")
-            if route53.get("HostedZoneId") is None:
+            if route53.get("HostedZoneId") is None and route53.get("HostedZoneName") is None:
                 raise InvalidResourceException(
                     self.logical_id, "HostedZoneId is required to enable Route53 support on Custom Domains."
```
praneetap: nit: "HostedZoneId or HostedZoneName is required to enable Route53 support on Custom Domains."

timoschilling:
```suggestion
                    self.logical_id, "HostedZoneId or HostedZoneName is required to enable Route53 support on Custom Domains."
```

examples/2016-10-31/custom_domains_with_route53_hosted_zone_name/template.yaml
mbarneyjr: This is an example, not saying these things are required, right?

samtranslator/model/api/api_generator.py
mbarneyjr: Missed that

examples/2016-10-31/custom_domains_with_route53_hosted_zone_name/template.yaml
timoschilling: The line below says `Everything below here is optional` but `IpV6` is already optional so it should be under this comment

mbarneyjr: Ahhh, I copied this from the `custom_domains_with_route53` example directory

praneetap: @timoschilling Ah, that comment was meant to point the difference between using your own distribution vs apigateway one. Not as "this property is optional".

mbarneyjr: I actually don't see `IpV6` in the SAM spec, but from looking at the code behind this, it looks like it is optional

timoschilling: Sorry didn't read the full comment, you both are right

samtranslator/model/api/api_generator.py
mbarneyjr: Added

mbarneyjr:
@praneetap I added the test cases and fixed a bug in the LogicalID generation thatI found because of it, thanks for the guidance!

praneetap:
@mbarneyjr please run make black, it will auto format for you. That should fix the build.

keetonian:
Looks like it's just a black formatting error. I submitted #1419 to update the black version that our tests use, this PR should be good to go once that is merged. I don't see any other test issues.

keetonian:
Closed/re-opened PR to kick off a new travis run. The black formatting issues should now be resolved.

keetonian:
@mbarneyjr Thank you for this contribution!

PR #1409, opened 2020-01-23
Added EventBridgePutEventsPolicy template

jmnarloch:
*Issue #, if available:* N/A

*Description of changes:* Adds a dedicated template policy for EventBridge granting the permissions to publish events to EventBus either custom ones or partner events.

*Description of how you validated changes:* Unit tests for transformed had been added to validate that the additional policy had been output.

*Checklist:*

- [X] Write/update tests
- [X] `make pr` passes
- [ ] Update documentation
- [X] Add/update example to `examples/2016-10-31`

### DIFF
--- a/examples/2016-10-31/policy_templates/all_policy_templates.yaml
+++ b/examples/2016-10-31/policy_templates/all_policy_templates.yaml
@@ -107,3 +107,6 @@ Resources:
         - TextractDetectAnalyzePolicy: {}
 
         - TextractGetResultPolicy: {}
+
+        - EventBridgePutEventsPolicy:
+            EventBusName: name
--- a/samtranslator/policy_templates_data/policy_templates.json
+++ b/samtranslator/policy_templates_data/policy_templates.json
@@ -1904,6 +1904,32 @@
           }
         ]
       }
+    },
+    "EventBridgePutEventsPolicy": {
+      "Description": "Gives permissions to send events to EventBridge",
+      "Parameters": {
+        "EventBusName": {
+          "Description": "Name of the EventBridge EventBus"
+        }
+      },
+      "Definition": {
+        "Statement": [
+          {
+            "Effect": "Allow",
+            "Action": "events:PutEvents",
+            "Resource": {
+              "Fn::Sub": [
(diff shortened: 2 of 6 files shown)


chrisoverzero:
It doesn’t look as though this overcomes the blockers described in [the previous attempt][] to add such a policy. 

[the previous attempt]: https://github.com/awslabs/serverless-application-model/pull/1203

examples/2016-10-31/policy_templates/all_policy_templates.yaml
```diff
@@ -101,3 +101,5 @@ Resources:
 
         - CodeCommitReadPolicy:
             RepositoryName: name
+
+        - EventBridgePutEventsPolicy: {}
```
jlhood: As discussed in the corresponding issue, let's add an EventBus parameter to this policy template so the permissions can be scoped to a specific event bus. See other policy templates that take a parameter as an example.

samtranslator/policy_templates_data/policy_templates.json
```diff
@@ -1849,6 +1849,22 @@
...
+    "EventBridgePutEventsPolicy": {
+      "Description": "Gives permissions to send events to EventBridge",
+      "Parameters": {},
+      "Definition": {
+        "Statement": [
+          {
+            "Effect": "Allow",
+            "Action": [
+              "events:PutEvents",
+              "events:PutPartnerEvents"
```
jlhood: I don't think it makes sense to include `PutPartnerEvents` in this policy template. The point of the policy template is to provide a shorthand syntax for common policies while allowing you to scope them down to a specific resource. So when we add EventBus as a parameter to this template, `PutPartnerEvents` no longer makes sense since that API doesn't allow you to specify a specific EventBus.

You could create a separate `EventBridgePutPartnerEventsPolicy` policy template if you think it's an important one to have.

jmnarloch:
I had updated the pull request. Changes made added EventBus name as parameter and I removed the policy on PutPartnerEvents.

samtranslator/policy_templates_data/policy_templates.json
jmnarloch: At the moment I will only limit the policy to put events, if there will be future need I think the `EventBridgePutPartnerEventsPolicy` can be added separately.

Issue #1432, opened 2020-02-04
Usage Plans: support Ref


keetonian:
**Description:**

In #1179, support was added for Usage Plans. However, it is not possible to reference these plans (and associated resources) without knowing their exact logical ids.

**Steps to reproduce the issue:**
1. Create an api using usage plans
2. Try to reference usage plan using `ApiLogicalId.UsagePlan`

**Observed result:**
Error

**Expected result:**
Should work (and reference usage plan/other resources)

ShreyaGangishetty:
This is released in [v1.22.0](https://github.com/awslabs/serverless-application-model/releases/tag/v1.22.0)

PR #1458, opened 2020-02-20
fix: make usageplan properties referable

ShreyaGangishetty:
*Issue #, if available:*
#1432 

*Description of changes:*
made usageplan resources a referable property to api

*Checklist:*

- [x] Write/update tests
- [x] `make pr` passes
- [ ] Update documentation

### DIFF
--- a/samtranslator/model/sam_resources.py
+++ b/samtranslator/model/sam_resources.py
@@ -10,7 +10,14 @@ from .api.http_api_generator import HttpApiGenerator
 from .s3_utils.uri_parser import construct_s3_location_object
 from .tags.resource_tagging import get_tag_list
 from samtranslator.model import PropertyType, SamResourceMacro, ResourceTypeResolver
-from samtranslator.model.apigateway import ApiGatewayDeployment, ApiGatewayStage, ApiGatewayDomainName
+from samtranslator.model.apigateway import (
+    ApiGatewayDeployment,
+    ApiGatewayStage,
+    ApiGatewayDomainName,
+    ApiGatewayUsagePlan,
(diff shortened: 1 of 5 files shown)


samtranslator/model/sam_resources.py
```diff
@@ -764,6 +771,9 @@ class SamApi(SamResourceMacro):
         "Stage": ApiGatewayStage.resource_type,
         "Deployment": ApiGatewayDeployment.resource_type,
         "DomainName": ApiGatewayDomainName.resource_type,
+        "UsagePlan": ApiGatewayUsagePlan.resource_type,
+        "UsagePlanKey": ApiGatewayUsagePlanKey.resource_type,
```
keetonian: `Ref` does nothing for UsagePlanKeys, we shouldn't include it here: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-apigateway-usageplankey.html

ShreyaGangishetty: `Ref` for `ApiGatewayUsagePlanKey` gives the physical id which has the mapping `ApiKeyPhysicalId:UsagePlanPhysicalId`. I think it was a miss in the docs

keetonian: Sounds good, this looks good to me then

ShreyaGangishetty:
> Please add docs on referable properties

I will submit another PR for docs on referable properties that will also have other referable properties documented for `AWS::Serverless::Function`, `AWS::Serverless::Api` and `AWS::Serverless::HttpApi`

PR #1481, opened 2020-03-03
explicit UpdateReplacePolicy

PatMyron:
fixes https://github.com/aws-cloudformation/cfn-python-lint/issues/1265

*Description of changes:* [`UpdateReplacePolicy`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-updatereplacepolicy.html) currently defaults to `Delete` anyway, so this change should be a no-op explicitly acknowledging the current default [`UpdateReplacePolicy`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-updatereplacepolicy.html) value by making it explicit. I'd encourage following up and re-evaluating what [`UpdateReplacePolicy`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-updatereplacepolicy.html) value really makes sense here, as [CloudFormation Linter rules](https://github.com/aws-cloudformation/cfn-python-lint/blob/master/docs/rules.md) [I3011](https://github.com/aws-cloudformation/cfn-python-lint/blob/master/src/cfnlint/rules/resources/UpdateReplacePolicyDeletionPolicyOnStatefulResourceTypes.py) and [W3011](https://github.com/aws-cloudformation/cfn-python-lint/blob/master/src/cfnlint/rules/resources/BothUpdateReplacePolicyDeletionPolicyNeeded.py) are really just meant to encourage explicit policies because the default values can lead to disastrous surprises

---

63 AssertionErrors to fix 😳

```shell
serverless-application-model $ gsed -i -r 's/(DeletionPolicy.*)/\1\n      "UpdateReplacePolicy": "Delete",/' tests/translator/output/**/*.json
```

---

### DIFF
--- a/samtranslator/model/__init__.py
+++ b/samtranslator/model/__init__.py
@@ -43,7 +43,7 @@ class Resource(object):
     property_types = None
     _keywords = ["logical_id", "relative_id", "depends_on", "resource_attributes"]
 
-    _supported_resource_attributes = ["DeletionPolicy", "UpdatePolicy", "Condition"]
+    _supported_resource_attributes = ["DeletionPolicy", "UpdateReplacePolicy", "UpdatePolicy", "Condition"]
 
     # Runtime attributes that can be qureied resource. They are CloudFormation attributes like ARN, Name etc that
     # will be resolvable at runtime. This map will be implemented by sub-classes to express list of attributes they
--- a/samtranslator/model/sam_resources.py
+++ b/samtranslator/model/sam_resources.py
@@ -677,6 +677,7 @@ class SamFunction(SamResourceMacro):
         if attributes is None:
             attributes = {}
         attributes["DeletionPolicy"] = "Retain"
+        attributes["UpdateReplacePolicy"] = "Delete"
 
         lambda_version = LambdaVersion(logical_id=logical_id, attributes=attributes)
         lambda_version.FunctionName = function.get_runtime_attr("name")
@@ -1104,6 +1105,7 @@ class SamLayerVersion(SamResourceMacro):
         if attributes is None:
             attributes = {}
         attributes["DeletionPolicy"] = retention_policy_value
+        attributes["UpdateReplacePolicy"] = "Delete"
 
         old_logical_id = self.logical_id
         new_logical_id = logical_id_generator.LogicalIdGenerator(old_logical_id, self.to_dict()).gen()
--- a/tests/translator/output/aws-cn/basic_layer.json
+++ b/tests/translator/output/aws-cn/basic_layer.json
@@ -10,6 +10,7 @@
   "Resources": {
     "LayerWithCondition7c655e10ea": {
       "DeletionPolicy": "Retain", 
+      "UpdateReplacePolicy": "Delete",
       "Type": "AWS::Lambda::LayerVersion", 
       "Properties": {
         "Content": {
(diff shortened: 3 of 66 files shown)


keetonian:
Clarification questions-
1. Should this be added to *every* resource that SAM generates?
1. Can you verify that adding this policy does result in a no-op (if I deploy a template, then add this policy and deploy again, CFN does not create/execute a new changeset)?

PatMyron:
> Clarification questions-
> 
> 1. Should this be added to _every_ resource that SAM generates?

The main idea behind [CloudFormation Linter rule W3011](https://github.com/aws-cloudformation/cfn-python-lint/blob/master/src/cfnlint/rules/resources/BothUpdateReplacePolicyDeletionPolicyNeeded.py) is that if either one of [`DeletionPolicy`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-deletionpolicy.html) / [`UpdateReplacePolicy`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-updatereplacepolicy.html) are set to a value to protect resources from deletion, the other policy should at least be explicitly set to something because many don't know about the other policy and someone who has opinions about protecting resources in one situation probably has opinions about protecting resources in the other situation

This PR should at least cover the situations where the [`AWS::Serverless` transform](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/transform-aws-serverless.html) was protecting resources from deletion with [`DeletionPolicy`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-deletionpolicy.html) but not setting any [`UpdateReplacePolicy`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-updatereplacepolicy.html) which caused noise [`AWS::Serverless` transform](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/transform-aws-serverless.html) customers couldn't fix

PatMyron:
> 2. Can you verify that adding this policy does result in a no-op (if I deploy a template, then add this policy and deploy again, CFN does not create/execute a new changeset)?

There is a CloudFormation deployment between these templates (whether change sets are used or not):

```yaml
Resources:
  Resource:
    Type: AWS::SNS::Topic
```
```yaml
Resources:
  Resource:
    Type: AWS::SNS::Topic
    UpdateReplacePolicy: Delete
```

but there are no stack events concerning the resources themselves, and the default [`UpdateReplacePolicy`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-updatereplacepolicy.html) was already `Delete` anyway. Is the no-op CloudFormation deployment an issue?

keetonian:
A no-op deployment is fine. This looks great, thank you for the contribution.

ShreyaGangishetty:
This change is updating the hash values of logical ids to some of the `Lambda:LayerVersion` and `Lambda:Version` resources which will cause re deployments to the customer templates. We are removing this change from the upcoming release(v1.23.0) and try to schedule this change for next release.

PatMyron:
> This change is updating the hash values of logical ids to some of the `Lambda:LayerVersion` and `Lambda:Version` resources which will cause re deployments to the customer templates. We are removing this change from the upcoming release(v1.23.0) and try to schedule this change for next release.

think [setting `UpdateReplacePolicy` after generating the new logical ID](https://github.com/PatMyron/serverless-application-model/commit/e1c7f922bdfd9f45b8eb8bebd265ea0bf77e3e76) should fix this

Issue #1486, opened 2020-03-03
Release tracker for SAM v1.22.0 release


ShreyaGangishetty:
### Changelog:

1. fix: add periods to error messages, fix quotes (#1430) 
1. feat(policy-templates): add EventBridgePutEventsPolicy (#1409) 
1. fix: use isinstance check instead of type (#1411) 
1. feat: add TimeoutInMillis to HttpApi events (#1450) 
1. docs: document support for HTTP API Access Log Settings (#1448) 
1. docs: fix markdown style links in not-markdown (#1439) 
1. feat: add Tags support to Http Api (#1459) 
1. docs: switch Definition to Model (#1423) 
1. chore: merge pull request #1463 from awslabs/master 
1. fix: make usageplan properties referable (#1458) 
1. feat: add support for DefaultRouteSettings in HttpApi (#1461)
1. feat: Trigger auto publish alias on env var/memorysize change (#1310) 
1. feat: cors httpapi (#1381)
1. chore: bump translator version (#1483) 
1. docs: add referable properties of sam resources (#1468) 
1. feat: custom domains in HTTP API (#1472) 
1. chore: merge pull request #1485 from awslabs/develop 
1. fix: remove OIDC auth from Http Api (#1491)
1. feat: propagate tags to DomainName, Stage resources (#1492)
1. docs: Add StageVariables to docs (#1493)
1. feat: support RouteSettings for HttpApi (#1490)
1. feat: add DynamoDBWritePolicy and S3WritePolicy (#1374)
1. Fix type of user pool tags (#1460)

adrai:
Any update on this?
What is the expected ETA?

ShreyaGangishetty:
@adrai This release is in Stages state now. Please track the status of this release here: https://github.com/awslabs/serverless-application-model/projects/3#card-34033303

ShreyaGangishetty:
This is released in [v1.22.0](https://github.com/awslabs/serverless-application-model/releases/tag/v1.22.0)

PR #1519, opened 2020-03-16
Release/v1.22.0

ShreyaGangishetty:
*Issue #, if available:*
#1486 

*Description of changes:*
Release v1.22.0

*Checklist:*

- [ ] Write/update tests
- [ ] `make pr` passes
- [ ] Update documentation


mhart:
So this release introduced a breaking change and really shouldn't have been a minor release.

This can easily be observed by all the tests that were changed from `"payloadFormatVersion": "1.0"` to `"payloadFormatVersion": "2.0"` – if your tests require that change, then everyone's code is also going to require a change.

We managed to fix this within a few hours of deployment, but it was very hard to find – especially when defaults change from under your feet.

Can we have some sort of assurance that breaking changes like this will be avoided in the future – and if they are made, then a new SAM transform version is used?

sapessi:
Hi @mhart, 

I apologize for the impact.  Our process does require breaking changes like this be accompanied by a version change and appropriate communication, but it’s clear we did not follow it here. We are root causing what happened internally and will add additional controls to prevent it from recurring again. Thank you again for the feedback.

Issue #1327, opened 2019-12-13
Add VpcEndpointIds to EndpointConfiguration Parameter


matmertz25:
**Description:**

Currently Cloudformation supports both Types and VpcEndpointIds as parameters on EndpointConfiguration. For endpoints types of PRIVATE a VPC endpoint is required to generate Route53 Aliases from which you can use to direct to specific APIs on apigateway. Without the Aliases the API id must be supplied in the header on every request (x-apigw-api-id). As a workaround the Api and VPCe can be linked through the console/sdks but every time SAM is deployed it overwrites the previous configuration.

Adding a VpcEndpointIds as a list of strings parameter on the EndpointConfiguration would enable the association of Apis on apigateway and vpce's

colinbjohnson:
**Steps to reproduce the issue:**

Create a template with the following content:

```
AWSTemplateFormatVersion: '2010-09-09'

Transform: AWS::Serverless-2016-10-31

Description: >
  sam-app

  Sample SAM Template for sam-app

Resources:

  AWSApiGateway:
    Type: AWS::Serverless::Api
    Properties:
      # resolves an issue with the following error:
      # Private REST API doesn't have a resource policy attached to it (Service: AmazonApiGateway;
      # Status Code: 400; Error Code: BadRequestException; Request ID: acbe714e-970b-4150-b042-f5624a8b3ff4)
      Auth:
        ResourcePolicy:
          IpRangeWhitelist: 10.0.0.0/24
      Name: Private AWS Api Gateway
      # EndpointConfiguration: Private
      EndpointConfiguration: 
        Types: 
          - PRIVATE
        VpcEndpointIds:
          - !Ref ApiGatewayVPCEndpoint
      StageName: Staging

  HelloWorldFunction:
    Type: AWS::Serverless::Function
    Properties:
      CodeUri: hello_world/
      Handler: app.lambda_handler
      Runtime: python3.8
      Events:
        HelloWorld:
          Type: Api
          Properties:
            Path: /hello
            Method: get
            RestApiId: !Ref AWSApiGateway

Outputs: {}
```
Where the `!mport` statement returns the id of a `AWS::EC2::VPCEndpoint` type resource.

Then run the following command: `sam build --use-container`

**Observed result:**

I receive the following error:

```
[...]

PR #1524, opened 2020-03-20
feat: add support for VPCEndpointIds in EndpointConfiguration

scbrown:
*Issue #1327
*Description of changes:*
Adds support for configuring a PRIVATE  RestApi Endpoint
added unit tests with expected input and output
*Checklist:*

- [x] Write/update tests
- [x] `make pr` passes
- [x] Update documentation

### DIFF
--- a/samtranslator/model/api/api_generator.py
+++ b/samtranslator/model/api/api_generator.py
@@ -927,8 +927,11 @@ class ApiGenerator(object):
         :param string/dict value: Value to be set
         """
         if isinstance(value, dict):
-            rest_api.EndpointConfiguration = {"Types": [value.get("Types", value)]}
             rest_api.Parameters = {"endpointConfigurationTypes": value.get("Types", value)}
+            if isinstance(value.get("Types"), list):
+                rest_api.EndpointConfiguration = {"Types": value.get("Types", value)}
+            else:
+                rest_api.EndpointConfiguration = {"Types": [value.get("Types", value)]}
             if "VPCEndpointIds" in value.keys():
                 rest_api.EndpointConfiguration["VpcEndpointIds"] = value.get("VPCEndpointIds")
         else:
--- a/samtranslator/model/api/api_generator.py
+++ b/samtranslator/model/api/api_generator.py
@@ -293,7 +293,8 @@ class ApiGenerator(object):
         elif endpoint not in ["EDGE", "REGIONAL", "PRIVATE"]:
             raise InvalidResourceException(
                 self.logical_id,
-                "EndpointConfiguration for Custom Domains must be" " one of {}.".format(["EDGE", "REGIONAL", "PRIVATE"]),
+                "EndpointConfiguration for Custom Domains must be"
+                " one of {}.".format(["EDGE", "REGIONAL", "PRIVATE"]),
             )
 
         if endpoint == "REGIONAL":
@@ -933,4 +934,3 @@ class ApiGenerator(object):
         else:
             rest_api.EndpointConfiguration = {"Types": [value]}
             rest_api.Parameters = {"endpointConfigurationTypes": value}
-
--- a/samtranslator/model/api/api_generator.py
+++ b/samtranslator/model/api/api_generator.py
@@ -290,10 +290,10 @@ class ApiGenerator(object):
         if endpoint is None:
             endpoint = "REGIONAL"
             self.domain["EndpointConfiguration"] = "REGIONAL"
-        elif endpoint not in ["EDGE", "REGIONAL"]:
+        elif endpoint not in ["EDGE", "REGIONAL", "PRIVATE"]:
             raise InvalidResourceException(
                 self.logical_id,
-                "EndpointConfiguration for Custom Domains must be" " one of {}.".format(["EDGE", "REGIONAL"]),
+                "EndpointConfiguration for Custom Domains must be" " one of {}.".format(["EDGE", "REGIONAL", "PRIVATE"]),
             )
 
         if endpoint == "REGIONAL":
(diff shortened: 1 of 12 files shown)


scbrown:
sorry for all the commits :) 

In the expected output JSON I added, I'm unsure of how the Parameters section of AWS::ApiGateway::RestApi is supposed to look. Could someone point me in the right direction?

scbrown:
AWS::ApiGateway::RestApi Parameters only allows simple string:string mappings, so I've kept it so that EndpointConfiguration only allows specifying 1 Type. I've tested this in my environment and it works, but I'm still not sure of the significance of setting these custom headers.

qmg-drettie:
would be great to get this merged! we are going to start using private gateways soon and require this

mjdouble:
Any ETA for this being released?

sriram-mv:
This looks good overall, taking a deeper look to make sure there are breaking changes from the current implementation.

markdwags:
Any update on this PR @sriram-mv? I'm currently running into the issue this PR will address, which is, in  `AWS::Serverless::Api` not only being able to assign the endpoint as PRIVATE, but to be able to assign a VPC Endpoint to it as well to avoid this manual step.

Thanks!

zkhan96:
Would a command such as:
```
aws apigateway update-rest-api --rest-api-id <api-id> --patch-operations "op='add',path='/endpointConfiguration/vpcEndpointIds',value='vpce-<vpce-id>'" --region <region>
```
After a deployment be a viable workaround until this is merged?

sriram-mv:
Is is possible to move `VPCEndpointIds ` to be its own property, because the way this is structured, changes the [EndpointConfiguration](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-resource-api.html#sam-api-endpointconfiguration) to be more than a string, and this could be a breaking change.

scbrown:
That may be possible if someone wants to start over, but this implementation was @ShreyaGangishetty recommendation from #1327

jfuss:
@sriram-mv I am just trying to understand your push back here? Personally, I think how this PR is laid out now is ok. My main thought process behind this is: EndpointConfiguration maps to https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-apigateway-restapi-endpointconfiguration.html 

Long ago, we made the decision to make EndpointConfiguration should be a string and for a couple years the only property of EndpointConfiguration was the "Types". In SAM, we simplified that definition to be just the value of EndpointConfiguration. Hindsight is 20/20, but now that there is another property I think it makes sense to keep this mapping to it. My main reason for that is, if tomorrow API Gateway adds another feature under EndpointConfiguration, where do we put it? I don't think we want to keep endlessly deduping these properties. Honestly, there is a lot of SAM that can just be pass through and can get out of the way, instead of us trying to always keep up with CloudFormation (which we know is hard).

So if we have a strong reason to not put this under EnpointConfiguration in SAM, lets come up with a way to do that and not go down this endless "if another Property is added, we need to add it to `AWS::Serverless::Api.Properties.NewPropName`. 

@scbrown Thanks you for getting us started here! I will work with @sriram-mv to come to a resolution and update this PR accordingly.

sriram-mv:
@jfuss My primary concern was with backward compatibility, but if endpoint configuration being private under SAM was not supported anyway and the remaining endpoint configurations are backward compatible, I'm ok moving forward with this. I also 100% agree to be conformant with the CFN spec and not continuously add properties as needed.

sriram-mv:
According to https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-apigateway-restapi-endpointconfiguration.html, Types is a list, but I cant find a real world use case where it actually is a list, whereas compared to SAM its just a string. With that in mind, I'm approving this PR.

saurabhshri-evive:
This is great news! 🎉 Thanks everyone for working on this, subscribed to 'watch' releases.

ariftahir:
Excellent news, been waiting for this 🥳

denisbr:
> According to https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-apigateway-restapi-endpointconfiguration.html, Types is a list, but I cant find a real world use case where it actually is a list, whereas compared to SAM its just a string. With that in mind, I'm approving this PR.

I appreciate your work, but this is just very very wrong.
The documentation https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-resource-api.html#sam-api-endpointconfiguration says : 
```AWS CloudFormation compatibility: This property is passed directly to the EndpointConfiguration property of an AWS::ApiGateway::RestApi resource. ```

In reality this is not true as per your implementation, and we've been struggling to get our _list_ of VPCEndpoints through the SAM transform, but no luck. If it wasn't for checking the issue tracker, I probably would never have discovered this "feature. 

Why is it that you think not implementing according to specification is a good idea in this case?

File: samtranslator/model/api/api_generator.py [hunk 1]
```diff
@@ -290,10 +290,11 @@ def _construct_api_domain(self, rest_api):
         if endpoint is None:
             endpoint = "REGIONAL"
             self.domain["EndpointConfiguration"] = "REGIONAL"
-        elif endpoint not in ["EDGE", "REGIONAL"]:
+        elif endpoint not in ["EDGE", "REGIONAL", "PRIVATE"]:
             raise InvalidResourceException(
                 self.logical_id,
-                "EndpointConfiguration for Custom Domains must be" " one of {}.".format(["EDGE", "REGIONAL"]),
+                "EndpointConfiguration for Custom Domains must be"
+                " one of {}.".format(["EDGE", "REGIONAL", "PRIVATE"]),
```
saurabhshri, line 297: @sriram-mv / @scbrown Custom Domains are not supported on private APIs, right?

- https://docs.aws.amazon.com/apigateway/latest/developerguide/how-to-custom-domains.html
- https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-apigateway-domainname-endpointconfiguration.html

samtranslator/model/api/api_generator.py
```diff
@@ -925,6 +926,11 @@ def _set_endpoint_configuration(self, rest_api, value):
         :param rest_api: RestApi resource
         :param string/dict value: Value to be set
         """
-
-        rest_api.EndpointConfiguration = {"Types": [value]}
-        rest_api.Parameters = {"endpointConfigurationTypes": value}
+        if isinstance(value, dict) and value.get("Type"):
+            rest_api.Parameters = {"endpointConfigurationTypes": value.get("Type")}
+            rest_api.EndpointConfiguration = {"Types": [value.get("Type")]}
+            if "VPCEndpointIds" in value.keys():
```
saurabhshri, line 932: I didn't see this earlier, but I wished it was `VpcEndpointIds` instead of `VPCEndpointIds`, just for consistency. :) I did not realise and kept trying with `VpcEndpointIds` and could not get it to work (there's no error, so it deployed successfully, without VPCe).

esukirno: I agree with @saurabhshri. Would  be better if the name can be made consistence as per cloudformation EndpointConfiguration properties.

This one confuses us, especially, because the documentation says the wrong thing.
https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-resource-api.html#sam-api-endpointconfiguration

PR #1557, opened 2020-04-13
feat: add end to end tests for features in SAM

ShreyaGangishetty:
*Issue #, if available:*

*Description of changes:*
Added basic setup for being able to write end to end tests for SAM

Ran the end to end tests using `make e2e-tests`

*Checklist:*

- [x] Write/update tests
- [x] `make pr` passes
- [x] Update documentation


end-to-end-tests/tst/helpers/base_test.py
```diff
@@ -0,0 +1,86 @@
...
+                outputs[output.get("OutputKey")] = output.get("OutputValue")
+            return outputs
+
+    def delete_stack(self):
+        self.cloud_formation_client.delete_stack(StackName=self.stack_name)
+
+    def get_sam_transformed_template(self, template_path):
+        try:
+            subprocess.call(
+                ["bin/sam-translate.py", "--template-file", template_path, "--o", self.stack_name + ".json"]
```
jfuss, line 78: Why are we shelling out here? Can't we just run the transform?

end-to-end-tests/tst/helpers/base_test.py
```diff
@@ -0,0 +1,86 @@
...
+import boto3
+import random
+import os
+import subprocess
+from resources import Resources
+
+
+class BaseTest(unittest.TestCase):
+    def setUp(self):
+        self.stack_name = "sam-temp-stack" + str(random.randint(0, 10000))
```
jfuss, line 11: instead of `random.randint` you should use `uuid`

end-to-end-tests/tst/helpers/resource_status.py
```diff
@@ -0,0 +1,22 @@
+from enum import Enum
+
+# contains all the the cloudformation stack resource status
+class ResourceStatus(Enum):
```
jfuss, line 4: What's the value of this being an Enum here?

end-to-end-tests/tst/helpers/base_test.py
```diff
@@ -0,0 +1,86 @@
...
+        self.cloud_formation_client = boto3.client("cloudformation")
+
+    def make_stack(self, capabilties, template_path):
```
praneetap, line 14: SHouldn't we raise an exception if this fails

end-to-end-tests/tst/helpers/base_test.py
```diff
@@ -0,0 +1,86 @@
...
+        stack_outputs = (
+            self.cloud_formation_client.describe_stacks(StackName=self.stack_name).get("Stacks")[0].get("Outputs")
+        )
+        outputs = dict()
+        if stack_outputs is not None:
+            for output in stack_outputs:
+                outputs[output.get("OutputKey")] = output.get("OutputValue")
+            return outputs
+
+    def delete_stack(self):
```
praneetap, line 72: raise an exception if delete fails

end-to-end-tests/tst/helpers/resources.py
```diff
@@ -0,0 +1,129 @@
...
+        if expected_resources.resources is not None:
+            diff_resources.update(expected_resources.resources)
+
+        return diff_resources
+
+    def __eq__(self, other):
+        if not isinstance(other, Resources):
+            return False
+        # checking if the keys are same
+        if self.resources.keys() != other.resources.keys():
```
jfuss, line 80: Can't we just compare the dictionaries? ` self.resources == other.resources`?

end-to-end-tests/tst/helpers/resources.py
```diff
@@ -0,0 +1,129 @@
...
+            return False
+        for id, resource in self.resources.items():
+            if not (resource == other.resources.get(id)):
+                return False
+        return True
+
+    """
+    Validates and Cleans up LogicalIDs generated by SAM translator.
```
jfuss, line 88: Random doc strings? Can this be placed right under the function/class/etc definitions?

end-to-end-tests/tst/helpers/resources.py
```diff
@@ -0,0 +1,129 @@
...
+                    )
+                )
+            return logical_id_to_clean[:-10]  # remove last 10 digits and return
+        return logical_id_to_clean
+
+
+class Resource:
+    # We only care about the following properties when comparing for equality because rest are dynamically
+    # generated by CloudFormation: LogicalResourceId, ResourceStatus, ResourceType
+    logical_id = str()
```
jfuss, line 113: Why are these defined globally on Resource?

end-to-end-tests/tst/helpers/resources.py
```diff
@@ -0,0 +1,129 @@
+from resource_types import ResourceTypes
+from resource_status import ResourceStatus
+
+import re
+
+
+class Resources:
```
jfuss, line 7: Can we add docstrings to all public functions? It will make it easier to understand everything.

end-to-end-tests/tst/test_http_api_cors.py
```diff
@@ -0,0 +1,47 @@
...
+        .created("ServerlessHttpApiApiGatewayDefaultStage", ResourceTypes.APIGWV2_STAGE)
+    )
+
+    def tearDown(self):
+        self.delete_stack()
+
+    def test_cors(self):
+        input_template = os.path.join(os.getcwd(), "end-to-end-tests/input", "http_api_cors", "template.yaml")
+        capabilities = ["CAPABILITY_IAM"]
+        stack_name = self.make_and_verify_stack(input_template, capabilities, self.http_api_with_cors_resources)
```
jfuss, line 24: Can we split the make and verify steps into separate functions? It will make it easier to reason about when each step does one operation.

end-to-end-tests/tst/test_rest_api_with_authorizers.py
```diff
@@ -0,0 +1,131 @@
...
+            lambda_token_authorizer.get("identityValidationExpression"),
+            None,
+            "lambdaTokenAuthorizer: validation expression must not be set",
+        )
+        self.assertEquals(
+            lambda_token_authorizer.get("authorizerUri"),
+            lambda_authorizer_uri,
+            "lambdaTokenAuthorizer: authorizer URI must be the Lambda Function Authorizer's URI",
+        )
+        self.assertEquals(
```
jfuss, line 74: `self.assertEquals` should be `self.assertEqual` (this is how it is done in python3).

jfuss:
What Python version are we running these tests in? Can we make it so they only run in Python3? It will make our lives much much easier to not port more code into Python3 after Python2 finally dies.

DEVELOPMENT_GUIDE.rst
```diff
@@ -70,6 +70,13 @@ drops below 95%.
...
+End to end tests
+~~~~~~~~~~~~~~~~
+
+``make e2e tests`` command will run all end to end tests. This command is configured to fail when code coverage for package drops below 95%.
+
+``pytest end-to-end-tests/tst/test.py`` command will run the integration test `test.py`
+
 Pull Requests
 -------------
 Before sending pull requests make sure to run ``make pr`` command. This will run unit tests, linters, and static
```
jfuss, line 82: Do we still only want `make pr` to work? Should running integ tests be apart of `make pr`?

end-to-end-tests/tst/test_rest_api_with_authorizers.py
praneetap: +1. This should come up as a warning in build, please check if there are any other warnings as well.

DEVELOPMENT_GUIDE.rst
```diff
@@ -70,6 +70,13 @@ drops below 95%.
 
 ``pytest -k "TestMyClass"`` command will run all unit tests within the `TestMyClass` class.
 
+End to end tests
+~~~~~~~~~~~~~~~~
+
+``make e2e tests`` command will run all end to end tests. This command is configured to fail when code coverage for package drops below 95%.
```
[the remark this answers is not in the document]
ShreyaGangishetty: agreed, removed the code coverage check in the next revision

end-to-end-tests/tst/helpers/base_test.py
ShreyaGangishetty: Updated the code with exceptions where ever applicable

end-to-end-tests/tst/helpers/base_test.py
ShreyaGangishetty: Cloudformation doesn't throw an exception for delete stack. https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/cloudformation.html#CloudFormation.Client.delete_stack
I will try to add additional check to throw exception when delete fails

end-to-end-tests/tst/helpers/base_test.py
ShreyaGangishetty: I could not import sam-translate even after adding `__init__.py` to bin folder. hence i used subprocess

end-to-end-tests/tst/helpers/base_test.py
ShreyaGangishetty: thank you, updated the code with `uuid`

end-to-end-tests/tst/helpers/resource_status.py
ShreyaGangishetty: To state that the ResourceStatus is a constant so I used Enum

end-to-end-tests/tst/helpers/resources.py
ShreyaGangishetty: Thank you, updated the code to just compare the dictionaries directly and it works as expected

end-to-end-tests/tst/helpers/resources.py
ShreyaGangishetty: Thank you, I have updated the code with doc strings for class and method definitions

end-to-end-tests/tst/helpers/resources.py
ShreyaGangishetty: Good catch! We don't need these attributes to be present globally. I have updated the code

end-to-end-tests/tst/helpers/resources.py
ShreyaGangishetty: Good suggestion! added docstrings to all functions and classes

end-to-end-tests/tst/test_http_api_cors.py
ShreyaGangishetty: Agreed. I have added three functions: `make_stack`, `verify_stack` and `make_and_verify_stack`. We can either use make step then verify or use a single function `make_and_verify_stack` now

end-to-end-tests/tst/test_rest_api_with_authorizers.py
ShreyaGangishetty: Initially I had written the code in Python2. Updated the code to work in python3

ShreyaGangishetty:
> What Python version are we running these tests in? Can we make it so they only run in Python3? It will make our lives much much easier to not port more code into Python3 after Python2 finally dies.

Initially I was running the tests in Python2. I have updated the code to run tests in Python3

DEVELOPMENT_GUIDE.rst
ShreyaGangishetty: I have added [here](https://github.com/awslabs/serverless-application-model/pull/1557/files#diff-195a635ad245ded9076330e31134bd80R12) but forgot to add in this page. I will update it
I didn't add a way to running integ tests as part of build but we will eventually want integ tests to be a part of make pr.

end_to_end_tests/input/rest_api_with_authorizers/template.yaml
```diff
@@ -0,0 +1,134 @@
...
+            Identity:
+              QueryStrings:
+                - authorization
+
+  MyFunction:
+    Type: AWS::Serverless::Function
+    Properties:
+      Handler: index.handler
+      Runtime: nodejs12.x
+      InlineCode: |
```
jfuss, line 31: Should one of these tests not use inline so we can see how that is setup? I would assume that InlineCode is not going to be the majority of things we do with these tests.

ShreyaGangishetty: Good idea! I will update one of the tests without `InlineCode`

end_to_end_tests/test/helpers/resources.py
```diff
@@ -0,0 +1,247 @@
...
+        :return: Resources object by adding the resource to list of resources
+        :rtype: Resources
+        """
+        if stack_resource is not None:
+            # LogicalID of ApiGateway::Deployment resource gets suffixed with SHA.
+            # This suffix changes on every run of the test. To keep tests simple, we will validate and strip
+            # out this suffix here. Tests can continue to access and verify the resource using the LogicalId
+            # specified in the template
+            if stack_resource.get("LogicalResourceId") is None:
+                raise ValueError("LogicalResourceId should not be empty for stack resource {}".format(stack_resource))
```
jfuss, line 84: specific errors

end_to_end_tests/test/test_rest_api_with_authorizers.py
```diff
@@ -0,0 +1,199 @@
+import boto3
+import os
+from .helpers.base_test import BaseTest
+from .helpers.resources import Resources
+from .helpers.resource_types import ResourceTypes
+import urllib3
```
jfuss, line 6: Any reason we choice urllib3 over requests?

pytest.ini
```diff
@@ -1,7 +1,4 @@
 [pytest]
-# Fail if coverage falls below 95%
-# NOTE: If debug breakpoints aren't working, comment out the code coverage line below
-addopts = --cov samtranslator --cov-report term-missing --cov-fail-under 95
```
jfuss, line 4: Is the coverage in all the commands we need it to be on the makefile?

requirements/dev.txt
```diff
@@ -10,6 +10,10 @@ pytest>=3.0.7
 mock>=2.0.0
 parameterized>=0.6.1
 
+# Requirements for integration tests
+certifi>=2020.4.5.1
+urllib3
```
jfuss, line 15: Wait, we already have a dependency on requests but add urllib3? We should just be able to use requests for calls.

end_to_end_tests/test/helpers/base_test.py
```diff
@@ -0,0 +1,199 @@
...
+        :raise: InvalidTemplateException if the template is not transformed successfully
+        :return: transformed template
+        :rtype: str
+        """
+
+        try:
+            output_file = self.stack_name + ".json"
+            with open(output_file, "w") as fp:
+                pass
+            subprocess.call(["python", "bin/sam-translate.py", "--template-file", template_path, "--o", output_file])
```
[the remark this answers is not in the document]
ShreyaGangishetty: I was not able to import the function even though i added `__init__.py` for bin to make that as a module. Hence, kept subprocess

pytest.ini
ShreyaGangishetty: Yes, right now we have only `make pr` which needs coverage and --cov is present in the Makefile already for `make pr`. Currently, we don't have 95% test coverage for integ tests. Removing these lines was necessary for `make e2e-tests` to pass

end_to_end_tests/test/test_rest_api_with_authorizers.py
ShreyaGangishetty: No specific reason. I will update the code to use `requests` module

end_to_end_tests/test/helpers/base_test.py
```diff
@@ -0,0 +1,199 @@
...
+        except IOError:
+            raise IOError("The input template path: {} is incorrect.".format(template_path))
+        else:
+            if os.path.exists(self.stack_name + ".json"):
+                template_body = open(self.stack_name + ".json").read()
+                # delete the temporary file created
```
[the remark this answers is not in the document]
ShreyaGangishetty: Thank you, I am removing `subprocess` to call the transform function, So I wont be needing to create temp files anymore.

end_to_end_tests/test/helpers/base_test.py
```diff
@@ -0,0 +1,199 @@
+import unittest
+import boto3
+import uuid
+import os
+import subprocess
+from .resources import Resources
+from samtranslator.model.exceptions import InvalidTemplateException
+
+
+class BaseTest(unittest.TestCase):
```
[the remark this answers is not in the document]
ShreyaGangishetty: Thanks you. I will update in the next revision

end-to-end-tests/tst/helpers/base_test.py
ShreyaGangishetty: I will update in the next revision to use transform code

end_to_end_tests/input/rest_api_with_authorizers/template.yaml
ShreyaGangishetty: Updated the `template.yaml` for `http_api_cors`integration test

requirements/dev.txt
ShreyaGangishetty: Updated the code to use `requests`

end_to_end_tests/test/helpers/base_test.py
```diff
@@ -0,0 +1,199 @@
...
+        template_body = self.get_sam_transformed_template(template_path=template_path)
+        try:
+            stack_id = self.cloud_formation_client.create_stack(
+                StackName=self.stack_name, TemplateBody=template_body, Capabilities=capabilities
+            )
+            waiter = self.cloud_formation_client.get_waiter("stack_create_complete")
+            waiter.wait(StackName=self.stack_name)
+            return stack_id
+        except Exception as e:
+            raise Exception("Unable to create stack for template {}. Exception: ".format(template_path, e))
```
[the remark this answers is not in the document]
ShreyaGangishetty: removed the generic exception here and not catching any other exception as the error from CloudFormation would be self explanatory

ShreyaGangishetty:
> We need to be very careful about exceptions. We should not catch broad ones (if ever). So things like `except Exception` need to be scoped. This can lead to hiding actually errors which will give the tests a false sense of security.
> 
> We should avoid raising ValueError or other generic exceptions as well. If python ends up throwing one of this, we think it is one error but it is a completely different one. This can be very dangerous for integration tests that we rely on for deployments.

Thank you for the suggestion, it was really useful. Removed all the generic Exceptions and updated with user defined exceptions where ever applicable.

PR #1580, opened 2020-04-28
fix: resource policy generation for {path+}

sriram-mv:
Why is this change necessary?

* To transfrom the arn in the api gateway resource policy to be "*"
instead of the name of the path itself.

How does it address the issue?

* Allow to address the appropriate resource.

What side effects does this change have?

* None that are known at this point.

*Issue #, if available:*

https://github.com/awslabs/serverless-application-model/issues/1549

*Description of changes:*

* Transform path of kinds, {path+}, {proxy+} into "*"

- deployed the template without errors as referenced in https://github.com/awslabs/serverless-application-model/issues/1549

*Checklist:*

- [x] Write/update tests
- [x] `make pr` passes
- [ ] Update documentation

### DIFF
--- a/samtranslator/swagger/swagger.py
+++ b/samtranslator/swagger/swagger.py
@@ -1144,4 +1144,5 @@ class SwaggerEditor(object):
 
     @staticmethod
     def get_path_without_trailing_slash(path):
-        return re.sub(r"{([a-zA-Z0-9._-]+|proxy\+)}", "*", path)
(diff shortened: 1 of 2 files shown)


jfuss:
@sriram-mv 

From your description: Transform path of three kinds, {path+}, {proxy+} and {path} into "*"

Should `{path}` really be resolved into `*`?

sriram-mv:
@jfuss Good catch on the PR description, will update that.

Issue #1506, opened 2020-03-11
Immutable AWS::Cognito::UserPool properties are not supported in the SAM translator.


phsstory:
**Description:**

Immutable AWS::Cognito::UserPool properties are not supported in the SAM translator.

Reference Documentation links:
* https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-cognito-userpool.html#cfn-cognito-userpool-usernameconfiguration
* https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-cognito-userpool.html#cfn-cognito-userpool-accountrecoverysetting

Expected missing keys in property_types in `samtranslator/model/cognito.py`

**Steps to reproduce the issue:**

Use the following template snippet in a deploy

    AWSTemplateFormatVersion: '2010-09-09'
    Transform: AWS::Serverless-2016-10-31
    Resources:
      CognitoUserPool:
        Type: AWS::Cognito::UserPool
        Properties:
          UsernameConfiguration:
            CaseSensitive: False
          AccountRecoverySetting:
            RecoveryMechanisms:
              - Name: verified_email
                Priority: 1

**Observed result:**

Failed to create the changeset: Waiter ChangeSetCreateComplete failed: Waiter encountered a terminal failure state Status: FAILED.

Transform AWS::Serverless-2016-10-31 failed with: Invalid Serverless Application Specification document. Resource with id [CognitoUserPool] is invalid. property UsernameConfiguration not defined for resource of type AWS::Cognito::UserPool

Transform AWS::Serverless-2016-10-31 failed with: Invalid Serverless Application Specification document. Resource with id [CognitoUserPool] is invalid. property AccountRecoverySetting not defined for resource of type AWS::Cognito::UserPool

**Expected result:**

Successful deploy

ShreyaGangishetty:
`AWS::Cognito::UserPool` is a passthrough property in SAM. 
Thanks for reporting the issue! We would really appreciate community contributions on this issue. 
All the properties of UserPool resource need to be updated [here](https://github.com/awslabs/serverless-application-model/blob/develop/samtranslator/model/cognito.py#L27) for SAM support.

dipak1296:
The above template mentioned by @phsstory now seems to be working but there are still a few cases where the serverless transform fails with the same error. Example:

`
AWSTemplateFormatVersion: 2010-09-09
Transform: 'AWS::Serverless-2016-10-31'
Resources:
[...]

PR #1581, opened 2020-04-29
fix : missing UserPool properties (#1506)

jtaylor00:
*Issue #, if available:*
#1506 
*Description of changes:*
Added missing property types

*Checklist:*

- [ ] Write/update tests
- [x] `make pr` passes
- [ ] Update documentation


jfuss:
We need a better way to deal with these kinds of things. We shouldn't have to update a model in the code every time something new is release on the base CloudFormation Resources.

Thanks for the contribution @jtaylor00

carlnordenfelt:
@jfuss Do you know when this will be available? Doesn't look like it made it into 1.25.0
Thanks

PR #1588, opened 2020-05-08
document IpV6 option on Domain Configuration object

timoschilling:
*Issue #, if available:*

*Description of changes:*

*Checklist:*

- [ ] Write/update tests
- [ ] `make pr` passes
- [x] Update documentation


jfuss:
@timoschilling Change looks fine, was this just something we missed initially and has always been supported? It seems like that might be the case, looking through the code. Wanted to double check to make sure this is just a doc issue and not something we also need to implement.

timoschilling:
Yes it's supported since a while.

jfuss:
@timoschilling Thanks for the patch. I will make sure this gets added to our docs.aws site as well.

Issue #1271, opened 2019-11-28
Function Event Schedule - Can't use conditional for "Enabled" property


muzehyun:
**Description:**
I have created Schedule event for a function.
And I wanted to enable it conditionally
Whatever conditional I have used, it's always enabled.
(I already verified with inverted true/false - conditional might wrong - still always enabled)
I have confirmed through cloudwatch log too.
If I use `Enabled: false` directly, it works well.
- similar issue - #1142 

```
Conditions:
  IsDev:
    Fn::Equals:
      - Ref: Stage
      - dev

SomeFunction:
  Type: AWS::Serverless::Function
  Properties:
    ...
    Events:
      warming:
        Type: Schedule
        Properties:
          Enabled:
            Fn::If:
              - IsDev
              - false
              - true
          Schedule: "cron(* * * * ? *)"
```

**Observed result:**
![Screen Shot 2019-11-28 at 12 40 26 AM](https://user-images.githubusercontent.com/3071771/69790785-fbb5d680-1177-11ea-8b4b-e04b5d5ebe9d.png)

**Expected result:**
Properly "Disabled" based on conditional.

jfuss:
This is a duplicate of #1329 

Closing this one in favor of that since #1329 has a bunch of comments.

Issue #1329, opened 2019-12-16
Allow to pass parameter value for 'Enabled' property in ScheduledEvent of 'AWS::Serverless::Function'


jeswanthamazon:
When you reference a parameter for 'Enabled' property in ScheduledEvent of 'AWS::Serverless::Function', it is not honoring the parameter value passed and setting the "State": "ENABLED" for Events::Rule in the processed template. 

'Enabled' property in ScheduledEvent of 'AWS::Serverless::Function' currently only accepts 'true' or 'false' bool values hard-coded as string in template. 

in the background when composing the processed template, 'AWS::Serverless-2016-10-31' transform sees:

- if 'Enabled' property is set in the ScheduledEvent, 
- if it is set, 'State' Is set to 'DISABLED' only if the 'Enabled' property is 'False'. If it is anything else, 'State' Is set to 'ENABLED'.
You can reference the transform code here: 
https://github.com/awslabs/serverless-application-model/blob/93b6ac0afb6f322e047d6e4c4887f0d829e54b88/samtranslator/model/eventsources/push.py#L113

 Currently Serverless transform is not capable of reading the parameter value to set the 'State' of Event::Rule. 

 Please add support to parse the parameter value and set the 'State' property value to 'DISABLED' /'Enabled' accordingly.

SydneyUniLibrary-Jim:
This is related to #1360

praneetap:
Thanks for the issue report! We have passed this along to our product team for possible prioritization. Please +1 on the feature to help with prioritization.

RichieRunner:
When is this going to be fixed?

jwalsh2me:
@jfuss is this still under review?

RichieRunner:
this is such a shame. you have a core feature broken, and nobody who's paid to work on it even cares.

z0ph:
Same issue here, lost a few hours on troubleshooting this...

RichieRunner:
Since it's abundantly clear now that AWS and the maintainers of this codebase are refusing to prioritize this to be a critical bug, what is everyone's recommendation for disabling the `Enabled` flag if you are building resources in NP?

I am trying to achieve this kind of conditional logic:
![image](https://user-images.githubusercontent.com/20289967/100397949-496e5480-3001-11eb-8c2a-8b8bdcf572b8.png)

z0ph:
[...]

PR #1666, opened 2020-07-27
Fix EventSource Schedule passes Enabled to State

lightpriest:
*Issue*

I stumbled upon this but noticed there's an open issue: #1271 

*Description of changes:*

In the documentation, it is stated that Enabled is passed directly to State,
and also that intrinsic functions are supported in all of the Schedule
EventSource properties.

Added the expected "passing" of the value, while maintaining the previous
behavior of `True -> ENABLED` and `False -> DISABLED`.

Wrote missing tests for `Schedule` event source, and for this specific test.

*Checklist:*

- [x] Write/update tests
- [x] `make pr` passes
- [x] Update documentation

*Examples?*


lightpriest:
I couldn't verify this change on a real template, how should I do this? I tried running the script under `bin` but it always failed with some different error (on packaged template, local, built, etc.) and I didn't have time to dig into each.

How do I translate a template that I have locally?

lightpriest:
Got it to work. A fully working template renders successfully.

lightpriest:
Any comments? Something? :)

Schwusch:
I would like to see this PR merged. We currently have no workaround other than to look at other solutions like CDK.

samtranslator/model/eventsources/push.py
```diff
@@ -115,7 +115,10 @@ def to_cloudformation(self, **kwargs):
 
         events_rule.ScheduleExpression = self.Schedule
         if self.Enabled is not None:
-            events_rule.State = "ENABLED" if self.Enabled else "DISABLED"
+            if isinstance(self.Enabled, bool):
+                events_rule.State = "ENABLED" if self.Enabled else "DISABLED"
+            else:
+                events_rule.State = self.Enabled
```
Schwusch, line 121: Can `self.Enabled` be an [instrinsic function](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference.html) at this point?
If so, is there a way to evaluate it?

benkehoe:
I'd also love to see this PR merged. Ran into this today and had to explode a couple schedule event sources into constituent rules, IAM roles, and Lambda permissions.

samtranslator/model/eventsources/push.py
lightpriest: I think that if it's passed directly to CloudFormation, it becomes an intrinsic function, isn't it so?

lightpriest:
Anyone? 🥺

milesgranger:
Ping as a plus one for this, quite unexpected behavior and uneconomic if we're talking about 100s of events that need to be toggled via parameters. :+1: 
Thanks for the work here @lightpriest :clap:

samtranslator/model/eventsources/push.py
```diff
@@ -96,6 +96,7 @@ class Schedule(PushEventSource):
         "Schedule": PropertyType(True, is_str()),
         "Input": PropertyType(False, is_str()),
         "Enabled": PropertyType(False, is_type(bool)),
+        "State": PropertyType(False, is_str()),
```
hawflau, line 99: Why do we want to add the `State` property? 

Essentially we just want to add intrinsic support (at least for `Fn::If` and `!Ref`). So, what we actually should do is:
- If `Enabled` is an intrinsic, just assign it to `events_rule.State`
- If `Enabled` is a string, keep it as how we handle it today (for sake of not introducing breaking change)

hawflau (changes requested):

samtranslator/model/eventsources/push.py
hawflau: okay, looking at this [comment](https://github.com/aws/serverless-application-model/issues/1329#issuecomment-1157759679), and realizing adding intrinsic support to `Enabled` also introduces breaking change, I think adding `State` property makes sense to me now.

samtranslator/model/eventsources/push.py
```diff
@@ -122,8 +123,18 @@ def to_cloudformation(self, **kwargs):
...
+
+        if self.State and self.Enabled:
+            raise InvalidEventException(self.relative_id, "State and Enabled Properties cannot both be specified.")
+
+        if self.State:
+            events_rule.State = self.State
+
         if self.Enabled is not None:
-            events_rule.State = "ENABLED" if self.Enabled else "DISABLED"
+            if isinstance(self.Enabled, bool):
```
hawflau, line 134: If we are introducing a new property to get the behavior right with intrinsic support, why don't we just leave this as it was before?

samtranslator/model/eventsources/push.py
```diff
@@ -122,8 +123,18 @@ def to_cloudformation(self, **kwargs):
         resources.append(events_rule)
 
         events_rule.ScheduleExpression = self.Schedule
+
+        if self.State and self.Enabled:
```
hawflau, line 127: Seems like we should explicitly check `self.Enabled` is not `None`. Consider:
```
Enabled: false # bool
State: ENABLED
```

hawflau (changes requested):

samtranslator/model/eventsources/push.py
jfuss: I meant to revert this back but looks like it got lost in me putting this task up and down. Will update.

hawflau (approved):
LGTM!

mildaniel (approved):

PR #1671, opened 2020-07-29
Fix: Top SAM crashes

prenx4x:
*Issue #, if available:*

*Description of changes:*
Fixed some common SAM crashes

Changes validated using unit tests

*Checklist:*

- [x] Write/update tests
- [x] `make pr` passes
- [ ] Update documentation

*Examples?*

### DIFF
--- a/samtranslator/swagger/swagger.py
+++ b/samtranslator/swagger/swagger.py
@@ -622,8 +622,9 @@ class SwaggerEditor(object):
                 # We want to ensure only a single ApiKey security entry exists while keeping everything else
                 for security in existing_security:
                     if not isinstance(security, dict):
-                        raise InvalidDocumentException(["Invalid entry {security} in 'security'"
-                                                       .format(security=security)])
+                        raise InvalidDocumentException(
+                            ["Invalid entry {security} in 'security'".format(security=security)]
+                        )
                     if apikey_security_names.isdisjoint(security.keys()):
                         existing_non_apikey_security.append(security)
                     else:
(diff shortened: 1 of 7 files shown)


samtranslator/swagger/swagger.py
```diff
@@ -621,6 +621,10 @@ def set_path_default_apikey_required(self, path):
                 # (e.g. sigv4 (AWS_IAM), authorizers, NONE (marker for ignoring default authorizer))
                 # We want to ensure only a single ApiKey security entry exists while keeping everything else
                 for security in existing_security:
+                    if not isinstance(security, dict):
+                        raise InvalidDocumentException(
+                            ["Invalid entry {security} in 'security'".format(security=security)]
```
jfuss, line 626: Why is it invalid?

samtranslator/swagger/swagger.py
```diff
@@ -830,10 +834,10 @@ def add_models(self, models):
             model_properties = schema.get("properties")
 
             if not model_type:
-                raise ValueError("Invalid input. Value for type is required")
+                raise InvalidDocumentException(["Invalid input. Value for type is required"])
 
             if not model_properties:
-                raise ValueError("Invalid input. Value for properties is required")
+                raise InvalidDocumentException(["Invalid input. Value for properties is required"])
```
jfuss, line 840: Same here

samtranslator/swagger/swagger.py
```diff
@@ -846,6 +850,8 @@ def add_resource_policy(self, resource_policy, path, api_id, stage):
         """
         if resource_policy is None:
             return
+        if not isinstance(resource_policy, dict):
+            raise InvalidDocumentException(["Invalid resource policy"])
```
jfuss, line 854: Make sure to provide some details about where this is in the template. We generally do this through adding the logical id, so the customer at least knows what resource they need to dig into. This goes for all the exceptions, not just this one.

jfuss (changes requested):
Can we add additional test cases in `test_transform_invalid_document` to verify we have fixed the issue.

prenx4x:
We can close this PR as I am breaking it in favor of #1757 and others to come.

PR #1681, opened 2020-08-11
Fix: OptIn SNS Topic assignment results in incorrect function resourc…

prenx4x:
OptIn SNS Topic assignment results in incorrect function resource policy principal

*Issue #, if available:*

*Description of changes:*
Added logic to check if SNS region is in opt-in region or not, and updated principal accordingly.

Ran all tests and manual verification

*Checklist:*

- [ ] Write/update tests
- [x] `make pr` passes
- [ ] Update documentation

*Examples?*

### DIFF
--- a/samtranslator/utils/get_service_regions.py
+++ b/samtranslator/utils/get_service_regions.py
@@ -1,0 +1,37 @@
+import boto3
+
+
+def get_all_available_regions(service_name):
+    """
+    Wrapper function to get all available regions including ones that are not regional endpoints
+
+    :param service_name: Name of service to get available regions like 's3' etc.
+    :return:Returns a list of endpoint names (e.g., ["us-east-1"]).
+    """
+
+    return boto3._get_default_session().get_available_regions(service_name, allow_non_regional=True)
+
+
+def get_available_serviceable_regions(service_name):
+    """
+    Wrapper function to get available regions excluding ones that are not regional endpoints
+
+    :param service_name: Name of service to get available regions like 's3' etc.
+    :return:Returns a list of endpoint names (e.g., ["us-east-1"]).
+    """
+
+    return boto3._get_default_session().get_available_regions(service_name, allow_non_regional=False)
+
+
+def get_available_nonserviceable_regions(service_name):
+    """
+    Wrapper function to get only non serviceable regions
+
+    :param service_name: Name of service to get available regions like 's3' etc.
+    :return:Returns a list of endpoint names (e.g., ["us-east-1"]).
+    """
+
+    all_regions = boto3._get_default_session().get_available_regions(service_name, allow_non_regional=True)
+    s_regions = boto3._get_default_session().get_available_regions(service_name, allow_non_regional=False)
+
+    return list(set(all_regions) - set(s_regions))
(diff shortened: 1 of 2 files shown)


prenx4x:
We might need to write some tests for this.

jfuss:
> We might need to write some tests for this.

We should always have tests for any change.

sriram-mv:
@prenx4x could you rebase this? and I can take a look at this. Thanks!

samtranslator/utils/get_service_regions.py
```diff
@@ -0,0 +1,37 @@
...
+    :param service_name: Name of service to get available regions like 's3' etc.
+    :return:Returns a list of endpoint names (e.g., ["us-east-1"]).
+    """
+
+    return boto3._get_default_session().get_available_regions(service_name, allow_non_regional=False)
+
+
+def get_available_nonserviceable_regions(service_name):
+    """
+    Wrapper function to get only non serviceable regions
```
sriram-mv, line 28: can we explain what serviceable and non-serviceable regions are in the docstring

prenx4x: Added more explanation. Let me know if it suffices.

samtranslator/utils/get_service_regions.py
```diff
@@ -0,0 +1,38 @@
...
+def get_available_nonserviceable_regions(service_name):
+    """
+    Wrapper function to get only non serviceable or AWS opt-in regions.
+    AWS opt-in regions require users to enable them manually in order to use them.
+
+    :param service_name: Name of service to get available regions like 's3' etc.
+    :return:Returns a list of endpoint names (e.g., ["us-east-1"]).
+    """
+
+    all_regions = boto3._get_default_session().get_available_regions(service_name, allow_non_regional=True)
```
sriram-mv, line 35: you could reuse the above helper functions here.

The docstring still needs clarity, maybe you can clarify by saying which regions are opt-in regions vs which regions are not.

sriram-mv (changes requested):

samtranslator/utils/get_service_regions.py
prenx4x: I think opt-in regions keeps changing, right? initially a region is opt-in only, but then becomes GA as per my knowledge

awood45: There are a few possible issues here, one of which is that a new region would require that `boto3` has that region listed, or else it will fail to function. What happens if a region is valid but is not returned by this?

tests/model/eventsources/test_sns_event_source.py
```diff
@@ -77,3 +78,20 @@ def test_to_cloudformation_when_sqs_subscription_disable(self):
         self.assertEqual(subscription.Endpoint, "arn:aws:lambda:mock")
         self.assertIsNone(subscription.Region)
         self.assertIsNone(subscription.FilterPolicy)
+
+    def test_to_cloudformation_when_region_is_optin(self):
```
awood45, line 82: We should define what happens when a region is not returned at all by `boto3`, and then test that behavior holds in the same way.

awood45 (changes requested):
Some possible race conditions with this implementation approach, will discuss directly.

PR #1686, opened 2020-08-14
feat: Allow Boto3 Session to be configured

jfuss:
*Issue #, if available:*
https://github.com/awslabs/aws-sam-cli/issues/442

*Description of changes:*
This is the first part in solving https://github.com/awslabs/aws-sam-cli/issues/442. Specifically, SAM uses boto to get a region and then pick the associated partition it lives within. This is mainly for two things (that I can tell):
1. Get the partition so it can be added to a generated arn in the generated template
2. Add pseudo values as Parameters so SAM can do some basic resolving. 

For the first, we shouldn't need to explicitly put the partition into the template anywhere but instead use the CloudFormation pseudo value when writing the template. I attempted to do this initially but required changing every single output.json test file we have and the pr started to really grow. So I would like to handle that outside of this PR and get the bug addressed in SAM CLI first, and then come back and see how we can do somethings better.

For the second, we still need to find the partition and since there is no way to do this in the sdk, basing it on the region is the way to go.

This change adds an optional parameter into the transform call (effectively the 'public' api to translate the template). This optional parameter is currently only used to get the region but one could image this would be something SAM might be able to use in the future to be able to create any AWS Clients through Boto.

This will help solve the above issue because currently SAM creates a new session and has no way to know what profile or region was passed into SAM CLI. Because of this, locally customers see command failures with stacktraces from SAM for cases that should work. If approved, I will update SAM CLI to pass in a Boto Session which will end up resolving the issue linked above.

Added unit tests and since this is an optional parameter, it is backwards compatible.

*Checklist:*

- [ ] Write/update tests
- [ ] `make pr` passes
- [ ] Update documentation

*Examples?*

### DIFF
--- a/samtranslator/translator/arn_generator.py
+++ b/samtranslator/translator/arn_generator.py
@@ -1,7 +1,13 @@
 import boto3
 
 
+class NoRegionFound(Exception):
+    pass
+
+
 class ArnGenerator(object):
+    class_boto_session = None
+
     @classmethod
     def generate_arn(cls, partition, service, resource, include_account_id=True):
         if not service or not resource:
@@ -43,7 +49,17 @@ class ArnGenerator(object):
         if region is None:
             # Use Boto3 to get the region where code is running. This uses Boto's regular region resolution
             # mechanism, starting from AWS_DEFAULT_REGION environment variable.
-            region = boto3.session.Session().region_name
+
+            if ArnGenerator.class_boto_session is None:
+                region = boto3.session.Session().region_name
+            else:
+                region = ArnGenerator.class_boto_session.region_name
+
+        # If region is still None, then we could not find the region. This will only happen
+        # in the local context. When this is deployed, we will be able to find the region like
+        # we did before.
+        if region is None:
+            raise NoRegionFound("AWS Region cannot be found")
 
         # setting default partition to aws, this will be overwritten by checking the region below
         partition = "aws"
--- a/tests/sdk/test_parameter.py
+++ b/tests/sdk/test_parameter.py
@@ -1,9 +1,10 @@
 from parameterized import parameterized, param
 
-import pytest
 from unittest import TestCase
 from samtranslator.sdk.parameter import SamParameterValues
-from mock import patch
+from mock import patch, Mock
+
+from samtranslator.translator.arn_generator import NoRegionFound
 
 
 class TestSAMParameterValues(TestCase):
@@ -101,3 +102,10 @@ class TestSAMParameterValues(TestCase):
         sam_parameter_values = SamParameterValues(parameter_values)
         sam_parameter_values.add_pseudo_parameter_values()
         self.assertEqual(expected, sam_parameter_values.parameter_values)
+
+    def test_add_pseduo_parameter_values_raises_NoRegionFound(self):
+        boto_session_mock = Mock()
+        boto_session_mock.region_name = None
+        sam_parameter_values = SamParameterValues({})
+        with self.assertRaises(NoRegionFound):
+            sam_parameter_values.add_pseudo_parameter_values(session=boto_session_mock)
--- a/tests/sdk/test_parameter.py
+++ b/tests/sdk/test_parameter.py
@@ -103,7 +103,7 @@ class TestSAMParameterValues(TestCase):
         sam_parameter_values.add_pseudo_parameter_values()
         self.assertEqual(expected, sam_parameter_values.parameter_values)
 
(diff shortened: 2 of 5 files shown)


tests/translator/test_arn_generator.py
```diff
@@ -0,0 +1,34 @@
+from unittest import TestCase
```
mgrandis, line 1: This file has moved to `tests/unit/translator/test_arn_generator.py`

tests/sdk/test_parameter.py
```diff
@@ -101,3 +102,10 @@ def test_add_pseudo_parameter_values_aws_partition_not_override(self):
         sam_parameter_values = SamParameterValues(parameter_values)
         sam_parameter_values.add_pseudo_parameter_values()
         self.assertEqual(expected, sam_parameter_values.parameter_values)
+
+    def test_add_pseduo_parameter_values_raises_NoRegionFound(self):
```
mgrandis, line 106: nit: `pseduo` -> `pseudo`

samtranslator/translator/arn_generator.py
```diff
@@ -42,7 +48,17 @@ def get_partition_name(cls, region=None):
...
+            if ArnGenerator.class_boto_session is None:
+                region = boto3.session.Session().region_name
+            else:
+                region = ArnGenerator.class_boto_session.region_name
+
+        # If region is still None, then we could not find the region. This will only happen
+        # in the local context. When this is deployed, we will be able to find the region like
+        # we did before.
+        if region is None:
+            raise NoRegionFound("AWS Region cannot be found")
```
mgrandis, line 61: We could maybe add `if running locally, is AWS_DEFAULT_REGION set?` at the end of the message?

mgrandis (changes requested):
@jfuss Is this still relevant?
If so, can we move forward and merge it?
This would help with the associated issue.
We would then also have a clearer error message.

samtranslator/translator/arn_generator.py
jfuss: Customer's shouldn't have to set an Env Var. With this change, SAM will accept a Boto Session. Which then SAM can used to create clients with those creds. SAM CLI will then pass this information to SAM, which will resolve the failures SAM CLI sees.

tests/sdk/test_parameter.py
jfuss: Done

tests/translator/test_arn_generator.py
jfuss: The file didn't move but looks like there are two. SAM has some weird testing structures.

jfuss:
@mgrandis Yes this is still relevant, just slipped off my radar.

mgrandis (approved):

tests/translator/test_arn_generator.py
```diff
@@ -0,0 +1,34 @@
...
+
+    @parameterized.expand(
+        [("us-east-1", "aws"), ("cn-east-1", "aws-cn"), ("us-gov-west-1", "aws-us-gov"), ("US-EAST-1", "aws")]
+    )
+    def test_get_partition_name(self, region, expected):
+        actual = ArnGenerator.get_partition_name(region)
+
+        self.assertEqual(actual, expected)
+
+    def test_get_partition_name_raise_NoRegionFound(self):
```
mgrandis, line 20: You may need a `@patch("boto3.session.Session.region_name", "")` here

samtranslator/translator/arn_generator.py
```diff
@@ -43,7 +49,17 @@ def get_partition_name(cls, region=None):
         if region is None:
```
mgrandis, line 49: Shall we move to `if not X` tests in here to also catch empty strings?

tests/translator/test_arn_generator.py
jfuss: Ya. I couldn't remember where I was running into the issues before. Added.

samtranslator/translator/arn_generator.py
jfuss: My goal was not to adjust the current behavior in this pr but instead focus on providing a better interface for tools (like SAM CLI), who get a profile and region through a command line. There is probably a case to validate the region in this manner but we aren't changing that expectation here at all. I don't think boto3 has a way to do this, so we would have to keep a list of regions. For now, I would prefer to keep the behavior in line with the previous (default to aws partition on non-region strings) and later add in this further validation if needed.

PR #1689, opened 2020-08-24
add new arguments to allow vpc/vpce endpoints as intrinsics

wchengru:
*Issue #, if available:*

*Description of changes:*

*Checklist:*

- [x] Write/update tests
- [x] `make pr` passes
- [x] Update documentation

*Examples?*

### DIFF
--- a/samtranslator/swagger/swagger.py
+++ b/samtranslator/swagger/swagger.py
@@ -3,8 +3,8 @@ import json
 import re
 from six import string_types
 
-from samtranslator.model.intrinsics import ref, is_intrinsic_no_value
-from samtranslator.model.intrinsics import make_conditional, fnSub, is_intrinsic_if
+from samtranslator.model.intrinsics import ref
+from samtranslator.model.intrinsics import make_conditional, fnSub
 from samtranslator.model.exceptions import InvalidDocumentException, InvalidTemplateException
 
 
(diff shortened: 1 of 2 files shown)


tests/swagger/test_swagger.py
```diff
@@ -1200,7 +1208,90 @@ def test_must_add_vpc_deny(self):
...
+                },
+            ],
+        }
+
+        self.assertEqual(deep_sort_lists(expected), deep_sort_lists(self.editor.swagger[_X_POLICY]))
+
+
+    def test_must_add_vpc_allow_string_and_instrinic(self):
+
+        resourcePolicy = {
```
awood45, line 1224: A couple changes:
- Please make sure the explicit intrinsic entries do NOT have "VPC" or "VPCE" in their strings, to ensure we don't regress on requiring regex matching.
- Please include more than one to ensure it's handling the full list.

Can we also test both whitelist and blocklist in a single test, to ensure both are applied?

samtranslator/swagger/swagger.py
```diff
@@ -980,33 +1002,42 @@ def _add_ip_resource_policy_for_method(self, ip_list, conditional, resource_list
...
+                    vpc_list.append(endpoint)
+            if vpc_list:
+                condition.setdefault("aws:SourceVpc", []).extend(vpc_list)
+            if vpce_list:
+                condition.setdefault("aws:SourceVpce", []).extend(vpce_list)
+        if intrinsic_vpc_endpoint_list is not None:
+            condition.setdefault("aws:SourceVpc", []).extend(intrinsic_vpc_endpoint_list)
+        if intrinsic_vpce_endpoint_list is not None:
+            condition.setdefault("aws:SourceVpce", []).extend(intrinsic_vpce_endpoint_list)
+        if not condition:
```
awood45, line 1038: What is this new clause doing?

awood45 (changes requested):
Test request, as well as a question.

samtranslator/swagger/swagger.py
wchengru: The plan here is adding a sanity check that it doesn't write empty endpoint lists into this section. It is similar as the original code does at [this line](https://github.com/aws/serverless-application-model/blob/develop/samtranslator/swagger/swagger.py#L989). 
I just realized that although the three lists in the `endpoint_dict` shouldn't be empty, but the regex classified vpc/vpce list may still be empty. So I modified this condition to better verify that the endpoint allow/deny lists are not empty.

tests/swagger/test_swagger.py
wchengru: Updated intrinsic entries and added more test cases.

awood45 (approved):

PR #1698, opened 2020-09-03
updating CodeDeploy lambda managed policy

vrr-21:
*Issue #, if available:*

*Description of changes:* The managed policy `AWSCodeDeployRoleForLambda` used for Lambda deployments has broad permissions, providing publish access to all SNS topics within the user's account.
This change replaces that with a new policy `AWSCodeDeployRoleForLambdaLimited` which removes those permissions.
To handle customers who may have triggers set up for SNS notifications on CodeDeploy, this change also attaches the `AmazonSNSFullAccess` managed policy when the deployment preferences include trigger configurations.

- `make test` run passes.
- Used the translator to transform a SAM template to a CloudFormation template and then deployed the same (Using instructions in the Development Guide). Then tried a deployment with the newly created resources successfully.
- Also verified that only the `AWSCodeDeployRoleForLambdaLimited` policy is attached when the SAM template does not contain any Trigger Configurations. When it does contain Trigger Configurations, the `AmazonSNSFullAccess` managed policy is also attached.

*Checklist:*

- [x] Write/update tests
- [x] `make pr` passes
- [ ] Update documentation

*Examples?*

### DIFF
--- a/samtranslator/model/preferences/deployment_preference_collection.py
+++ b/samtranslator/model/preferences/deployment_preference_collection.py
@@ -102,7 +102,7 @@ class DeploymentPreferenceCollection(object):
             ],
         }
         iam_role.ManagedPolicyArns = [
-            ArnGenerator.generate_aws_managed_policy_arn("service-role/AWSCodeDeployRoleForLambda")
+            ArnGenerator.generate_aws_managed_policy_arn("service-role/AWSCodeDeployRoleForLambdaLimited")
         ]
 
         return iam_role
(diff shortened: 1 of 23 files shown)


Helen1987:
@pfeilbr , @me2resh @szilveszter can it be reviewed?

me2resh (approved):
LGTM

samtranslator/model/preferences/deployment_preference_collection.py
```diff
@@ -102,7 +102,7 @@ def _codedeploy_iam_role(self):
             ],
         }
         iam_role.ManagedPolicyArns = [
-            ArnGenerator.generate_aws_managed_policy_arn("service-role/AWSCodeDeployRoleForLambda")
+            ArnGenerator.generate_aws_managed_policy_arn("service-role/AWSCodeDeployRoleForLambdaLimited")
```
moelasmar, line 105: there is some conflict in this part, we already applied this change as following:
https://github.com/aws/serverless-application-model/blob/579353c793b9c779c0f869f7bb7e4a16b16f8767/samtranslator/model/preferences/deployment_preference_collection.py#L112-L121
my understanding is Policy AWSCodeDeployRoleForLambda will be deprecated only in MVP regions

vrr-21: Thank you for reviewing this,

No, we will be deprecating AWSCodeDeployRoleForLambda policy in all regions.

samtranslator/model/preferences/deployment_preference_collection.py
```diff
@@ -135,13 +135,16 @@ def deployment_group(self, function_logical_id):
 
         deployment_group.DeploymentStyle = {"DeploymentType": "BLUE_GREEN", "DeploymentOption": "WITH_TRAFFIC_CONTROL"}
 
+        if deployment_preference.trigger_configurations:
```
hoffa, line 138: A comment on why this is necessary will be helpful.

jfuss:
Closing. As written this is a breaking change and we avoid this for customers. We add the new managed policy in specific cases, which can be found [here](https://github.com/aws/serverless-application-model/blob/develop/samtranslator/model/preferences/deployment_preference_collection.py#L116)

PR #1723, opened 2020-09-24
request a new policy template to be added --- CognitoReadOnly

waynehong789:
request a new policy template to be added --- CognitoReadOnly

*Issue #, if available:*

*Description of changes:*

*Checklist:*

- [ ] Write/update tests
- [ ] `make pr` passes
- [ ] Update documentation

*Examples?*

### DIFF
--- a/samtranslator/policy_templates_data/policy_templates.json
+++ b/samtranslator/policy_templates_data/policy_templates.json
@@ -2306,6 +2306,36 @@
           }
         ]
       }
+    },
+"CognitoReadOnlyPolicy": {
+      "Description": "Gives permission to read Cognito identity",
+      "Parameters": {
+        
+      },
+      "Definition": {
+        "Statement": [
+          {
+                  "Effect": "Allow",
+                  "Action": [
+                      "cognito-identity:Describe*",
+                      "cognito-identity:Get*",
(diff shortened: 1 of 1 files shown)


awood45:
This is a really broad policy template, it would provide full access to all Cognito resources in an entire account. What are you looking for this to accomplish that is distinct from the `arn:aws:iam::aws:policy/AmazonCognitoReadOnly` managed policy?

waynehong789:
@awood45 Thanks for your reply. I am looking for the way to setup the policy for a lambda/stateMachine in SAM template which will allow us to query the user/account details on Cognito. So, it will be perfect if something like arn:aws:iam::aws:policy/AmazonCognitoReadOnly can be added into the SAM template policy. BTW, AmazonCognitoReadOnly is not on the page: https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/serverless-policy-templates.html

jfuss:
Closing the PR as in current state not something we can accept.

@waynehong789 Moving forward, please create a Github issue for the request first so we can discuss the use-case before adding policies. Typically, we look for policies that are restricted to resources and fit a use-case need for the community. You can always provide your own Role or Policies to `AWS::Serverless::Function` for your specific need.

Issue #1637, opened 2020-06-30
Unable to use cross-account token lambda authorizer with SAM template


xiananfan:
**Description:**

We have 2 AWS accounts, let's say account A and account B. Account A contains a lambda authorizer we want to use in Account B with API Gateway as a token authorizer. Using the classic cloud formation template, we will be able to create the authorizer reference in account B, and it will not attempt to modify the authorizer in account A to update its resource-based policy to allow cross-account access. The CFT deployment will be able to complete without problems.

However, using SAM syntax like below will cause the CFT deployment to fail:
```
# template.yaml, intended to be deployed in Account B
    TestAuthServiceGateway:
        Type: AWS::Serverless::Api
        Properties:
            StageName: !Ref Environment
            OpenApiVersion: '3.0.0'
            Auth:
                DefaultAuthorizer: MyAuthorizerInAccountA
                Authorizers:
                    MyAuthorizerInAccountA:
                        FunctionArn: arn:aws:lambda:us-west-2:<AccountA>:function:dev-us-west-2-StandardAuthorizer
                        FunctionPayloadType: 'TOKEN'
                        Identity:
                            ReauthorizeEvery: 0 # disable cache
                            ValidationExpress: ^Bearer [a-zA-Z0-9\-_]+?\.[a-zA-Z0-9\-_]+?\.([a-zA-Z0-9\-_]+)?$
```
The reason is the user who triggered this deployment is currently in Account B and does not have permission to modify the authorizer in Account A.

The error CFT gives looks like below:
```
User: arn:aws:iam::<AccountB>:user/CICD is not authorized to perform: lambda:AddPermission on resource: arn:aws:lambda:us-west-2:<AccountA>:function:dev-us-west-2-StandardAuthorizer (Service: AWSLambda; Status Code: 403; Error Code: AccessDeniedException; Request ID: 964a7be1-fb95-4759-8c2d-11476c74630b)
```

My questions are:
1. Is there a way to setup IAM role/policy/trust relationship, such that user in Account B, at cloudformation deployment time, can have the permission to modify cross-account authorizer's resource-based policy? This [wiki](https://docs.amazonaws.cn/en_us/lambda/latest/dg/access-control-resource-based.html) does not suggest "lambda:AddPermission" is a supported API in resourced-based policy. I also gave it a try, and could not get it to work.
[...]

PR #1735, opened 2020-10-06
Only create LambdaPermission for authorizer if invoke role is blank

icj217:
*Issue #, if available:* #1637 

*Description of changes:* Made the creation of a `AWS::Lambda:LambdaPermission` for lambda authorizer functions dependent on the presence of the `FunctionInvokeRole` parameter. When a role is provided, it indicates that resource-based permissions on the Lambda function are insufficient for the API to invoke the function (see documentation of the `AuthorizerCredentials` parameter [here](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-apigateway-authorizer.html)). **This will allow cross-account authorizer functions to be specified and used in SAM resources. Currently specifying a cross-account lambda function will fail because Cloudformation tries to create a LambdaPermission on the remote function.**

*Description of how you validated changes:* Added a unit test that confirms a LambdaPermission is not created when the `FunctionInvokeRole` parameter is populated in an Api resource's authorizer config.

*Checklist:*

- [x] Write/update tests
- [x] `make pr` passes
- [ ] Update documentation

*Examples?*


icj217:
I don't think this is a workable solution without adding some new property to control the generation of the lambda permission resource. Going to close this. See workaround/mitigation in #1637

PR #1744, opened 2020-10-15
chore: Upgrade outdated dependencies in base.txt and dev.txt

aahung:
*Issue #, if available:*

*Description of changes:*

Update all dependencies versions in requirement txts to match their latest available versions.
Since we are either using `~=x.y` or `>=x.y.z`, this PR does not have effective change.

I left `pylint` untouched because it is not being used right now. I created another PR #1745 for enabling it.

*Checklist:*

- [ ] Write/update tests
- [x] `make pr` passes
- [ ] Update documentation

*Examples?*

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -1,6 +1,6 @@
 pyrsistent~=0.16.0; python_version<"3"
 boto3~=1.5
 enum34~=1.1; python_version<"3.4"
-jsonschema~=3.0
-six~=1.11
+jsonschema~=3.2
+six~=1.15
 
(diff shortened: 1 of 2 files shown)


requirements/base.txt
```diff
@@ -1,6 +1,6 @@
 pyrsistent~=0.16.0; python_version<"3"
 boto3~=1.5
 enum34~=1.1; python_version<"3.4"
-jsonschema~=3.0
-six~=1.11
+jsonschema~=3.2
```
sriram-mv, line 4: have we verified that this does not break any tests?

sriram-mv (changes requested):

requirements/base.txt
aahung: These changes do not do anything in fact. `~=3.0` and `~=3.2` both spec to `3.x`

sriram-mv (approved):

qingchm (approved):

PR #1748, opened 2020-10-22
Manage black version using requirement file

aahung:
*Issue #, if available:*

*Description of changes:*

update black version to 20.8b1 and make it managed by requirement dev.txt.

This change is a counterpart of https://github.com/aws/aws-sam-cli/pull/2314 

**Note: Since black does not support Python 2.7, nor the later black releases contains a binary artifact, I turned off black check for Python 2.7.** (https://github.com/psf/black/issues/1669)

*Checklist:*

- [ ] Write/update tests
- [x] `make pr` passes
- [ ] Update documentation

*Examples?*

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,4 +1,4 @@
-# https://github.com/psf/black/blob/master/.pre-commit-config.yaml
+# https://pre-commit.com/#repository-local-hooks
 repos:
 - repo: local
   hooks:
@@ -6,5 +6,4 @@ repos:
       name: black
       entry: black
       language: system
-      require_serial: true
       types: [python]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,7 +1,10 @@
+# https://github.com/psf/black/blob/master/.pre-commit-config.yaml
 repos:
-  - repo: https://github.com/python/black
-    rev: 19.3b0
-    hooks:
-      - id: black
-        language_version: python3.7
-        exclude_types: ['markdown', 'ini', 'toml', 'rst']
+- repo: local
+  hooks:
+    - id: black
+      name: black
+      entry: black
+      language: system
+      require_serial: true
+      types: [python]
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,12 +21,6 @@ matrix:
 
 install:
 # Install the code requirements
-- mkdir $HOME/bin-black
-- wget -O $HOME/bin-black/black https://github.com/python/black/releases/download/19.10b0/black
-- chmod +x $HOME/bin-black/black
-- export PATH=$PATH:$HOME/bin-black
-- black --version
-
 - make init
 
 # Install Docs requirements
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,15 +21,6 @@ matrix:
 
 install:
 # Install the code requirements
-- >
-  if [ $TOXENV = py27 ]; then
-    mkdir $HOME/bin-black
-    # make sure we are using the same black version as in dev.txt
-    wget -O $HOME/bin-black/black https://github.com/python/black/releases/download/20.8b1/black
-    chmod +x $HOME/bin-black/black
-    export PATH=$PATH:$HOME/bin-black
-    black --version
-  fi
 - make init
 
 # Install Docs requirements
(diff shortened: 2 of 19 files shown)


.travis.yml
```diff
@@ -21,12 +21,6 @@ matrix:
 
 install:
 # Install the code requirements
-- mkdir $HOME/bin-black
-- wget -O $HOME/bin-black/black https://github.com/python/black/releases/download/19.10b0/black
```
hoffa, line 25: Nice, good to get rid of this 😩

Makefile
```diff
@@ -20,6 +20,9 @@ dev: test
 # Verifications to run before sending a pull request
 pr: black-check init dev
 
+# Verifications to run before sending a pull request, skipping black check because black requires Python 3.6+
+pr2.7: init dev
```
hoffa, line 24: Nit: assuming existing code is `black`-compliant, would be nice if we didn't have to introduce new Python 2-specific cruft, but I'm assuming it doesn't play too well with `pip` et al. 🤒

hoffa (approved):

.pre-commit-config.yaml
```diff
@@ -1,7 +1,10 @@
+# https://github.com/psf/black/blob/master/.pre-commit-config.yaml
```
hawflau, line 1: Just clicked into the url and saw this line:
```
# Note: don't use this config for your own repositories. Instead, see
# "Version control integration" in README.md.
```
Can you make sure it is okay to follow its config?

aahung: interesting. I missed it..

aahung: Nice catch! Thanks! I did a bit research, although they don't suggest to use it, it does suit our use case (somehow):

in pre-commit doc: https://pre-commit.com/#repository-local-hooks

The doc states when we should use `local` repo instead of what black's "Version control integration" suggests:

> Hooks require state that is only present in a built artifact of your repository (such as your app's virtualenv for pylint).

And the example they gave (for `pylint`) is quite similar to what we are using right now, except for `require_serial` (if true this hook will execute using a single process instead of in parallel), which does not make too much a difference (I just removed it)

```
    -   id: pylint
        name: pylint
        entry: pylint
        language: system
        types: [python]
```

Makefile
aahung: When we run CI, there are multiple env including Python2.7, 3.6, 3.7 and 3.8. The code style will be checked in Python3 .X though. Here we just skip the check in one of the env, I think it should be ok 🤔

aahung: FYI, why the old way (download black binary directly) won't work:

https://github.com/psf/black/issues/1669

DEVELOPMENT_GUIDE.rst
```diff
@@ -28,16 +28,44 @@ Setup Python locally using `pyenv`_
...
+Since black is installed in virtualenv, when you follow `this
+instruction <https://black.readthedocs.io/en/stable/editor_integration.html>`__,
+``which black`` might give you this
+
+::
+
+   (samtranslator37) $ where black
+   /Users/<username>/.pyenv/shims/black
+
+However, IDEs such PyChaim (using FileWatcher) will have a hard time
```
hawflau, line 52: typo I guess?

hawflau: PyChaim -> PyCharm

aahung: fixed thanks!

.pre-commit-config.yaml
hawflau: looks good to me :)

DEVELOPMENT_GUIDE.rst
```diff
@@ -28,16 +28,44 @@ Setup Python locally using `pyenv`_
...
+Since black is installed in virtualenv, when you follow `this
+instruction <https://black.readthedocs.io/en/stable/editor_integration.html>`__,
+``which black`` might give you this
+
+::
+
+   (samtranslator37) $ where black
+   /Users/<username>/.pyenv/shims/black
+
+However, IDEs such PyCharm (using FileWatcher) will have a hard time
```
hawflau, line 52: I guess such -> such as?

aahung: omg. Thank you

hawflau (approved):

PR #1757, opened 2020-10-27
Added fixes for SAM crashes with tests

prenx4x:
*Issue #, if available:*

*Description of changes:*
Fixed some common SAM crashes

Changes validated using unit tests

*Checklist:*

- [x] Write/update tests
- [x] `make pr` passes
- [ ] Update documentation

*Examples?*


samtranslator/model/eventsources/push.py
```diff
@@ -657,6 +659,15 @@ def _add_swagger_integration(self, api, function, intrinsics_resolver):
                             ),
                         )
 
+                    if not isinstance(method_authorizer, str):
+                        raise InvalidEventException(
+                            self.relative_id,
+                            "Unable to set Authorizer [{authorizer}] on API method [{method}] for path [{path}] "
```
hoffa, line 665: More of a general note, but it seems this whole function could benefit from refactoring. It's large, repetitive and consists of multiple intertwined responsibilities (validation and processing).

samtranslator/model/eventsources/push.py
```diff
@@ -218,6 +218,8 @@ class S3(PushEventSource):
     def resources_to_link(self, resources):
         if isinstance(self.Bucket, dict) and "Ref" in self.Bucket:
             bucket_id = self.Bucket["Ref"]
+            if not isinstance(bucket_id, str):
+                raise InvalidEventException(self.relative_id, "'Ref' value in S3 events is not a valid string.")
```
aahung, line 222: The error messages seems not match the condition. `bucket_id` here is not a str, while the error message makes me feel like it is a str, just not a valid one. Does the situation here mean the `Ref` key is missing?

prenx4x: Ref is a parameter name which should have some value. like Ref: somevalue . Bucket_id is that value. Therefore we are checking bucket_id which should be a string i.e. value of 'Ref' in the template should be a string. Make sense?

aahung: will `bucket_id` be other types other than `str` and `None`?

samtranslator/model/eventsources/push.py
prenx4x: Thats a good point and I totally agree. however it feels like a different task as this refactoring may be needed at several places, not just this function.

samtranslator/model/eventsources/push.py
prenx4x: No. I should only be string.

samtranslator/model/eventsources/push.py
hoffa: Yep, agreed! Meant it as a general call-out; PR itself looks good (forgot to approve!).

hoffa (approved):

samtranslator/model/eventsources/push.py
hoffa: What about Python 2 where a string can be `unicode`?

aahung: > No. It should only be string.

aahung: > No. It should only be string.

Can I assume if `not isinstance(bucket_id, str)` is `True`, it means the user forgets to put a `Ref`, and as long as `Ref` exists, the `not isinstance(bucket_id, str)` is `False`?

prenx4x: no. The user can have other types like list or dict as well as value for Ref. e.g.
Ref:
  -abc
or
Ref:
  abc:123

prenx4x: we mainly care for hashable values here.

samtranslator/model/eventsources/push.py
```diff
@@ -218,6 +218,8 @@ class S3(PushEventSource):
     def resources_to_link(self, resources):
         if isinstance(self.Bucket, dict) and "Ref" in self.Bucket:
             bucket_id = self.Bucket["Ref"]
+            if not isinstance(bucket_id, str):
```
hoffa, line 221: As discussed offline, we should use `six` for string types to deal with Python 2's `unicode` et al.

samtranslator/model/eventsources/push.py
aahung: Oh got it, thanks for the explanation!

aahung: resolving this. python 2 related unicode discussion is here https://github.com/aws/serverless-application-model/pull/1757#discussion_r513794030

aahung (approved):
Except the python 2 unicode concern by @hoffa, LGTM on my side

hoffa (approved):
Bueno 👌

jfuss (approved):
LGTM

PR #1765, opened 2020-11-02
Fix: SAM Crashes Invalid swagger models exception

prenx4x:
*Issue #, if available:*

*Description of changes:*

*Checklist:*

- [x] Write/update tests
- [x] `make pr` passes
- [ ] Update documentation

*Examples?*

### DIFF
--- a/samtranslator/swagger/swagger.py
+++ b/samtranslator/swagger/swagger.py
@@ -852,6 +852,8 @@ class SwaggerEditor(object):
         """
         if resource_policy is None:
             return
+        if not isinstance(resource_policy, dict):
+            raise InvalidDocumentException([InvalidTemplateException("Resource Policy is not a valid dictionary.")])
 
         aws_account_whitelist = resource_policy.get("AwsAccountWhitelist")
         aws_account_blacklist = resource_policy.get("AwsAccountBlacklist")
--- a/samtranslator/swagger/swagger.py
+++ b/samtranslator/swagger/swagger.py
@@ -836,10 +836,10 @@ class SwaggerEditor(object):
             model_properties = schema.get("properties")
 
             if not model_type:
-                raise ValueError("Invalid input. Value for type is required")
+                raise InvalidDocumentException([InvalidTemplateException("'Models' schema is missing 'type'.")])
 
             if not model_properties:
-                raise ValueError("Invalid input. Value for properties is required")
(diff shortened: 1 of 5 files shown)


samtranslator/swagger/swagger.py
```diff
@@ -852,6 +852,8 @@ def add_resource_policy(self, resource_policy, path, api_id, stage):
         """
         if resource_policy is None:
             return
+        if not isinstance(resource_policy, dict):
+            raise InvalidDocumentException([InvalidTemplateException("Resource Policy is not a valid dictionary.")])
```
aahung, line 856: LGTM, this is the fix and others are just changing the exception types right?

aahung (approved):

samtranslator/swagger/swagger.py
c2tarun, line 856: Curious, what is benefit of wrapping an array of exceptions in  another exception?

aahung: This is how it is meant to be used
```
    Attributes:
        message -- explanation of the error
        causes -- list of errors which caused this document to be invalid
    """
```

wchengru: +1 on this question. Also, not sure if resource_policy here is allowed to be an instrinsic expression, like `!Ref: sth`. If so, maybe we need to consider this.

c2tarun (approved):

samtranslator/swagger/swagger.py
prenx4x: No idea why. This is what InvalidDocumentException needs. I saw other examples and this is what they did. I tried other ways but none works.

prenx4x: that and adding tests so to not regress in future.

prenx4x: afaict, InvalidDocumentException is meta exception, which counts all sub exceptions and then produces a single exception message combining them all.

prenx4x: @wchengru I am not sure sure what you mean. Can you explain further?

prenx4x: As per @wchengru 's suggestion tried a template with resourcepolicy: !ref but did not work. We probably do not support it here.

PR #1767, opened 2020-11-03
fix: use newer policy name in gov & cn regions for xray

mndeveci:
*Description of changes:*
Previously used `AWSXrayWriteOnlyAccess` policy name is deprecated and not available in gov & cn regions. With this change we are going to assign newer policy name (`AWSXRayDaemonWriteAccess `) for those regions.

Updated unit tests and validated different region has different policy names.

*Checklist:*

- [x] Write/update tests
- [x] `make pr` passes
- [ ] Update documentation

*Examples?*


jfuss (approved):
LGTM! Thanks for the patch 🎉

c2tarun (approved):

PR #1794, opened 2020-11-12
Fix: Regex issue and Keyerror crash

prenx4x:
*Issue #, if available:*

*Description of changes:*

Fix: Regex issue when basepath is /
Fix: keyerror in open_api when method is ANY

*Checklist:*

- [x] Write/update tests
- [x] `make pr` passes
- [ ] Update documentation

*Examples?*

### DIFF
--- a/samtranslator/model/api/http_api_generator.py
+++ b/samtranslator/model/api/http_api_generator.py
@@ -308,11 +308,16 @@ class HttpApiGenerator(object):
                 invalid_regex = r"[^0-9a-zA-Z\/\-\_]+"
                 if re.search(invalid_regex, path) is not None:
                     raise InvalidResourceException(self.logical_id, "Invalid Basepath name provided.")
-                # ignore leading and trailing `/` in the path name
-                m = re.search(r"[a-zA-Z0-9]+[\-\_]?[a-zA-Z0-9]+", path)
-                path = m.string[m.start(0) : m.end(0)]
-                if path is None:
-                    raise InvalidResourceException(self.logical_id, "Invalid Basepath name provided.")
+
+                if path == "/":
+                    path = ""
+                else:
+                    # ignore leading and trailing `/` in the path name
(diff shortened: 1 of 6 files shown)


samtranslator/model/api/http_api_generator.py
```diff
@@ -308,11 +308,16 @@ def _construct_basepath_mappings(self, basepaths, http_api):
...
-                path = m.string[m.start(0) : m.end(0)]
-                if path is None:
-                    raise InvalidResourceException(self.logical_id, "Invalid Basepath name provided.")
+
+                if path == "/":
+                    path = ""
+                else:
+                    # ignore leading and trailing `/` in the path name
+                    m = re.search(r"[a-zA-Z0-9]+[\-\_]?[a-zA-Z0-9]+", path)
+                    path = m.string[m.start(0) : m.end(0)]
```
aahung, line 317: nit: since we are just ignoring the leading and trailing `/`, I think we can simplify the code as
```py
path = path.strip('/')

...

if not path:
    raise ...
```

aahung (approved):
LGTM!

CoshUS (approved):

samtranslator/model/api/http_api_generator.py
prenx4x: Thanks, it does look better, however even with this i won't be able to get rid of the if else as i would still have to check if path is "" before I do re.search. so I don't think it will be that useful.

hoffa, line 317: It's not clear from reading the code what `path` is, and why it's reassigned to in such a way (e.g. `foo-bar-baz` becomes `foo-bar`... _why_?).

If it's generating a logical ID from an arbitrary string, we should refactor in such a way; it seems to be a fairly generic use-case.

prenx4x: That's a good point. I am not sure about the reason behind creating logical_id like this. But we can try to understand where we use the logical id further and modify the logic accordingly.

I will merge this PR for now and can track this later.

PR #1797, opened 2020-11-14
test: Add simple integration tests

mingkun2020:
*Issue #, if available:*

*Description of changes:*

*Checklist:*

- [ ] Write/update tests
- [ ] `make pr` passes
- [ ] Update documentation

*Examples?*


tests_integ/helpers/base_test.py
```diff
@@ -0,0 +1,108 @@
...
+
+    @classmethod
+    def tearDownClass(cls) -> None:
+        BaseTest._clean_bucket()
+
+    @classmethod
+    def _clean_bucket(cls):
+        s3_client = boto3.client("s3")
+        response = s3_client.list_objects(Bucket=BaseTest.s3_bucket_name)
+        for content in response["Contents"]:
```
awood45, line 48: This is possibly okay because of the fixed number of resources we add, but if this request is paginated you will not properly delete all bucket objects. However, the [S3 Bucket Resource](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#S3.Bucket.objects) has methods that can iterate over all objects and handle pagination automatically, which is a better choice for this method.

tests_integ/helpers/helpers.py
```diff
@@ -0,0 +1,89 @@
...
+    return sorted(resources, key=lambda d: d["LogicalResourceId"])
+
+
+def create_bucket(bucket_name, region=None):
+    """Create an S3 bucket in a specified region
+
+    copy code from boto3 doc example
+    MG: removed the try so that the exception bubbles up and interrupts the test
+
+    If a region is not specified, the bucket is created in the S3 default
```
awood45, line 74: It feels for our test purposes that we should require a region to be specified, most likely. Have we tested what happens for tests overall if you don't specify a region here?

tests_integ/single/test_basic_api.py
```diff
@@ -0,0 +1,18 @@
+import os
+
```
awood45, line 2: Generally - why so many different classes? These all appear to do basically the same thing, why not one class with a long parameterized list unless there are material differences?

tests_integ/single/test_basic_api.py
```diff
@@ -0,0 +1,18 @@
...
+from parameterized import parameterized
+from tests_integ.helpers.base_test import BaseTest
+
+
+class TestBasicApi(BaseTest):
+    @parameterized.expand(
+        [
+            "basic_api_inline_openapi",
+            "basic_api_inline_swagger",
+            # "basic_api_with_cache", # no test case in java code base
```
awood45, line 12: If we don't have a model for this and aren't making one, let's delete this.

awood45 (changes requested):
Looks like a great start! Some comments attached

tests_integ/helpers/helpers.py
```diff
@@ -0,0 +1,89 @@
...
+    :return: True if bucket created, else False
+    """
+
+    # Create bucket
+    if region is None:
+        s3_client = boto3.client("s3")
+        s3_client.create_bucket(Bucket=bucket_name)
+    else:
+        s3_client = boto3.client("s3", region_name=region)
+        location = {"LocationConstraint": region}
```
hoffa, line 88: Is this necessary even though the region is already specified?

tests_integ/helpers/helpers.py
```diff
@@ -0,0 +1,89 @@
...
+    If a region is not specified, the bucket is created in the S3 default
+    region (us-east-1).
+
+    :param bucket_name: Bucket to create
+    :param region: String region to create bucket in, e.g., 'us-west-2'
+    :return: True if bucket created, else False
+    """
+
+    # Create bucket
+    if region is None:
```
hoffa, line 83: Would using `region_name=region` in all cases (even when `None`) work, so we can avoid the extra branch?

tests_integ/helpers/base_test.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def setUp(self):
+        self.cloudformation_client = boto3.client("cloudformation")
+        self.deployer = Deployer(self.cloudformation_client, changeset_prefix="sam-integ-")
+
+    def create_and_verify_stack(self, file_name):
+        input_file_path = str(Path(BaseTest.template_dir, file_name + ".yaml"))
+        self.output_file_path = str(Path(BaseTest.output_dir, "cfn_" + file_name + ".yaml"))
+        expected_resource_path = str(Path(BaseTest.expected_dir, file_name + ".json"))
+        self.stack_name = STACK_NAME_PREFIX + file_name.replace("_", "-") + generate_suffix()
```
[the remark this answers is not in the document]
mgrandis: You mean replace the hyphens with underscores?
Stack names can only contain [alphanumeric and hyphen characters](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-using-console-create-stack-parameters.html).
We could maybe switch from underscores to hyphens in the input file names

tests_integ/helpers/base_test.py
```diff
@@ -0,0 +1,108 @@
...
+from samcli.lib.deploy.deployer import Deployer
+from tests_integ.helpers.helpers import transform_template, verify_stack_resources, generate_suffix, create_bucket
+
+STACK_NAME_PREFIX = "sam-integ-stack-"
+S3_BUCKET_PREFIX = "sam-integ-bucket-"
+CODE_KEY_TO_FILE_MAP = {"codeuri": "code.zip", "contenturi": "layer1.zip", "definitionuri": "swagger1.json"}
+
+
+class BaseTest(TestCase):
+    @classmethod
```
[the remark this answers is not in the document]
mgrandis: Indeed we don't use the "cls" parameter so let's remove it

tests_integ/helpers/base_test.py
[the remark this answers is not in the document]
hoffa: Interesting, wasn't aware! In that case I don't mind either, and it's probably fine as-is (especially if the convention is to use underscores in the filenames).

tests_integ/helpers/base_test.py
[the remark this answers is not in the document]
mgrandis: I find it weird however, for readability, to have a "setUpClass" method with a "self" parameter instead of a "cls".
What do you think?

tests_integ/helpers/helpers.py
mgrandis: If you are using the us-east-1 region you should **not** specify the region or you get an error.
For all the other regions, you should explicitly specify it with the LocationConstraint parameter.

tests_integ/helpers/helpers.py
mgrandis: See my reply above

tests_integ/helpers/base_test.py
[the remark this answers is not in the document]
hoffa: Indeed that's how it's even used in the docs: https://docs.python.org/3/library/unittest.html#unittest.TestCase.setUpClass

So it's fine to keep it here. In `_clean_bucket` however it seems to be unnecessary.

tests_integ/helpers/helpers.py
hoffa: Just checked boto3 docks; [it's true](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/s3-example-creating-buckets.html#create-an-amazon-s3-bucket). D:

That's pretty terrible design, oh well.

tests_integ/single/test_basic_api.py
mingkun2020: For sure.

tests_integ/single/test_basic_api.py
mingkun2020: For this initial version, we just test all the templates with `create_and_verify_stack` function. We will add specific steps for tests later and that's why we group them into different classes.

tests_integ/helpers/helpers.py
mingkun2020: We use the credentials in local aws configuration for testing. If we don't specified any region in local, we will get connection error.

tests_integ/helpers/base_test.py
mgrandis: The list_objects will return 1000 objects per page, I'm not sure we'll ever reach this amount?
Also, we should maybe use list_objects_v2 instead.

mingkun2020: I test the bucket.objects.all() method in [S3 Bucket Resource](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#S3.Bucket.objects), this one should work without the constraint of paginated.

mgrandis: It seems to fetch all the objects at once and then paginate on the client which, performance wise, isn't ideal.

mgrandis: Alex confirmed objects.all() streams pages of 1000 elements and doesn't load everything at once.
Good for me!

tests_integ/helpers/base_test.py
```diff
@@ -0,0 +1,108 @@
...
+        BaseTest.output_dir = BaseTest.tests_integ_dir
+        BaseTest.expected_dir = Path(BaseTest.resources_dir, "expected", "single")
+        code_dir = Path(BaseTest.resources_dir, "code")
+
+        BaseTest.s3_bucket_name = S3_BUCKET_PREFIX + generate_suffix()
+        session = boto3.session.Session()
+        my_region = session.region_name
+        create_bucket(BaseTest.s3_bucket_name, region=my_region)
+
+        s3_client = boto3.client("s3")
```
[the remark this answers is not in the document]
mgrandis: Yes indeed

tests_integ/helpers/base_test.py
```diff
@@ -0,0 +1,345 @@
...
+        ----------
+        file_name : string
+            Template file name
+        """
+        input_file_path = str(Path(self.template_dir, file_name + ".yaml"))
+        updated_template_path = str(Path(self.output_dir, "sub_" + file_name + ".yaml"))
+        with open(input_file_path, "r") as f:
+            data = f.read()
+        for key, _ in CODE_KEY_TO_FILE_MAP.items():
+            data = data.replace("${{{}}}".format(key), self.get_code_key_s3_uri(key))
```
hoffa, line 281: > `${{{}}}` 

What's going on here?

tests_integ/helpers/base_test.py
```diff
@@ -0,0 +1,345 @@
...
+        resource_name: string
+            resource name
+        property_name: string
+            property name
+        value
+            value
+        """
+        with open(self.sub_input_file_path, "r") as f:
+            data = f.read()
+        yaml_doc = yaml.load(data, Loader=yaml.FullLoader)
```
hoffa, line 304: Can we refactor this away? Reading a YAML is a common use case and we're repeating it.

tests_integ/helpers/base_test.py
```diff
@@ -0,0 +1,345 @@
...
+
+        return yaml_doc["Resources"][resource_name]["Properties"][property_name]
+
+    def deploy_stack(self, parameters=[]):
+        """
+        Deploys the current cloud formation stack
+        """
+        with open(self.output_file_path, "r") as cfn_file:
```
hoffa, line 321: nit: the default mode is already `r`

tests_integ/helpers/helpers.py
```diff
@@ -0,0 +1,91 @@
...
+        expected_resources = _sort_resources(json.load(expected_data))
+    parsed_resources = _sort_resources(stack_resources["StackResourceSummaries"])
+
+    if len(expected_resources) != len(parsed_resources):
+        return False
+
+    for i in range(len(expected_resources)):
+        exp = expected_resources[i]
+        parsed = parsed_resources[i]
+        if not re.fullmatch(exp["LogicalResourceId"] + "([0-9a-f]{10})?", parsed["LogicalResourceId"]):
```
hoffa, line 52: Why `{10}`?

tests_integ/helpers/base_test.py
mgrandis: The end result we need is "${mykeyvalue}", I'll add a comment

appveyor.yml
```diff
@@ -2,16 +2,27 @@ version: 1.0.{build}
...
+    PYTHON_VERSION: '3.6'
   - TOXENV: py37
+    PYTHON_VERSION: '3.7'
   - TOXENV: py38
+    PYTHON_VERSION: '3.8'
 
 build: off
 
 install:
+- sh: "source ${HOME}/venv${PYTHON_VERSION}/bin/activate"
```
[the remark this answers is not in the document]
hoffa: Also it seems `sh:` is the default, so we can get rid of it.

tests_integ/helpers/base_test.py
```diff
@@ -0,0 +1,345 @@
...
+        else:
+            output_key_to_value = {}
+            output_list = self.stack_description["Stacks"][0]["Outputs"]
+            for output in output_list:
+                output_key_to_value[output["OutputKey"]] = output["OutputValue"]
+            return output_key_to_value
+
+    def get_resource_status_by_logical_id(self, logical_id):
+        if not self.stack_resources:
+            return ""
```
hoffa, line 229: Any reason for returning an empty string instead of `None`?

tests_integ/helpers/base_test.py
```diff
@@ -0,0 +1,345 @@
...
+        if not resources:
+            return []
+
+        return self.api_client.get_stages(restApiId=resources[0]["PhysicalResourceId"])["item"]
+
+    def get_stack_stages_v2(self):
+        resources = self.get_stack_resources("AWS::ApiGatewayV2::Api")
+
+        if not resources:
+            return None
```
hoffa, line 205: Why do we return `None` here but `[]` for v1 above?

tests_integ/helpers/base_test.py
mgrandis: Added functions to load/write yaml

tests_integ/helpers/base_test.py
mgrandis: Good catch, changing to []

tests_integ/helpers/base_test.py
mgrandis: Changing to None

tests_integ/helpers/helpers.py
mgrandis: That's the length of the random string being added to logical IDs for some resources (not all) once deployed 
I moved it to a constant

tests_integ/helpers/base_test.py
```diff
@@ -0,0 +1,370 @@
...
+        cls.resources_dir = Path(cls.tests_integ_dir, "resources")
+        cls.template_dir = Path(cls.resources_dir, "templates", "single")
+        cls.output_dir = Path(cls.tests_integ_dir, "tmp")
+        cls.expected_dir = Path(cls.resources_dir, "expected", "single")
+        cls.code_dir = Path(cls.resources_dir, "code")
+        cls.s3_bucket_name = S3_BUCKET_PREFIX + generate_suffix()
+        cls.session = boto3.session.Session()
+        cls.my_region = cls.session.region_name
+        cls.s3_client = boto3.client("s3")
+        cls.api_client = boto3.client("apigateway", cls.my_region)
```
awood45, line 36: why does only this client specify a region?

tests_integ/helpers/base_test.py
```diff
@@ -0,0 +1,370 @@
...
+                LOG.error(
+                    "Unable to delete object %s from bucket %s", object_summary.key, cls.s3_bucket_name, exc_info=e
+                )
+        try:
+            cls.s3_client.delete_bucket(Bucket=cls.s3_bucket_name)
+        except ClientError as e:
+            LOG.error("Unable to delete bucket %s", cls.s3_bucket_name, exc_info=e)
+
+    @classmethod
+    def _upload_resources(cls):
```
awood45, line 73: Should we allow resources to be passed in? Perhaps something like:

```python
def _upload_resource(cls, file_to_s3_map=FILE_TO_S3_URI_MAP):
```

awood45: This may also useful if we want to move this to a config file and out of source.

tests_integ/helpers/base_test.py
```diff
@@ -0,0 +1,370 @@
...
+            return "s3://{}/{}".format(cls.s3_bucket_name, file_name)
+
+        if cls.my_region == "us-east-1":
+            return "https://s3.amazonaws.com/{}/{}".format(cls.s3_bucket_name, file_name)
+        if cls.my_region == "us-iso-east-1":
+            return "https://s3.us-iso-east-1.c2s.ic.gov/{}/{}".format(cls.s3_bucket_name, file_name)
+        if cls.my_region == "us-isob-east-1":
+            return "https://s3.us-isob-east-1.sc2s.sgov.gov/{}/{}".format(cls.s3_bucket_name, file_name)
+
+        return "https://s3-{}.amazonaws.com/{}/{}".format(cls.my_region, cls.s3_bucket_name, file_name)
```
awood45, line 106: Is this correct for China regions? We may want something to determine partition and feed that into the URL determination.

tests_integ/helpers/base_test.py
```diff
@@ -0,0 +1,370 @@
...
+    def get_s3_uri(self, file_name):
+        """
+        Returns the S3 URI of a resource file
+
+        Parameters
+        ----------
+        file_name : string
+            Resource file name
+        """
+        return FILE_TO_S3_URI_MAP[file_name]["uri"]
```
awood45, line 155: This feels like it should be an instance variable that we seed, for flexibility.

tests_integ/helpers/base_test.py
```diff
@@ -0,0 +1,370 @@
...
+    def get_code_key_s3_uri(self, code_key):
+        """
+        Returns the S3 URI of a code key for template replacement
+
+        Parameters
+        ----------
+        code_key : string
+            Template code key
+        """
+        return FILE_TO_S3_URI_MAP[CODE_KEY_TO_FILE_MAP[code_key]]["uri"]
```
awood45, line 166: This also feels like it should be an instance variable that we seed, for flexibility - both values, arguably.

tests_integ/helpers/base_test.py
```diff
@@ -0,0 +1,370 @@
...
+
+    def get_stack_deployment_ids(self):
+        resources = self.get_stack_resources("AWS::ApiGateway::Deployment")
+        ids = []
+        for res in resources:
+            ids.append(res["LogicalResourceId"])
+
+        return ids
+
+    def get_stack_stages(self):
```
awood45, line 197: I'd prefer a more descriptive name like `api_stack_stages` for this and the below to avoid confusion since this special cases a single resource type.

tests_integ/helpers/deployer/exceptions/exceptions.py
```diff
@@ -0,0 +1,93 @@
...
+    def __init__(self, stack_name, msg):
+        self.stack_name = stack_name
+        self.msg = msg
+
+        message_fmt = "Failed to create/update the stack: {stack_name}, {msg}"
+
+        super(DeployFailedError, self).__init__(message=message_fmt.format(stack_name=self.stack_name, msg=msg))
+
+
+class GuidedDeployFailedError(UserException):
```
awood45, line 46: Can we consider trimming errors like this that are SAM CLI specific down to only errors relevant to us?

tests_integ/helpers/deployer/utils/table_print.py
```diff
@@ -0,0 +1,137 @@
+"""
```
awood45, line 1: Do we need this? (Or the colors class?)

awood45 (changes requested):
This looks like a great start - I added some comments, but one big one would be that we may have copied over a lot more of `Deployer` than we need and could benefit from trimming out functionality that we aren't using. For example, we shouldn't use or need deployment detail printing for these tests.

tests_integ/helpers/deployer/exceptions/exceptions.py
mgrandis: Yes, removing the ones we don't use

tests_integ/helpers/base_test.py
mgrandis: Renaming those

tests_integ/helpers/deployer/utils/table_print.py
mgrandis: It can be useful to debug a test error and it isn't displayed if a test succeeds

tests_integ/helpers/base_test.py
mgrandis: I could have sworn I had an error without but just ran the tests again without and it worked
Removed!

hoffa:
I just realized this, but if we're using `pytest` to run the tests, shouldn't we use `pytest` throughout? e.g. `assert` instead of the `self.assert*` methods, etc.

tests_integ/helpers/base_test.py
mingkun2020: Update: allowed resources to be passed in as variable. We have file_resources.py as a config file to store all the info about resources, do we want to change it to a json/yaml file?

tests_integ/helpers/base_test.py
mingkun2020: Yes, I have updated it to use instance variable.

tests_integ/helpers/base_test.py
mingkun2020: Yes, I have updated it to use instance variables.

tests_integ/helpers/base_test.py
mgrandis: Let's discuss this and review the S3 doc, I migrated this from the Java code, it seems some of it might be outdated

appveyor.yml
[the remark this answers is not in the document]
mingkun2020: This file is directly copy from develop branch. Not sure how this work, it has been added by this commit :https://github.com/aws/serverless-application-model/commit/6b00fe40d403182cea44de37d65b11b58b743685.

Makefile
```diff
@@ -6,13 +6,17 @@ init:
 	pip install -e '.[dev]'
 
 test:
-	pytest --cov samtranslator --cov-report term-missing --cov-fail-under 95 tests
+	pytest --cov samtranslator --cov-report term-missing --cov-fail-under 95 tests/*
+
+# Need to specify *.py or pytest attempts to execute the README.md file
+test-integ:
+	pytest --no-cov tests_integ/**/*.py
```
hoffa, line 13: `**` is not supported in POSIX `sh`; which shell is used to run this Makefile?

hoffa: At least for [GNU Make](https://www.gnu.org/software/make/manual/html_node/Choosing-the-Shell.html):

> The program used as the shell is taken from the variable SHELL. If this variable is not set in your makefile, the program /bin/sh is used as the shell.

Which suggests this isn't correct.

Why do we need to specify in that way? Instead of `pytest --no-cov tests_integ`.

awood45: Would tend to concur, we should just be able to run on the folder.

awood45 (approved):

Makefile
mgrandis: Fixed

tests_integ/helpers/deployer/utils/time.py
```diff
@@ -0,0 +1,127 @@
...
+        Value to be converted to unix epoch. This must be without any timezone identifier
+
+    Returns
+    -------
+    int
+        Unix timestamp of the given time
+    """
+
+    # `total_seconds()` returns elaped microseconds as a float. Get just milliseconds and discard the rest.
+    return int((some_time - datetime.datetime(1970, 1, 1)).total_seconds() * 1000.0)
```
hoffa, line 65: Why not [`datetime.timestamp`](https://docs.python.org/3/library/datetime.html#datetime.datetime.timestamp)?

tests_integ/helpers/deployer/utils/table_print.py
```diff
@@ -0,0 +1,137 @@
...
+    from itertools import izip_longest as zip_longest
+import textwrap
+from functools import wraps
+
+import click
+
+MIN_OFFSET = 20
+
+
+def pprint_column_names(format_string, format_kwargs, margin=None, table_header=None, color="yellow"):
```
hoffa, line 18: What does this do?

I assume it's to improve legibility? Do we expect changes to this? It's worth considering the long-term maintenance impact.

tests_integ/helpers/helpers.py
```diff
@@ -0,0 +1,148 @@
...
+import boto3
+from botocore.exceptions import ClientError, NoRegionError
+
+from samtranslator.model.exceptions import InvalidDocumentException
+from samtranslator.translator.managed_policy_translator import ManagedPolicyLoader
+from samtranslator.translator.transform import transform
+from samtranslator.yaml_helper import yaml_parse
+
+# Length of the suffix sometimes added by CFN to logical IDs
+LOGICAL_RES_SUFFIX_LENGTH = 10
```
hoffa, line 17: Is this length documented?

tests_integ/helpers/deployer/utils/time.py
```diff
@@ -0,0 +1,127 @@
...
+Date & Time related utilities
+"""
+
+import datetime
+import dateparser
+
+from dateutil.tz import tzutc
+
+
+def timestamp_to_iso(timestamp):
```
hoffa, line 11: Should mention all timestamps here are milliseconds (why?); docstring suggests second-precision. Python timestamps are usually second-precision floats.

tests_integ/helpers/deployer/utils/time.py
mgrandis: It doesn't produce the right value:
```python
>>> d = datetime.datetime(2020,12,18)
>>> d
datetime.datetime(2020, 12, 18, 0, 0)
>>> d.timestamp() * 1000
1608278400000.0 # 2020-12-18 at 8 AM
# Current code:
>>> int((d - datetime.datetime(1970, 1, 1)).total_seconds() * 1000.0)
1608249600000 # 2020-12-18 at 12 AM
# From the docs https://docs.python.org/3/library/datetime.html#datetime.datetime.timestamp:
>>> ((d - datetime.datetime(1970, 1, 1)) / datetime.timedelta(seconds=1)) * 1000
1608249600000.0 # 2020-12-18 at 12 AM
```

tests_integ/helpers/helpers.py
mgrandis: After a little digging, this is actually us (SAM) that add this suffix to the Logical IDs when translating.

It's located in `logical_id_generator.py`, I changed the code to remove this constant and use `LogicalIdGenerator.HASH_LENGTH` instead.

tests_integ/helpers/deployer/utils/time.py
hoffa: Ah fair enough, and that's mentioned in the docs.

hoffa (approved):

appveyor.yml
```diff
@@ -20,4 +20,4 @@ install:
 - make init
 
 test_script:
-- tox
+- tox
```
jfuss, line 23: Why are we running tox here? We have the matrix above which is doing what tox would do anyways.

mgrandis: No idea, this was here before.
The only change here is removing the last newline which I'm not sure is a good idea or necessary, I'm going to revert it.

Makefile
```diff
@@ -6,13 +6,16 @@ init:
 	pip install -e '.[dev]'
 
 test:
-	pytest --cov samtranslator --cov-report term-missing --cov-fail-under 95 tests
+	pytest --cov samtranslator --cov-report term-missing --cov-fail-under 95 tests/*
+
+test-integ:
```
jfuss, line 11: Nit: can we match the naming with SAM CLI? https://github.com/aws/aws-sam-cli/blob/develop/Makefile#L17 That way we are not trying to remember which project does which thing?

Makefile
```diff
@@ -6,13 +6,16 @@ init:
...
+	pytest --no-cov tests_integ/*
```
jfuss, line 12: Same nit here: I would prefer if we standardize our naming and structure across all our repos we own.

requirements/dev.txt
```diff
@@ -11,6 +11,11 @@ pytest~=4.6.11; python_version < '3.6' # pytest dropped python 2 support after 4
 mock>=3.0.5,<4.0.0  # 4.0.0 drops Python 2 support
 parameterized~=0.7.4
 
+# Integration tests
+pathlib2>=2.3.5; python_version < '3'
+click~=7.1
```
jfuss, line 16: Why do we need click for our integration tests?

tests_integ/helpers/base_test.py
```diff
@@ -0,0 +1,379 @@
...
+        cls.tests_integ_dir = Path(__file__).resolve().parents[1]
+        cls.resources_dir = Path(cls.tests_integ_dir, "resources")
+        cls.template_dir = Path(cls.resources_dir, "templates", "single")
+        cls.output_dir = Path(cls.tests_integ_dir, "tmp")
+        cls.expected_dir = Path(cls.resources_dir, "expected", "single")
+        cls.code_dir = Path(cls.resources_dir, "code")
+        cls.s3_bucket_name = S3_BUCKET_PREFIX + generate_suffix()
+        cls.session = boto3.session.Session()
+        cls.my_region = cls.session.region_name
+        cls.s3_client = boto3.client("s3")
```
jfuss, line 36: Should we be creating clients based on the session instead? I think this is typically the better practice, though it probably doesn't impact anything here. Just caught my eye is all.

tests_integ/helpers/base_test.py
```diff
@@ -0,0 +1,379 @@
...
+        cls.api_client = boto3.client("apigateway")
+        cls.lambda_client = boto3.client("lambda")
+        cls.iam_client = boto3.client("iam")
+        cls.api_v2_client = boto3.client("apigatewayv2")
+        cls.sfn_client = boto3.client("stepfunctions")
```
jfuss, line 41: Do we really need each of these clients for every integ test? Shouldn't we only create the client the individual tests need?

tests_integ/helpers/base_test.py
```diff
@@ -0,0 +1,379 @@
...
+    @classmethod
+    def tearDownClass(cls):
+        cls._clean_bucket()
+
+    @classmethod
+    def _clean_bucket(cls):
+        """
+        Empties and deletes the bucket used for the tests
+        """
+        s3 = boto3.resource("s3")
```
jfuss, line 60: Why use resource here but client everywhere else? Seems like resource is a much better way to handle things in boto3 in general?

tests_integ/helpers/base_test.py
```diff
@@ -0,0 +1,379 @@
...
+        bucket = s3.Bucket(cls.s3_bucket_name)
+        object_summary_iterator = bucket.objects.all()
+
+        for object_summary in object_summary_iterator:
+            try:
+                cls.s3_client.delete_object(Key=object_summary.key, Bucket=cls.s3_bucket_name)
+            except ClientError as e:
+                LOG.error(
```
jfuss, line 68: Should this cause a test to fail? If there is data left around, will that cause issue for the next run of the tests?

tests_integ/helpers/base_test.py
```diff
@@ -0,0 +1,379 @@
...
+    def _get_s3_uri(cls, file_name, uri_type):
+        if uri_type == "s3":
+            return "s3://{}/{}".format(cls.s3_bucket_name, file_name)
+
+        if cls.my_region == "us-east-1":
+            return "https://s3.amazonaws.com/{}/{}".format(cls.s3_bucket_name, file_name)
+        if cls.my_region == "us-iso-east-1":
+            return "https://s3.us-iso-east-1.c2s.ic.gov/{}/{}".format(cls.s3_bucket_name, file_name)
+        if cls.my_region == "us-isob-east-1":
+            return "https://s3.us-isob-east-1.sc2s.sgov.gov/{}/{}".format(cls.s3_bucket_name, file_name)
```
jfuss, line 112: Is there no way to get this from S3 itself? Seems like a maintenance burden otherwise.

tests_integ/helpers/helpers.py
```diff
@@ -0,0 +1,147 @@
+import json
```
jfuss, line 1: Do we have a better name then "helpers"? We should try to be explicit (single principle) and in my experience, these types of classes/files just become dumping grounds

jfuss (changes requested):
This generally looks good. I have a few small comments about keeping consistent between our repos. The other main things for me is there is a lot of borrowed code from SAM CLI to help deploy these stacks. Which I think is ok but do we need all of it? More code here, means more to maintain. I wonder if it is worth stripping this down to reduce the amount of code needed for our integ tests.

Makefile
mgrandis: Just so I understand correctly, does this mean moving the whole content of the current `tests` dir under `tests\unit` and moving the integration tests under `tests\integration`?

tests_integ/helpers/base_test.py
mgrandis: I'm not sure cleanup of a test is part of the test itself, so not sure it should fail.
What we could do is attempt to clean it again and if it fails set this resource name aside and attempt to clean it again at the end of the run.

This shouldn't cause an issue for the next run as we are adding a simple random suffix to each of the objects, stacks and buckets, that we create (see `generate_suffix()`). I also added this suffix so that we could run the tests in parallel but for now it's not an option as we get throttled.

tests_integ/helpers/base_test.py
mgrandis: Yep this could be better indeed, to be sure we are using the same session throughout

tests_integ/helpers/base_test.py
mgrandis: I was thinking of this as a way to provide a simple toolbox so that every test could access any service easily, we'll check if creating all the client for each test takes too many resources (memory or cpu time)

Note that the session is not thread safe so instantiating all the clients once for the whole test suite may not be a good idea.

tests_integ/helpers/base_test.py
mgrandis: It seems `resource` is more high level that `client`, we'll check to see if we can use resource everywhere.
We needed the `resource` here to the files list in a paginated way.

tests_integ/helpers/helpers.py
mgrandis: Agree, also `helpers/helpers.py` is redundant.
We could maybe move the `transform_template` function to `helpers\template.py` and the rest to `helpers\resource.py`.

tests_integ/helpers/base_test.py
mgrandis: It seems you can only get a [presigned temporary URL](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#S3.Client.generate_presigned_url) from the client.

tests_integ/helpers/base_test.py
mingkun2020: By the boto3 documentation: "Boto3 acts as a proxy to the default session. This is created automatically when you create a low-level client." So all the clients are created by using the default session and during the tests we don't use any customized session.

Makefile
mgrandis: Just moving the `tests_integration` content in `tests/integration` will make the unit tests also run the integration tests so we may have to move everything.

tests_integ/helpers/helpers.py
mingkun2020: Updated

Makefile
mingkun2020: Updated

tests_integ/helpers/helpers.py
mingkun2020: Updated

tests_integ/helpers/base_test.py
mingkun2020: Updated with a Client_provider class with lazy initialization.

requirements/dev.txt
mingkun2020: Click is used by the deployer class copied from samcli.

tests_integ/helpers/base_test.py
mingkun2020: ```client``` is low level and has typically maps 1:1 with the AWS service API while ```resource``` is high level but does not provide 100% API coverage of AWS services. Most of the case we use ```client```, here we use ```resource``` since it provides a easier way to lists all the objects compared to the ```client```.

jfuss (approved):
LGTM

PR #1798, opened 2020-11-14
fix: propagate condition to sqs queue policy for sqssubscription

hawflau:
*Issue #, if available:*

*Description of changes:*

This fixes a bug of not having `Condition` propagated to the generated `AWS::SQS::QueuePolicy` resource if a SNS event source is used with `SqsSubscription` in a `AWS::Serverless::Function` resource.

Also updated the black commands in `Makefile` to check only `.py` files

*Checklist:*

- [x] Write/update tests
- [x] `make pr` passes
- [ ] Update documentation

*Examples?*

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,10 +9,10 @@ test:
 	pytest --cov samtranslator --cov-report term-missing --cov-fail-under 95 tests
 
 black:
-	black setup.py samtranslator/**/*.py tests/**/*.py bin/*.py
+	black setup.py samtranslator/* tests/* bin/*.py
 
 black-check:
-	black --check setup.py samtranslator/**/*.py tests/**/*.py bin/*.py
+	black --check setup.py samtranslator/* tests/* bin/*.py
 
 # Command to run everytime you make changes to verify everything works
 dev: test
--- a/Makefile
+++ b/Makefile
@@ -9,10 +9,10 @@ test:
 	pytest --cov samtranslator --cov-report term-missing --cov-fail-under 95 tests
 
 black:
-	black setup.py samtranslator/* tests/* bin/*.py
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -9,10 +9,10 @@ test:
 	pytest --cov samtranslator --cov-report term-missing --cov-fail-under 95 tests
 
 black:
-	black setup.py samtranslator/* tests/* bin/*.py
+	black setup.py samtranslator/**/*.py tests/**/*.py bin/*.py
```
hoffa, line 12: Why is this change necessary? This also assumes the shell supports `**` (which isn't in POSIX `sh`), which e.g. in Bash is only enabled if `globstar` is set. AFAIK `make` won't necessarily default to Bash.

Tangentially, can we instead simplify this to `black setup.py samtranslator tests bin` and change include/exclude behavior in the config?

hawflau: I intended to skip checking the .pyc files. But looking at the doc, yeah you are right, it's not necessary.
Will revert this commit

hoffa (approved):

wchengru (approved):

tests/translator/input/function_event_conditions.yaml
```diff
@@ -79,6 +79,22 @@ Resources:
...
+        SNSTopicWithSQSSubscription:
+          Type: SNS
+          Properties:
+            Topic:
+              Ref: Notifications
+            SqsSubscription:
+              QueueArn: !GetAtt Queue.Arn
+              QueueUrl: !Ref Queue
+
+        AnotherSNSWithSQSSubscription:
```
elbayaaa, line 91: I'd prefer a more descriptive name to immediately figure out what is the difference between this one and the one above, what each of them tests.

elbayaaa (approved):

PR #1802, opened 2020-11-17
Fix: SAM crashes method_definition for path is invalid

prenx4x:
*Issue #, if available:*

*Description of changes:*
Fix: SAM crashes method_definition for path is invalid
Fix: SAM crashes when mappings is null
Fix: Swagger security not a dict

*Checklist:*

- [x] Write/update tests
- [x] `make pr` passes
- [ ] Update documentation

*Examples?*

### DIFF
--- a/samtranslator/intrinsics/resolver.py
+++ b/samtranslator/intrinsics/resolver.py
@@ -1,6 +1,7 @@
 # Help resolve intrinsic functions
 
 from samtranslator.intrinsics.actions import Action, SubAction, RefAction, GetAttAction
+from samtranslator.model.exceptions import InvalidTemplateException, InvalidDocumentException
 
 # All intrinsics are supported by default
 DEFAULT_SUPPORTED_INTRINSICS = {action.intrinsic_name: action() for action in [RefAction, SubAction, GetAttAction]}
@@ -17,7 +18,9 @@ class IntrinsicsResolver(object):
         """
 
         if parameters is None or not isinstance(parameters, dict):
-            raise TypeError("parameters must be a valid dictionary")
+            raise InvalidDocumentException(
+                [InvalidTemplateException("'Mappings' or 'Parameters' is either null or not a valid dictionary.")]
+            )
 
         if not isinstance(supported_intrinsics, dict) or not all(
             [isinstance(value, Action) for value in supported_intrinsics.values()]
--- a/samtranslator/swagger/swagger.py
+++ b/samtranslator/swagger/swagger.py
@@ -534,6 +534,14 @@ class SwaggerEditor(object):
                 for method_definition in self.get_method_contents(self.get_path(path)[normalized_method_name]):
 
                     # If no integration given, then we don't need to process this definition (could be AWS::NoValue)
+                    if not isinstance(method_definition, dict):
+                        raise InvalidDocumentException(
+                            [
+                                InvalidTemplateException(
+                                    "{} for path {} is not a valid dictionary.".format(method_definition, path)
+                                )
+                            ]
+                        )
                     if not self.method_definition_has_integration(method_definition):
                         continue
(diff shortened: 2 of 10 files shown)


samtranslator/intrinsics/resolver.py
```diff
@@ -17,7 +18,9 @@ def __init__(self, parameters, supported_intrinsics=DEFAULT_SUPPORTED_INTRINSICS
         """
 
         if parameters is None or not isinstance(parameters, dict):
-            raise TypeError("parameters must be a valid dictionary")
```
mndeveci, line 20: Is `TypeError` causes 500 errors?

prenx4x: sorry, what?

mndeveci: If we raise `TypeError` instead of `InvalidDocumentException`, will it cause 500 server error on lambda execution? Is that the reason we changed the type of the exception here?

prenx4x: ohh i see. yes TypeError causes 500 server error.

mndeveci (approved):

samtranslator/swagger/swagger.py
```diff
@@ -534,6 +534,14 @@ def set_path_default_authorizer(
                 for method_definition in self.get_method_contents(self.get_path(path)[normalized_method_name]):
 
                     # If no integration given, then we don't need to process this definition (could be AWS::NoValue)
+                    if not isinstance(method_definition, dict):
```
hoffa, line 537: Can we start refactoring these branches? The patterns are essentially identical, but takes 8 LoC for a single check.

aahung: irrelevant: we should have mypy to end this manual type check nightmare!

aahung (approved):

samtranslator/swagger/swagger.py
prenx4x: @hoffa I dont understand. What (or how) do you mean refactor?
@aahung These errors are coming from user template. How will mypy know what is valid and what is not?

aahung: if we define the possible types of `method_definition`, and mypy will scream if you don't have this check.

prenx4x: This is the challenge right now and itself could be a lot of work (of course worth doing) which is to assess and note every parameter's data type. If we can do that then it would solve all current and future SAM crashes as well.

hoffa: @prenx4x `grep InvalidDocumentException . -RC2` gives me a bunch of very similar checks: if value doesn't match type, throw the exception. It's noisy and repetitive, would be good to start refactoring these so we can simplify common patterns and reduce cognitive load, otherwise it won't be maintainable.

hoffa: And more generally, this sort of "lazy" checking doesn't seem to be ideal; maybe we can validate up-front what needs to be validated, and use [EAFP](https://docs.python.org/3.9/glossary.html#term-eafp) elsewhere. This would reduce complexity and improve maintainability.

prenx4x: I see. One thing to note here is that the errors could be similar, but the message (or cause) is very different. This message is seen by the user and it should be clear enough to point out the issue in their template. We we find same errors with same cause at multiple places we can refactor it, but if causes are different then they should be separate exceptions.

hoffa: Right, and that's fine; messages can be custom. But we should still be continuously improving our codebase.

This doesn't mean refactoring everything at once, but if we recognize common patterns that can benefit from refactoring in _new_ code, we should go ahead and do it, and ensure that in the long term our quality keeps increasing.

Issue #1739, opened 2020-10-12
Impossible to use intrinsic functions in S3 triggering event 


MaxTeiger:
**Description:**

The use of intrinsic functions in S3 triggering event does not work when deploying in AWS.
I tried to build and deploy my application from the following template:

```yaml
AWSTemplateFormatVersion: '2010-09-09'
Transform: AWS::Serverless-2016-10-31

Parameters:
  S3EventParam:
    Type: CommaDelimitedList
    Default: s3:ObjectCreated:*
    Description: The S3 events on which the function must be triggered 
  PrefixParam:
    Type: String
    Default: '*'
    Description: The prefix in the name of each file for which the function must trigger
  SuffixParam:
    Type: String
    Default: '*'
    Description: The suffix in the name of each file for which the function must trigger

Globals:
  Function:
    Timeout: 3
    Runtime: nodejs12.x
    CodeUri: functions
    MemorySize: 128

Resources:
  S3EventFunction:
    Type: 'AWS::Serverless::Function'
    Properties:
      Handler: app.handler
      Events:
        BucketEvent1:
          Type: S3
          Properties:
            Bucket:
              Ref: SrcBucket
            Events:
              !Ref S3EventParam
            Filter:
              S3Key:
                Rules:
                  - Name: prefix
                    Value: !Ref PrefixParam
                  - Name: suffix
                    Value: !Ref SuffixParam
  LambdaInvokePermission:
    Type: 'AWS::Lambda::Permission'
    Properties:
      FunctionName: !GetAtt S3EventFunction.Arn
      Action: 'lambda:InvokeFunction'
      Principal: 's3.amazonaws.com'
      SourceAccount: !Sub ${AWS::AccountId}
      SourceArn: !GetAtt SrcBucket.Arn
  SrcBucket:
    Type: AWS::S3::Bucket
    Properties:
      BucketEncryption: 
        ServerSideEncryptionConfiguration:
          - ServerSideEncryptionByDefault:
              SSEAlgorithm: aws:kms
```

**Steps to reproduce the issue:**
Use `sam build` and `sam deploy` with the previous template and a dummy hello world function in NodeJS

I am using SAM CLI version 1.2.0 for Windows 10

**Observed result:**

I get the error ` The event is not supported for notifications`  when SAM deploys the bucket. In AWS Console, the generated cloudformation template looks like this:

```json
        "NotificationConfiguration": {
          "LambdaConfigurations": [
            {
              "Function": {
[...]

PR #1824, opened 2020-11-21
Fix: Impossible to use intrinsic functions in S3 triggering event

fbn-roussel:
*Issue #, if available:*
#1739

*Description of changes:*
Updating `_inject_notification_configuration` function to handle `Ref` intrinsic function for S3 triggering event types.

Added a test case using `Ref` intrinsic function.
The fix I made was suggested by this comment : https://github.com/aws/serverless-application-model/issues/1739#issuecomment-716693880

*Checklist:*

- [x] Write/update tests
- [x] `make pr` passes
- [ ] Update documentation

*Examples?*


Jacco:
I have tried to deploy the template that was generated using the code in this PR. But I get an `CREATE_FAILED` error with the message:

`Value of property Event must be of type String`

jfuss:
Cleaning up old PRs that are no longer active in to help us get the PR backlog in better shape. Closing this PR as the "from" repo is no longer. 

Intrinsic support is hit or miss throughout our repo to this point. Mostly because it requires us to re-implement how CloudFormation resolving works. This does not appear to be that and instead looks like we would need to pass along the value correctly to CloudFormation. Happy to still accept a PR in this area. The team has been working on process improvements in looking at PRs faster, so the lingering reviews from maintainers shouldn't happen for new PRs.

PR #1858, opened 2020-12-11
fix: Support new CodeDeploy ManagedPolicy

jfuss:
CodeDeploy is migrating from AWSCodeDeployRoleForLambda to AWSCodeDeployRoleForLambdaLimited.
Some partitions do not support AWSCodeDeployRoleForLambda and therefore we need to use the newer
one in those partitions. We cannot widely update to AWSCodeDeployRoleForLambdaLimited since this
can cause customer's stacks to fail unexpectedly.

*Issue #, if available:*

*Description of changes:*

*Checklist:*

- [ ] Write/update tests
- [ ] `make pr` passes
- [ ] Update documentation

*Examples?*

### DIFF
--- a/samtranslator/model/preferences/deployment_preference_collection.py
+++ b/samtranslator/model/preferences/deployment_preference_collection.py
@@ -101,9 +101,17 @@ class DeploymentPreferenceCollection(object):
                 }
             ],
         }
-        iam_role.ManagedPolicyArns = [
-            ArnGenerator.generate_aws_managed_policy_arn("service-role/AWSCodeDeployRoleForLambda")
-        ]
+
+        # CodeDeploy has a new managed policy. We cannot update any existing partitions, without customer reach out
+        # that support AWSCodeDeployRoleForLambda since this could regress stacks that are currently deployed.
+        if ArnGenerator.get_partition_name() in ["aws-iso", "aws-iso-b"]:
+            iam_role.ManagedPolicyArns = [
+                ArnGenerator.generate_aws_managed_policy_arn("service-role/AWSCodeDeployRoleForLambdaLimited")
+            ]
+        else:
+            iam_role.ManagedPolicyArns = [
+                ArnGenerator.generate_aws_managed_policy_arn("service-role/AWSCodeDeployRoleForLambda")
+            ]
 
         return iam_role
 
--- a/samtranslator/region_configuration.py
+++ b/samtranslator/region_configuration.py
@@ -7,8 +7,6 @@ class RegionConfiguration(object):
     class abstracts all region/partition specific configuration.
     """
 
-    partitions = {"govcloud": "aws-us-gov", "iso": "aws-iso", "isob": "aws-iso-b", "china": "aws-cn"}
-
     @classmethod
     def is_apigw_edge_configuration_supported(cls):
         """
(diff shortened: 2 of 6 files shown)


samtranslator/translator/arn_generator.py
```diff
@@ -47,6 +47,10 @@ def get_partition_name(cls, region=None):
         region_string = region.lower()
         if region_string.startswith("cn-"):
             return "aws-cn"
+        elif region_string.startswith("us-iso-"):
```
mgrandis, line 50: nit: You don't need elifs/else here

samtranslator/model/preferences/deployment_preference_collection.py
```diff
@@ -101,9 +101,17 @@ def _codedeploy_iam_role(self):
                 }
             ],
         }
-        iam_role.ManagedPolicyArns = [
-            ArnGenerator.generate_aws_managed_policy_arn("service-role/AWSCodeDeployRoleForLambda")
-        ]
+
+        # CodeDeploy has a new managed policy. We cannot update any existing partitions, without customer reach out
+        # that support AWSCodeDeployRoleForLambda since this could regress stacks that are currently deployed.
+        if ArnGenerator.get_partition_name() in ["aws-iso", "aws-iso-b"]:
```
mgrandis, line 107: We are handling a lot of strings, can't we put those in constants?

samtranslator/region_configuration.py
```diff
@@ -18,4 +18,9 @@ def is_apigw_edge_configuration_supported(cls):
         :return: True, if API Gateway does not support Edge configuration
         """
 
-        return ArnGenerator.get_partition_name() not in [cls.partitions["govcloud"], cls.partitions["china"]]
+        return ArnGenerator.get_partition_name() not in [
```
wchengru, line 21: minor: Looks like the list can be replaced by cls.partitions.values(), so we don't change two places.

mgrandis: Isn't it cls.partitions.keys()?

hawflau: +1 for using `cls.partitions.values()`

it should be `cls.partitions.values()` instead of `cls.partitions.keys()`, because for example we are checking whether `"aws-iso"` is in the list, but not `"iso"`.

mgrandis: Yes indeed, I removed my comment :)

jfuss: Good catch. Not sure why this is defined this way in general. Will simplify.

samtranslator/model/preferences/deployment_preference_collection.py
jfuss: Fair call out. I want to try and limit changes (aka fix all the mentions like this). Let me play with this and see what we can get to for setting a better pattern in the code base.

mgrandis: Yes, let's merge this one this way and think about it in another PR

hawflau:
Do we also need to update `tests/translator/test_translator.py` and `tests/translator/output` to test templates for iso regions?

samtranslator/region_configuration.py
jfuss: There was no need for the dictionary that I could find. So I just pushed everything into the `in` statement.

jfuss:
> Do we also need to update `tests/translator/test_translator.py` and `tests/translator/output` to test templates for iso regions?

I added unit test around this so we don't need to do the full functional testing. We also have integ tests we run when deploying the service. I would prefer not to keep replicating these tests per region/partition.

hawflau (approved):
LGTM!

mgrandis (approved):

PR #1870, opened 2020-12-17
fix(AppConfig): Updated AppConfig boto3 Client Timeout

CoshUS:
*Description of changes:*
boto3 client has default 60 seconds timeout for both read and connection.
The worst case would be 360 seconds from 120 seconds timeout * 3 retires.
SAM translator lambda function has 120 seconds timeout which can be exceeded if AppConfig experiences a long response delay issue.

Tested with 900ms network latency. (1800ms roundt trip)

*Checklist:*

- [ ] Write/update tests
- [x] `make pr` passes
- [ ] Update documentation

*Examples?*

### DIFF
--- a/samtranslator/feature_toggle/feature_toggle.py
+++ b/samtranslator/feature_toggle/feature_toggle.py
@@ -110,7 +110,7 @@ class FeatureToggleAppConfigConfigProvider(FeatureToggleConfigProvider):
         try:
             # Lambda function has 120 seconds limit
             # (5 + 25) * 2, 60 seconds maximum timeout duration
-            client_config = Config(connect_timeout= 5, read_timeout=25, retries={"total_max_attempts": 2})
+            client_config = Config(connect_timeout=5, read_timeout=25, retries={"total_max_attempts": 2})
             self.app_config_client = boto3.client("appconfig", config=client_config)
             response = self.app_config_client.get_configuration(
                 Application=application_id,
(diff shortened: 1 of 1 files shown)


aahung (approved):

hoffa (approved):

samtranslator/feature_toggle/feature_toggle.py
```diff
@@ -105,8 +107,11 @@ class FeatureToggleAppConfigConfigProvider(FeatureToggleConfigProvider):
 
     def __init__(self, application_id, environment_id, configuration_profile_id):
         FeatureToggleConfigProvider.__init__(self)
-        self.app_config_client = boto3.client("appconfig")
         try:
+            # Lambda function has 120 seconds limit
+            # (5 + 25) * 2, 60 seconds maximum timeout duration
+            client_config = Config(connect_timeout=5, read_timeout=25, retries={"total_max_attempts": 2})
+            self.app_config_client = boto3.client("appconfig", config=client_config)
```
hoffa, line 114: Why was this moved under the `try`? If this fails the error will be silently ignored.

CoshUS: We shouldn't fail a request just because an error is caused by feature toggle. Proceed with empty config on errors should be the default.

hoffa (approved):

Issue #1539, opened 2020-03-26
Defining two Apis domains in the same hosted zone


NickDobsonMO:
**Description:**

When you define two `AWS::Serverless::Api` resources in the same template which have domains in the same Route 53 hosted zone the resulting Cloudformation template contains one `AWS::Route53::RecordSetGroup` which only includes the record set for the second API. The recordset for the first API is not present in the output and is lost.

During the transform, the logical id for the `AWS::Route53::RecordSetGroup` is generated using a hash of either the hosted zone name or hosted zone id. As the same hosted zone is used for both `AWS::Serverless::Api` resources domains, the logic name is the same and the `AWS::Route53::RecordSetGroup` created for the second `AWS::Serverless::Api` domain overwrites the first. 

**Steps to reproduce the issue:**
1. Create a template containing two gateways with domains in the same hosted zone:

```yaml
AWSTemplateFormatVersion: "2010-09-09"
Transform: AWS::Serverless-2016-10-31

Resources:
    ApiGatewayOne:
        Type: AWS::Serverless::Api
        Properties:
            StageName: Prod
            Domain:
                CertificateArn: CertificateArn
                DomainName: api-one.domain.com
                Route53:
                    HostedZoneName: HostedZoneName

    ApiGatewayTwo:
        Type: AWS::Serverless::Api
        Properties:
            StageName: Prod
            Domain:
                CertificateArn: CertificateArn
                DomainName: api-two.domain.com
                Route53:
                    HostedZoneName: HostedZoneName
```
2. Transform the SAM template to Cloudformation using the CLI or console.

**Observed result:**
```yaml
AWSTemplateFormatVersion: '2010-09-09'
Resources:
  ApiGatewayOne:
    Properties:
      Body:
        info:
          title:
            Ref: AWS::StackName
          version: '1.0'
        paths: {}
        swagger: '2.0'
    Type: AWS::ApiGateway::RestApi
  ApiGatewayOneDeploymentf9e2e3fce7:
    Properties:
      Description: 'RestApi deployment id: f9e2e3fce7ff39be96068344e44c14271beacd7e'
      RestApiId:
        Ref: ApiGatewayOne
      StageName: Stage
    Type: AWS::ApiGateway::Deployment
  ApiGatewayOneProdStage:
    Properties:
      DeploymentId:
        Ref: ApiGatewayOneDeploymentf9e2e3fce7
      RestApiId:
        Ref: ApiGatewayOne
      StageName: Prod
[...]

PR #1871, opened 2020-12-19
fix: RecordSetGroups resource overwrites the first

travisdieckmann:
*Issue #, if available:* #1539 

*Description of changes:* Refactor API Custom Domain RecordSetGroup Logical ID to be unique to the `AWS::Serverless::Api` resource

*Description of how you validated changes:* As the logical ID of the API resource must be unique, using this as the prefix to the translated RecordSetGroup ensures uniqueness as well as provides contextual association to the API in question.

*Checklist:*

- [X] Write/update tests
- [X] `make pr` passes
- [x] Update documentation

*Examples?*

### DIFF
--- a/samtranslator/model/api/api_generator.py
+++ b/samtranslator/model/api/api_generator.py
@@ -381,12 +381,8 @@ class ApiGenerator(object):
                     self.logical_id,
                     "HostedZoneId or HostedZoneName is required to enable Route53 support on Custom Domains.",
                 )
-            logical_id = logical_id_generator.LogicalIdGenerator(
-                "", route53.get("HostedZoneId") or route53.get("HostedZoneName")
-            ).gen()
-            record_set_group = Route53RecordSetGroup(
-                "RecordSetGroup" + logical_id, attributes=self.passthrough_resource_attributes
-            )
+            logical_id = "{}{}".format(self.logical_id, "DomainRecordSetGroup")
+            record_set_group = Route53RecordSetGroup(logical_id, attributes=self.passthrough_resource_attributes)
             if "HostedZoneId" in route53:
                 record_set_group.HostedZoneId = route53.get("HostedZoneId")
             if "HostedZoneName" in route53:
--- a/tests/translator/output/api_with_custom_domain_route53.json
+++ b/tests/translator/output/api_with_custom_domain_route53.json
@@ -1,50 +1,50 @@
 {
   "Parameters": {
     "ACMCertificateArn": {
-      "Default": "cert-arn-in-us-east-1", 
+      "Default": "cert-arn-in-us-east-1",
       "Type": "String"
-    }, 
+    },
     "DomainName": {
-      "Default": "example.com", 
+      "Default": "example.com",
       "Type": "String"
     }
-  }, 
+  },
   "Resources": {
     "MyFunction": {
-      "Type": "AWS::Lambda::Function", 
+      "Type": "AWS::Lambda::Function",
       "Properties": {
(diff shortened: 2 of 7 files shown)


samtranslator/model/api/api_generator.py
```diff
@@ -381,12 +381,8 @@ def _construct_api_domain(self, rest_api):
                     self.logical_id,
                     "HostedZoneId or HostedZoneName is required to enable Route53 support on Custom Domains.",
                 )
-            logical_id = logical_id_generator.LogicalIdGenerator(
-                "", route53.get("HostedZoneId") or route53.get("HostedZoneName")
-            ).gen()
-            record_set_group = Route53RecordSetGroup(
-                "RecordSetGroup" + logical_id, attributes=self.passthrough_resource_attributes
-            )
+            logical_id = "{}{}".format(self.logical_id, "DomainRecordSetGroup")
```
CoshUS, line 384: Thanks for the PR!
What's the reason for appending `"DomainRecordSetGroup"` to `self.logical_id` instead of passing `self.logical_id` as prefix to LogicalIdGenerator constructor?

TechnoNerdDad: Absolutely! I love SAM so being able to contribute back is my pleasure.

I have 2 assertions for using this format (examples below using `SomeApiGateway` as the API logical ID and `1e8f0a9f3b` as a generic hash):

a.) This follows the similar format to [ApiGatewayDeployment](https://github.com/aws/serverless-application-model/blob/58919bbf05c68183d3f76fcc8a01f4998a4512e5/samtranslator/model/api/api_generator.py#L230) or ApiGatewayBasePathMapping (as well as several others) formatting resulting in `SomeApiGatewayDeployment1e8f0a9f3b`
b.) When reviewing stack events, a logical ID of `SomeApiGateway` and `SomeApiGateway1e8f0a9f3b` does not intuitively distinguish the 2 different resources and requires cross-referencing when debugging/troubleshooting.

CoshUS (dismissed):

samtranslator/model/api/api_generator.py
CoshUS: After discussing with the team, we found that changing the Logical ID will force redeployments of existing RecordSetGroup resources which might break references to those ones already deployed.
One solution is to keep the logical ID for the first RecordSetGroup resource as the current implementation and update subsequent ones.

TechnoNerdDad: That is a fair point that it could break existing deployments and appreciate the desire to mitigate that risk. This is an interesting idea to alter the logical ID generation based on count-of-occurrence of `HostedZoneId` or `HostedZoneName`. Perhaps the team could provide some input/suggestions on the proposed method of achieving this before going through the effort of engineering the solution:

- Inspection of the calling object. This would break Python encapsulation and could add significant fragility.
- Pass a handle to the calling object as an argument to all Generator classes enabling inspection of the generated resources to allow the class method `to_cloudformation` to handle merging of duplicate logical IDs within the class instance.
- Create a class variable as a `dict` or `list` of previously used `HostedZoneId` and `HostedZoneName`. As a dict, other meta could be shared across class instances. This has a potential negative consequence of unpredictability in tests but nothing that some regex couldn't handle.

I'm sure I'm missing other plausible solutions. :) I'm happy to implement any proposed solution but would like some input as any one of the above seems appear to add some tech-debt, in my opinion.

jfuss:
Closing this as it was solved by a new PR we ended up merging: https://github.com/aws/serverless-application-model/pull/2408

PR #1884, opened 2021-01-07
Fix: Description in AWS::Serverless::HttpApi

hawflau:
*Issue #, if available:*

*Description of changes:*
This PR fixes how the `Description` property is handled for a `AWS::Serverless::HttpApi` resource:
When `Description` is set, one of the followings will happen:
* if `DefinitionBody` is not set, an error will be thrown
* if `DefinitionBody` is set and `description` is already set in the OpenAPI definition, an error will be thrown
* if `DefinitionBody` is set and no `description` is set in the OpenAPI definition, `description` will be set in the OpenAPI definition.

Before the changes, deploying a `AWS::Serverless::HttpApi` resource with `Description` set will result in an error:
```
Redundant fields [description="description"] provided when either Body or
BodyS3Location is not empty (Service: null; Status Code: 0; Error Code: null;
Request ID: null; Proxy: null) (Service: null; Status Code: 404; Error Code:
BadRequestException; Request ID: null; Proxy: null)  
```
Note that the same error does not occur for `AWS::Serverless::Api`, but only `AWS::Serverless::HttpApi`.

Updated and added tests to cover all changes made
```
pytest tests/openapi/test_openapi.py
pytest tests/model/test_sam_resources.py
```
Verified transformed template deploys without error.

*Checklist:*

- [x] Write/update tests
- [x] `make pr` passes
- [ ] Update documentation

*Examples?*

### DIFF
--- a/samtranslator/model/api/http_api_generator.py
+++ b/samtranslator/model/api/http_api_generator.py
@@ -112,6 +112,8 @@ class HttpApiGenerator(object):
         if self.disable_execute_api_endpoint is not None:
             self._add_endpoint_configuration()
 
+        self._set_description()
+
         if self.definition_uri:
             http_api.BodyS3Location = self._construct_body_s3_dict()
         elif self.definition_body:
@@ -123,9 +125,6 @@ class HttpApiGenerator(object):
                 "'AWS::Serverless::HttpApi'. Add a value for one of these properties or "
                 "add a 'HttpApi' event to an 'AWS::Serverless::Function'.",
             )
-
-        if self.description:
-            http_api.Description = self.description
 
         return http_api
 
@@ -585,6 +584,27 @@ class HttpApiGenerator(object):
         stage.RouteSettings = self.route_settings
 
         return stage
+
+    def _set_description(self):
+        """Sets Description to DefinitionBody"""
+        if not self.description:
+            return
+
+        if not self.definition_body:
+            raise InvalidResourceException(
+                self.logical_id,
+                "Description works only with inline OpenApi specified in the 'DefinitionBody' property.",
+            )
+        if self.definition_body.get("info", {}).get("description"):
+            raise InvalidResourceException(
+                self.logical_id,
+                "Unable to set Description because it is already defined within inline OpenAPI specified in the "
+                "'DefinitionBody' property.",
+            )
+
+        open_api_editor = OpenApiEditor(self.definition_body)
+        open_api_editor.set_description(self.description)
+        self.definition_body = open_api_editor.openapi
 
     def to_cloudformation(self):
         """Generates CloudFormation resources from a SAM HTTP API resource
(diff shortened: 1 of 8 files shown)


samtranslator/model/api/http_api_generator.py
```diff
@@ -586,6 +585,27 @@ def _construct_stage(self):
 
         return stage
 
+    def _set_description(self):
```
mndeveci, line 588: nit: when I read `set_xyz` I was expecting a setter which will get a value and sets the value of `xyz`. Can we change the method name to something else?

mndeveci (approved):

samtranslator/model/api/http_api_generator.py
hawflau: @mndeveci how about `_add_description`? It only sets `description` when it's not present, so essentially it's "add" instead of "set". It also matches other methods within the class like `_add_cors(...)` and `_add_endpoint_configuration(...)`

aahung (approved):

samtranslator/model/api/http_api_generator.py
hawflau: changed to `_add`

mndeveci (approved):

samtranslator/model/api/http_api_generator.py
```diff
@@ -586,6 +585,27 @@ def _construct_stage(self):
...
+        """Sets Description to DefinitionBody"""
+        if not self.description:
+            return
+
+        if not self.definition_body:
+            raise InvalidResourceException(
+                self.logical_id,
+                "Description works only with inline OpenApi specified in the 'DefinitionBody' property.",
+            )
+        if self.definition_body.get("info", {}).get("description"):
```
moelasmar, line 598: is description value with empty string accepted?

samtranslator/model/api/http_api_generator.py
```diff
@@ -586,6 +585,27 @@ def _construct_stage(self):
 
         return stage
 
+    def _set_description(self):
+        """Sets Description to DefinitionBody"""
+        if not self.description:
```
moelasmar, line 590: shall we do not do this validation if the description is an empty string?

samtranslator/model/api/http_api_generator.py
hawflau: `description` can be set as empty string in the OpenAPI definition, but APIGateway won't update the API Description. On the contrary, if `description` is not empty in the OpenAPI definition, APIGateway will update the API Description with the value of `description`. 
I think the idea is - if `description` is an empty string and the `Description` property is set in the SAM template, we probably want to use the value of `Description` to overwrite that of `description` in the OpenAPi definition. Does that make sense?

moelasmar (approved):

PR #1891, opened 2021-01-14
Fix: Replaced invalid AMQ managed policy by providing policy statements

qingchm:
*Issue #, if available:*

*Description of changes:*

*Checklist:*

- [X] Write/update tests
- [X] `make pr` passes
- [X] Update documentation

*Examples?*


samtranslator/model/eventsources/pull.py
```diff
@@ -181,11 +203,69 @@ class MSK(PullEventSource):
...
+                self.relative_id,
+                "SourceAccessConfigurations for ActiveMQ only supports single configuration entry.",
+            )
+        if not self.SourceAccessConfigurations[0].get("URI"):
+            raise InvalidEventException(
+                self.relative_id,
+                "No URI property specified in SourceAccessConfigurations for ActiveMQ.",
+            )
+        document = {
+            "PolicyName": "AMQPolicy",
```
mndeveci, line 241: nit: the name of the policy could be more specific to us, something like `SamAutogeneratedAMQPolicy`

samtranslator/model/eventsources/pull.py
```diff
@@ -181,11 +203,69 @@ class MSK(PullEventSource):
...
 
     resource_type = "MQ"
 
     def get_policy_arn(self):
-        return ArnGenerator.generate_aws_managed_policy_arn("service-role/AWSLambdaAMQExecutionRole")
+        return None
+
+    def get_policy_statements(self):
+        if not self.SourceAccessConfigurations:
+            raise InvalidEventException(
```
mndeveci, line 220: Can we check how `InvalidEventException` been handled? I remember we are using something like `InvalidDocumentException` for validations, and wondering if `InvalidEventException` will result in `4xx` error.

qingchm: File "/Users/qingchm/serverless-application-model/bin/../samtranslator/parser/parser.py", line 12, in parse
    self._validate(sam_template, parameter_values)
  File "/Users/qingchm/serverless-application-model/bin/../samtranslator/parser/parser.py", line 30, in _validate
    raise InvalidDocumentException([InvalidTemplateException("'Resources' section is required")])
samtranslator.model.exceptions.InvalidDocumentException: [InvalidTemplateException("'Resources' section is required")]

samtranslator/model/eventsources/pull.py
qingchm: This is a good suggestion!

samtranslator/model/eventsources/pull.py
hawflau: In both `pull.py` and `push.py`, `InvalidEventException` was used instead of `InvalidDocumentException`. 
I think it's ambiguous, but `InvalidEventException` might refer to an exception in an event source of a function

mndeveci: So looking at the code, `InvalidEventException` have been caught in here and `InvalidResourceException` have been thrown; https://github.com/aws/serverless-application-model/blob/develop/samtranslator/model/sam_resources.py#L193

And `InvalidResourceException` have been caught here, and errors have been appended; https://github.com/aws/serverless-application-model/blob/develop/samtranslator/translator/translator.py#L145

Looking at these, it should work as expected. And I guess this `InvalidEventException` is created specific to function event sources.

hawflau: Agreed. And I think the docstring of `InvalidEventException` should be better written.

mndeveci (approved):

hawflau (approved):

PR #1901, opened 2021-01-26
fix: prevent IndexError with integration URIs without .Arn

hoffa:
### Issue #, if available

N/A

### Description of changes

Mitigates issue where transforming a template such as the included `function_alias_with_http_api_events.yaml` causes an uncaught `IndexError`.

This only happens when a function alias has at least two events, one of which uses the default method/path. `get_integration_function_logical_id` is only used here for an optimization for events with default method/path (see integration test):

https://github.com/aws/serverless-application-model/blob/3bf8f611d7386d309c37bf8429e3d93692b07883/samtranslator/model/eventsources/push.py#L1074-L1079

If a function has an `HttpApi` event with default method and path, it corresponds to `*/*` in the `AWS::Lambda::Permission`s [`SourceArn` for executing APIs](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-control-access-using-iam-policies-to-invoke-api.html#api-gateway-calling-api-permissions), which encompasses _every_ method and path. The optimization is skipping all other `Permission`s since they're already encompassed by the `*/*`, and the permissions are always the same (`lambda:InvokeFunction`).

Ideally the same logic could be applied to function aliases. I didn't want to extend the regex further as it's already a hack that relies on a specific `Fn::Sub` format. The logical ID could maybe be tracked elsewhere, although this might be difficult if the input is the OpenAPI document.

Note: fixed `code.zip` as the function didn't run (was `c.done()` when it should've been `c.succeed()`).

### Side effects

- SAM Translator won't crash with `function_alias_with_http_api_events.yaml`.
- The optimization used for `Function`s won't be used if it includes `AutoPublishAlias` (i.e. it'll create a separate `AWS::Lambda::Permission` for each `HttpApi` event).

```shell
pytest --no-cov integration/single/test_basic_function.py -vv -k test_function_with_http_api_events
```

### Checklist

- [x] Write/update tests
- [x] `make pr` passes
- [ ] Update documentation

### Examples?


samtranslator/open_api/open_api.py
```diff
@@ -101,7 +101,14 @@ def get_integration_function_logical_id(self, path_name, method_name):
 
         # Extract lambda integration (${LambdaName.Arn}) and split ".Arn" off from it
         regex = "([A-Za-z0-9]+\.Arn)"
-        match = re.findall(regex, arn)[0].split(".Arn")[0]
+        matches = re.findall(regex, arn)
+        # For Functions with AutoPublishAlias the integration URI will contain
+        # ${LambdaName}, not ${LambdaName.Arn} (both resolve to the AWS::Lambda::Alias ARN).
```
jfuss, line 106: It is not the LogicalId of the Lambda Resource but the LogicalId of the `AWS::Lambda::Alias` right? If you do `LambdaName.Arn` you get the ARN of LASTEST which is not the same ARN as from a `Ref: LambdaAliasResource`.

samtranslator/open_api/open_api.py
```diff
@@ -101,7 +101,14 @@ def get_integration_function_logical_id(self, path_name, method_name):
...
+        # Without the .Arn suffix, parsing the URI is more ambiguous, so instead of
+        # introducing more brittle hacks, return gracefully.
+        if not matches:
+            return False
```
jfuss, line 110: Does this mean the integration is not setup then? Can we add and End to End tests to verify this actually fixes the issues and produces a deploy-able template?

jfuss (changes requested):
It's hard for me to tell if this actually addresses the failure without adding additional tests.

samtranslator/open_api/open_api.py
hoffa: Added

samtranslator/open_api/open_api.py
hoffa: Correct. The comment was misleading since `Alias` doesn't have `.Arn` like `Function` does (but `AliasLogicalId` and `FunctionLogicalId.Arn` both resolve to their respective ARNs).

jfuss (approved):

mgrandis (approved):

Issue #866, opened 2019-03-22
RFC: Support Events on functions within AWS::Serverless::Application


jlhood:
**Description:**
Per https://github.com/awslabs/serverless-application-model/issues/349#issuecomment-475723188, creating a separate RFC for a SAM syntax that would allow functions provided by nested applications (`AWS::Serverless::Application`) to be attached to event triggers via the same Events syntax as `AWS::Serverless::Function` resources.

High-Level Approach: Add a new SAM resource type: `AWS::Serverless::FunctionReference`. It doesn't create a new `AWS::Lambda::Function` resource. Instead, it allows you to specify the name and ARN of an existing function and then attach event sources to it using the exact same Events syntax as `AWS::Serverless::Function`.

The nice thing about this approach is it's not strictly tied to nested apps. It could be used with [Fn::ImportValue](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-importvalue.html) for example, if you want to add triggers to a function created in a separate stack in your account.

Example:

```yaml
  # Example nested SAR application that creates a function and outputs the function's name and ARN as template output variables: NestedFunctionName and NestedFunctionArn
  NestedApp:
    Type: AWS::Serverless::Application
    Properties:
      Location:
        ApplicationId: <app id>
        SemanticVersion: 1.0.0

  # example of attaching events to a function from a nested app/stack
  NestedFunctionExample:
    Type: AWS::Serverless::FunctionReference
    Properties:
      FunctionName: !GetAtt NestedApp.Outputs.NestedFunctionName
      FunctionArn: !GetAtt NestedApp.Outputs.NestedFunctionArn
      Events: 
        GetRoot:
          Type: Api
          Properties:
            Path: /
            Method: GET

  # example of attaching events to a function whose name/arn are exported by a different stack
  ImportValueExample:
    Type: AWS::Serverless::FunctionReference
    Properties:
      FunctionName: !ImportValue ExportedFunctionName
      FunctionArn: !ImportValue ExportedFunctionArn
      Events: 
        SQSTrigger:
          Type: SQS
          Properties:
            Queue: !GetAtt MyQueue.Arn
            BatchSize: 10
```

Questions:
1. Does the SAM transform need more than function name/ARN for any of the supported Events?
[...]

PR #1917, opened 2021-02-07
feat: Support adding events to existing functions (take 2)

Jacco:
*Issue #866:*
Support adding events to existing functions

*Description of changes:*
- class SamEventsResource(SamResourceMacro) to share events handling with SamFunction, SamApplication and SamFunctionReference
- class SamFunction now inherits from SamEventsResource 
- class SamStackFunctionReference(SamResourceMacro) helper function reference for sar
- class SamFunctionReference(SamEventsResource) introduced for use as new sam resource that supports events
- class SamApplication now inherits from SamEventsResource so events can be used on it

Added two translator tests and two integration tests

*Checklist:*

- [x] Write/update tests
- [x] `make pr` passes
- [ ] Update documentation

*Examples?*

### DIFF
--- a/integration/helpers/file_resources.py
+++ b/integration/helpers/file_resources.py
@@ -4,6 +4,7 @@ FILE_TO_S3_URI_MAP = {
     "swagger1.json": {"type": "s3", "uri": ""},
     "swagger2.json": {"type": "s3", "uri": ""},
     "template.yaml": {"type": "http", "uri": ""},
+    "stack.yaml": {"type": "http", "uri": ""},
 }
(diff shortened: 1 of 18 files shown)


jfuss:
@Jacco Just wanted to first say, thank you for spending the time on this. Given how old this is and conflicting files, I am going to close it.

I think this still could be powerful but it is unclear how we will want to achieve this. We are looking into what is next for SAM (#2319), which I think this could be a good candidate to make sure we solve (at the very least) within that context/issue.

Issue #1624, opened 2020-06-15
DeploymentPreference Alarms and Fn::If


keetonian:
I took a look into the Alarm error from #350.

I replicated the reported issue by using an `Fn::If` intrinsic in the `Alarms` section of my template:

```yaml
...
        Alarms:
          Fn::If:
            - IsBetaStage:
            - - { Ref: IncreasedFaults }
              - { Ref: IncreasedLatency }
            - - { Ref: IncreasedFaults2 }
              - { Ref: IncreasedLatency2 }
...
```

This resulted in the following JSON definition:

```json
        "AlarmConfiguration": {
          "Alarms": [
            {
              "Name": "Fn::If"
            }
          ], 
          "Enabled": true
        }, 
```

It looks like SAM is trying to provide shorthand syntax for supporting alarms; when SAM transforms this section, it needs to transform the alarms to be `Name: <alarm-name>`.

SAM does this transformation here: https://github.com/awslabs/serverless-application-model/blob/master/samtranslator/model/preferences/deployment_preference_collection.py#L104

A possible fix might be to explicitly add support for `Fn::If` for this property. If `Fn::If` is encountered here, we need to do the `Name: <alarm-name>` transformation on all items inside of the conditional. This would also need to be aware of other values common to `Fn::If`, like `AWS::NoValue`.

_Originally posted by @keetonian in https://github.com/awslabs/serverless-application-model/issues/350#issuecomment-496623239_

sriram-mv:
PR has been released.

PR #1923, opened 2021-02-11
feat: Handle Fn::If intrinsic for Alarms

mgrandis:
**Issue #, if available:**
#1624 

**Description of changes:**
Handles `Fn::If` in `DeploymentPreference/Alarms` where we would only allow a list before.

**Description of how you validated changes:**
- Unit tests
- E2E tests
- Deployed the resulting CFN template and verified the stack content
- Integration test

**Checklist:**
- [x] Write/update tests
- [x] `make pr` passes
- [ ] Update documentation

**Examples?**

### DIFF
--- a/samtranslator/model/preferences/deployment_preference_collection.py
+++ b/samtranslator/model/preferences/deployment_preference_collection.py
@@ -3,7 +3,13 @@ from samtranslator.model.codedeploy import CodeDeployApplication
 from samtranslator.model.codedeploy import CodeDeployDeploymentGroup
 from samtranslator.model.exceptions import InvalidResourceException
 from samtranslator.model.iam import IAMRole
-from samtranslator.model.intrinsics import fnSub, is_intrinsic, is_intrinsic_if, is_intrinsic_no_value
+from samtranslator.model.intrinsics import (
+    fnSub,
+    is_intrinsic,
+    is_intrinsic_if,
+    is_intrinsic_no_value,
+    is_valid_intrinsic_if_items,
+)
 from samtranslator.model.update_policy import UpdatePolicy
 from samtranslator.translator.arn_generator import ArnGenerator
 import copy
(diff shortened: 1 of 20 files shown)


samtranslator/model/preferences/deployment_preference_collection.py
```diff
@@ -152,6 +152,67 @@ def deployment_group(self, function_logical_id):
...
+        group : dict
+            Deployment group
+
+        Raises
+        ------
+        ValueError
+            If Alarms is in the wrong format
+        """
+        if not preference.alarms or is_intrinsic_no_value(preference.alarms):
+            return
```
elbayaaa, line 172: what if it is in the following of: [Fn::If, condition, AWS::NoValue, someValue] and the condition may or maynot evaluates to true?

mgrandis: We won't enter this condition and proceed to line 174

samtranslator/model/preferences/deployment_preference_collection.py
```diff
@@ -152,6 +152,67 @@ def deployment_group(self, function_logical_id):
 
         return deployment_group
 
+    def _process_alarms(self, preference, group):
```
elbayaaa, line 155: I see "group" is used only as an output parameter(lines 181,  & 184) I would prefer to remove it and make the call that way:
```
group.AlarmConfiguration = self._process_alarms(deployment_preference);
```

mgrandis: Indeed and return None at line 172

elbayaaa: in this case, you might want to rename "process_alarms" to something more descriptive

elbayaaa (approved):

samtranslator/model/preferences/deployment_preference_collection.py
```diff
@@ -152,6 +152,69 @@ def deployment_group(self, function_logical_id):
...
+            If Alarms is in the wrong format
+        """
+        if not preference_alarms or is_intrinsic_no_value(preference_alarms):
+            return None
+
+        if is_intrinsic_if(preference_alarms):
+            processed_alarms = copy.deepcopy(preference_alarms)
+            alarms_list = processed_alarms.get("Fn::If")
+            if not isinstance(alarms_list, list) or not len(alarms_list) == 3:
+                raise ValueError("Fn::If requires 3 arguments")
```
hawflau, line 181: nit: it looks like a generic check on `Fn::If`. I think it's better to refactor this part to intrinsics.py

hawflau (approved):

PR #1878, opened 2021-01-02
Add a built-in AWS_IAM authorizer for HTTP APIs

harrisonhjones:
This is a work in progress and does not yet contain tests. I am sending it out at this stage to get feedback on my approach. This is an **alternate** approach from the other PR I sent out, #1876. The main difference between that PR and this one is that this code change adds a new special `Authorizer` built in called `AWS_IAM` that **cannot** be defined in the `Authorizers` section of an HTTP API.

*Issue #, if available:* #1731

*Description of changes:*

This code change does one thing:
1. Adds a new built-in authorizer, `AWS_IAM`, to be used by HTTP APIs.

*Description of how you validated changes:*

I created the following test template:

```yml
AWSTemplateFormatVersion: "2010-09-09"
Description: AWS SAM template with a simple API definition
Transform: AWS::Serverless-2016-10-31
Resources:
  #######
  # Serverless functions that use the automatically-created AWS::Serverless::HttpApi called "ServerlessHttpApi".
  #######
  # Should have no auth set.
  HttpApiFunctionDefaultApiDefaultAuth:
    Type: AWS::Serverless::Function
    Properties:
      Events:
        ApiEvent:
          Type: HttpApi
          Properties:
            Path: /default-api/default-auth
            Method: GET
      Handler: index.handler
      InlineCode:
        "def handler(event, context):\n    return {'body':
        'HttpApiFunctionDefaultApiDefaultAuth', 'statusCode': 200}\n"
      Runtime: python3.8
  # Should have IAM auth set.
  HttpApiFunctionDefaultApiIAMAuth:
    Type: AWS::Serverless::Function
    Properties:
      Events:
        ApiEvent:
          Type: HttpApi
          Properties:
            Path: /default-api/iam-auth
            Method: GET
            Auth:
              Authorizer: AWS_IAM
      Handler: index.handler
      InlineCode:
        "def handler(event, context):\n    return {'body':
        'HttpApiFunctionDefaultApiIAMAuth', 'statusCode': 200}\n"
      Runtime: python3.8
  #######
  # Serverless functions that use a manually-created AWS::Serverless::HttpApi with the default auth set to the IAM authorizer.
  #######
  # Should have IAM auth set because the default authorizer is the IAM authorizer.
  HttpApiFunctionCustomApiWithDefaultIamAuthDefaultAuth:
    Type: AWS::Serverless::Function
    Properties:
      Events:
        ApiEvent:
          Type: HttpApi
          Properties:
[...]

PR #1876, opened 2020-12-27
WIP - Add support for IAM authorizers to AWS::Serverless::HttpApi

harrisonhjones:
This is a work in progress and does not yet contain tests. I am sending it out at this stage to get feedback on my approach.

*Issue #, if available:* #1731

*Description of changes:*

This code change does two things:
1. Adds support for IAM authorizers to AWS::Serverless::HttpApi
1. Adds an IAM authorizer to the automatically generated ServerlessHttpApi

I'd be happy to split this into two separate PRs if people feel that's more appropriate. It's already separate commits.

*Description of how you validated changes:*

I created the following test template:

```yml
AWSTemplateFormatVersion: "2010-09-09"
Description: AWS SAM template with a simple API definition
Transform: AWS::Serverless-2016-10-31
Resources:
  #######
  # Serverless functions that use the automatically-created AWS::Serverless::HttpApi called "ServerlessHttpApi".
  #######
  # Should have no auth set.
  HttpApiFunctionDefaultApiDefaultAuth:
    Type: AWS::Serverless::Function
    Properties:
      Events:
        ApiEvent:
          Type: HttpApi
          Properties:
            Path: /default-api/default-auth
            Method: GET
      Handler: index.handler
      InlineCode:
        "def handler(event, context):\n    return {'body':
        'HttpApiFunctionDefaultApiDefaultAuth', 'statusCode': 200}\n"
      Runtime: python3.8
  # Should have IAM auth set.
  HttpApiFunctionDefaultApiIAMAuth:
    Type: AWS::Serverless::Function
    Properties:
      Events:
        ApiEvent:
          Type: HttpApi
          Properties:
            Path: /default-api/iam-auth
            Method: GET
            Auth:
              Authorizer: AWS_IAM
      Handler: index.handler
      InlineCode:
        "def handler(event, context):\n    return {'body':
        'HttpApiFunctionDefaultApiIAMAuth', 'statusCode': 200}\n"
      Runtime: python3.8
  #######
  # Serverless functions that use a manually-created AWS::Serverless::HttpApi with the default auth set to an IAM authorizer.
  #######
  # Should have IAM auth set because the default authorizer is an IAM authorizer.
  HttpApiFunctionCustomApiWithDefaultIamAuthDefaultAuth:
    Type: AWS::Serverless::Function
    Properties:
      Events:
        ApiEvent:
          Type: HttpApi
          Properties:
            Path: /custom-api-with-default-iam-auth/default-auth
            Method: GET
[...]

Issue #1731, opened 2020-09-30
Unable to use  DefaultAuthorizer: IAM for AWS::Serverless::HttpApi


Kiranjith83:
**Description:**

  IAM Authorization configuration is supported by API GW [1], however not seems to be supported by CFN/SAM [2] yet. 

**Steps to reproduce the issue:**
1.
Sample template:

###  Sample SAM Template for sam-app
```
AWSTemplateFormatVersion: '2010-09-09'
Transform: AWS::Serverless-2016-10-31
Description: >
  sam-app
Globals:
  Function:
    Timeout: 3

Resources:
  HttpApiType:
    Type: AWS::Serverless::HttpApi
    Properties:
      FailOnWarnings: true
  HelloWorldFunction:
    Type: AWS::Serverless::Function # More info about Function Resource: https://github.com/awslabs/serverless-application-model/blob/master/versions/2016-10-31.md#awsserverlessfunction
    Properties:
      CodeUri: hello_world/
      Handler: app.lambda_handler
      Runtime: python3.7
      Events:
        HelloWorld:
          Type: HttpApi # More info about API Event Source: https://github.com/awslabs/serverless-application-model/blob/master/versions/2016-10-31.md#api
          Properties:
            Path: /hello
            Method: POST
            Auth:
              #Authorizer: AWS_IAM
              DefaultAuthorizer: AWS_IAM
            PayloadFormatVersion: "2.0" # aws-serverless-express doesn't support v2.0 yet
            ApiId:
              Ref: HttpApiType
```
2. ```sam deploy --guided```
3.

**Observed result:**
 Waiting for changeset to be created..
Error: Failed to create changeset for the stack: samappiam1, ex: Waiter ChangeSetCreateComplete failed: Waiter encountered a terminal failure state Status: FAILED. Reason: Transform AWS::Serverless-2016-10-31 failed with: Internal transform failure.

**Expected result:**
 Have the default authorizer configured as AWS IAM for HttpApi, created by SAM. 
Reference Links:
 [1] https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-access-control-iam.html 
 [2] https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-apigatewayv2-authorizer.html#cfn-apigatewayv2-authorizer-authorizertype

CrayTexas2018:
Any progress on this? This feature is very needed for one of my projects.

harrisonhjones:
[...]

PR #1924, opened 2021-02-11
feat: Add a built-in AWS_IAM authorizer for HTTP APIs

harrisonhjones:
Issue #, if available: #1731

*Description of changes:*

This is another approach for adding AWS_IAM support to HTTP APIs created by SAM. Other approaches can be found in #1876 and #1878. This approach is based on a conversation with @jfuss.

In this version a new property is added to the Serverless HTTP API object called `EnableIamAuthorizer`. When set to true an `AWS_IAM` authorizer is automatically added to the HTTP API's (internal) list of authorizers under the key "AWS_IAM". This authorizer is setup to us IAM auth. If enabled (It is not enabled by default) it can be set as the HTTP API's default authorizer and set as the authorizer for any Serverless functions that use the API.

I created the following test template:

```yaml
Description: AWS SAM template with a simple API definition
Globals:
  HttpApi:
    Auth:
      EnableIamAuthorizer: true
Resources:
  #######
  # Serverless functions that use the automatically-created AWS::Serverless::HttpApi called "ServerlessHttpApi".
  #######
  # Should have no auth set because there is no default authorizer defined on the API.
  HttpApiFunctionDefaultApiDefaultAuth:
    Properties:
      Events:
        ApiEvent:
          Type: HttpApi
          Properties:
            Path: /default-api/default-auth
            Method: GET
      Handler: index.handler
      InlineCode:
        "def handler(event, context):\n    return {'body':
        'HttpApiFunctionDefaultApiDefaultAuth', 'statusCode': 200}\n"
      Runtime: python3.8
  # Should have IAM auth set because the IAM authorizer is enabled (see Globals) and set.
  HttpApiFunctionDefaultApiIAMAuth:
    Properties:
      Events:
        ApiEvent:
          Type: HttpApi
          Properties:
            Path: /default-api/iam-auth
            Method: GET
            Auth:
              Authorizer: AWS_IAM
      Handler: index.handler
      InlineCode:
        "def handler(event, context):\n    return {'body':
        'HttpApiFunctionDefaultApiIAMAuth', 'statusCode': 200}\n"
      Runtime: python3.8
  #######
  # Serverless functions that use a manually-created AWS::Serverless::HttpApi with the IAM authorizer enable and the default auth set to the IAM authorizer.
  #######
  # Should have IAM auth set because the default authorizer is the IAM authorizer.
  HttpApiFunctionCustomApiWithDefaultIamAuthDefaultAuth:
    Properties:
      Events:
        ApiEvent:
          Type: HttpApi
          Properties:
            Path: /custom-api-with-default-iam-auth/default-auth
            Method: GET
            ApiId: !Ref CustomServerlessHttpApiWithDefaultIamAuth
      Handler: index.handler
      InlineCode:
        "def handler(event, context):\n    return {'body':
        'HttpApiFunctionCustomApiWithDefaultIamAuthDefaultAuth', 'statusCode':
        200}\n"
      Runtime: python3.8
  # Should have no auth (override the default authorizer).
  HttpApiFunctionCustomApiWithDefaultIamAuthNoneAuth:
    Properties:
      Events:
        ApiEvent:
          Type: HttpApi
          Properties:
            Path: /custom-api-with-default-iam-auth/none-auth
            Method: GET
            ApiId: !Ref CustomServerlessHttpApiWithDefaultIamAuth
            Auth:
              Authorizer: NONE
      Handler: index.handler
      InlineCode:
        "def handler(event, context):\n    return {'body':
        'HttpApiFunctionCustomApiWithDefaultIamAuthNoneAuth', 'statusCode':
        200}\n"
      Runtime: python3.8
  # Should have IAM auth set because the authorizer is set to the IAM authorizer.
  HttpApiFunctionCustomApiWithDefaultIamAuthIamAuth:
    Properties:
      Events:
        ApiEvent:
          Type: HttpApi
          Properties:
            Path: /custom-api-with-default-iam-auth/iam-auth
            Method: GET
            ApiId: !Ref CustomServerlessHttpApiWithDefaultIamAuth
            Auth:
              Authorizer: AWS_IAM
      Handler: index.handler
      InlineCode:
        "def handler(event, context):\n    return {'body':
        'HttpApiFunctionCustomApiWithDefaultIamAuthIamAuth', 'statusCode':
        200}\n"
      Runtime: python3.8
  # Serverless Http Api with an IAM authorizer (default) and a None authorizer.
  CustomServerlessHttpApiWithDefaultIamAuth:
    Properties:
      FailOnWarnings: true
      Auth:
        EnableIamAuthorizer: true
        DefaultAuthorizer: AWS_IAM

  #######
  # Serverless functions that use a manually-created AWS::Serverless::HttpApi with the IAm authorizer enabled but the default auth unset.
  #######
  # Should have no auth set because the default authorizer is not set.
  HttpApiFunctionCustomApiWithNoDefaultAuthDefaultAuth:
    Properties:
      Events:
        ApiEvent:
          Type: HttpApi
          Properties:
            Path: /custom-api-with-no-default-auth/default-auth
            Method: GET
            ApiId: !Ref CustomServerlessHttpApiWithNoDefault
      Handler: index.handler
      InlineCode:
        "def handler(event, context):\n    return {'body':
        'HttpApiFunctionCustomApiWithNoDefaultAuthDefaultAuth', 'statusCode':
        200}\n"
      Runtime: python3.8
  # Should have IAM auth set because the authorizer is set to the IAM authorizer.
  HttpApiFunctionCustomApiWithNoDefaultIamAuth:
    Properties:
      Events:
        ApiEvent:
          Type: HttpApi
          Properties:
            Path: /custom-api-with-no-default-auth/iam-auth
            Method: GET
            ApiId: !Ref CustomServerlessHttpApiWithNoDefault
            Auth:
              Authorizer: AWS_IAM
      Handler: index.handler
      InlineCode:
        "def handler(event, context):\n    return {'body':
        'HttpApiFunctionCustomApiWithNoDefaultIamAuth', 'statusCode': 200}\n"
      Runtime: python3.8
  # Serverless Http Api with the IAM authorizer enabled but no default authorizer set.
  CustomServerlessHttpApiWithNoDefault:
    Properties:
      FailOnWarnings: true
      Auth:
        EnableIamAuthorizer: true
```

I then compiled it using:

python ./bin/sam-translate.py --template-file=../project/template.yml
I then manually deployed it using CloudFormation in my personal AWS account.

Once deployed I validated all the routes had the expected auth type:

*Default API*

1. https://ptwccx1qv4.execute-api.us-east-1.amazonaws.com/default-api/default-auth - has no auth
1. https://ptwccx1qv4.execute-api.us-east-1.amazonaws.com/default-api/iam-auth - has IAM auth

*Custom API with no default auth*

1. https://warmewfp00.execute-api.us-east-1.amazonaws.com/custom-api-with-no-default-auth/default-auth - has no auth
1. https://warmewfp00.execute-api.us-east-1.amazonaws.com/custom-api-with-no-default-auth/iam-auth - has IAM auth

*Custom API with default IAM auth*

1. https://fj5vzsu863.execute-api.us-east-1.amazonaws.com/custom-api-with-default-iam-auth/default-auth - has IAM auth
1. https://fj5vzsu863.execute-api.us-east-1.amazonaws.com/custom-api-with-default-iam-auth/iam-auth - has IAM auth
1. https://fj5vzsu863.execute-api.us-east-1.amazonaws.com/custom-api-with-default-iam-auth/none-auth - has no auth

*Checklist:*

- [x] Write/update tests
- [x] `make pr` passes
- [ ] Update documentation

*Examples?*


jfuss:
@harrisonhjones I think this is probably the best option from the 3 you have proposed. I didn't do a deep review of the code, but I think the setup you have is good. I am having another member of my team (@hawflau) look at this as well, just so we have some agreement on the approach.

hawflau:
@harrisonhjones This option looks good to me.
I compared with the other two options you provided, but it seems they require altering things that customers might currently rely on. We should avoid this.
The code implementation in this option also looks fine. We just probably need more tests. And when we roll this out, we need docs and examples to let customers know how to use IAM auth

harrisonhjones:
> @harrisonhjones This option looks good to me.

:+1:

> I compared with the other two options you provided, but it seems they require altering things that customers might currently rely on. We should avoid this.

I agree.

> The code implementation in this option also looks fine. We just probably need more tests.

Got it. Will add tests.

> And when we roll this out, we need docs and examples to let customers know how to use IAM auth

Where do I add that? In this package or elsewhere? I see an "Update Documentation" checkbox but after a quick scan of the repo I don't see how documentation is supposed to be updated.

harrisonhjones:
# 2/18/2021 Update:

- ➕ Added tests
- ✅ `make pr` passes
- ✅ integration test: `pytest --no-cov integration/single/test_function_with_http_api_and_auth.py` passes
- ✅existing integration test: `pytest --no-cov integration/single/test_basic_http_api.py` passes
- ⚠️ confirmed I don't like writing in Python.

samtranslator/model/api/http_api_generator.py
```diff
@@ -489,14 +489,21 @@ def _set_default_authorizer(self, open_api_editor, authorizers, default_authoriz
                 path, default_authorizer, authorizers=authorizers, api_authorizers=api_authorizers
             )
 
-    def _get_authorizers(self, authorizers_config, default_authorizer=None):
+    def _get_authorizers(self, authorizers_config, enable_iam_authorizer=False):
```
jfuss, line 492: Just making sure I understand here. It seems like `default_authorizer` was never used in this function so it should be safe to updated this. I just want to double check what I am seeing/understanding to make sure there is no accidentally side affect here.

samtranslator/model/api/http_api_generator.py
```diff
@@ -489,14 +489,21 @@ def _set_default_authorizer(self, open_api_editor, authorizers, default_authoriz
...
         """
         Returns all authorizers for an API as an ApiGatewayV2Authorizer object
         :param authorizers_config: authorizer configuration from the API Auth section
-        :param default_authorizer: name of the default authorizer
+        :param enable_iam_authorizer: if true add an "AWS_IAM" authorizer
         """
         authorizers = {}
 
+        if enable_iam_authorizer:
```
jfuss, line 500: I haven't got to the test yet but we should make sure we testing that handles different arguments. Python objects are all truthy, so what I want to make sure we validate is that the end overall experience handles things like "empty resource value, true is turned into a bool from the template, same with false, what happens when a customer sends a string instead of a bool type. 

My main concern is how this might behave. SAM runs before CFN will do anything to a template, so we get a lot of 'raw' data and can lead to small bugs that are harder to unwind/support.

samtranslator/model/eventsources/push.py
```diff
@@ -1167,36 +1167,51 @@ def _add_auth_to_openapi_integration(self, api, editor):
...
-        if method_authorizer != "NONE" and not api_authorizers:
-            raise InvalidEventException(
-                self.relative_id,
-                "Unable to set Authorizer [{authorizer}] on API method [{method}] for path [{path}] "
-                "because the related API does not define any Authorizers.".format(
-                    authorizer=method_authorizer, method=self.Method, path=self.Path
-                ),
-            )
+        # The IAM authorizer is built-in and not defined as a regular Authorizer.
+        enable_iam_authorizer = api_auth and api_auth.get("EnableIamAuthorizer")
```
jfuss, line 1171: If `EnabledIamAuthorizer` is `yes` or `"true"` or `FN:if...` this expression would trigger to be true and cause us to enable the `IAM_AUTH`. Somewhere we should have validation on the value of `EnbableIamAuthorizer`. I think @mgrandis is looking/thinking about this area already. I would like to make sure we consider any validation we may need to do on this pr before merging. I don't have the best understanding of what should be done in this area. 

@mgrandis Could you take a look through this and suggest how we might add this so we keep things consistent with what you are working through now?

tests/model/api/test_http_api_generator.py
```diff
@@ -76,6 +77,57 @@ def test_auth_intrinsic_default_auth(self):
         with pytest.raises(InvalidResourceException):
             HttpApiGenerator(**self.kwargs)._construct_http_api()
 
+    def test_auth_iam_enabled(self):
```
jfuss, line 80: This goes back to my validation comment. This might not be the right place to assert that happens but we should have some negative test cases (string passed, dict, etc). The main reason is that CFN send json data to a CFN Macro, which is how SAM runs, and CFN doesn't do any validation on that input. So we need to make sure we are guarding our code properly from input way may receive in that json data.

jfuss (changes requested):

jfuss:
@hawflau or @mgrandis Could one of take a look through this as well?

jfuss:
@harrisonhjones 

> Where do I add that? In this package or elsewhere? I see an "Update Documentation" checkbox but after a quick scan of the repo I don't see how documentation is supposed to be updated.

I would need to double check, but I believe we generate docs based on https://github.com/aws/serverless-application-model/blob/develop/versions/2016-10-31.md. I will handle the communication with our doc writer about getting these things added, but updating that file will help in our process.

samtranslator/model/api/http_api_generator.py
harrisonhjones: yea. I was a bit surprised when I noticed it appeared to be unused. As far as I can tell it was unused.

mgrandis:
Sorry about the late review @harrisonhjones!

I have mixed feelings about adding a new property `EnableIamAuthorizer` to `Auth` rather than a new value for `Authorizers`.

Couldn't we add a new `IamAuthorizer` with that `EnableIamAuthorizer` property in it instead?
It would be future proof if we were to add a new option/property to configure the `IamAuthorizer` in the future.

jfuss:
@mgrandis The main reason @harrisonhjones and I landed on this design was because the IAM Auth is a static configuration. Forcing a customer to specify `Authorizers` seemed like unnecessary set. `Authorizers` is also a set a different objects, which can be confusing. So we elected to push this up to an `Enable` Property under Auth. We played with just setting the `DefaultAuthorizer: AWS_IAM` or `Auth.Authorizer: AWS_IAM` but forcing an Property to set what Authorizers are available made checking for Auth in other tools (like SAM CLI or CFN Lint) easier and more explicit.

Do you still have concerns in this approach?

jfuss (approved):
The looks good to me.

mgrandis:
All right, let's move on with this approach.

samtranslator/model/api/http_api_generator.py
mgrandis: Should maybe switch to `enable_iam_authorizer == True`

samtranslator/model/api/http_api_generator.py
```diff
@@ -489,14 +489,21 @@ def _set_default_authorizer(self, open_api_editor, authorizers, default_authoriz
...
+            authorizers["AWS_IAM"] = ApiGatewayV2Authorizer(is_aws_iam_authorizer=True)
+
+        # If all the customer wants to do is enable the IAM authorizer the authorizers_config will be None.
+        if authorizers_config == None:
```
mgrandis, line 504: nit: Think it's better to use `is` instead of `==`

samtranslator/open_api/open_api.py
```diff
@@ -415,7 +415,8 @@ def _set_method_authorizer(self, path, method_name, authorizer_name, authorizers
             security_dict = dict()
             security_dict[authorizer_name] = []
 
-            if authorizer_name != "NONE":
+            # Neither the NONE nor the AWS_IAM built-in authorizers support authorization scopes.
+            if authorizer_name != "NONE" and authorizer_name != "AWS_IAM":
```
mgrandis, line 419: nit: You can use `not in ["NONE", "AWS_IAM"]` here

samtranslator/model/api/http_api_generator.py
mgrandis: Used in `ApiGenerator` so maybe it was copied from there.

samtranslator/model/api/http_api_generator.py
harrisonhjones: ok, so `if authorizers_config is None` ? Why is `is` preferred over `==`?

samtranslator/open_api/open_api.py
harrisonhjones: Fancy; seems like a reasonable change.

samtranslator/model/api/http_api_generator.py
harrisonhjones: Why that over the implied `== True`? (Genuinely asking; Python is not my strong suit)

samtranslator/model/api/http_api_generator.py
mgrandis: `==` is for equality, `is` is for identity.
It's a good practice when checking that something is not set, to use `is None`
You may actually want to use `if not authorizers_config:` here, as this will also check for values like empty list `[]` or empty object `{}` (which I think could be good checks here)
About is/==: https://stackoverflow.com/a/48120163

samtranslator/model/api/http_api_generator.py
mgrandis: We could pass `"False"` in this value for example and it would pass the if condition.
```
>>> a = "False"
>>> "foo" if a else "bar"
'foo'
```
We should actually use `is True` instead of `== True`:
```
>>> 1 == True
True
```

samtranslator/model/eventsources/push.py
mgrandis: We could maybe go with:
`enable_iam_authorizer = api_auth and api_auth.get("EnableIamAuthorizer", False) is True`

If EnableIamAuthorizer is missing, default to False and check that the value is indeed a Boolean True

tests/model/api/test_http_api_generator.py
mgrandis: Agree, on a general note we need to test against many more bad/wrong inputs

harrisonhjones: Do y'all want me to write more tests? If so, I'd appreciate some guidance on which tests to write.

samtranslator/model/eventsources/push.py
harrisonhjones: Works for me

tests/model/api/test_http_api_generator.py
mgrandis: You could add a `test_auth_iam_not_enabled_with_unsupported_values(self)`
It would loop on an array of values and check for each one that no IAM auth is added.
Examples of non boolean values that we could get as input:
`unsupported_values = [1, "", [], {}, { "Ref": "MyVar" }, "True", None]`

samtranslator/model/apigatewayv2.py
```diff
@@ -174,6 +178,14 @@ def generate_openapi(self):
         """
         authorizer_type = self._get_auth_type()
 
+        if authorizer_type == "AWS_IAM":
+            openapi = {
```
mgrandis, line 182: We should maybe return early here and in the `JWT` case.
So that we don't have to scan the rest of the function to see what we're doing with `openapi`.

samtranslator/model/api/http_api_generator.py
```diff
@@ -489,14 +489,21 @@ def _set_default_authorizer(self, open_api_editor, authorizers, default_authoriz
...
+            return authorizers
+
         if not isinstance(authorizers_config, dict):
             raise InvalidResourceException(self.logical_id, "Authorizers must be a dictionary.")
```
mgrandis, line 509: Sorry not able to add a comment down below, but we should maybe add a condition in the `for` loop to ignore or fail on finding an `AWS_IAM` object in `authorizers_config.items()` so that we don't overwrite the one defined by `enable_iam_authorizer`

harrisonhjones: I think that's fair. I was originally thinking we'd want to allow an overwrite but that was when this change was opt-out. Now that it is opt-in we shouldn't let users shoot themselves in the foot.

samtranslator/model/api/http_api_generator.py
harrisonhjones: Got it. Fairly easy change.

samtranslator/model/api/http_api_generator.py
harrisonhjones: Got it. Fairly easy change.

samtranslator/model/apigatewayv2.py
harrisonhjones: Yea. I didn't do it that way because we didn't do it that way for any of the other authorizers. I would prefer to make this change, if desired, in another smaller PR.

tests/model/api/test_http_api_generator.py
harrisonhjones: Got it. Fairly easy change.

AndersAsa:
Any timeline for this feature being merged?

tests/model/api/test_http_api_generator.py
```diff
@@ -1,6 +1,7 @@
 from unittest import TestCase
 from mock import patch
 import pytest
+import json
```
mgrandis, line 4: nit: I think you can remove this as it's not used anymore

mgrandis:
@harrisonhjones It looks like there are only a few minor updates to do, do you have time to work on them?
Thanks!

wolfeidau:
@mgrandis looks like the author isn't able to address this feedback, can we merge this and follow up with another PR?

I would love to see this merged as it has been an issue for a while.

CrayTexas2018:
Any timeline on when this PR will be merged?

harrisonhjones:
> @mgrandis looks like the author isn't able to address this feedback, can we merge this and follow up with another PR?
> 
> I would love to see this merged as it has been an issue for a while.

I can try to take a look at fixing this next week. Sorry for the delay getting back to y'all.

AndersAsa:
> > @mgrandis looks like the author isn't able to address this feedback, can we merge this and follow up with another PR?
> > I would love to see this merged as it has been an issue for a while.
> 
> I can try to take a look at fixing this next week. Sorry for the delay getting back to y'all.

Thank you!!!! This will be super helpful for me

ben-elsen:
> > @mgrandis looks like the author isn't able to address this feedback, can we merge this and follow up with another PR?
> > I would love to see this merged as it has been an issue for a while.
> 
> I can try to take a look at fixing this next week. Sorry for the delay getting back to y'all.

Hi @harrisonhjones, will you have some time in the near future to finish this? 
This would extremely helpful!

tests/model/api/test_http_api_generator.py
harrisonhjones: Turns out this caught an issue so great suggestion!

harrisonhjones:
I have addressed all PR feedback and added a few new tests. `make pr` passes

samtranslator/model/api/http_api_generator.py
```diff
@@ -24,8 +24,8 @@
 )
 CorsProperties.__new__.__defaults__ = (None, None, None, None, None, False)
 
-AuthProperties = namedtuple("_AuthProperties", ["Authorizers", "DefaultAuthorizer"])
-AuthProperties.__new__.__defaults__ = (None, None)
+AuthProperties = namedtuple("_AuthProperties", ["Authorizers", "DefaultAuthorizer", "EnableIamAuthorizer"])
+AuthProperties.__new__.__defaults__ = (None, None, None)
```
harrisonhjones, line 28: Looking at this now: should the last `None` here actually be `False`?

samtranslator/model/api/http_api_generator.py
```diff
@@ -494,14 +494,21 @@ def _set_default_authorizer(self, open_api_editor, authorizers, default_authoriz
                 path, default_authorizer, authorizers=authorizers, api_authorizers=api_authorizers
             )
 
-    def _get_authorizers(self, authorizers_config, default_authorizer=None):
+    def _get_authorizers(self, authorizers_config, enable_iam_authorizer=False):
         """
         Returns all authorizers for an API as an ApiGatewayV2Authorizer object
         :param authorizers_config: authorizer configuration from the API Auth section
-        :param default_authorizer: name of the default authorizer
+        :param enable_iam_authorizer: if true add an "AWS_IAM" authorizer
```
harrisonhjones, line 501: s/true/True ?

CrayTexas2018:
> 

Thank you! Hopefully this gets merged! Really appreciate the effort.

tests/model/api/test_http_api_generator.py
```diff
@@ -1,6 +1,7 @@
 from unittest import TestCase
 from unittest.mock import patch
 import pytest
+import json
```
harrisonhjones, line 4: reminder to self to actually remove this...

harrisonhjones (commented):
Three small comments; two for more experienced Python folks.

samtranslator/model/api/http_api_generator.py
jfuss: `None` is falsy so they both work the same. It would only matter if `None` had some value to be considered different. For this, I don't think it strictly matters.

samtranslator/model/api/http_api_generator.py
jfuss: Technically yes but personally don't think it strictly matters. The communication of the comment is the same.

samtranslator/model/api/http_api_generator.py
harrisonhjones: Would it be better to use False instead of None largely to make it clear we expect a boolean value?

samtranslator/model/eventsources/push.py
```diff
@@ -1218,38 +1218,53 @@ def _add_auth_to_openapi_integration(self, api, editor):
...
-        if method_authorizer != "NONE" and not api_authorizers.get(method_authorizer):
-            raise InvalidEventException(
-                self.relative_id,
-                "Unable to set Authorizer [{authorizer}] on API method [{method}] for path [{path}] "
-                "because it wasn't defined in the API's Authorizers.".format(
-                    authorizer=method_authorizer, method=self.Method, path=self.Path
-                ),
-            )
+        if method_authorizer == "AWS_IAM":
+            if not enable_iam_authorizer:
```
jfuss, line 1225: Why fail if IAM auth is not enabled?

tests/translator/input/error_http_api_invalid_iam_auth.yaml
```diff
@@ -0,0 +1,37 @@
+Resources:
```
jfuss, line 1: I know you added integration tests (which thank you!) but could we add some of these input/output tests for valid cases? It is much easier to run these tests as quick validations of the translations.

jfuss (changes requested):
Just a couple small things.

samtranslator/model/eventsources/push.py
harrisonhjones: I guess we don't need to. In fact if they have a custom authorizer named "AWS_IAM" this code change might actually break them. I'll remove it.

tests/translator/input/error_http_api_invalid_iam_auth.yaml
harrisonhjones: I'll look into that sure

harrisonhjones:
Addressed PR feedback; added five new tests, removed an unnecessary error test, and updated and existing multi-auth test to include the new iam authorizer.

CoshUS (commented):
Thank you for the contribution! One small question and everything else looks good to me.

samtranslator/model/api/http_api_generator.py
harrisonhjones: Yea. if they already have an authorizer called "AWS_IAM" SAM doesn't clobber/overwrite it. The user's authorizer is used (though it likely won't help them).

samtranslator/model/apigatewayv2.py
harrisonhjones: I'm open to that. As a Go developer I prefer early returns but if elif is more python-style I'm open to it. Can we _not_ do it as part of this PR?

CoshUS (approved):
LGTM! Thanks for addressing the comments!

harrisonhjones:
@CoshUS you are welcome :) What are the next steps to getting this merged? Anything I need to do?

harrisonhjones:
Rebased off of develop, fixed merge conflicts, updated new util function that was breaking tests, squashed into a single commit, ran `make pr` successfully, and force-pushed.

samtranslator/model/apigatewayv2.py
moelasmar: I prefer moving the openapi object creation to separate private functions, but it is Ok for now.

moelasmar (approved):

tests/translator/output/aws-cn/http_api_custom_iam_auth.json
```diff
@@ -0,0 +1,137 @@
...
+      },
+      "Type": "AWS::IAM::Role"
+    },
+    "MyApi": {
+      "Properties": {
+        "Body": {
+          "components": {
+            "securitySchemes": {
+              "AWS_IAM": {
+                "type": "oauth2",
```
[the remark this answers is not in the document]
harrisonhjones: See the description in tests/translator/input/http_api_custom_iam_auth.yaml:

```
# This test-case tests what happens when an AWS_IAM authorizer is defined on an HttpApi but not enabled anywhere else.
# While the defined authorizer isn't really a true Iam authorizer (it's just a poorly-named OAuth authorizer) this shouldn't cause an error.
```

jfuss (approved):

jfuss:
There are some things we need to handle internally (docs for example). I need to kick off that processes before merging.

claudiocidade:
great work guys this is fantastic. any estimates as to when it's merged/released? I have a bit of a life/death situation here that would greatly benefit from this one

brysontyrrell:
I was just talking to my TAM about this. 👀

Issue #1851, opened 2020-12-08
Cannot set HttpApi Event Type on AWS::Serverless::StateMachine


davingreen:
I'm attempting to configure step functions to be triggered from an HttpApi (rather than a REST api). Here is my template:

```
  APIGatewayDeployment:
    Type: AWS::Serverless::HttpApi
    Properties:
      StageName: !Ref pStage
    DefinitionBody:
        info:
          title: !Ref AWS::StackName
          version: "1.0"
        openapi: 3.0.1

  MyStateMachine:
    Type: AWS::Serverless::StateMachine
    Properties:
      DefinitionUri: statemachines/MyStateMachine.asl.json
      DefinitionSubstitutions:
        MyFunction: !GetAtt MyFunction.Arn
      Events:
        HttpAPIGateway:
          Type: HttpApi
          Properties:
            Path: /inbound
            Method: post
            ApiId: !Ref APIGatewayDeployment
      Policies: 
        - LambdaInvokePolicy:
            FunctionName: !Ref MyFunction
```

This builds, but fails validation or deployment with the following:
`Error: Failed to create changeset for the stack: , ex: Waiter ChangeSetCreateComplete failed: Waiter encountered a terminal failure state Status: FAILED. Reason: Transform AWS::Serverless-2016-10-31 failed with: Invalid Serverless Application Specification document. Number of errors found: 1. Resource with id [MyStateMachine] is invalid. Event with id [HttpAPIGateway] is invalid. Resource dict has missing or invalid value for key Type. Event Type is: HttpApi.`

If I've done something wrong, let me know! I've been unable to find helpful documentation about using HTTP rather than REST with step functions.

peterwmwong:
This would help our team aswell if `HttpApi` was a supported event source.
We tried to work around this by defining the necessary CloudFormation resources manually (extending the example above)...

```yaml
MyStateMachine:
  Type: AWS::Serverless::StateMachine
  # ... The State Machine configuration

MyStateMachineIntegrationRole:
  Type: AWS::IAM::Role
  Properties:
    AssumeRolePolicyDocument:
      Version: 2012-10-17
      Statement:
        - Sid: MyStateMachineIntegrationStartExecuteStateMachines
          Effect: Allow
          Principal:
            Service:
              - apigateway.amazonaws.com
          Action:
            - states:StartExecution

MyStateMachineIntegration:
  Type: AWS::ApiGatewayV2::Integration
  Properties:
[...]

PR #1925, opened 2021-02-11
feat: Add HttpApi Event as new event type for AWS::Serverless::StateMachine

Jacco:
*Issue #1851*

*Description of changes:*
- integration test
- unit test(s)
- added a full role to also support stop and startSync
- added action property to event source (can be start, stop, startSync)
- error added when you try to add stop to EXPRESS state machine
- error added when you try to add startSync to STANDARD state machine
- refactored adding auth to HttpApi from `eventsources/push.py` and `stepfunctions/events.py`
- added TimeoutMillis support

- used the integration test to verify deployed httpapi
- visually verified that all the integrations were added properly
- exported openapi yaml and checked all is there
- tested if the API responded correctly using Postman

*Checklist:*

- [x] Write/update tests
- [x] `make pr` passes
- [ ] Update documentation

*Examples?*

See integration test that was added

*TODO*
- [x] test with various auth options
- [x] Error, Cause, Input, Name,  TraceHeader parameters for action
- [x] discuss role naming/scope
- [x] discuss Action property
- [x] discuss possibility of cross region integration api (eu-west-1) -> statemachine (us-east-1)
- [x] timeOutMilles is missing
- [x] fix code coverage (get open_api.py) green
- [ ] discuss output parameters

### DIFF
--- a/integration/helpers/file_resources.py
+++ b/integration/helpers/file_resources.py
@@ -4,6 +4,7 @@ FILE_TO_S3_URI_MAP = {
     "swagger1.json": {"type": "s3", "uri": ""},
     "swagger2.json": {"type": "s3", "uri": ""},
     "template.yaml": {"type": "http", "uri": ""},
+    "statemachine.json": {"type": "s3", "uri": ""},
 }
 
 CODE_KEY_TO_FILE_MAP = {
@@ -11,4 +12,5 @@ CODE_KEY_TO_FILE_MAP = {
     "contenturi": "layer1.zip",
     "definitionuri": "swagger1.json",
     "templateurl": "template.yaml",
+    "statemachineurl": "statemachine.json",
 }
--- a/integration/resources/code/statemachine.json
+++ b/integration/resources/code/statemachine.json
@@ -1,0 +1,16 @@
+{
+    "Comment": "A Hello World example of the Amazon States Language using Pass states",
+    "StartAt": "Hello",
+    "States": {
+      "Hello": {
+        "Type": "Pass",
+        "Result": "Hello",
+        "Next": "World"
+      },
+      "World": {
+        "Type": "Pass",
+        "Result": "World",
+        "End": true
+      }
+    }
+  }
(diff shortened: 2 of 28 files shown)


Jacco:
Discussion about role naming:
======================

It used to be:

`StateMachineName + EventName + 'Role'`

I changed role naming to:

`ApiName + StateMachineName + 'Role'`

If one StateMachine has more than one events to an API the same role will be reused (it will be generated multiple times, so it is exported only once), instead of a new one for every event. Still with this naming a new Role will be created per API. So I propose a improvement:

I think it would even be better `StateMachineName + 'StartStopRole'`

Then only one role would be created. This role can then also be (re)used for other event sources than API.

@jfuss

Jacco:
Discussion about new "Action" property on EventSource:
=========================================
So I added an "Action" property to the event source. There are three possible values:

- `start` for IntegrationSubtype `StepFunctions-StartExecution`
- `stop` for IntegrationSubtype `StepFunctions-StopExecution`
-  `startSync` for IntegrationSubtype `StepFunctions-StartSyncExecution`

If no "Action" property is specified `start` is assumed.

Maybe it makes sense that for EXPRESS StateMachines the default would be `startSync`?

Jacco:
Discussion extra parameters
====================

Name - for ExecutionName in start & startSync
TraceHeader - for startSync
Input - for start & startSync
Error - for stop
Cause - for stop

To support these parameter I added:

Parameters optional property that is a dict (it is not check but passed to CloudFormation)

Jacco:
Discussion cross region HttpApi -> StateMachine
===================================

By **default** the API integration will assume the specified ARN is in the same region as the API is, but the integrations support calling a StateMachine in a different region.

If you specify the Region parameter (which cannot be specified from the request.header/body/path) the API will look for the specified ARN in the Region specified. This means the region in the ARN should match the Region specified otherwise `com.amazonaws.swf.service.v2.model#StateMachineDoesNotExist` is returned.

I think the way to support this with SAM is as follows:

``` yaml
Resources:
  MyStateMachine:
    Type: AWS::Serverless::StateMachineReference
    Properties:
      Name: MyStateMachine
      Region: us-east-1
      Account: 123454567 # is cross account supported
      Events:
        SyncPost:
          Type: HttpApi
          Properties:
            Path: /sync
            Method: post
            Action: startSync    # static checking if resource is EXPRESS??
```

The integration would then use the StateMachine ARN it can contruct from the StateMachineReference properties and also specify the Region in the integration so the integration looks in the correct region for the StateMachine.

Jacco:
Discussion documentation
===================

Where is the documentation :-)

mndeveci:
Closing this due to inactivity. To reopen this please create an issue first to get it prioritized with the SAM team.

PR #1935, opened 2021-02-19
test: Add service exclusion by region

mingkun2020:
*Description of changes:*
Add region partition test for foss integration tests

*Checklist:*

- [ ] Write/update tests
- [x] `make pr` passes
- [x] Update documentation

*Examples?*


integration/helpers/resource.py
```diff
@@ -107,3 +114,50 @@ def create_bucket(bucket_name, region):
...
+        the resource to be tested in the current testing region
+
+    Returns
+    -------
+    Boolean
+        If skip return true otherwise false
+    """
+
+    session = boto3.session.Session()
+    my_region = session.region_name
```
mgrandis, line 154: nit: you can rename `my_region` to `region`

integration/helpers/resource.py
```diff
@@ -107,3 +114,50 @@ def create_bucket(bucket_name, region):
...
+    -------
+    Object
+        Yaml object
+    """
+    with open(file_path) as f:
+        data = f.read()
+    return yaml.load(data, Loader=yaml.FullLoader)
+
+
+def should_exclude_test_in_region(exclude_resource):
```
mgrandis, line 138: I would rename this to `current_region_does_not_support(service)` so that it reads like a sentence:

`@skipIf(current_region_does_not_support("ServerlessRepo"), ...)`

mgrandis (changes requested):

integration/helpers/resource.py
mgrandis: Also should this accept a list instead of a single item?

integration/helpers/resource.py
```diff
@@ -107,3 +114,50 @@ def create_bucket(bucket_name, region):
         s3_client = boto3.client("s3", region_name=region)
         location = {"LocationConstraint": region}
         s3_client.create_bucket(Bucket=bucket_name, CreateBucketConfiguration=location)
+
+
+def load_yaml(file_path):
```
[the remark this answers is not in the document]
mingkun2020: I just create a yaml_util.py file and move both the load and dump method there. It seems these 2 methods are not sharing resources so I decide not to create a class.

integration/config/region_exclude_resources.yml
```diff
@@ -0,0 +1,103 @@
+regions:
```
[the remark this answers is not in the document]
mingkun2020: Yes, this naming style makes more senses.

integration/helpers/resource.py
mingkun2020: Updated.

integration/helpers/resource.py
mingkun2020: Good idea, updated.

moelasmar: @mgrandis .. if it will support list, so shall it return True if all services in the list are excluded, or just any of them

integration/single/test_basic_layer_version.py
```diff
@@ -1,6 +1,10 @@
+from unittest.case import skipIf
+
 from integration.helpers.base_test import BaseTest
+from integration.helpers.resource import should_exclude_test_in_region
 
 
+@skipIf(should_exclude_test_in_region("Layers"), "Layers is not supported in this testing region")
```
moelasmar: I think it is a good idea to move the skip if decorator to the testing methods, so any new test case will be added to this class will not be skipped by default, and if it is needed to be skipped, the developer should explicitly add the skip if decorator

integration/config/region_exclude_resources.yml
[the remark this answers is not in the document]
moelasmar: I think this file will not contain just contain AWS services or resources, I think it is mainly for some use cases, like I below in the code, we check if "LambdaEnvVars" is supported or not, and I think "LambdaEnvVars" is a use case (or functionality) provided by the Lambda service.

integration/single/test_basic_function.py
```diff
@@ -15,6 +17,7 @@ class TestBasicFunction(BaseTest):
             "basic_function_openapi",
         ]
     )
+    @skipIf(should_exclude_test_in_region("LambdaEnvVars"), "LambdaEnvVars is not supported in this testing region")
```
moelasmar: I could not find the "LambdaEnvVars" in the exclusion file. Does it mean that it should be supported by all regions?
Also, does it mean that all the test cases, should define the functionality name that it got tested by it regardless it will be added to the exclusion file or not?

integration/helpers/resource.py
mgrandis: I would say any.
If the test relies on 2 services and only 1 is not available in a region it shouldn't run.

integration/helpers/base_test.py
```diff
@@ -349,24 +343,6 @@ def verify_stack(self):
...
-        Returns
-        -------
-        Object
-            Yaml object
-        """
-        with open(file_path) as f:
-            data = f.read()
-        return yaml.load(data, Loader=yaml.FullLoader)
-
     def _dump_yaml(self, file_path, yaml_doc):
```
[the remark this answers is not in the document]
mgrandis: You may be viewing an older version of the file

integration/config/region_exclude_resources.yml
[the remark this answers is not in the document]
mgrandis: What about `feature` then?

integration/single/test_basic_function.py
mingkun2020: I will remove the "LambdaEnvVars" since it we don't use it in the exclusion file. If we don't want to skip a test in certain region, we don't need to define a service/functionality name in the test case within skipIf decorator.

integration/single/test_basic_layer_version.py
mingkun2020: Agree, will update.

mgrandis (approved):

moelasmar (approved):

PR #1938, opened 2021-02-23
Raise an InvalidEventException if api-event's properties is not a dic…

elbayaaa:
…tionalry

*Issue #, if available:*

*Description of changes:*

*Checklist:*

- [√] Write/update tests
- [√] `make pr` passes
- [ ] Update documentation

*Examples?*


samtranslator/plugins/api/implicit_http_api_plugin.py
```diff
@@ -57,6 +57,10 @@ def _process_api_events(self, function, api_events, template, condition=None):
         for logicalId, event in api_events.items():
             # api_events only contains HttpApi events
             event_properties = event.get("Properties", {})
+
+            if event_properties and not isinstance(event_properties, dict):
+                raise InvalidEventException(logicalId, "Event Properties is invalid.")
```
sriram-mv, line 62: should we add a hint of what type it needs to be?

sriram-mv (changes requested):

samtranslator/plugins/api/implicit_http_api_plugin.py
elbayaaa: That is the part I'm struggling with, what a clear message could be. I don't want to use the python-specific term "dict/dictionary". Do you have a good suggestion?

sriram-mv (approved):

wchengru (approved):

hawflau (commented):
nit: Can `Properties` for other event sources be a string at all? I think it should be a dict object for all event sources. Should we extend this check for all event sources?

elbayaaa:
> nit: Can `Properties` for other event sources be a string at all? I think it should be a dict object for all event sources. Should we extend this check for all event sources?

We have a `on_before_transform_template` hook that scans lambda functions and step function for API events to merge them to the one serverless-api resource the template has/should have. during this hook we didn't handle the case if the Properties is not a dict. This PR is to solve this issue.

During the Lambda function resource transform i.e. after the "on_before_transform_template" hook, we process all the event-sources of a given function(API, Schedule, EventBridgeRule ...etc) and during this we handle such an issue(Properties not a dict) by a simple try..except. see [here](https://github.com/aws/serverless-application-model/blob/a79e4d689e1b3af13963258855614a559e090322/samtranslator/model/sam_resources.py#L563)

Issue #1728, opened 2020-09-30
Adding X-Ray Tracing to State Machine Does Not Update Step Function IAM Role (X-Ray Permissions)


gekart:
Step functions have been extended with X-Ray support: https://aws.amazon.com/blogs/compute/introducing-aws-x-ray-new-integration-with-aws-step-functions/

In SAM, adding (updating)

```
      Tracing:
        Enabled: True
```
correctly enables X-Ray for a State Machine, but does not modify the corresponding IAM role (X-Ray permissions). 

However, adding

```
      Tracing: Active
```

to functions does modify IAM correspondingly:

```
CloudFormation events from changeset
---------------------------------------------------------------------------------------------------------------------------------
ResourceStatus                   ResourceType                     LogicalResourceId                ResourceStatusReason           
---------------------------------------------------------------------------------------------------------------------------------
UPDATE_IN_PROGRESS               AWS::IAM::Role                   StockSellerFunctionRole          -                              
UPDATE_IN_PROGRESS               AWS::IAM::Role                   StockBuyerFunctionRole           -                              
UPDATE_IN_PROGRESS               AWS::IAM::Role                   StockCheckerFunctionRole         -                              
UPDATE_COMPLETE                  AWS::IAM::Role                   StockBuyerFunctionRole           -                              
UPDATE_COMPLETE                  AWS::IAM::Role                   StockCheckerFunctionRole         -                              
UPDATE_COMPLETE                  AWS::IAM::Role                   StockSellerFunctionRole          -                              
UPDATE_IN_PROGRESS               AWS::Lambda::Function            StockBuyerFunction               -                              
UPDATE_COMPLETE                  AWS::Lambda::Function            StockCheckerFunction             -                              
UPDATE_COMPLETE                  AWS::Lambda::Function            StockBuyerFunction               -                              
UPDATE_IN_PROGRESS               AWS::Lambda::Function            StockCheckerFunction             -                              
UPDATE_COMPLETE                  AWS::Lambda::Function            StockSellerFunction              -                              
UPDATE_IN_PROGRESS               AWS::Lambda::Function            StockSellerFunction              -                              
UPDATE_IN_PROGRESS               AWS::StepFunctions::StateMachi   StockTradingStateMachine         -                              
                                 ne                                                                                               
UPDATE_COMPLETE                  AWS::StepFunctions::StateMachi   StockTradingStateMachine         -                              
                                 ne                                                                                               
[...]

PR #1941, opened 2021-02-24
Fix missing xray policy for State Machine resource

hawflau:
*Issue #, if available:*
#1728 

*Description of changes:*
If `tracing` is enabled and `Policies` is used in the State Machine resource, the created IAM role for the State Machine resource will include a correct X Ray managed policy (depending on partition). 

Add tests to cover case when `tracing` is enabled and `Policies` is used in the State Machine resource in a template.

*Checklist:*

- [x] Write/update tests
- [x] `make pr` passes
- [ ] Update documentation

*Examples?*


samtranslator/model/sam_resources.py
```diff
@@ -455,11 +456,7 @@ def _construct_role(self, managed_policy_map, event_invoke_policies):
         if self.Tracing:
             # use previous (old) policy name for regular regions
```
mgrandis, line 457: You can move these two comments to the new `get_xray_managed_policy_name()` function

mgrandis (approved):

samtranslator/model/sam_resources.py
hawflau: done

wchengru (approved):

PR #1945, opened 2021-02-26
fix(policy-templates): Correct AMIDescribePolicy

### LINKED ISSUE #1944: Format of Generated ARN for Policy Template `AMIDescribePolicy` Is Incorrect
### Description:

When adding the SAM Policy Template `AMIDescribePolicy` to a Function, it is unable to describe images. The policy template uses as the input to `!Sub` the value `arn:${AWS::Partition}:ec2:${AWS::Region}:${AWS::AccountId}:image/*`. This is wrong in two ways:

1. AWS documentation claims that the format of an image's ARN is this: `arn:${Partition}:ec2:${Region}::image/${ImageId}`, without an account ID.
2. `ec2:DescribeImages` can't take a resource anyway; it has to be `*`.

### Steps to reproduce:

Create a Lambda Function which writes to an SSM parameter with data type `aws:ec2:image`. (Writing a valid AMI as the value, natch.) Provide this Function with permissions via the SAM Policy Template `AMIDescribePolicy`.

Upon execution, note that the client (boto3, say) confirms that the write succeeds. Watch CloudTrail for the parameter validation event. The event will indicate a failure with "Client.UnauthorizedOperation: You are not authorized to perform this operation."

Modify the Function to recreate `AMIDescribePolicy` manually:

```yaml
- Effect: Allow
  Action: ec2:DescribeImages
  Resource: '*'
```

Execute with the same input, note that the write succeeds, and note that the validation succeeds.

### Observed result:

The validation fails and the parameter is rolled back to its previous version. (Or, I guess, the provisional new version never becomes current.)

### Expected result:

The provisional new version becomes current. 

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

1. OS: macOS Catalina 10.15.7
2. If using SAM CLI, `sam --version`: 1.19.1
3. AWS region: eu-west-1

`Add --debug flag to any SAM CLI commands you are running`

chrisoverzero:
The action underlying this policy does not take a particular resource.

*Issue #, if available:*

Resolves #1944

*Description of changes:*

Removes templated ARN from `AMIDescribePolicy`, replacing it with a static `*`.

Aside from the translator tests, created a Lambda Function using these permissions and successfully performed the action.

*Checklist:*

- [x] Write/update tests
- [x] `make pr` passes
- [x] ~Update documentation~ _The documentation appears not to be in the repo – wherever that is will have to be updated. Maybe it's generated?_

*Examples?*

### DIFF
--- a/samtranslator/policy_templates_data/policy_templates.json
+++ b/samtranslator/policy_templates_data/policy_templates.json
@@ -478,9 +478,7 @@
             "Action": [
               "ec2:DescribeImages"
             ],
-            "Resource": {
-              "Fn::Sub": "arn:${AWS::Partition}:ec2:${AWS::Region}:${AWS::AccountId}:image/*"
-            }
+            "Resource": "*"
           }
         ]
       }
--- a/tests/translator/output/all_policy_templates.json
+++ b/tests/translator/output/all_policy_templates.json
@@ -336,9 +336,7 @@
                   "Action": [
                     "ec2:DescribeImages"
                   ], 
-                  "Resource": {
-                    "Fn::Sub": "arn:${AWS::Partition}:ec2:${AWS::Region}:${AWS::AccountId}:image/*"
-                  }, 
+                  "Resource": "*", 
                   "Effect": "Allow"
                 }
               ]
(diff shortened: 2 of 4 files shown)


hawflau (approved):
LGTM!

mgrandis (approved):

PR #1952, opened 2021-03-03
Percentage-based Enablement for Feature Toggle

hawflau:
*Issue #, if available:*

*Description of changes:*
Add support for Percentage-based Enablement for Feature Toggle.
When `enabled-%` is set (instead of `enabled`) under a region config, `is_enabled_for_account_in_region(...)` will determine whether the given `account_id` falls in the percentage range for feature enablement.

Added more tests

*Checklist:*

- [x] Write/update tests
- [x] `make pr` passes
- [ ] Update documentation

*Examples?*

### DIFF
--- a/samtranslator/feature_toggle/dialup.py
+++ b/samtranslator/feature_toggle/dialup.py
@@ -17,13 +17,13 @@ class BaseDialup(object):
         return self.__class__.__name__
 
 
-class NeverEnabledDialup(BaseDialup):
+class DisabledDialup(BaseDialup):
     """
     A dialup that is never enabled
     """
 
     def __init__(self, region_config, **kwargs):
-        super(NeverEnabledDialup, self).__init__(region_config)
+        super(DisabledDialup, self).__init__(region_config)
 
     def is_enabled(self):
         return False
--- a/samtranslator/feature_toggle/dialup.py
+++ b/samtranslator/feature_toggle/dialup.py
@@ -1,0 +1,65 @@
+import hashlib
+
+
+class BaseDialup(object):
+    def __init__(self, region_config, **kwargs):
+        self.region_config = region_config
+
+    def is_enabled(self):
+        raise NotImplementedError
+
+    def __str__(self):
+        return self.__class__.__name__
+
+
+class NeverEnabledDialup(BaseDialup):
+    """
+    A dialup that is never enabled
+    """
+
+    def __init__(self, region_config, **kwargs):
+        super(NeverEnabledDialup, self).__init__(region_config)
+
+    def is_enabled(self):
+        return False
+
+
+class ToggleDialup(BaseDialup):
+    """
+    A simple toggle Dialup
+    Example of region_config: { "type": "toggle", "enabled": True }
+    """
+
+    def __init__(self, region_config, **kwargs):
+        super(ToggleDialup, self).__init__(region_config)
+        self.region_config = region_config
+
+    def is_enabled(self):
+        return self.region_config.get("enabled", False)
+
+
+class SimpleAccountPercentileDialup(BaseDialup):
+    """
+    Simple account percentile dialup, enabling X% of
+    Example of region_config: { "type": "account-percentile", "enabled-%": 20 }
+    """
+
+    def __init__(self, region_config, account_id, feature_name, **kwargs):
+        super(SimpleAccountPercentileDialup, self).__init__(region_config)
+        self.account_id = account_id
+        self.feature_name = feature_name
+
+    def _get_account_percentile(self):
+        """
+        Get account percentile based on sha256 hash of account ID and feature_name
+
+        :returns: integer n, where 0 <= n < 100
+        """
+        m = hashlib.sha256()
+        m.update(self.account_id.encode())
+        m.update(self.feature_name.encode())
+        return int(m.hexdigest(), 16) % 100
+
+    def is_enabled(self):
+        target_percentile = self.region_config.get("enabled-%", 0)
+        return self._get_account_percentile() < target_percentile
--- a/samtranslator/feature_toggle/dialup.py
+++ b/samtranslator/feature_toggle/dialup.py
@@ -2,10 +2,15 @@ import hashlib
 
 
 class BaseDialup(object):
+    """BaseDialup class to provide an interface for all dialup classes"""
+
     def __init__(self, region_config, **kwargs):
         self.region_config = region_config
 
     def is_enabled(self):
+        """
+        Returns a bool on whether this dialup is enabled or not
+        """
         raise NotImplementedError
 
     def __str__(self):
@@ -61,5 +66,9 @@ class SimpleAccountPercentileDialup(BaseDialup):
         return int(m.hexdigest(), 16) % 100
 
     def is_enabled(self):
+        """
+        Enable when account_percentile falls within target_percentile
+        Meaning only (target_percentile)% of accounts will be enabled
+        """
         target_percentile = self.region_config.get("enabled-%", 0)
         return self._get_account_percentile() < target_percentile
--- a/samtranslator/feature_toggle/feature_toggle.py
+++ b/samtranslator/feature_toggle/feature_toggle.py
@@ -37,6 +37,16 @@ class FeatureToggle:
             LOG.info("Stage '{}' not enabled for Feature '{}'.".format(stage, feature_name))
             return False
         region_config = stage_config.get(region, {}) if region in stage_config else stage_config.get("default", {})
+        if "enabled-%" in region_config:
+            LOG.warning(
+                """
+            Percentage-based enablement is configured for Feature '{}' in Stage '{}' in Region '{}'. 
+            Please use is_enabled_for_account_in_region() instead. Feature not enabled.
+            """.format(
+                    feature_name, stage, region
+                )
+            )
+            return False
         is_enabled = region_config.get("enabled", False)
         LOG.info("Feature '{}' is enabled: '{}'".format(feature_name, is_enabled))
         return is_enabled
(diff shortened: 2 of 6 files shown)


samtranslator/feature_toggle/feature_toggle.py
```diff
@@ -57,11 +67,23 @@ def is_enabled_for_account_in_region(self, feature_name, stage, account_id, regi
...
+            account_config = stage_config.get(account_id)
+            region_config = (
+                account_config.get(region, {}) if region in account_config else account_config.get("default", {})
+            )
+        else:
+            region_config = stage_config.get(region, {}) if region in stage_config else stage_config.get("default", {})
+
+        if "enabled-%" in region_config:
+            # Percentage-based enablement
+            bound = region_config.get("enabled-%")
```
c2tarun, line 81: Probably add few lines of comments explaining what bound and partition mean here.

c2tarun (changes requested):

samtranslator/feature_toggle/feature_toggle.py
```diff
@@ -57,11 +67,23 @@ def is_enabled_for_account_in_region(self, feature_name, stage, account_id, regi
...
-        account_config = stage_config.get(account_id) if account_id in stage_config else stage_config.get("default", {})
-        region_config = (
-            account_config.get(region, {}) if region in account_config else account_config.get("default", {})
-        )
-        is_enabled = region_config.get("enabled", False)
+
+        if account_id in stage_config:
+            account_config = stage_config.get(account_id)
+            region_config = (
+                account_config.get(region, {}) if region in account_config else account_config.get("default", {})
```
hoffa, line 74: The first `get` is misleading/unused as `region in account_config`, so it's probably better to `account_config[region]`. Same for `stage_config`.

samtranslator/feature_toggle/feature_toggle.py
```diff
@@ -57,11 +67,23 @@ def is_enabled_for_account_in_region(self, feature_name, stage, account_id, regi
...
+            )
+        else:
+            region_config = stage_config.get(region, {}) if region in stage_config else stage_config.get("default", {})
+
+        if "enabled-%" in region_config:
+            # Percentage-based enablement
+            bound = region_config.get("enabled-%")
+            partition = int(account_id) % 100
```
hoffa, line 82: Unless account IDs are uniformly distributed, this will give biased results. Should include a comment as to why this was chosen.

hawflau: Updated

samtranslator/feature_toggle/feature_toggle.py
hawflau: Removed unnecessary uses of dict.get method

samtranslator/feature_toggle/feature_toggle.py
```diff
@@ -18,8 +18,22 @@ class FeatureToggle:
...
     """
 
-    def __init__(self, config_provider):
+    def __init__(self, config_provider, stage="beta", account_id=None, region="default"):
         self.feature_config = config_provider.config
+        self.stage = stage
+        self.account_id = account_id
+        self.region = region
+
+    def is_enabled(self, feature_name):
```
hoffa, line 27: Why is this method needed? e.g. instead of moving the `if` to the actual call?

hawflau: I renamed `is_enabled_for_account_in_region` to this one, removed `is_enabled_for_stage_in_region`. What `is_enabled_for_stage_in_region` does is already done in `is_enabled_for_account_in_region`.

I also made `stage`, `account_id`, `region` required when instanciating `FeatureToggle`

hoffa: Nice! Turned out really nice! 👍

hoffa (approved):

samtranslator/feature_toggle/feature_toggle.py
sriram-mv: Can we confirm that account ids are infact uniformly distributed?

sriram-mv (changes requested):

samtranslator/feature_toggle/feature_toggle.py
```diff
@@ -18,50 +18,53 @@ class FeatureToggle:
...
+            region_config = stage_config[region] if region in stage_config else stage_config.get("default", {})
+
+        if "enabled-%" in region_config:
+            # Percentage-based enablement
+            # Assumption: account_id is uniformly distributed
+            # account_id is calculated into one of 100 partitions (0-99)
+            # if partition < enabled_percent, we consider the feature is enabled for this given account_id
+            enabled_percent = region_config["enabled-%"]
+            partition = int(account_id) % 100
+            is_enabled = partition < enabled_percent
```
sriram-mv, line 64: Can we automate this dial up or enablement by specifiying start and end time in UTC. this way its automatically controlled dialup once the config is on appconfig, without needing any pushes for dialup.

sriram-mv (changes requested):

samtranslator/feature_toggle/feature_toggle.py
hoffa: If we have plans for dial-up, we _could_ make this generic from the get-go. The percentage is linear between start/end percentages during the time interval, and an always-on dial is an open interval (e.g. end is infinite).

hawflau: Let's just handle the simple percentage-based enablement in this PR first. We can consider automated dial up later

samtranslator/feature_toggle/feature_toggle.py
```diff
@@ -18,52 +19,65 @@ class FeatureToggle:
...
-        region_config = stage_config.get(region, {}) if region in stage_config else stage_config.get("default", {})
-        is_enabled = region_config.get("enabled", False)
+
+        if account_id in stage_config:
+            account_config = stage_config[account_id]
+            region_config = account_config[region] if region in account_config else account_config.get("default", {})
+        else:
+            region_config = stage_config[region] if region in stage_config else stage_config.get("default", {})
+
+        if "enabled-%" in region_config:
```
hoffa, line 58: Would it make sense to refactor this if-else block away so that it's easier to reason about the core logic and adapt to more complex toggles (e.g. linear)?

hawflau: Yeah that makes sense

samtranslator/feature_toggle/feature_toggle.py
```diff
@@ -18,51 +19,76 @@ class FeatureToggle:
...
-            LOG.info("Stage '{}' not enabled for Feature '{}'.".format(stage, feature_name))
-            return False
-        account_config = stage_config.get(account_id) if account_id in stage_config else stage_config.get("default", {})
-        region_config = (
-            account_config.get(region, {}) if region in account_config else account_config.get("default", {})
-        )
-        is_enabled = region_config.get("enabled", False)
-        LOG.info("Feature '{}' is enabled: '{}'".format(feature_name, is_enabled))
+        m = hashlib.sha256()
+        m.update(self.account_id.encode())
```
sriram-mv, line 71: this may have been pre-validated, so its more of a question. will we always have an account_id and feature_name defined by the time we get here?

samtranslator/feature_toggle/feature_toggle.py
```diff
@@ -18,51 +19,76 @@ class FeatureToggle:
...
+        else:
+            region_config = stage_config[region] if region in stage_config else stage_config.get("default", {})
+
+        is_enabled = self._is_feature_enabled_for_region_config(feature_name, region_config)
+
         LOG.info("Feature '{}' is enabled: '{}'".format(feature_name, is_enabled))
         return is_enabled
 
-    def is_enabled_for_account_in_region(self, feature_name, stage, account_id, region="default"):
+    def _get_account_percentile(self, feature_name):
```
sriram-mv, line 63: can we make this more generic such as a class? and then we can have different mechanisms (classes) to to do the dial up potentially.

Potential names for each dialup class.

```
SimplePercentageDialup
TimeIntervalLinearDialup
```

sriram-mv (changes requested):

samtranslator/feature_toggle/feature_toggle.py
hawflau: yes, both of them are defined when we get here. `account_id` is required when instantiating the `FeatureToggle` class, and `feature_name` is a required parameter for this method.

samtranslator/feature_toggle/dialup.py
```diff
@@ -0,0 +1,65 @@
...
+        self.region_config = region_config
+
+    def is_enabled(self):
+        raise NotImplementedError
+
+    def __str__(self):
+        return self.__class__.__name__
+
+
+class NeverEnabledDialup(BaseDialup):
```
sriram-mv, line 20: DisabledDialup maybe?

samtranslator/feature_toggle/feature_toggle.py
```diff
@@ -19,12 +24,35 @@ class FeatureToggle:
...
+        get the right dialup instance
+        if no dialup type is provided or the specified dialup is not supported,
+        an instance of NeverEnabledDialup will be returned
+
+        :param region_config: region config
+        :param feature_name: feature_name
+        :return: an instance of
+        """
+        dialup_type = region_config.get("type")
+        if dialup_type in FeatureToggle.DIALUP_RESOLVER:
```
sriram-mv, line 49: is FeatureToggle.DIALUP_RESOLVER.keys() required?

sriram-mv (approved):
Just some nits, otherwise looks good.

samtranslator/feature_toggle/feature_toggle.py
hawflau: no, just using `if key in dict_obj:` is good

samtranslator/feature_toggle/dialup.py
hawflau: Updated

PR #1957, opened 2021-03-08
Update state machine generator to use Definition object instead of DefinitionString

jormello:
*Description of changes:*
Step Functions recently released a new CloudFormation property called Definition, which allows users to specify their state machine definitions in object format. Since SAM already supported users providing their state machine definitions as an object, we would convert that object to a DefinitionString and use that in the generated CloudFormation template. With this update to the state machine resource schema, we no longer need to do this conversion in SAM, and can simply pass the Definition object directly to the CFN template.

A request for documentation updates has been made to note that SAM Definition property is now passed directly to CFN Definition property.

- Update to unit test suite for State Machine resource. 
- Manual testing of generated CloudFormation template 

*Checklist:*

- [X] Write/update tests
- [X] `make pr` passes
- [X] Update documentation

*Examples?*
No changes are needed for the state machine examples. 

### DIFF
--- a/samtranslator/model/stepfunctions/generators.py
+++ b/samtranslator/model/stepfunctions/generators.py
@@ -124,7 +124,7 @@ class StateMachineGenerator(object):
                     self.state_machine.DefinitionSubstitutions.update(substitutions)
                 else:
                     self.state_machine.DefinitionSubstitutions = substitutions
-            self.state_machine.DefinitionString = self._build_definition_string(processed_definition)
+            self.state_machine.Definition = processed_definition
         elif self.definition_uri:
             self.state_machine.DefinitionS3Location = self._construct_definition_uri()
         else:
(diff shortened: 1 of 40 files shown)


jormello:
Will these changes cause any impact to customers who have created State Machine resources using SAM? If a customer provides the same SAM template, they will now generate a different CFN template (using Definition property instead of DefinitionString), meaning there will potentially be a CFN update even if there is no change to their template. Is there typically customer wording provided by the service team to help communicate that these template changes are expected and do not necessarily indicate a change in the underlying resource?

wchengru (approved):
The PR looks good to me. I only have a couple of questions about the change.
1. If the CFN template changes, how will the resource "AWS::StepFunctions::StateMachine" get updated? Will they get deployed again, or the resource will be deleted and created again? Will there be any other changes related to this update?
2. Will CFN announce this change?

jormello:
1. The expected CFN template change is that the DefinitionString property will no longer be defined, and instead the state machine definition will be specified in the Definition property. Given that the CFN template is changing, customers should expect to see an UpdateStateMachine call to be performed even if they have not changed the definition in their template in any way. The definitionString value in the UpdateStateMachineRequest could potentially be different (if sibling properties are listed in a different order, or if there is a different number of indentation spaces used). These definitionString values will be functionally identical, meaning although there is a call to UpdateStateMachine, the resource itself should have no changes.

SimonCMoore:
Not quite sure why this stalled, but the number one concern I would have is causing an unexpected redeploy for existing stacks without there being any other change to the statemachine.  I think we need to understand potential impact for existing customers (e.g. if they are under load would they experience an outage /drop traffic and for how long).  Secondly I am not clear if the resolution would always be identical, so there may be other unexpected impact. An alternative approach might be to add a new parameter for this so it can be backward compatible but users can use the SFN native approach.

hoffa:
We could potentially have a separate `Definition` property, but I'm not sure what the benefit is. Is this an alternative to `DefinitionString` and `DefinitionSubstitutions`?

PR #1972, opened 2021-03-26
fix: Added SAR Support Check

CoshUS:
*Issue #, if available:*

*Description of changes:*
SAR does not provide service in af-south-1. SAM translator will throw a 5xx error instead of a 4xx.

*Checklist:*

- [ ] Write/update tests
- [ ] `make pr` passes
- [ ] Update documentation

*Examples?*

### DIFF
--- a/samtranslator/plugins/application/serverless_app_plugin.py
+++ b/samtranslator/plugins/application/serverless_app_plugin.py
@@ -12,6 +12,7 @@ from samtranslator.public.sdk.resource import SamResourceType
 from samtranslator.public.sdk.template import SamTemplate
 from samtranslator.intrinsics.resolver import IntrinsicsResolver
 from samtranslator.intrinsics.actions import FindInMapAction
+from samtranslator.region_configuration import RegionConfiguration
 
 LOG = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


wchengru (approved):

samtranslator/plugins/application/serverless_app_plugin.py
```diff
@@ -104,6 +105,10 @@ def on_before_transform_template(self, template_dict):
 
             if key not in self._applications:
                 try:
+                    if not RegionConfiguration().is_sar_supported():
```
aahung, line 108: Do we need to create a class instance to use class method?

samtranslator/region_configuration.py
```diff
@@ -22,3 +24,9 @@ def is_apigw_edge_configuration_supported(cls):
             "aws-iso-b",
             "aws-cn",
         ]
+
+    @classmethod
+    def is_sar_supported(cls):
+        return boto3.Session().region_name not in [
```
aahung, line 30: Can we have a docstring containing the URL in case `af-south-1` supports it we can quickly find the reference?

aahung (approved):
LGTM

samtranslator/plugins/application/serverless_app_plugin.py
mgrandis: Indeed, `is_apigw_edge_configuration_supported` is used like this `RegionConfiguration.is_apigw_edge_configuration_supported()`

CoshUS: Updated.

samtranslator/region_configuration.py
CoshUS: Added

PR #1975, opened 2021-03-29
fix the request parameter parsing, the value can contain dots

moelasmar:
*Issue #, if available:*
As per [this documentation](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-property-function-api.html#sam-function-api-requestparameters), Request Parameters names can contain dots.

fix the request parameter parsing, the value can contain dots
*Description of changes:*

*Checklist:*

- [X] Write/update tests
- [X] `make pr` passes
- [ ] Update documentation

*Examples?*

### DIFF
--- a/samtranslator/swagger/swagger.py
+++ b/samtranslator/swagger/swagger.py
@@ -1134,7 +1134,10 @@ class SwaggerEditor(object):
 
                 parameter_name = request_parameter["Name"]
                 location_name = parameter_name.replace("method.request.", "")
-                location, name = location_name.split(".")
+
+                parameter_name_parts = location_name.split(".")
+                location = parameter_name_parts[0]
+                name = ".".join(parameter_name_parts[1:])
 
                 if location == "querystring":
                     location = "query"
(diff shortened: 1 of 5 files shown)


samtranslator/swagger/swagger.py
```diff
@@ -1134,7 +1134,10 @@ def add_request_parameters_to_method(self, path, method_name, request_parameters
 
                 parameter_name = request_parameter["Name"]
                 location_name = parameter_name.replace("method.request.", "")
-                location, name = location_name.split(".")
+
+                parameter_name_parts = location_name.split(".")
+                location = parameter_name_parts[0]
+                name = ".".join(parameter_name_parts[1:])
```
aahung, line 1140: `split()` has a 2nd argument `maxsplit` which can be used here: https://docs.python.org/3/library/stdtypes.html#str.split

aahung (approved):

mgrandis (approved):

PR #1997, opened 2021-04-22
fix: mutable default values in method definitions

mndeveci:
*Checklist:*

- [x] Write/update tests
- [x] `make pr` passes
- [ ] Update documentation

*Examples?*


jfuss (approved):
Thanks for updating this. We should track getting pylint enabled in this repo. General linting could have caught some of this, but looks like we never put the effort into doing it in this repo :(

hoffa (approved):

sriram-mv (approved):

PR #1999, opened 2021-04-23
fix: Fail when Intrinsics are in SourceVPC list instead of IntrinsicsSourceVPC

jfuss:
*Issue #, if available:*
N/A

*Description of changes:*
When SAM is running and an Intrinsic is the value of SourceVPC* under the API Auth, SAM translation will fail. This is due to the code expecting SourceVPC* being a list of strings and operating on it with a regex, which eventually fails because you can't do a regex on a dictionary. SAM introduced a property called IntrinsicsSourceVPC to support this use-case but never guarded against them passed into SourceVPC.

This PR also removes unneeded if statements. We were doing a collection of `or` statements which means the functions we called within that block needed to handle `None`'s aways, so simplified the code paths to remove the unneeded checks.

* Ran tests locally
* Validated that before these changes, the templates tests added failed. 

*Checklist:*

- [ ] Write/update tests
- [ ] `make pr` passes
- [ ] Update documentation

*Examples?*

### DIFF
--- a/samtranslator/swagger/swagger.py
+++ b/samtranslator/swagger/swagger.py
@@ -877,6 +877,8 @@ class SwaggerEditor(object):
         ip_range_blacklist = resource_policy.get("IpRangeBlacklist")
         source_vpc_whitelist = resource_policy.get("SourceVpcWhitelist")
         source_vpc_blacklist = resource_policy.get("SourceVpcBlacklist")
+
+        # Intrinsic's supported in these properties
         source_vpc_intrinsic_whitelist = resource_policy.get("IntrinsicVpcWhitelist")
         source_vpce_intrinsic_whitelist = resource_policy.get("IntrinsicVpceWhitelist")
         source_vpc_intrinsic_blacklist = resource_policy.get("IntrinsicVpcBlacklist")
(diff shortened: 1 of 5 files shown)


samtranslator/swagger/swagger.py
```diff
@@ -898,31 +900,38 @@ def add_resource_policy(self, resource_policy, path, api_id, stage):
...
-            or (source_vpce_intrinsic_blacklist is not None)
-        ):
-            blacklist_dict = {
-                "StringEndpointList": source_vpc_blacklist,
-                "IntrinsicVpcList": source_vpc_intrinsic_blacklist,
-                "IntrinsicVpceList": source_vpce_intrinsic_blacklist,
-            }
-            resource_list = self._get_method_path_uri_list(path, api_id, stage)
-            self._add_vpc_resource_policy_for_method(blacklist_dict, "StringEquals", resource_list)
+        if source_vpc_blacklist is not None and not all(isinstance(x, string_types) for x in source_vpc_blacklist):
```
prenx4x, line 903: Does Inclusive language apply here in this source code? - https://w.amazon.com/bin/view/Inclusive_Language_Goal/

hoffa, line 903: Do we have any generic methods to check list types? I assume this could be a fairly common operation.

hoffa: In fact just noticed it's repeated underneath. Can we refactor that into its own function?

jfuss: @hoffa What's the proposal here? I don't really understand the need to make this into it's own function. To me that function will just return a bool and keep the if, so we can keep the context in one place. From that, it really doesn't make sense to me to drop it into a method but could be missing an alternative solution.

Ideally, this should be all in one place (the spec). Since we don't or that it's a much bigger change. I am proposing to keep this as is and stop the failures/crashing on SAM. We can later build this into the spec to solve it at the root.

hoffa: My point is that `if source_vpc_blacklist is not None and not all(isinstance(x, string_types) for x in source_vpc_blacklist)` is verbose and adds cognitive load to understand for what's a simple and generic check ("are the values in `source_vpc_blacklist` strings?").

My proposal is to have something like:

```python
def is_iterable_type(xs: Optional[Iterable], types: Tuple[Type]):
    if not xs:
        return False
    return all(isinstance(x, types) for x in xs)
```

And then replace those checks by:

```python
if not is_iterable_type(source_vpc_blacklist, string_types):
```

samtranslator/swagger/swagger.py
```diff
@@ -898,31 +900,38 @@ def add_resource_policy(self, resource_policy, path, api_id, stage):
...
+            raise InvalidDocumentException(
```
mgrandis, line 904: We should maybe raise an `InvalidResourceException` so that we don't exit `translate` and also list the other potential errors in the template.

mgrandis:
Code coverage is decreasing, we may want to add some tests in `tests/swagger/test_swagger.py` for the new exceptions

samtranslator/swagger/swagger.py
mgrandis: Could we maybe reuse `list_of` from `samtranslator/model/types.py`?

jfuss: @hoffa Fair enough.

@mgrandis We can but honestly not sure if that is the right thing. The `types.py` is used as part of the Spec validation. Ideally, these things should be caught before we get into the generation of swagger, but we aren't doing validation on the Values for these properties (no idea why that is). If we use `types.py` here, my concern is that we start blending the validation logic. It could be the right thing, just not sure if it is. I know you have lots of thoughts on the validation parts. Do you have concerns in mixing this? 

It would look something like:
```
list_validator = list_of(is_str())
if not list_validator(source_vpc_blacklist, should_raise=False):
    raise
```

samtranslator/swagger/swagger.py
jfuss: This is a good callout but I was just following how the rest of the class handles the errors. `InvalidDocumentException` is thrown. My guess after looking into it further is  `InvalidResourceException` needs a LogicalId and it doesn't seem like the swagger class/file current has access. I am not sure if there are other reasons for this either, just what I currently understand just looking at things.

Let me spend a little time on this to see how good or bad this will be. My initially thinking is to just keep it as that is what we currently do for the rest of the errors in this category but could be worth trying to get this updated to the `InvalidResourceException`.

mgrandis: Yes indeed, this class is not working with Logical IDs/resources.

Ideally we should throw an `InvalidResourcePolicyException` and catch it in the caller to raise the appropriate exception:
- `InvalidResourceException` in [api_generator.py](https://github.com/aws/serverless-application-model/blob/develop/samtranslator/model/api/api_generator.py#L588)
- `InvalidEventException` in [push.py](https://github.com/aws/serverless-application-model/blob/develop/samtranslator/model/eventsources/push.py#L749) and [events.py](https://github.com/aws/serverless-application-model/blob/develop/samtranslator/model/stepfunctions/events.py#L415)

jfuss: So I agree though this will lead to confusion to customers. The way SAM deals with APIs that are defined though the `Events` Property on a `AWS::Lambda::Function` is by first creating a [`AWS::Serverless::API` through a 'plugin'](https://github.com/aws/serverless-application-model/blob/develop/samtranslator/plugins/api/implicit_api_plugin.py). So by the time we hit this code path, there are a couple possible logical ids. We don't track the Logical Id(s) that produced this resource, so we would output the "default name logical id" we create. Customers do not have this resource (strictly) within their template so it will be very confusing. This could be why its an `InvalidDocumentException` (though I am just guessing as I don't really know).

tests/translator/output/error_api_invalid_source_vpc_blacklist.json
```diff
@@ -0,0 +1,3 @@
+{
+  "errorMessage": "Invalid Serverless Application Specification document. Number of errors found: 1. SourceVpcBlacklist must be a list of strings. Use IntrinsicVpcBlacklist instead for values that use Intrinsic Functions"
```
sriram-mv, line 2: ```Invalid Serverless Application Specification document``` is that a copy paste error?

wchengru (approved):

tests/translator/output/error_api_invalid_source_vpc_blacklist.json
sriram-mv: 👍🏽

sriram-mv (approved):

PR #2001, opened 2021-04-26
fix: Crash when using an invalid method in open api

jfuss:
When customers use auth and define an invalid method in the open api
definition, SAM would return a 'server error'. This was actually
due to SAM attempting to get the method from the path. If the method
was not a supported method and non-lowercase, SAM would attempt to fetch
the lower case method and crash with a KeyError. This PR addresses that
by checking for the valid methods supported.

*Issue #, if available:*

*Description of changes:*

*Checklist:*

- [ ] Write/update tests
- [ ] `make pr` passes
- [ ] Update documentation

*Examples?*

### DIFF
--- a/tests/translator/input/error_api_with_invalid_path_object.yaml
+++ b/tests/translator/input/error_api_with_invalid_path_object.yaml
@@ -1,15 +1,13 @@
 Globals:
   Api:
     Name: "some api"
-    CacheClusterEnabled: True
-    CacheClusterSize: "1.6"
+    Variables:
+      SomeVar: Value
     Auth:
       DefaultAuthorizer: MyCognitoAuth
       Authorizers:
         MyCognitoAuth:
           UserPoolArn: !GetAtt MyUserPool.Arn
-    Variables:
-      SomeVar: Value
 
 Resources:
   ImplicitApiFunction:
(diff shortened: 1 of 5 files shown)


tests/translator/input/error_api_with_invalid_path_object.yaml
```diff
@@ -0,0 +1,47 @@
...
+      Runtime: nodejs12.x
+
+  ExplicitApi:
+    Type: AWS::Serverless::Api
+    Properties:
+      StageName: SomeStage
+      DefinitionBody:
+        swagger: 2.0
+        paths:
+          "/a":
```
sriram-mv, line 27: the httpMethod and paths dont line up with the test itself.

sriram-mv (changes requested):

tests/translator/input/error_api_with_invalid_path_object.yaml
jfuss: I swear this worked locally. Good catch.

sriram-mv (approved):
🎉

mgrandis:
Coverage is decreasing, I think you need to add 1 test in `tests/swagger/test_swagger.py` to cover the new exception throwing.

mndeveci (approved):

jfuss:
@mgrandis Added.

Side note: I don't really understand why we have codecov on this repo. Our builds will fail if code coverage falls below 95%: https://github.com/aws/serverless-application-model/blob/develop/Makefile#L9 which will include coverage from all the "functional" tests in `test_translator.py`.

mgrandis (approved):

PR #2023, opened 2021-05-14
feat: Swagger api operation simple validation

elbayaaa:
*Issue #, if available:*

*Description of changes:*
The method `method_definition_has_integration` in `SwaggerEditor` class expects a `dict` object, it raise 5XX if it receives a String instead. This change is to raise 4XX error instead as this is an invalid User template.
Unit testing
*Checklist:*

- [x] Write/update tests
- [x] `make pr` passes
- [ ] Update documentation

*Examples?*


mgrandis (approved):

samtranslator/swagger/swagger.py
```diff
@@ -620,15 +624,15 @@ def set_path_default_apikey_required(self, path):
         :param string path: Path name
         """
 
+        valid_non_http_method_sections = ["parameters", "summary", "description", "$ref", "servers"]
```
jfuss, line 627: This seems to match the OpenAPI 3.0. Looking at references, this class seems to be specific to Rest APIs. Rest APIs support both [OpenAPI 2.0](https://swagger.io/specification/v2/) and 3.0: https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-export-api.html

I think we may need to think deeper about this and the recent revert: https://github.com/aws/serverless-application-model/pull/2021 as a swagger validation overhaul.

jfuss (changes requested):

samtranslator/swagger/swagger.py
elbayaaa: This is a block-list while #2021 was an allow-list. I mean, we know as a **fact** that `["parameters", "summary", "description", "$ref", "servers"]` are not API operations (HTTP methods)

jfuss: Maybe asking this a different way. If a customer is using OpenAPI 2.0 and specifies "servers" in the path object, should SAM fail with this being invalid? I know this just ignores the values and moves on, but not sure where we should fail this.

If SAM doesn't fail in the case above, then API Gateway should reject it. Which could be enough, just poking at this to see if we really need or want to do these deeper level of validation or not.

elbayaaa: Are you asking whether we should validate that the API definition complies with the OpenApi specification or not?
If so, I think this is not the responsibility of SAM. Here, we are trying to do our best to raise 4xx error instead of 5xx when applicable. So, if the API definition is invalid but it doesn't cause SAM to fail, I believe it shouldn't be a concern for SAM. Agree?

jfuss: I assumed this was in part validation we do.

Your reasoning makes sense to me.

jfuss (changes requested):
Change looks good. 

Can we add test that do the translator so we can build out our test coverage? https://github.com/aws/serverless-application-model/tree/develop/tests/translator/input

aahung (approved):

hoffa (approved):

samtranslator/swagger/swagger.py
```diff
@@ -620,15 +625,15 @@ def set_path_default_apikey_required(self, path):
...
         for method_name, _ in self.get_path(path).items():
-            # Excluding parameters section
-            if method_name == "parameters":
+            # Excluding non-http-method sections https://swagger.io/specification/
+            if method_name in valid_non_http_method_sections:
                 continue
 
             normalized_method_name = self._normalize_method_name(method_name)
             # It is possible that the method could have two definitions in a Fn::If block.
             for method_definition in self.get_method_contents(self.get_path(path)[normalized_method_name]):
```
mgrandis, line 636: We should use `method_name` instead of `normalized_method_name` here and at line 539.
If a customer defines the method as `Get` for example, we are trying to get the normalized version `get`of this and this will fail with a `KeyError` exception.

mgrandis (changes requested):

samtranslator/swagger/swagger.py
elbayaaa: There is a separate PR for that: https://github.com/aws/serverless-application-model/pull/2031/files

jfuss:
I think we already solved this in a different way (looking at the conflicts). Therefore closing this pr.

Issue #1403, opened 2020-01-21
Support Request Validator to the method on API Event


scarlier:
**Description:** Defining api model to required=true will not add the Request Validator to the method.

with following sam template and sam cli version 0.40

```
AWSTemplateFormatVersion : "2010-09-09"
Transform: "AWS::Serverless-2016-10-31"

Globals:
  Function:
    AutoPublishAlias: live
    Runtime: python3.8
    Handler: lambda_handler.handler

Resources:

  restGateway:
    Type: "AWS::Serverless::Api"
    Properties:
      StageName: default
      Models:
        SomeModel:
          type: object
          properties:
            keyname:
              type: string
          required:
            - keyname

  byIamcToken:
    Type: "AWS::Serverless::Function"
    Properties:
      CodeUri: src.zip
      Events:
        HttpGet:
          Type: Api
          Properties:
            Path: '/request-path'
            Method: post
            RestApiId: !Ref restGateway
            RequestModel:
              Model: SomeModel
              Required: true
```

**Observed result:**
Request Validator in method settings has value "NONE"

**Expected result:**
Request Validator in method settings has value "Validate body..."

praneetap:
Thanks for the feature request! We have passed this along to our product team for possible prioritization. Please +1 on the feature to help with prioritization.

brysontyrrell:
@praneetap this has been raised before. I opened issue #1232 and it was quickly closed in favor of #931

asyba:
+1

g-pelletier:
A temporary solution could be to add a partial DefinitionBody in your ApiGateway resource :
```      
   DefinitionBody:
        swagger: 2.0
        x-amazon-apigateway-request-validators:
          basic:
            validateRequestBody: true
            validateRequestParameters: true
        x-amazon-apigateway-request-validator: basic
        paths:
          /request-path:
            post:
              x-amazon-apigateway-integration:
                httpMethod: POST
                type: aws_proxy
                uri:
                  Fn::Sub: 'arn:aws:apigateway:${AWS::Region}:lambda:path/2015-03-31/functions/${byIamcToken.Arn}/invocations'
```
This will add a validator named basic and will add it to every function in your apigateway, thus will enforce validation of the body and parameters (tweak it as you need)

brysontyrrell:
[...]

PR #2026, opened 2021-05-14
feat: Add ValidateBody/ValidateParameters attribute to setup API model validators

Rondineli:
*Issue #, if available:* https://github.com/aws/serverless-application-model/issues/1403

*Description of changes:* Adding a request validator to the method when a model is defined and set required to True

Create a sam template like the following:
```bash
Resources:
  HtmlFunction:
    Properties:
      CodeUri: s3://sam-demo-bucket/member_portal.zip
      Handler: index.gethtml
      Runtime: nodejs12.x
      Events:
        GetHtml:
          Type: Api
          Properties:
            RestApiId: HtmlApi
            Path: /
            Method: get
            RequestModel:
              Model: User
              Required: true
              ValidateBody: true
              ValidateParameters: true

  HtmlApi:
    Type: AWS::Serverless::Api
    Properties:
      StageName: Prod
      Models:
        User:
          type: object
          properties:
            username:
              type: string
```
Then on the swagger definitions it should create a validation definition with the `path-method-validator` format and set it on the integration. Then on the console, we should see this flagged:
Screenshot from 2021-05-14 21-40-42

*Checklist:*

- [ x ] Write/update tests
- [ x ] `make pr` passes
- [ ] Update documentation
- [ x ] Verify transformed template deploys and application functions as expected

*Examples?*


gfechio:
+1

AllanOricil:
How would the model validator work with a GET? I thought we could use it only to validate body on POST and PATCH I never really passed a body to a GET so that is why I am asking.

Rondineli:
> @Rondineli How would the model validator work with a GET? I thought we could use it only to validate body on POST and PATCH I never really passed a body to a GET so that is why I am asking.

Exist a possibility of passing a body request trouh a get, yes: example:
```bash
curl -XGET https://<apigw id>.execute-api.us-east-1.amazonaws.com/test/ -H "Content-Type: application/json" -d '{"message": "foo"}'
```

get-method-request

Then on lambda event:
```bash
{'resource': '/', 'path': '/', 'httpMethod': 'GET', 'headers': {'accept': '*/*', 'content-type': 'application/json', ..., 'body': '{"message": "foo"}', 'isBase64Encoded': False}
```

If you define a model you can validate the body on those methods.

jfuss:
@Rondineli Unfortunately we cannot accept this in the current state. The `Required` key is not meant to also add parameter validation. It will only add the keys to make the model required. 

In order to not break existing customers (magically add validation into their API), this needs to be a separate property within the Spec. Are you still interesting in contributing this? We will want to understand the design of the property at a high level, could be done in a comment here, and then we can go with code changes once we agree on the Spec design.

Rondineli:
@jfuss Got it! Under the issue comments there are comments around maybe inside of the Model block, adding two new parameters: `ValidateBody` and `ValidateRequest` options, this way we can keep things as it is and add those 2 new fields to set the validators. I would be happy to change my PR to achieve that if you are ok :)

Rondineli:
would be something like:
```bash
Resources:
  HtmlFunction:
    Properties:
      CodeUri: s3://sam-demo-bucket/member_portal.zip
      Handler: index.gethtml
      Runtime: nodejs12.x
      Events:
        GetHtml:
          Type: Api
          Properties:
            RestApiId: HtmlApi
            Path: /
            Method: get
            RequestModel:
              Model: User
              Required: true
              ValidateBody: true
              ValidateRequest: true
```
if they are not present on the config, I don't add the validator.

jfuss:
@Rondineli 
```
...
RequestModel:
              Model: User
              Required: true
              ValidateBody: true
              ValidateRequest: true
```

That looks reasonable to me. I think that makes sense. It's possible to condense this into one property too.

```
RequestModel:
              Model: User
              Required: true
              Validate: # Accepts ValidateBody, ValidateRequest, NoValue
```

I think I like your a little better. Less conditional logic to handle and makes intrinsic a little cleaner.

Side note:
Super top of mind to me. SAM doesn't have good support for intrinsics. Mostly because we would have to reimplement CloudFormation's behavior and CloudFormation doesn't resolve intrinsics before SAM is invoked. So in either way, we will want to make sure we have tests around the case for a customer doing "Ref: Param" or "FindInMap: ...", etc. It's ok to explicitly not support intrinsics but we will have to do some basic checks to ensure thing work properly. If the values of the Properties are pass throughs to the underlying CloudFormation resources, that is best (as things will just work). This may impact implementation so trying to bring in up early.

Rondineli:
> @Rondineli
> 
> ```
> ...
> RequestModel:
>               Model: User
>               Required: true
>               ValidateBody: true
>               ValidateRequest: true
> ```
> 
> That looks reasonable to me. I think that makes sense. It's possible to condense this into one property too.
> 
> ```
> RequestModel:
>               Model: User
>               Required: true
>               Validate: # Accepts ValidateBody, ValidateRequest, NoValue
> ```
> 
> I think I like your a little better. Less conditional logic to handle and makes intrinsic a little cleaner.
> 
> Side note:
> Super top of mind to me. SAM doesn't have good support for intrinsics. Mostly because we would have to reimplement CloudFormation's behavior and CloudFormation doesn't resolve intrinsics before SAM is invoked. So in either way, we will want to make sure we have tests around the case for a customer doing "Ref: Param" or "FindInMap: ...", etc. It's ok to explicitly not support intrinsics but we will have to do some basic checks to ensure thing work properly. If the values of the Properties are pass throughs to the underlying CloudFormation resources, that is best (as things will just work). This may impact implementation so trying to bring in up early.

ok! Will make a PR with those 2 fields and to set only booleans, I like the idea of not having an intrinsics better as there is a way of setting up customised validators trough swagger body definitions. I think set those two fields (as the validators only supports those two anyways) the only case I would see set a custom validator is to change the name of it, and it can be done trough swagger definitions.

Ok, will change my PR to achieve this then! Thanks!

Rondineli:
@jfuss Is ready to a re-review. :)

codepay4999 (approved):

codepay4999 (commented):
would be great to have this reviewed and merged

jfuss:
@Rondineli Thanks for updating the PR. I haven't gotten a chance to review yet but on my list to do by the end of the week (6/4).

tests/translator/output/api_request_model.json
```diff
@@ -27,7 +18,16 @@
               }
             }
           ]
-        }
+        },
```
jfuss, line 21: Looks like IDE auto formatting. Can you revert these changes?

This really doesn't hurt anything, but we do use changes in "unrelated" outputs as an indication of possible regressions. Just makes sure we (as reviewers) are understanding the changes correctly and the PRs are organized to show all changes for the implementation.

tests/translator/input/api_request_model_with_validator.yaml
```diff
@@ -0,0 +1,85 @@
...
+            RestApiId: HtmlApi
+            Path: /mixin
+            Method: get
+            RequestModel:
+              Model: User
+              Required: true
+              ValidateBody: true
+              ValidateRequest: false
+
+  HtmlFunctionNotDefinedValidation:
```
jfuss, line 59: Can you add a function or path in one of the existing ones here that adds the following cases:
* only define ValidateBody
* only define ValidateRequest

samtranslator/swagger/swagger.py
```diff
@@ -781,6 +782,42 @@ def _set_method_apikey_handling(self, path, method_name, apikey_required):
...
+
+        :param string path: Path name
+        :param string method_name: Method name
+        :param bool validate_body: Add validator parameter on the body
+        :param bool validate_request: Validate request
+        """
+
+        normalized_method_name = self._normalize_method_name(method_name)
+        normalized_path_name = SwaggerEditor.get_path_name_normalized(path)
+        validator_name = "{normalized_path_name}-{normalized_method_name}-validator".format(
```
jfuss, line 797: So looking at this now, I wonder if this is the right design for the spec. 

Thought process:
This will add N number of validators to the `x-amazon-apigateway-request-validators` part of the doc. But in reality there are only 3 combinations to `x-amazon-apigateway-request-validators`: both request body and request param, only request body, only request param. Sure it is possible for API Gateway to add other validators here but my assumption is that it is still finite and small. This leads to larger template files and OpenApi definitions when it isn't needed. I wonder if we should have a way to define a set of "validator combinations" instead. 

Proposal/discussion:
I think all the main logic is mostly the same, but if we agree deduping is better maybe what we really need to do here is:
* Add a dict of Validators to the `AWS::Serverless::Api` (and maybe Http?) that accepts a dict of key to validators. Bascially this dict of key values would be added directly to `x-amazon-apigateway-request-validators` in the OpenApi Definition.
* This new property would be added to Globals to support not having to write a `AWS::Serverless::Api` resources at all and let SAM generate this from the API Events in `AWS::Serverless::Function` (implicit API).
* Add a `ValidatorName` type property that will be used to specify in the API Event to key into the new property added above.

I think this might better align with what API Gateway already defines in OpenAPI Definitions. Happy to have an open dialog about this and sorry this is happening now vs catching it during the proposal. I didn't realize API Gateway defined it in this manner, which is my bad.

samtranslator/model/eventsources/push.py
```diff
@@ -788,6 +788,35 @@ def _add_swagger_integration(self, api, function, intrinsics_resolver):
...
+                if not isinstance(validate_body, type(None)) or not isinstance(validate_request, type(None)):
+
+                    # as we are setting two different fields we are here setting as default False
+                    # In case one of them are not defined
+                    validate_body = False if type(validate_body) is None else validate_body
+                    validate_request = False if type(validate_request) is None else validate_request
+
+                    # If not type None but any other type it should explicitly invalidate the Spec
+                    # Those fields should be only a boolean
+                    if not isinstance(validate_body, bool) or not isinstance(validate_request, bool):
```
jfuss, line 804: I think this will fail if `validate_body` or `validate_request` is set to a more complex intrinsic like an If. I made a comment in the tests about this, so that should prove what happens one way or another.

samtranslator/model/eventsources/push.py
```diff
@@ -788,6 +788,35 @@ def _add_swagger_integration(self, api, function, intrinsics_resolver):
                     path=self.Path, method_name=self.Method, request_model=self.RequestModel
                 )
 
+                validate_body = self.RequestModel.get("ValidateBody")
```
jfuss, line 791: You may still need to do a `None` check but if you default this to `false`. I can't think if there is a way to get `None` if you default (probably because it is late). Though you should be able to prove that through testing. Could simply the below logic.

You can test for both false then and not add any validators. Maybe that is confusing if you want to audit the OpenApi Definition later and the validators disappear even if they were both set to `false`. So maybe what you have is better anyways. Sorry thinking while I type, but leaving it to strike discussion.

jfuss (changes requested):
Overall, this looks really good. I (and the rest of the team) appreciate the time you spend so far on this and the level of detail you have in adding intrinsic related tests! 

There is really only one "larger" comment about the spec. I noticed it only after reading the code and doing more digging on API Gateway. Please let me know what you think about it.

samtranslator/swagger/swagger.py
Rondineli: Hi @jfuss I actually thought in the way let it explicitly to the path/method has a validator, but haven't thought about the small limit of combinations (sorry the correction, but would be 4 combinations). I think, maybe defining a pre-defined combination on a constant definition, and add only when called, ie:

FULL = (request, body)
BODY = (body, !request)
REQUEST = (!body, request)
NO_VALIDATE =(!request, !body)

If not defined
NONE = ()

Each time we have a iteration, I would try to get which one is being requested then add it to the document. A second time, if the definition already exist I just re-use it don't add a new one. If is not defined, I do not add it to the spec.

Tbh, I think leaving those 2 options simple as a boolean would be easier to understand and to use. Let me know what you think about it!

Rondineli:
> Overall, this looks really good. I (and the rest of the team) appreciate the time you spend so far on this and the level of detail you have in adding intrinsic related tests!
> 
> There is really only one "larger" comment about the spec. I noticed it only after reading the code and doing more digging on API Gateway. Please let me know what you think about it.

Hey, thanks for your review! I appreciate the comments and will work to fix each one of them. The larger comment I added a comment about maybe one better fix (quicker and still having it on the path/method level using a simple boolean), tbh I think that still the better way to go (at least for a first implementation). Let me know your thoughts anyway I would be happy to fix/do it :)

samtranslator/swagger/swagger.py
```diff
@@ -781,6 +782,42 @@ def _set_method_apikey_handling(self, path, method_name, apikey_required):
...
+        )
+
+        # Creating validator
+        request_validator_definition = {
+            validator_name: {"validateRequestBody": validate_body, "validateRequestParameters": validate_request}
+        }
+        if self._doc.get(self._X_APIGW_REQUEST_VALIDATOR):
+            self._doc[self._X_APIGW_REQUEST_VALIDATOR].update(request_validator_definition)
+        else:
+            self._doc[self._X_APIGW_REQUEST_VALIDATOR] = request_validator_definition
```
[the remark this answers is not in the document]
Rondineli: Done!

tests/translator/input/error_api_request_model_with_intrinsics_validator.yaml
```diff
@@ -0,0 +1,44 @@
...
+        GetHtml:
+          Type: Api
+          Properties:
+            RestApiId: HtmlApi
+            Path: /
+            Method: get
+            RequestModel:
+              Model: User
+              Required: true
+              ValidateBody: !Ref validateRequestBody
```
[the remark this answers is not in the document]
Rondineli: Done!

tests/translator/input/error_api_request_model_with_intrinsics_validator.yaml
```diff
@@ -0,0 +1,44 @@
+Parameters:
+  validateRequestBody:
+    Type: String
+  validateRequestParameters:
+    Type: String
+
+Resources:
+  RequestValidator:
```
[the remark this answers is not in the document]
Rondineli: Done! Deleted it from the definition!

tests/translator/output/api_request_model.json
Rondineli: Removed it from the git, sorry, didn't realize it had changed it.

tests/translator/input/api_request_model_with_validator.yaml
Rondineli: Added those two cases for each scenario, added only a body for true and false and same for request

samtranslator/swagger/swagger.py
Rondineli: So, about this issue, I have changed my PR to that limit of combinations. Where if the spec has been defined already, I don't add a new one with a diff name instead just add what is already defined. I think that makes more sense, would simplify implementations and let easier to the client point of view as adding it to the validatorName is already possible doing Body Definitions, which is a bit more complex than only add those 2 parameters.

I added a testcase where If the combination is repeated I am not adding a new one, but re-using an existent one already defined on the swagger doc.

Let me know your thinking around that.

samtranslator/model/eventsources/push.py
Rondineli: I added an If case, same behavior would be expected as setting an intrinsic definition.

samtranslator/model/eventsources/push.py
Rondineli: Yes. My first thought is when we want explicitly False. If is not defined I am treating it as a None, but if it is False, then I would consider it as a defined field and even if it is False *I think* we should set it, as it will be explicitly on the configs saying we don't want validate anything.

Rondineli:
@jfuss Sorry this being on a Saturday other than during the week, but I am anxious to try solve this (with all your concerns of course and all of them are valid). 

I added on my PR a solution where we are setting only a max of combinations for those 2 fields instead of pollute the spec with all those repeated validators names. Instead, I am setting pre-defined names and re-using them if the combination is already defined on the swagger doc, if they are not, then I will add it. Let me know your thoughts on that approach.

samtranslator/swagger/swagger.py
jfuss: @rohandalvi You are correct, there are 4 combinations. I ignored the `NO_VALIDATE` case and assumed having `False` set to both Validates was equivalent. I can see value in having this still to allow customers the ability to audit the template and ensure all paths have this set.

What you have/proposed does work, though my main concern is around keeping this updated in the future. SAM becomes the bottle neck in this solution. While this might not ever happen, if API Gateway adds another "validate" property, SAM needs to find a way to handle this. We could keep the updating/adding these pre-defined configuration but SAM will always be behind and we risk creating a more confusing Spec. For example `FULL` would need to be expanded have any of the "new validation" which would be a breaking change for SAM (an updated will change the behavior of customer's template without any new being added in the Spec). Now, we could not do that but then it's hard to tell what `FULL` is really for. This will also exponentially increase the number of combinations to support, which is fine in small numbers but could become large as time goes on.

Given that, my opinion right now is to enable this is a way that SAM can be out of the way in the future. I like the simplicity of your solution but the possible maintenance is what I fear. Let me ponder this a little more and bring it up with my team to see what we collectively think. I think this comes down to a balancing act between simplicity and longevity/maintenance.

samtranslator/model/eventsources/push.py
```diff
@@ -788,6 +788,35 @@ def _add_swagger_integration(self, api, function, intrinsics_resolver):
...
+                validate_request = self.RequestModel.get("ValidateRequest")
+
+                # Checking if any of the fields are defined as it can be false we are checking if the field are not None
+                if not isinstance(validate_body, type(None)) or not isinstance(validate_request, type(None)):
+
+                    # as we are setting two different fields we are here setting as default False
+                    # In case one of them are not defined
+                    validate_body = False if type(validate_body) is type(None) else validate_body
```
mgrandis, line 799: Same here and below, you can use `validate_body is None`.

tests/swagger/test_swagger.py
```diff
@@ -831,6 +831,108 @@ def test_must_add_body_parameter_to_method_openapi_required_true(self):
...
+
+    @parameterized.expand(
+        [
+            param(True, False, "BODY"),
+            param(True, True, "FULL"),
+            param(False, True, "REQUEST"),
+            param(False, False, "NO_VALIDATE"),
+        ]
+    )
+    def test_must_normalize_validator_names(self, validate_body, validate_request, normalized_name):
```
mgrandis, line 920: You should maybe replace the `normalize` name with `return` here, no normalization is done.

mgrandis (changes requested):
Thanks @Rondineli for the PR and also, seconding Jacob, the extensive tests with intrinsics/odd values!

Added a few remarks.

samtranslator/swagger/swagger.py
```diff
@@ -781,6 +782,41 @@ def _set_method_apikey_handling(self, path, method_name, apikey_required):
...
+        }
+        if not self._doc.get(self._X_APIGW_REQUEST_VALIDATOR):
+            self._doc[self._X_APIGW_REQUEST_VALIDATOR] = {}
+
+        if not self._doc[self._X_APIGW_REQUEST_VALIDATOR].get(validator_name):
+            # Adding only if the validator hasn't been defined already
+            self._doc[self._X_APIGW_REQUEST_VALIDATOR].update(request_validator_definition)
+
+        # It is possible that the method could have two definitions in a Fn::If block.
+        for method_definition in self.get_method_contents(self.get_path(path)[normalized_method_name]):
```
qingchm, line 810: Question here: would the swagger doc ever cause self.get_path(path)[normalized_method_name] to error out? Any checks needed?

tests/swagger/test_swagger.py
Rondineli: It seems it was left from the first review, nice catch 👍

samtranslator/model/eventsources/push.py
Rondineli: Done, sometimes being fancy is not good idea, tbh I always use `is None`. I think due the size of this project I tried to be too much fancy 😅. Jokes aside, fixed!

samtranslator/swagger/swagger.py
```diff
@@ -1265,6 +1301,26 @@ def get_path_without_trailing_slash(path):
...
+    def get_validator_name(validate_body, validate_request):
+        """
+        Get a readable path name to use as validator name
+
+        :param boolean validate_body: Boolean if validate body
+        :param boolean validate_request: Boolean if validate request
+        :return string: Normalized validator name
+        """
+        if validate_body and validate_request:
+            return "FULL"
```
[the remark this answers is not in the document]
Rondineli: I changed the names. (Namimg things are hard). I did on uppercase because I saw somewhere on the comments it being uppercase so I stick with ir. But all docs refer as lowercase you are right, fixed it!

samtranslator/swagger/swagger.py
```diff
@@ -23,6 +23,7 @@ class SwaggerEditor(object):
     _X_APIGW_GATEWAY_RESPONSES = "x-amazon-apigateway-gateway-responses"
     _X_APIGW_POLICY = "x-amazon-apigateway-policy"
     _X_ANY_METHOD = "x-amazon-apigateway-any-method"
+    _X_APIGW_REQUEST_VALIDATOR = "x-amazon-apigateway-request-validators"
```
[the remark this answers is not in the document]
Rondineli: You like details! I liked you! Fixed!

samtranslator/model/eventsources/push.py
```diff
@@ -788,6 +788,35 @@ def _add_swagger_integration(self, api, function, intrinsics_resolver):
                     path=self.Path, method_name=self.Method, request_model=self.RequestModel
                 )
 
+                validate_body = self.RequestModel.get("ValidateBody")
+                validate_request = self.RequestModel.get("ValidateRequest")
```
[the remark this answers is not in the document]
Rondineli: That makes sense. Fixed!

Rondineli:
> Thanks @Rondineli for the PR and also, seconding Jacob, the extensive tests with intrinsics/odd values!
> 
> Added a few remarks.

Thanks for reviewing this and pointing out the improvements. I fixed every comment!

samtranslator/swagger/swagger.py
Rondineli: Hi @qingchm tbh if that happen w9uld happen way before than set the validators. As the models and other methods setting apigw methods is doing it and that validations would happen before. I am not on laptop right now but I can chek on the code where it is being checked.

Rondineli: Hi @qingchm Sorry getting back to you only now. So, looking at this, the only way of this fails is if the Method is not defined, and it will fail waay before as I said, once per [this line](https://github.com/aws/serverless-application-model/blob/90edf3e1304f79bb5da12860e6fe6e6c776ff5a5/samtranslator/model/stepfunctions/events.py#L236) shows the model as required. The normalized method get is expect only a small conversion and the [get_path method](https://github.com/aws/serverless-application-model/blob/90edf3e1304f79bb5da12860e6fe6e6c776ff5a5/samtranslator/swagger/swagger.py#L52) would return its path dict (where the Method is required). :)

tking2096:
On the subject of "adding N validators" and not knowing what validation API Gateway might add in the future:

Why not at just add RequestValidatorId to the event and let the user reference a validator that's defined in the template? This way, if API Gateway adds to the RequestValidator's capabilities, SAM doesn't have to worry about keeping up. As you mentioned, there are limited combinations so the user would at most need to define 3 validator resources. What's the point of auto-generating the validator(s)?

samtranslator/swagger/swagger.py
```diff
@@ -781,6 +783,41 @@ def _set_method_apikey_handling(self, path, method_name, apikey_required):
...
+        }
+        if not self._doc.get(self._X_APIGW_REQUEST_VALIDATORS):
+            self._doc[self._X_APIGW_REQUEST_VALIDATORS] = {}
+
+        if not self._doc[self._X_APIGW_REQUEST_VALIDATORS].get(validator_name):
+            # Adding only if the validator hasn't been defined already
+            self._doc[self._X_APIGW_REQUEST_VALIDATORS].update(request_validator_definition)
+
+        # It is possible that the method could have two definitions in a Fn::If block.
+        for method_definition in self.get_method_contents(self.get_path(path)[normalized_method_name]):
```
jfuss, line 811: We have been moving away from this kind of access pattern and moving towards what is in this pr: https://github.com/aws/serverless-application-model/pull/2031/files

The reason is the normalization causes `KeyErrors` when methods like "any" are used, since that will get translated to `x-amazon-apigateway-any-method`. Can you match what is in the PR I linked? `for method_definition in self.get_method_contents(method):`

jfuss (changes requested):
@Rondineli I appreciate the patient from our end and working through everything. I have one additional comment (based upon some other changes going on). If you could address that, I think we will be in really good shape to get this approved on our side and work through our release process (doc updates, service updates, etc).

samtranslator/swagger/swagger.py
Rondineli: @jfuss done!

mgrandis (approved):

samtranslator/swagger/swagger.py
Rondineli: @jfuss thinking a bit more about this last change, it doesn't seems right to me:
```bash
for method_name, method in self.get_path(path).items():
            for method_definition in self.get_method_contents(method):
                # If no integration given, then we don't need to process this
```

So, it would be something like:
```bash
        for method_name, method in self.get_path(path).items():
            # Adding it to only path  
            if method_name == normalized_method_name:
                for method_definition in self.get_method_contents(method):
```
right?

Rondineli: without checking the method_name with the normalized_method_name it would add this to all methods for the given path.

samtranslator/swagger/swagger.py
```diff
@@ -781,6 +783,46 @@ def _set_method_apikey_handling(self, path, method_name, apikey_required):
...
+        if not self._doc[self._X_APIGW_REQUEST_VALIDATORS].get(validator_name):
+            # Adding only if the validator hasn't been defined already
+            self._doc[self._X_APIGW_REQUEST_VALIDATORS].update(request_validator_definition)
+
+        # It is possible that the method could have two definitions in a Fn::If block.
+        # for method_definition in self.get_method_contents(self.get_path(path)[normalized_method_name]):
+        for method_name, method in self.get_path(path).items():
+
+            # Adding it to only given method to the path
+            if method_name == normalized_method_name:
```
mgrandis, line 815: We may have an issue if the `method_name` is not normalized (in the template).

Also, this name is conflicting with the function's parameter `method_name` so I think we should find another name.

Rondineli: ok, just pushed what I think make more sense now. Normalizing the method_name that comes from the template before check if the method definition is the one we want to add the validator.

samtranslator/swagger/swagger.py
jfuss: Seems reasonable to me. I am more comfortable with this since it is a new code path (unlikely to accidentally break existing customers for something we are unaware of).

jfuss (approved):

jfuss:
@Rondineli Thank you for all the effort and time spent on the contribution. It is very much appreciated. 

As for next steps: I need to kick of some processes internally (docs for example). I will eventually merge this into develop (later this week). Just giving you the heads up and communicate no further action on you.

ChristianET-DS:
this will be only for the requestmodel or also for requestparameter ?

Rondineli:
> this will be only for the requestmodel or also for requestparameter ?

ValidateRequestBody and validateRequestParams:

```bash

ValidateBody: true
              ValidateParameters: true
```

AllanOricil:
Congratulations @Rondineli 
You did an amazing work.
Im looking forward to test it. How can I use the CLI using this aws:develop branch? Is there a tutorial somewhere that helps me to build the cli from the develop branch?

iRoachie:
Hey guys, if I understand this code correctly, no validator will be created if `Model` is not set correct?

Does this mean in a scenario where I just want to validate if a user is passing in a query param that I have to create a "fake" model anyway to create this?

e.g

```yaml
Properties:
  Method: get
  RequestParameters:
    - method.request.querystring.query:
        Required: true
  RequestModel:
    ValidateParameters: true
```

^ This creates the parameter but no validator.

Screenshot 2021-10-16 at 10 07 51 PM

If I create a dummy model, however, then it works. Is this something that should be added (open a new issue)? Just trying to gauge if it was the intention of this PR to support this.

iRoachie:
@Rondineli in the console or using cdk (which I ended up using), no model is needed to validate the presence of a query string parameter.

I should be able to perform the same in sam but the implementation of this PR requires a model to be created.

stevehogdahl:
I have a question regarding the implementation of Models. Our architecture uses the implicitly created AWS::Serverless::Api resource rather than creating the resource manually and adding every Function (70-80) to it. I'm trying to create a AWS::ApiGateway::Model resource and set the RestApiId to the implicit resource (ServerlessRestApi). This fails the sam validation with the following error "the related API does not contain valid Models". Is the only way to implement this to explicitly create a AWS::Serverless::Api resource, add the models, then add everything to the explicit resource rather than using the implicitly created one.

santiperone:
> https://github.com/aws/serverless-application-model/releases/tag/v1.39.0



> > Hey guys, if I understand this code correctly, no validator will be created if `Model` is not set correct?
> > Does this mean in a scenario where I just want to validate if a user is passing in a query param that I have to create a "fake" model anyway to create this?
> > e.g
> > ```yaml
> > Properties:
> >   Method: get
> >   RequestParameters:
> >     - method.request.querystring.query:
> >         Required: true
> >   RequestModel:
> >     ValidateParameters: true
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ^ This creates the parameter but no validator.
> > Screenshot 2021-10-16 at 10 07 51 PM
> > If I create a dummy model, however, then it works. Is this something that should be added (open a new issue)? Just trying to gauge if it was the intention of this PR to support this.
> 
> the validators supposed to work with models, the model is not "dummy", it is where you will set the type of data you are validating, ie: query is part of a querystring, but is it a number, string? The model needs to exist to then set a validator.

I'm having the same issue as @iRoachie. On a GET method you typically want to validate only requestParameters, not RequestModel. Nevertheless you have to define a Model in orther to activate de parameter validation. Unless you can somehow use a Model to define querystringparameters, which is unclear to me.

PR #2027, opened 2021-05-14
fix: Correct grammar in the waiting for changeset message

skyzyx:
*Description of changes:*

There is no such punctuation as `..`. There is a period (`.`). There is an ellipsis (`…` or `…`).

This corrects the grammar in the line to use punctuation which exists.


aahung (approved):
Thanks for the PR

mgrandis (approved):

PR #2031, opened 2021-05-16
fix: Don't attempt to refetch a swagger method object by its name as we already have it

elbayaaa:
*Issue #, if available:*
N/A

*Description of changes:*
[here](https://github.com/aws/serverless-application-model/blob/develop/samtranslator/swagger/swagger.py#L628) the method_name, which is a [key](https://github.com/aws/serverless-application-model/blob/develop/samtranslator/swagger/swagger.py#L623) in the path hash, is normalized (may be modified) then get [fetched](https://github.com/aws/serverless-application-model/blob/develop/samtranslator/swagger/swagger.py#L630) again from the same hash which will result in KeyError if the normalization modified the key

Not normalizing the key. It looks like the logic was just a copy/paste from another place where the normalization was needed. but here it is not needed at all.

*Checklist:*

- [ ] Write/update tests
- [x] `make pr` passes
- [ ] Update documentation

*Examples?*

### DIFF
--- a/samtranslator/open_api/open_api.py
+++ b/samtranslator/open_api/open_api.py
@@ -352,7 +352,7 @@ class OpenApiEditor(object):
                             )
                         ]
                     )
-                for method_definition in self.get_method_contents(self.get_path(path)[normalized_method_name]):
+                for method_definition in self.get_method_contents(method):
                     # If no integration given, then we don't need to process this definition (could be AWS::NoValue)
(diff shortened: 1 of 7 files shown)


mgrandis:
You may want to update line 534 too

mgrandis:
Added a fix for the same code in `samtranslator/open_api/open_api.py`

mgrandis (approved):

moelasmar (approved):

PR #2032, opened 2021-05-17
fear: Api validation with a new schema

mgrandis:
*Issue #, if available:*

*Description of changes:*
Update the `schema.json` which is not currently used (its result is not used) to validate an Api.
For now a `schema_new.json` has been added but will be removed as this PR is reviewed.
This may need individual passing tests for each property.

YAML templates.

*Checklist:*

- [x] Write/update tests
- [x] `make pr` passes
- [ ] Update documentation

*Examples?*


samtranslator/validator/validator.py
```diff
@@ -7,49 +9,236 @@
...
+
+        Parameters
+        ----------
+        schema_path : str, optional
+            Path to a schema to use for validation, by default None, the default schema.json will be used
+        """
+        if not schema_path:
+            schema_content = self._read_default_schema()
+        else:
+            schema_content = self._read_json(schema_path)
```
jfuss, line 34: Suggestion:

```
if schema_path is None:
   schema_path = "default path"

schema_content = self._read_json(schema_path)
```

Not sure if that works, but I like minimizing control statements where possible.

tests/translator/input/basic_function.yaml
```diff
@@ -50,37 +50,37 @@ Resources:
       Timeout: 60
       VpcConfig:
         SecurityGroupIds:
-        - sg-edcd9784
+          - sg-edcd9784
```
jfuss, line 53: If we get the previous version of this, will the validation fail?

jfuss (changes requested):
Giving some early feedback. Still want to spend some more time understanding things but wanted to at least share initial questions.

samtranslator/validator/sam_schema/definitions/intrinsics.json
```diff
@@ -0,0 +1,261 @@
+{
```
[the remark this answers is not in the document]
mgrandis: Handcrafted but an experiment for now, we might not want it for now.
The `if` / `else` part is added to only report the errors from this specific definition for example.

Not sure we should validate Intrinsic functions for now, just now that something is an intrinsic (through the `intrinsic` type) and allow/disallow it.

samtranslator/validator/validator.py
```diff
@@ -7,49 +9,236 @@
 
 
 class SamTemplateValidator(object):
-    @staticmethod
```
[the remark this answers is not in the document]
mgrandis: We could yes, calling the new one behind the scenes.

samtranslator/validator/validator.py
mgrandis: Yes!
Don't remember why I went with this form, I was maybe reusing `_read_default_schema` elsewhere before.

tests/translator/input/basic_function.yaml
mgrandis: It passes, just tested.
This was changed by the VSCode save auto-format.

samtranslator/parser/parser.py
```diff
@@ -55,4 +55,5 @@ def _validate(self, sam_template, parameter_values):
                     ]
                 )
 
-        SamTemplateValidator.validate(sam_template)
+        validator = SamTemplateValidator()
+        validation_errors = validator.validate(sam_template)
```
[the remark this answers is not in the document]
mgrandis: Added a `LOG.warn()`.
This will also be displayed when calling `sam validate` as the parser is called by the [SamTemplateValidator from the CLI](https://github.com/aws/aws-sam-cli/blob/develop/samcli/commands/validate/lib/sam_template_validator.py#L68).

This is how it looks when adding an Api with an empty `StageName` to a template and calling `sam validate`:
```
>> samdev validate
2021-06-29 16:28:42 Loading policies from IAM...
2021-06-29 16:28:46 Finished loading policies from IAM.
2021-06-29 16:28:46 Template schema validation reported the following errors: [Resources/MyApi/Properties/StageName] Must not be empty
Template provided at '/path/to/template.yaml' was invalid SAM Template.
Error: [InvalidResourceException('MyApi', "Missing required property 'StageName'.")] ('MyApi', "Missing required property 'StageName'.")
```

samtranslator/validator/sam_schema/definitions/intrinsics.json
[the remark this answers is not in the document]
jfuss: Got it. I think the thing we will need to be careful about is failing customers unexpectedly. The risk we will take in validating everything is if CFN updates or maybe a customer has some custom transform/custom resource that does something slightly differently than the defaults. Probably just overly cautious honestly, but just trying to think of ways this could hurt us.

samtranslator/validator/validator.py
```diff
@@ -7,49 +9,254 @@
...
 
 class SamTemplateValidator(object):
+    """
+    SAM template validator
+    """
+
+    # Useful to find a unicode prefixed string type to replace it with the non unicode version
+    # On Python2, the validator returns types in error messages prefixed with 'u'
+    # We remove them to be consistent between python versions
+    # Example: "u'integer'" -> "'integer'"
```
jfuss, line 19: @hawflau Could be worth a look here. I know you are looking into Python2 removal stuff, so want to flag this to you. Not sure if this will be an issue at all with some of the hashing work you are doing between python2 and python3.

samtranslator/validator/validator.py
```diff
@@ -7,49 +9,254 @@
...
+        # Otherwise the instance would call the static one
+        # Static one is kept for backward compatibility
+        self.validate = self._validate
+
+        if not schema:
+            schema = self._read_json(sam_schema.SCHEMA_NEW_FILE)
+
+        # Helps resolve the $Ref to external files
+        # schema content must be passed to resolve local (#/def...) references
+        resolver = jsonschema.RefResolver("file://" + sam_schema.SCHEMA_DIR + "/", schema)
```
jfuss, line 41: Will this work on windows? I see a "/", which might cause problems for SAM CLI?

samtranslator/validator/validator.py
```diff
@@ -7,49 +9,254 @@
...
+        if error is None:
+            return
 
-    @staticmethod
-    def _read_json(filepath):
+        if not error.context:
+            # We only display the leaves
+            # Format the message:
+            # [/Path/To/Element] Error message
+            error_path = "/".join([str(p) for p in error.absolute_path]) if error.absolute_path else "/"
```
jfuss, line 129: Should this be `.` instead of `/`? I assuming this is for paths in the template? I think its more consistent to use `.` instead of `/`

jfuss (changes requested):

samtranslator/validator/sam_schema/definitions/api.json
```diff
@@ -0,0 +1,789 @@
...
+            "$ref": "file:common.json#definitions/DependsOn"
+          },
+          "Metadata": {
+            "type": "object"
+          },
+          "Properties": {
+            "additionalProperties": false,
+            "properties": {
+              "AccessLogSetting": {
+                "$ref": "file:cfn.json#definitions/AWS::ApiGateway::Stage.AccessLogSetting"
```
CoshUS, line 33: Do we want to handle CFN objects?
Are we able to use [CFN schemas](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/resource-type-schemas.html) and link those fields to here with pointers?

samtranslator/validator/sam_schema/definitions/api.json
```diff
@@ -0,0 +1,789 @@
...
+              },
+              "Name": {
+                "type": [
+                  "string",
+                  "intrinsic"
+                ]
+              },
+              "OpenApiVersion": {
+                "type": "string",
+                "pattern": "^[2-3](\\.\\d)(\\.\\d)?$",
```
CoshUS, line 147: Should this be `^(2\.0|3\.\d\.\d)?$` to only match 2.0 and 3.x.x instead of 2.x.x and 3.x.x?

samtranslator/validator/sam_schema/definitions/api.json
```diff
@@ -0,0 +1,789 @@
...
+                "$ref": "file:cfn.json#definitions/AWS::ApiGateway::Stage.AccessLogSetting"
+              },
+              "Auth": {
+                "$ref": "#definitions/AWS::Serverless::Api.Auth"
+              },
+              "BinaryMediaTypes": {
+                "items": {
+                  "type": [
+                    "string",
+                    "intrinsic"
```
CoshUS, line 42: General question on the type "intrinsic". How are we determining what can be intrinsic or not? Is this done by testing each field manually or are we including all fields that gets directly passed to CFN?

samtranslator/validator/sam_schema/definitions/common.json
```diff
@@ -0,0 +1,180 @@
...
+    "UpdatePolicy": {
+      "enum": [
+        "Delete",
+        "Retain",
+        "Snapshot"
+      ],
+      "type": "string"
+    },
+    "UpdateReplacePolicy": {
+      "type": "string"
```
CoshUS, line 177: Should this also have enum?

samtranslator/validator/validator.py
CoshUS: For URIs, "/" works on Windows, but we should verify it nonetheless.

samtranslator/validator/validator.py
```diff
@@ -7,49 +9,254 @@
...
+
+    Returns
+    -------
+    [type]
+        [description]
+    """
+    return isinstance(instance, dict) and has_intrinsic_attr(instance)
+
+
+def has_intrinsic_attr(dict):
```
CoshUS, line 247: PyLint is probably going to complain overriding built-in dict. Even if it does not, another name might be better.

CoshUS (changes requested):
First round of review. Added a few comments and I'll dig a bit deeper later.

samtranslator/validator/sam_schema/definitions/api.json
mgrandis: I'm using these files as a base for building the schema:
- https://raw.githubusercontent.com/awslabs/goformation/master/schema/sam.schema.json
- https://raw.githubusercontent.com/awslabs/goformation/master/schema/cloudformation.schema.json

That's where the names come from.
Maybe we could drop the `cfn.json` file, rename these common properties and put them in `common.json`.
`AWS::ApiGateway::Stage.AccessLogSetting` would become `AccessLogSetting` for example.

samtranslator/validator/sam_schema/definitions/api.json
mgrandis: All fields that we directly pass to CFN could receive an intrinsic.
We also resolve some intrinsic `Ref` so these should also be validated.
If we need the value (we transform a list of strings to a list of objects for example), we currently don't support intrinsic for these properties.

samtranslator/validator/sam_schema/definitions/api.json
mgrandis: I think I was initially thinking of putting in this file the definitions for which we directly link to the Cloud Formation documentation in our SAM spec doc.

samtranslator/validator/validator.py
```diff
@@ -1,55 +1,262 @@
...
+
+        Returns
+        -------
+        list[str]
+            List of validation errors
         """
-        Reads the JSON Schema at given file path
+        validator = SamTemplateValidator(schema)
+
+        return ", ".join(validator.validate(template_dict))
```
jfuss, line 74: Should this be `validator._validate(template_dict)`? I don't see a `validate` method that isn't static here. If there is and I just don't see it (due to not enough coffee this morning), I would suggest making the static and instance method different names to help the reader not to think this is an infinite loop in calling itself.

tests/translator/input/unsupported_resources.yaml
```diff
@@ -7,8 +7,8 @@ Resources:
       StageName: Prod
       DefinitionUri: s3://sam-demo-bucket/webpage_swagger.json
 
-  InvalidResourceTypeShouldBeAccepted:
-    Type: AWS::Serverless::ABCD
+  UnknownTypeShouldBeAccepted:
+    Type: AWS::Other::ABCD
```
jfuss, line 11: Should we have tests for both (`AWS::Other::ABCD` and `AWS::Serverless::ABCD`)? I am assuming the `AWS::Serverless::ABCD` was tested for a specific reason.

tests/validator/input/api/error_definitionbody.yaml
```diff
@@ -0,0 +1,15 @@
+Transform: AWS::Serverless-2016-10-31
```
jfuss, line 1: Sorry if I asked this before and just don't remember. Is there a reason add all new `.yaml` files instead of bootstrapping off the list we already have? 

Just thinking about maintenance going forward. It's already a pain to add all the (what I call) functional tests that run the translator locally. My concern here is that we have to add a whole new set of files on top of the other ones we have. At first it feels like duplicated effort but could be a very good reason to do it this way over what I expected.

jfuss (changes requested):
Just a couple small comments/questions. Generally, this looks really good and glad we are adding this level of detailed validation.

tests/translator/input/unsupported_resources.yaml
mgrandis: Those are passing tests for the Translator.

We were accepting `AWS::Serverless::ABCD` before but we decided to validate the `AWS::Serverless::` resources with an allow list in the [validator](https://github.com/aws/serverless-application-model/pull/2032/files#diff-d0a88e876cff26b717361261093ba7d883eadeda19cb9504a4e8a1fc8b5ac185R78).

tests/validator/input/api/error_definitionbody.yaml
mgrandis: This would have required to update all the output files for the affected translator files and would have made this PR even larger.

Some of the errors caught by the schema validation are also not caught by the translator and can lead to exceptions/crashes.

What we could consider however is to use the validator tests for the translator in a future PR.

samtranslator/validator/validator.py
hawflau: It seems this is only used to remove the prefix `u` in error messages. The error messages are not used as part of the hash generation (both directly and indirectly), so I don't think it will cause any issue.

We will not need this once we completely remove support for Python2. For now, it makes the error messages generated in a Python2 runtime and a Python3 runtime consistent.

samtranslator/validator/validator.py
mgrandis: Calling a private function on an instance in generally not a good practice and frowned upon by pylint.
`validate` is overridden in the constructor with a call to the private one.
We kept the static one for backward compatibility, added a Deprecated notice in the doc.
Not sure about a good new name for the instance validate, if we want to change it.

samtranslator/validator/sam_schema/definitions/api.json
mgrandis: @jfuss Your opinion on this?
As we link to the CFN doc in our doc, the naming would be more consistent with the doc:
- https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-resource-api.html

Links to
- https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-apigateway-stage.html#cfn-apigateway-stage-accesslogsetting
- https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-apigateway-stage-accesslogsetting.html

tests/translator/input/unsupported_resources.yaml
jfuss: I have mixed feelings about this. We aren't failing the Translator to start so fine place to start and see how things shake out. My worry is that someone is building a transform that uses `AWS::Serverless::` and while SAM doesn't support this, we risk failing in the future for something that would have worked.

tests/validator/input/api/error_definitionbody.yaml
jfuss: Maybe separate PRs? I am generally concerned that we duplicated tons of effort and makes it hard (that it is) to do these end to end style testing. 

Not blocking here on this, but I feel pretty strong and finding a way out of this in the future.

samtranslator/validator/validator.py
```diff
@@ -1,55 +1,259 @@
...
+import re
 
 import jsonschema
 from jsonschema.exceptions import ValidationError
 
 from . import sam_schema
 
 
-class SamTemplateValidator(object):
+class SamTemplateValidator:
```
CoshUS, line 10: Can we add type hinting to this file?

samtranslator/validator/validator.py
```diff
@@ -1,55 +1,259 @@
...
+        ----------
+        error : Error
+            Error message to clean
+
+        Returns
+        -------
+        str
+            Cleaned message
+        """
+        final_message = re.sub(self.UNICODE_TYPE_REGEX, r"\1", error.message)
```
CoshUS, line 154: A bit confused on this part. So `error.message` is a string that contains a unicode prefixed string?
`error.message == " Error message: u'inner string' "` 
Will this break anything if a non-ascii character is contained inside of the inner string in Python 2?

CoshUS (changes requested):

samtranslator/validator/validator.py
mgrandis: When run on Python2, the `jsonschema` validator returns error messages with `u` prefixed strings.
For example: `"[Resources] 3 is not of type u'object'"`.
On Python3 it returns: `"[Resources] 3 is not of type 'object'"`.
We need to remove the `u` to have tests results consistent between py2 and py3.
(`u`'s are also ugly)

samtranslator/validator/validator.py
mgrandis: I don't think we use type hinting in the translator project because of Python2.

samtranslator/validator/sam_schema/definitions/common.json
```diff
@@ -0,0 +1,181 @@
...
+    "TagsMap": {
+      "additionalProperties": false,
+      "patternProperties": {
+        "^[a-zA-Z0-9]{1,127}$": {
+          "maxLength": 255,
+          "minLength": 1,
+          "type": "string"
+        }
+      },
+      "patternError": "can only contain alphanumeric characters (max length 127)",
```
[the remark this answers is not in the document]
CoshUS: Unless we have a limit inside of SAM, we should probably get a doc update and change the max length here.
Everything else looks good to me. I'll approve the PR after this is addressed.

samtranslator/validator/validator.py
mgrandis: Renamed the instance validate to `get_errors` to remove the ambiguity.

CoshUS (approved):
LGTM!

jfuss (approved):

PR #2043, opened 2021-05-20
fix: Resources missing when named the same as implicit APIs

aahung:
*Issue #, if available:*

*Description of changes:*

`ServerlessRestApi` and `ServerlessHttpApi` are two implicit resource logical IDs created by SAM. However, customers could totally create their own resources with these two logical IDs. And when that happens, at the same time, if customers define some implicit api events in their templates, those resources will be removed, without customers knowing what happened.

This PR will adjust implicit api logical id when naming collision is detected. For example, if `ServerlessRestApi` already exists, the implicit api local id will become `ServerlessRestApi2`, if `ServerlessRestApi2` also exists, `ServerlessRestApi3` will be used, and so on so forth.

- Also updated the end to end tests. Previously, the test `implicit_api_with_serverless_rest_api_resource` template contains a function with name `ServerlessRestApi` and expected output does not contain that function. I think the test is not right. This PR also fixes it. 

*Checklist:*

- [x] Write/update tests
- [x] `make pr` passes
- [ ] Update documentation

*Examples?*

### DIFF
--- a/samtranslator/plugins/api/implicit_api_plugin.py
+++ b/samtranslator/plugins/api/implicit_api_plugin.py
@@ -38,7 +38,6 @@ class ImplicitApiPlugin(BasePlugin):
         """
         super(ImplicitApiPlugin, self).__init__(name)
 
-        self.existing_implicit_api_resource = None
         # dict containing condition (or None) for each resource path+method for all APIs. dict format:
         # {api_id: {path: {method: condition_name_or_None}}}
         self.api_conditions = {}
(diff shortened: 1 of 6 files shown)


samtranslator/plugins/api/implicit_api_plugin.py
```diff
@@ -59,15 +61,20 @@ def on_before_transform_template(self, template_dict):
...
 
+        # Adjust self.implicit_api_logical_id if the template already contains a resource with the same logical ID
+        # Appending a number (2) to it. If it is still not unique, increment the number until the logical ID is unique.
+        resource_logical_ids = set(template.resources.keys())
+        if self.implicit_api_logical_id in resource_logical_ids:
+            suffix = 2
+            while (self.implicit_api_logical_id + str(suffix)) in resource_logical_ids:
+                suffix = suffix + 1
+            self.implicit_api_logical_id = self.implicit_api_logical_id + str(suffix)
+            LOG.debug('Adjust implicit API logical ID to "%s"', self.implicit_api_logical_id)
```
hoffa, line 72: I assume `ServerlessRestApi` (and other reserved words) isn't hardcoded elsewhere that could break a reference?

Could also make this a `disambiguate(name: str, names: list[str]) -> str` if this is expected to be a common pattern.

aahung: yes, it is not hardcoded except in tests. 

> disambiguate(name: str, names: list[str]) -> str

sounds like a good idea

aahung:
As discussed, we will be looking for non-breaking fixes like adding warning and update docs.

PR #2046, opened 2021-05-22
test: New test group to test for side effects

mgrandis:
*Issue #, if available:*

*Description of changes:*
Added a new group test `test_transform_success_no_side_effect` in `tests/translator/test_translator.py` to test for side effects between translator executions

*Checklist:*

- [x] Write/update tests
- [x] `make pr` passes
- [ ] Update documentation

*Examples?*

### DIFF
--- a/tests/translator/test_translator.py
+++ b/tests/translator/test_translator.py
@@ -599,6 +599,62 @@ class TestTranslatorEndToEnd(AbstractTestTranslator):
 
         self.assertEqual(deep_sort_lists(output_fragment), deep_sort_lists(expected))
 
+    @parameterized.expand(
+        [
+            (
+                "usage_plans",
+                ("api_with_usageplans_shared_no_side_effect_1", "api_with_usageplans_shared_no_side_effect_2"),
+            ),
+        ]
+    )
+    @patch(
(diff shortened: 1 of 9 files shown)


mndeveci (approved):
LGTM

tests/translator/test_translator.py
```diff
@@ -599,6 +599,49 @@ def test_transform_success_resource_policy(self, testcase, partition_with_region
...
+        testcase : Tuple
+            Test name (unused) and Templates
+        templates : List
+            List of templates to transform
+        """
+        partition = partition_with_region[0]
+        region = partition_with_region[1]
+
+        for template in testcase[1]:
+            print(template, partition, region)
```
wchengru, line 639: Is that a debug left?

mgrandis: This is useful when a test fails and you want to know which template/partition/region was responsible.
We already print the output of the transform in `_compare_transform` but it doesn't show the name of the template.
Those are only printed if the test fails.

wchengru: Sounds good! I would also recommend to format this information better.

wchengru (approved):

PR #2061, opened 2021-06-18
fix: validate unknown method for a path in swagger definition for apigw

mndeveci:
*Issue #, if available:*
N/A

*Description of changes:*
When setting path for default authorizers, SAM Transform tries to access specific method information from path definition inside swagger in this line; https://github.com/aws/serverless-application-model/blob/develop/samtranslator/swagger/swagger.py#L534

Since it is trying to access this dictionary directly, it might raise exception when a method is defined in function's event definition but not in swagger definition.

Before setting up path for default authorizer, we first try to see if that path has the expected method definition, if not we are going to raise an invalid document exception rather than getting item not found exception from dictionary.

*Checklist:*

- [x] Write/update tests
- [ ] `make pr` passes
- [ ] Update documentation

*Examples?*

### DIFF
--- a/samtranslator/swagger/swagger.py
+++ b/samtranslator/swagger/swagger.py
@@ -531,7 +531,17 @@ class SwaggerEditor(object):
             if add_default_auth_to_preflight or normalized_method_name != "options":
                 normalized_method_name = self._normalize_method_name(method_name)
                 # It is possible that the method could have two definitions in a Fn::If block.
-                for method_definition in self.get_method_contents(self.get_path(path)[normalized_method_name]):
+                method_dict = self.get_path(path).get(normalized_method_name)
+                if method_dict is None:
+                    raise InvalidDocumentException(
+                        [
+                            InvalidTemplateException(
+                                "Can't find method definition dictionary for method '{}'".format(normalized_method_name)
+                            )
+                        ]
+                    )
+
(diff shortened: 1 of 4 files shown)


hawflau (approved):
LGTM!

wchengru (approved):

samtranslator/swagger/swagger.py
```diff
@@ -531,7 +531,17 @@ def set_path_default_authorizer(
             if add_default_auth_to_preflight or normalized_method_name != "options":
                 normalized_method_name = self._normalize_method_name(method_name)
                 # It is possible that the method could have two definitions in a Fn::If block.
-                for method_definition in self.get_method_contents(self.get_path(path)[normalized_method_name]):
+                method_dict = self.get_path(path).get(normalized_method_name)
```
sriram-mv, line 534: Some additional questions which may have already been considered.

can normalized_method_name be anything other than a string?
can path be anything other than a string? 

Could we have intrinsics which resolve each of them to a dict instead of a string?

mndeveci:
Thanks for the reviews everyone, I find out that Ahmed has an earlier PR which is addressing same issue in a different way. Since it is not polluting our code with extra validation and using what information we have there, I would suggest to move forward with that. You can check it here; https://github.com/aws/serverless-application-model/pull/2031

Closing this PR.

PR #2062, opened 2021-06-18
Adding support for metric publishing, manually tested and tests pending.

c2tarun:
*Issue #, if available:*

*Description of changes:*

*Checklist:*

- [x] Write/update tests
- [x] `make pr` passes
- [ ] Update documentation

*Examples?*

### DIFF
--- a/samtranslator/metrics/metrics.py
+++ b/samtranslator/metrics/metrics.py
@@ -1,0 +1,140 @@
+from botocore.exceptions import ClientError
+from collections import namedtuple
+import logging
+
+LOG = logging.getLogger(__name__)
+
+
+class MetricsPublisher:
+    """Interface for all MetricPublishers"""
+
+    def publish(self, namespace, metrics):
+        raise NotImplementedError
+
+
+class CWMetricsPublisher(MetricsPublisher):
+    def __init__(self, cloudwatch_client):
+        """
+        Constructor
+
+        :param cloudwatch_client: cloudwatch client required to publish metrics to cloudwatch
+        """
+
+        self.cloudwatch_client = cloudwatch_client
+
+    def publish(self, namespace, metrics):
+        """
+        Method to publish all metrics to Cloudwatch.
+
+        :param namespace: namespace applied to all metrics published.
+        :param metrics: list of metrics to be published
+        """
+        batch = []
+        for metric in metrics:
+            batch.append(metric)
+            # Cloudwatch recommends not to send more than 20 metrics at a time
+            if len(batch) == 20:
+                self._flush_metrics(namespace, batch)
+                batch = []
+        self._flush_metrics(namespace, batch)
+
+    def _flush_metrics(self, namespace, metrics):
+        """
+        Internal method to publish all provided metrics to cloudwatch, please make sure that array size of metrics is <= 20."""
+        metric_data = list(map(lambda m: m.get_metric_data(), metrics))
+        try:
+            self.cloudwatch_client.put_metric_data(Namespace=namespace, MetricData=metric_data)
+        except ClientError:
+            LOG.exception("Failed to report {} metrics".format(len(metric_data)), metrics)
+
+
+class DummyMetricsPublisher(MetricsPublisher):
+    def publish(self, namespace, metrics):
+        """Method to actually publish them metric"""
+
+
+class Unit:
+    Seconds = "Seconds"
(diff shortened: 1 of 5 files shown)


samtranslator/metrics/metrics.py
```diff
@@ -0,0 +1,156 @@
...
+    def _flush_metrics(self, namespace, metrics):
+        """
+        Internal method to publish all provided metrics to cloudwatch, please make sure that array size of metrics is <= 20.
+        """
+        metric_data = list(map(lambda m: m.get_metric_data(), metrics))
+        try:
+            self.cloudwatch_client.put_metric_data(Namespace=namespace, MetricData=metric_data)
+        except Exception as e:
+            LOG.error("Failed to report {} metrics".format(len(metric_data)))
+            LOG.exception(e)
```
mndeveci, line 53: nit: I think there is a parameter in the logging framework to pass the exception rather than doing it in 2 lines

samtranslator/metrics/metrics.py
```diff
@@ -0,0 +1,156 @@
...
+
+
+class DummyMetricsPublisher(MetricsPublisher):
+    def __init__(self):
+        MetricsPublisher.__init__(self)
+
+    def publish(self, namespace, metrics):
+        """Method to actually publish them metric"""
+        LOG.info("Dummy publisher ignoring {} metrices".format(len(metrics)))
```
mndeveci, line 62: Can we do nothing or change this to `debug`? My concern is coming from people who is using this as a dependency in their library, for instance we might see this message from time to time in SAM CLI too.

samtranslator/metrics/metrics.py
```diff
@@ -0,0 +1,156 @@
...
+        Create metric with unit Count.
+
+        :param name: metric name
+        :param value: value of metric
+        :param unit: unit of metric (try using values from Unit class)
+        :param dimensions: array of dimensions applied to the metric
+        """
+        self._record_metric(name, value, unit, dimensions)
+
+    def record_latency(self, name, value, unit=Unit.Milliseconds, dimensions=[]):
```
mndeveci, line 142: Given that these methods are specialized, can we make them easier by removing `unit` and `dimensions` parameters?

samtranslator/metrics/metrics.py
c2tarun: I'll change this to `debug`.

samtranslator/metrics/metrics.py
c2tarun: I'll remove `unit` we kind of need `dimensions`

samtranslator/metrics/metrics.py
c2tarun: Fixed!!

samtranslator/metrics/metrics.py
```diff
@@ -0,0 +1,156 @@
+import logging
```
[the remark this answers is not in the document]
c2tarun: Done

samtranslator/metrics/metrics.py
```diff
@@ -0,0 +1,158 @@
...
+            LOG.exception("Failed to report {} metrics".format(len(metric_data)), exc_info=e)
+
+
+class DummyMetricsPublisher(MetricsPublisher):
+    def __init__(self):
+        MetricsPublisher.__init__(self)
+
+    def publish(self, namespace, metrics):
+        """Method to actually publish them metric"""
+        LOG.debug("Dummy publisher ignoring {} metrices".format(len(metrics)))
```
mndeveci, line 64: Sorry I didn't catch this in the first round but can we add some documentation that this won't publish anything?
And also the doc on the `publish` method seems to be suggesting that this method will publish the metrics.

mndeveci (commented):
Thanks for addressing the feedbacks @c2tarun , left one comment about docs, and I was wondering if we need `tests/metrics/__init__.py `?

samtranslator/metrics/metrics.py
c2tarun: Addressed, please let me know if it makes sense now.

c2tarun:
> Thanks for addressing the feedbacks @c2tarun , left one comment about docs, and I was wondering if we need `tests/metrics/__init__.py `?

You are right, we don't, good catch :) remove it.

mndeveci (approved):

hoffa (approved):

PR #2072, opened 2021-06-28
initial barebones serverless canary implementation and testing

aaythapa:
*Issue #, if available:*
no issues

*Description of changes:*
Implementation of the barebones AWS::Serverless::Canary resource. Only transforms into AWS::Synthetics::Canary for now and assumes that users fill all the properties.

Unit tests to make sure CodeURI and InlineCode are transformed correctly.
Unit tests to make sure RunConfig properties of AWS::Synthetics::Canary is made correctly when the right properties in AWS::Serverless::Canary are defined.
Integration tests to make sure AWS::Serverless::Canary is deployed correctly.

*Checklist:*

- [x] Write/update tests
- [x] `make pr` passes
- [ ] Update documentation

*Examples?*


c2tarun:
@aaythapa, can you please make sure that `make pr` passes?

samtranslator/model/sam_resources.py
```diff
@@ -802,6 +803,114 @@ def _validate_deployment_preference_and_add_update_policy(
...
+        return resources
+
+    def _construct_synthetics_canary(self):
+        """Constructs a AWS::Synthetics::Canary resource."""
+        canary = SyntheticsCanary(
+            self.logical_id, depends_on=self.depends_on, attributes=self.get_passthrough_resource_attributes()
+        )
+        canary.ArtifactS3Location = self.ArtifactS3Location
+        canary.Code = self._construct_code_dict()
+        canary.ExecutionRoleArn = self.Role
```
c2tarun, line 851: Just adding a comment in case there are other eyes on PR. We'll add support for auto creation of Role in next PRs.

samtranslator/model/sam_resources.py
```diff
@@ -802,6 +803,114 @@ def _validate_deployment_preference_and_add_update_policy(
             )
 
 
+class SamCanary(SamResourceMacro):
+    """SAM canary macro."""
+
+    resource_type = "AWS::Serverless::Canary"
+    property_types = {
+        "FunctionName": PropertyType(True, is_str()),
```
c2tarun, line 811: I would rather not make `FunctionName` a mandatory field and probably use LogicalId in case this one is not provided.

samtranslator/model/sam_resources.py
```diff
@@ -802,6 +803,114 @@ def _validate_deployment_preference_and_add_update_policy(
...
+        "Handler": PropertyType(True, is_str()),
+        "Runtime": PropertyType(True, is_str()),
+        "CodeUri": PropertyType(False, one_of(is_str(), is_type(dict))),
+        "InlineCode": PropertyType(False, one_of(is_str(), is_type(dict))),
+        "MemorySize": PropertyType(False, is_type(int)),
+        "Tags": PropertyType(False, is_type(dict)),
+        "Tracing": PropertyType(False, is_type(bool)),
+        "Timeout": PropertyType(False, is_type(int)),
+        "Role": PropertyType(True, is_str()),
```
c2tarun, line 820: Do we want to make this mandatory, given we are going to resolve this if not provided?

samtranslator/model/sam_resources.py
```diff
@@ -802,6 +803,114 @@ def _validate_deployment_preference_and_add_update_policy(
...
+        canary.StartCanaryAfterCreation = self.StartCanaryAfterCreation
+        canary.SuccessRetentionPeriod = self.SuccessRetentionPeriod
+        canary.Tags = self._construct_tag_list(self.Tags)
+        canary.VPCConfig = self.VpcConfig
+
+        if self.Tracing or self.Environment or self.MemorySize or self.Timeout:
+            canary.RunConfig = self._construct_run_config()
+        return canary
+
+    def _construct_run_config(self):
```
c2tarun, line 866: Add some documentation to this method.

samtranslator/model/sam_resources.py
```diff
@@ -802,6 +803,114 @@ def _validate_deployment_preference_and_add_update_policy(
...
+        runconfig = {"ActiveTracing": self.Tracing, "MemoryInMB": self.MemorySize, "TimeoutInSeconds": self.Timeout}
+        if self.Environment:
+            runconfig["EnvironmentVariables"] = self.Environment["Variables"]
+
+        return {k: v for k, v in runconfig.items() if v is not None}
```
c2tarun, line 871: It took me couple of seconds to understand that you are filtering out `None` properties. This might be a useful method, can you probably extract out this code to a separate method like:

```
def extract_not_none_properties(self, dict):
    return {k: v for k, v in dict.items() if v is not None}
```

this way somebody who is reading the code understands it instantly and you can also unit test this logic in more ways.

samtranslator/model/sam_resources.py
```diff
@@ -802,6 +803,114 @@ def _validate_deployment_preference_and_add_update_policy(
...
+
+    def _construct_code_dict(self):
+        """Constructs Synthetics Canary Code Dictionary based on the accepted SAM artifact properties such
+        as `InlineCode` and `CodeUri` and also raises errors if more than one of them is
+        defined. `PackageType` determines which artifacts are considered.
```
c2tarun, line 876: Do we support `PackageType` for Canary?

samtranslator/model/sam_resources.py
```diff
@@ -802,6 +803,114 @@ def _validate_deployment_preference_and_add_update_policy(
...
+        # dispatch mechanism per artifact on how it needs to be transformed.
+        artifact_dispatch = {
+            "InlineCode": _construct_inline_code,
+            "CodeUri": construct_s3_location_object,
+        }
+
+        filtered_artifacts = dict(filter(lambda x: x[1] != None, artifacts.items()))
+
+        # if len(filtered_artifacts) == 0:
+        #     raise InvalidResourceException(self.logical_id, "Only one of 'InlineCode' or 'CodeUri' can be set.")
```
c2tarun, line 897: Probably delete it.

samtranslator/model/sam_resources.py
```diff
@@ -802,6 +803,114 @@ def _validate_deployment_preference_and_add_update_policy(
...
+        def _construct_inline_code(*args, **kwargs):
+            return {"Handler": self.Handler, "Script": self.InlineCode}
+
+        # dispatch mechanism per artifact on how it needs to be transformed.
+        artifact_dispatch = {
+            "InlineCode": _construct_inline_code,
+            "CodeUri": construct_s3_location_object,
+        }
+
+        filtered_artifacts = dict(filter(lambda x: x[1] != None, artifacts.items()))
```
c2tarun, line 894: Sorry I am not able to clearly understand what is going on here? Is it similar to logic above where you are filtering out fields from dict with value None?

samtranslator/model/sam_resources.py
```diff
@@ -802,6 +803,114 @@ def _validate_deployment_preference_and_add_update_policy(
...
+
+        # if len(filtered_artifacts) == 0:
+        #     raise InvalidResourceException(self.logical_id, "Only one of 'InlineCode' or 'CodeUri' can be set.")
+
+        filtered_keys = [key for key in filtered_artifacts.keys()]
```
c2tarun, line 899: probably an easier to read will be:

```
filtered_keys = list(filtered_artifacts.keys())
```

c2tarun: Also can you think of a better name than `filtered_keys`?

c2tarun (changes requested):

samtranslator/model/sam_resources.py
aaythapa: Made "Role" and "ArtifactS3Location" not required as they will be resolved if they are not provided

samtranslator/model/sam_resources.py
aaythapa: Typo, removed it 👍

samtranslator/model/sam_resources.py
aaythapa: It does the same thing, instead of using `dict(filter(lambda x: x[1] != None, artifacts.items()))`

samtranslator/model/sam_resources.py
```diff
@@ -802,6 +803,114 @@ def _validate_deployment_preference_and_add_update_policy(
...
+    resource_type = "AWS::Serverless::Canary"
+    property_types = {
+        "FunctionName": PropertyType(True, is_str()),
+        "Handler": PropertyType(True, is_str()),
+        "Runtime": PropertyType(True, is_str()),
+        "CodeUri": PropertyType(False, one_of(is_str(), is_type(dict))),
+        "InlineCode": PropertyType(False, one_of(is_str(), is_type(dict))),
+        "MemorySize": PropertyType(False, is_type(int)),
+        "Tags": PropertyType(False, is_type(dict)),
+        "Tracing": PropertyType(False, is_type(bool)),
```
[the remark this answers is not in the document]
aaythapa: In SAM Function it's apparently for Intrinsic functions like FN::Sub. Going to keep it the same for now just to be safe and will change it to `"InlineCode": PropertyType(False, is_str())` in the future if needed.

c2tarun (approved):

samtranslator/model/sam_resources.py
```diff
@@ -802,6 +803,126 @@ def _validate_deployment_preference_and_add_update_policy(
...
+    def _construct_synthetics_canary(self):
+        """Constructs a AWS::Synthetics::Canary resource."""
+        canary = SyntheticsCanary(
+            self.logical_id, depends_on=self.depends_on, attributes=self.get_passthrough_resource_attributes()
+        )
+        canary.ArtifactS3Location = self.ArtifactS3Location
+        canary.Code = self._construct_code_dict
+        canary.ExecutionRoleArn = self.Role
+        canary.FailureRetentionPeriod = self.FailureRetentionPeriod
+        canary.Name = self.FunctionName if self.FunctionName else self.logical_id
```
aahung, line 853: Can we put a comment about why we set canary's name to the function name?

samtranslator/model/sam_resources.py
```diff
@@ -802,6 +803,126 @@ def _validate_deployment_preference_and_add_update_policy(
...
+            set_artifact = "InlineCode"
+        elif "CodeUri" in filtered_artifact_keys:
+            set_artifact = "CodeUri"
+        else:
+            raise InvalidResourceException(self.logical_id, "Either 'InlineCode' or 'CodeUri' must be set.")
+
+        dispatch_function = artifact_dispatch[set_artifact]
+        code = dispatch_function(artifacts[set_artifact], self.logical_id, set_artifact)
+        code["Handler"] = self.Handler
+        return code
```
aahung, line 923: Seems these logics can be rearranged to make it clearer:

```suggestion
        filtered_artifacts = self._extract_not_none_properties(artifacts.items())
        filtered_artifact_keys = list(filtered_artifacts.keys())

        if "InlineCode" in filtered_artifact_keys:
            return {"Handler": self.Handler, "Script": self.InlineCode}
        elif "CodeUri" in filtered_artifact_keys:
            code = construct_s3_location_object(self.CodeUri, self.logical_id, "CodeUri")
            code["Handler"] = self.Handler
            return code
        else:
            raise InvalidResourceException(self.logical_id, "Either 'InlineCode' or 'CodeUri' must be set.")
```

aahung (commented):
It looks good to me in general!

samtranslator/model/sam_resources.py
aaythapa: do you mean why we set canary name to the logical id? we set it to function name because that's the name provided by the user in the FunctionName property

samtranslator/model/cloudwatch.py
```diff
@@ -0,0 +1,21 @@
+from samtranslator.model import PropertyType, Resource
+from samtranslator.model.types import is_type, is_str, list_of, one_of
+
+
+class SyntheticsCanary(Resource):
```
hoffa, line 5: For reference:

```suggestion
# https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-synthetics-canary.html
class SyntheticsCanary(Resource):
```

samtranslator/model/sam_resources.py
```diff
@@ -802,6 +803,117 @@ def _validate_deployment_preference_and_add_update_policy(
...
+        canary.StartCanaryAfterCreation = self.StartCanaryAfterCreation
+        canary.SuccessRetentionPeriod = self.SuccessRetentionPeriod
+        canary.Tags = self._construct_tag_list(self.Tags)
+        canary.VPCConfig = self.VpcConfig
+
+        if self.Tracing or self.Environment or self.MemorySize or self.Timeout:
+            canary.RunConfig = self._construct_run_config()
+        return canary
+
+    def _extract_not_none_properties(self, dict):
```
hoffa, line 868: Should be static (or in some utility module?). Also avoid shadowing standard symbols (`dict`) otherwise they won't be usable.

samtranslator/model/sam_resources.py
```diff
@@ -802,6 +803,117 @@ def _validate_deployment_preference_and_add_update_policy(
...
+        transformed AWS::Synthetics::Canary needs to be added. Note, for Environment property the syntaxt in AWS::Serverless::Canary is
+        Environment:
+            Variables:
+                Var1: Var2
+        while in AWS::Synthetics::Canary its
+        EnvironmentVariables:
+            Var1: Var2
+        so it needs to be transformed accordingly
+        """
+        runconfig = {"ActiveTracing": self.Tracing, "MemoryInMB": self.MemorySize, "TimeoutInSeconds": self.Timeout}
```
hoffa, line 886: Should we use names consistent with other serverless resources? e.g. [`AWS::Serverless::Function`](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-resource-function.html)

aahung: I believe this is spec from https://docs.aws.amazon.com/AmazonSynthetics/latest/APIReference/API_CanaryRunConfigOutput.html

hoffa (approved):

PR #2074, opened 2021-06-30
fix: fail with invalid resource, when RetentionPolicy has unresolved intrinsic value

mndeveci:
*Issue #, if available:*
N/A

*Description of changes:*
Fail with `InvalidResourceException` if `RetentionPolicy` value of `AWS::Serverless::LayerVersion` has un-resolved intrinsic function.

Added test cases that will cover invalid use-case, as well as an un-tested valid use case.

*Checklist:*

- [x] Write/update tests
- [x] `make pr` passes
- [ ] Update documentation

*Examples?*


samtranslator/model/sam_resources.py
```diff
@@ -1224,6 +1225,16 @@ def _get_retention_policy_value(self):
         :return: value for the DeletionPolicy attribute.
         """
 
+        if is_intrinsic(self.RetentionPolicy):
+            # RetentionPolicy attribute of AWS::Serverless::LayerVersion does set the DeletionPolicy
+            # attribute. And DeletionPolicy attribute does not support intrinsic values.
+            raise InvalidResourceException(
+                self.logical_id,
+                "'{}' does not accept intrinsic functions, please use one of the following options: {}".format(
+                    "RetentionPolicy", [self.RETAIN, self.DELETE]
```
mgrandis, line 1234: nit: I think you can directly put `RetentionPolicy` in the format string here

mgrandis (approved):

samtranslator/model/sam_resources.py
```diff
@@ -1224,6 +1225,15 @@ def _get_retention_policy_value(self):
         :return: value for the DeletionPolicy attribute.
         """
 
+        if is_intrinsic(self.RetentionPolicy):
+            # RetentionPolicy attribute of AWS::Serverless::LayerVersion does set the DeletionPolicy
+            # attribute. And DeletionPolicy attribute does not support intrinsic values.
+            raise InvalidResourceException(
```
jfuss, line 1231: Should we communicate that the customer could use the top level Attribute of DeletionPolicy? For customers needing or wanting the Intrinsic, this might be the only way forward for them.

tests/translator/input/basic_layer.yaml
```diff
@@ -35,4 +35,10 @@ Resources:
...
     Condition: TestCondition    
     Properties:
-      ContentUri: s3://sam-demo-bucket/layer.zip  
+      ContentUri: s3://sam-demo-bucket/layer.zip
+
+  LayerWithCaseInsensitiveRetentionPolicy:
+    Type: 'AWS::Serverless::LayerVersion'
+    Properties:
+      ContentUri: s3://sam-demo-bucket/layer.zip
+      RetentionPolicy: DeleTe
```
jfuss, line 44: ❤️

tests/translator/input/error_layer_invalid_properties.yaml
```diff
@@ -43,4 +46,10 @@ Resources:
...
       ContentUri: s3://sam-demo-bucket/layer.zip
-      RetentionPolicy: !Ref DeleteList
\ No newline at end of file
+      RetentionPolicy: !Ref DeleteList
+
+  LayerWithRetentionPolicyAsIntrinsic:
+    Type: 'AWS::Serverless::LayerVersion'
+    Properties:
+      ContentUri: s3://sam-demo-bucket/layer.zip
+      RetentionPolicy: !If [DeletePolicyIf, Retain, Delete]
```
jfuss, line 55: Do we have a test for Ref? I think that intrinsic is supported right?

jfuss (changes requested):
A couple small questions

samtranslator/model/sam_resources.py
mndeveci: Unfortunately, intrinsics are not supported for top level attributes; https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference.html

I tested with a template but I wasn't able to deploy it.

tests/translator/input/error_layer_invalid_properties.yaml
mndeveci: Yeap it is supported, and we already have tests in `basic_layer.yaml` file above.

jfuss (approved):

Issue #1911, opened 2021-02-05
Module name clashes


meyer1994:
### Description:

When using this library, or any library that uses it, there are name clashes
with common names used in projects. Like `utils`. This is the result of a
random `sys.path.insert` at `samtranslator/feature_toggle/feature_toggle.py`.

This was not breaking before, I do not know on which version it started. This
breaks CI with unpinned versions. Or, for example, it breaks when any
dependency uses this module. For example [moto][1]:

```
moto > cfn-lint > aws-sam-translator
```

Either way, modules should not modify `sys.path` willy-nilly.

### Steps to reproduce:

1. Create an `utils.py`
1. Create a `main.py` file that first imports something that reaches
`feature_toggle.py`
1. Try importing something from your `utils.py`

#### Example

```py
# utils.py
something = 'nice'

# main.py
from samtranslator.translator.translator import Translator
from utils import something  # ImportError
```

### Observed result:

`ImportError` is thrown when trying to import things from my `utils.py`.

### Expected result:

No `ImportError`.

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

1. OS: Any
2. If using SAM CLI, `sam --version`:
3. AWS region: Does not matter

`Add --debug flag to any SAM CLI commands you are running`

[1]: https://github.com/spulec/moto

wyattanderson:
Also running into this. Looks like it was introduced in #1737.

jfuss:
This was patched in [v1.46.0](https://github.com/aws/serverless-application-model/releases/tag/v1.46.0). Closing

PR #2076, opened 2021-07-05
fix: unexpected sys.path injection

wchengru:
*Issue #, if available:*
#1911
*Description of changes:*

*Checklist:*

- [ ] Write/update tests
- [ ] `make pr` passes
- [ ] Update documentation

*Examples?*

### DIFF
--- a/samtranslator/feature_toggle/feature_toggle.py
+++ b/samtranslator/feature_toggle/feature_toggle.py
@@ -12,9 +12,6 @@ from samtranslator.feature_toggle.dialup import (
     SimpleAccountPercentileDialup,
 )
 
-my_path = os.path.dirname(os.path.abspath(__file__))
-sys.path.insert(0, my_path + "/..")
-
 LOG = logging.getLogger(__name__)
 
 


samtranslator/feature_toggle/feature_toggle.py
```diff
@@ -12,9 +12,6 @@
     SimpleAccountPercentileDialup,
 )
 
-my_path = os.path.dirname(os.path.abspath(__file__))
```
jfuss, line 15: @c2tarun I believe you where the original author of Feature Toggle. Why was this developed in this way? If we remove this, are there any downstream side effects? I assume this was done for a reason, just not sure what it was.

c2tarun: As discussed offline, this is so that we can reference [this config file](https://github.com/aws/serverless-application-model/blob/develop/tests/feature_toggle/input/feature_toggle_config.json) from [sam-translate](https://github.com/aws/serverless-application-model/blob/7eb358731933152df41c56b2e8ce1916186460e1/bin/sam-translate.py#L102-L104) script for testing.

jfuss: If this is just for the script, I would rather break that than do patching like this in the lib. We shouldn't need FeatureToggle from that script anyways. Maybe I am missing something but I would error on not overriding `sys.path` as it's super dangerous.

jfuss:
Closing in favor of #2380

PR #2092, opened 2021-07-16
fix: Validate Trigger field and test Cognito properties with intrinsic functions

mildaniel:
*Issue #, if available:*

*Description of changes:*
SAM transform was allowing invalid values and types for Cognito Trigger to be transformed, resulting in invalid templates that won't deploy.

*Checklist:*

- [x] Write/update tests
- [x] `make pr` passes


moelasmar (approved):

samtranslator/model/eventsources/push.py
```diff
@@ -994,6 +995,16 @@ def to_cloudformation(self, **kwargs):
 
     def _inject_lambda_config(self, function, userpool):
         event_triggers = self.Trigger
+
+        if isinstance(event_triggers, dict):
```
mgrandis, line 999: Could you instead check setting the third parameter of `PropertyType` at line to `False`? (Forbidding intrinsic functions). See what kind of error message we get, if it makes sense for the customers. With this, also check the error that you get if you pass a dict that is not an intrinsic.

If it doesn't and we keep the current code, we should maybe change the `isinstance(event_triggers, dict)` to a `not list or string` test.

mgrandis (changes requested):
Let's see if we can use the `PropertyType` mechanism to exclude intrinsics instead.

samtranslator/model/eventsources/push.py
mildaniel: Setting this parameter returns the following for both intrinsics and generic dicts: 

```
ERROR:__main__:Invalid Serverless Application Specification document. Number of errors found: 1. Resource with id [ImplicitApiFunctionOneTrigger] is invalid. Type of property 'Trigger' is invalid.
ERROR:__main__:["Resource with id [ImplicitApiFunctionOneTrigger] is invalid. Type of property 'Trigger' is invalid."]
```

This seems to make enough sense for implementation so I'll go ahead and do it this way.

mgrandis (approved):

PR #2094, opened 2021-07-16
feat: Serverless canary creates s3 bucket if ArtifactS3Location property is not provided

aaythapa:
_Issue #, if available:_
no issues

_Description of changes:_
Implementing the making of AWS::S3::Resource if user doesn't provide the ArtifactS3Location property in AWS::Serverless::Canary resource.

_Description of how you validated changes:_
Unit tests to make sure an S3 resource was made whenever the ArtifactS3Location property wasn't filled.
Integration tests to make sure AWS::Serverless::Canary and AWS::S3::Resource deployed correctly.

*Checklist:*

- [x] Write/update tests
- [x] `make pr` passes
- [ ] Update documentation

*Examples?*


samtranslator/model/sam_resources.py
```diff
@@ -865,23 +917,41 @@ def _construct_synthetics_canary(self):
             canary.RunConfig = self._construct_run_config()
         return canary
 
+    # Need to construct canary name since the Name property is required in AWS::Synthetics::Canary and CloudFormation
+    # doesn't automatically generate one upon deployment
+    def _construct_canary_name(self):
+        # Synthetics Canary name is limited to 22 characters
```
c2tarun, line 923: That low 😲

samtranslator/model/sam_resources.py
```diff
@@ -865,23 +917,41 @@ def _construct_synthetics_canary(self):
...
+        max_logical_id_length = 11
+        max_unique_id_length = 5
+        prefix = "sam-"
```
c2tarun, line 926: Can these be moved out as constants?

samtranslator/model/sam_resources.py
```diff
@@ -865,23 +917,41 @@ def _construct_synthetics_canary(self):
...
+
+        # max_logical_id_length + max_unique_id_length + len(prefix) must be less than or equal to 22
+        logical_id_lowered = self.logical_id.lower()[:max_logical_id_length] + "-"
+        suffix = uuid.uuid4().hex[:max_unique_id_length]
+        return prefix + logical_id_lowered + suffix
```
c2tarun, line 931: I like the idea of adding unique suffix and certainly will work. However, say tomorrow Synthetics realized that 22 is too low and they increased it to say 128, we have to come here and update this right?

What happens if we do this:
1. Have an options for customers to override this?
2. Simply use the full logical id and let it fail with error from Synthetics. In that case customers will have option either to override the name generated by us or reduce the size of their logical id.

Thoughts?

tests/translator/output/canaries_with_no_artifact_location.json
```diff
@@ -0,0 +1,124 @@
...
+          {
+            "Key": "lambda:createdBy",
+            "Value": "SAM"
+          }
+        ]
+      }
+    },
+    "SecondCanaryWithNoArtifactLocationArtifactBucket": {
+      "Type": "AWS::S3::Bucket",
+      "DeletionPolicy": "Retain",
```
c2tarun, line 110: So this will create the bucket and will retain it after integration tests will clean up the Cloudformation stack. Can you ensure that this bucket gets deleted after you finish running integration tests? If not can you actually update integ tests to delete this bucket?

For more context the problem with leaving the bucket is that S3 have a 100 bucket limit per account. If your tests create 5 buckets per run, then we'll run out of limit in 20 or so runs of integration tests.

c2tarun (changes requested):

samtranslator/model/sam_resources.py
```diff
@@ -839,8 +841,58 @@ def to_cloudformation(self, **kwargs):
...
+            synthetics_canary.ArtifactS3Location = {"Fn::Join": ["", ["s3://", {"Ref": s3bucket.logical_id}]]}
+        else:
+            # Check if ArtifactS3Location includes "s3://" at the beginning. The rest of the string will be used when
+            # creating a role for this canary
+            artifact_prefix = "s3://"
+            s3string = synthetics_canary.ArtifactS3Location[: len(artifact_prefix)]
+            if s3string != artifact_prefix:
+                raise InvalidResourceException(
+                    self.logical_id, "Invalid ArtifactS3Location, must include s3:// at the beginning"
+                )
```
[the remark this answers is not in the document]
c2tarun: Also I was thinking more about it, this `ArtifactS3Location` has always have to be an S3 uri right? Can you add a method to validate type s3 URI in `types.py` and just declare it in resource attributes.

So in types file you'll add something like:
```
def s3_uri():
  ... your validation logic here, which will check string and prefix and whatever you want to check.
```

And then you'll modify Canary resource:

```
class SamCanary(SamResourceMacro):
    """SAM canary macro."""

    resource_type = "AWS::Serverless::Canary"
    property_types = {
......
        "StartCanaryAfterCreation": PropertyType(True, is_type(bool)),
        "ArtifactS3Location": PropertyType(False, s3_uri()),
        "FailureRetentionPeriod": PropertyType(False, is_type(int)),
.....
    }

```

integration/resources/templates/single/basic_canary.yaml
```diff
@@ -0,0 +1,80 @@
+Parameters:
+  CanaryName:
+    Type: String
+    Default: integration-test
+    MaxLength: 21
+Resources:
+  # role and role permission created for now so that canary will develop, won't be needed in the future when feature to
+  # automatically create role from serverless canary is added
+  CloudWatchSyntheticsRole:
```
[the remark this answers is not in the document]
aaythapa: yep, this PR doesn't generate role.

samtranslator/model/sam_resources.py
```diff
@@ -808,7 +810,7 @@ class SamCanary(SamResourceMacro):
 
     resource_type = "AWS::Serverless::Canary"
     property_types = {
-        "FunctionName": PropertyType(False, is_str()),
+        "FunctionName": PropertyType(False, one_of(is_str(), is_type(dict))),
```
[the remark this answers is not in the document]
aaythapa: yep, similar to how InlineCode can be a dict. Its so users can input intrinsic function

tests/translator/output/canaries_with_no_artifact_location.json
aaythapa: yep, the bucket gets deleted once the stack is verified.

samtranslator/model/sam_resources.py
aaythapa: If synthetics increases their character count for name then yes we have to update the max allowed length (we can increase the max logical id length so that the entire logical id is include)

1. Not sure what you mean by customers being able to override 
2. The reason the full logical id isn't included is because CloudFormation cuts the logical id as well when its generating names for resource so i was trying to emulate that. For example an S3 bucket name has a char limit of 63 and when auto-generating a name for a s3 resource Cloudformation follows the `<stack name>-<logical id>-<characters>` format. So if an s3 bucket resource is made with a logical id of 73 chars, when cloudformation is auto generating the name it cuts the logical id to keep the `<stack name>-<logical id>-<characters>` format and to fit the 63 characters char limit for s3 bucket names.

samtranslator/model/sam_resources.py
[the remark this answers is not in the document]
aaythapa: Adding a comment in case there are other eyes on PR. The validation for s3_uri() was dropped because cfn validates it when the template is deployed.

samtranslator/model/sam_resources.py
c2tarun: 1. By customers able to override I mean if a value to FunctionName is provided then it'll be used instead of our generated value (which I think may already be the case).
2. Okay, so I understand now why you did it this way. What I was wondering is like `s3://` URI validation can we also just use full `LogicalId` with some prefix/suffix to indicate that this is name generated by SAM and let CFN handle the length validation.

c2tarun: Discussed offline and we agreed that it is in favor of better UX to generate a name and have an option for users to override this.

c2tarun (approved):

integration/resources/templates/single/basic_canary_with_tags.yaml
```diff
@@ -33,7 +35,7 @@ Resources:
               - s3:PutObject
               - s3:GetBucketLocation
             Resource:
-              - Fn::Sub: arn:aws:s3:::${ResultsBucket}/*
+              - Fn::Sub: arn:aws:s3:::${SyntheticsCanaryArtifactBucket}/*
```
aahung, line 38: Where is the value of `SyntheticsCanaryArtifactBucket` coming from? I don't see the parameter or resources with such name

samtranslator/model/sam_resources.py
```diff
@@ -51,6 +53,11 @@
 from samtranslator.model.role_utils import construct_role_for_resource
 from samtranslator.model.xray_utils import get_xray_managed_policy_name
 
+# max_logical_id_length + max_unique_id_length + len(prefix) must be less than or equal to 22
```
aahung, line 56: Is there any reference to the number 22?

samtranslator/model/sam_resources.py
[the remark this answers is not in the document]
aahung: This is consistent with `"AWS::Serverless::Function"` (samtranslator/model/sam_resources.py:59)

samtranslator/model/sam_resources.py
```diff
@@ -839,8 +846,49 @@ def to_cloudformation(self, **kwargs):
...
+
+        :returns: the generated S3Bucket
+        :rtype: model.s3.S3Bucket
+        """
+
+        # Construct the LogicalId of S3Bucket by appending ArtifactBucket to the Canary LogicalId. Once deployed, the
+        # bucket name will be automatically generated by Cloudformation.
+        # https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-name.html
+
+        logical_id = self.logical_id + "ArtifactBucket"
```
aahung, line 869: When we append string to a logical ID, do we need to worry about char limit?

samtranslator/model/sam_resources.py
```diff
@@ -839,8 +846,49 @@ def to_cloudformation(self, **kwargs):
...
+
+        # Set the DeletionPolicy of the S3 resource to Retain to prevent CloudFormation from trying to delete the
+        # bucket when its not empty. This is necessary because if a user creates and runs a serverless canary without
+        # an artifact bucket, then tries to delete/replace that resource, CloudFormation will give an error as it
+        # tries to delete the artifact bucket made by SAM but can't since its not empty. Retaining the bucket will by
+        # pass this error
```
aahung, line 875: How about `UpdatePolicy`?

integration/resources/templates/single/basic_canary_with_tags.yaml
aaythapa: It'll be the logical id of the artifact bucket thats automatically generated (logical id of canary + "ArtifactBucket"). Since this PR doesn't automatically generate the roles I had to hardcode in the logical id of the bucket to get the tests running but in the PR with role creation implemented it won't be hardcoded

samtranslator/model/sam_resources.py
aaythapa: the char limit for logical id of a resource is 255 characters so it's pretty big. Also, in Serverless Function when a role is being created the logical id for that role is self.logical_id + "Role" and there is no check for char limit so i didn't add a check here. Think there is no check because CFN handles it when deploying.

samtranslator/model/sam_resources.py
aahung: resolved offline

samtranslator/model/sam_resources.py
aahung: Fair enough if we don't check in current code. Thanks for checking

samtranslator/model/sam_resources.py
aaythapa: updated documentation in code for better understanding

aahung (approved):

PR #1577, opened 2020-04-27
Issue 1508 remove check requiring identity to be required if ReauthorizeEvery equals zero

daftster:
Check removed to avoid must specify Identity with at least one of Headers, QueryStrings, StageVariables, or Context. error.  This is allowed to be removed from aws console.

*Issue 1508:*
https://github.com/awslabs/serverless-application-model/issues/1508

*Description of changes:*
- According to https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-property-api-lambdarequestauthorizer.html#sam-api-lambdarequestauthorizer-identity identity is optional
- Confirmed that I am able to remove all identities from aws console after API had been deployed using sam cli
- Remove validation of having identity when payload type is REQUEST.

*Description of how you validated changes:*

*Checklist:*

- [X] Write/update tests
- [X] `make pr` passes
- [ ] Update documentation
- [X] Verify transformed template deploys and application functions as expected
- [ ] Add/update example to `examples/2016-10-31`

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

Issue #1508, opened 2020-03-11
Api Authorizer must specify at least one identity even authorization  caching is disabled


respire:
### Description

I want to create a request authorizer which performs authentication if Authorization is existed in header. However, when i try to run sam deploy, it occured an error.

**Authorizer must specify Identity with at least one of Headers, QueryStrings, StageVariables, or Context**

but the official api docs said (identitySource)

> When the authorization caching is not enabled, this property is optional.

https://docs.aws.amazon.com/apigateway/api-reference/resource/authorizer/

so I set ReauthorizeEvery to zero so as to disable caching, but it has no effect.

### Steps to reproduce

```.yml
Auth:
      Authorizers:
        MyRequestAuthorizer:
          FunctionPayloadType: REQUEST
          FunctionArn: !GetAtt MyRequestAuthorizerFunction.Arn
          Identity:
            ReauthorizeEvery: 0
```

### Observed result

**Authorizer must specify Identity with at least one of Headers, QueryStrings, StageVariables, or Context**

### Expected result

should be able to create a request authorizer with empty identitySource when caching is disabled
noted that i can remove all the identitySources under api gateway dashboard.

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)
SAM CLI, version 0.44.0
macos 10.15.3

Elliot128:
I am running into this issue as well. Did you ever get a workaround besides using the console?

daftster:
PR #1577 opened for fixing this issue.

moelasmar:
PR got merged

mgrandis:
Released in [1.3.6](https://github.com/aws/serverless-application-model/releases/tag/v1.36.0) and [CLI 1.24.0](https://github.com/aws/aws-sam-cli/releases/tag/v1.24.0)

jfuss:
Re-opening. The original PR was reverted due to service impact when `ReauthorizeEvery` is an intrinsic function.

sukruavcuoglu:
Still seeing the same problem. Any future plan on this one?

Leigh-M:
Including  ReauthorizeEvery: 0  (as below)  fixed the issue for me:

```
      Auth:
        DefaultAuthorizer: AWS_IAM
        Authorizers:
          AuthFunction:
            FunctionPayloadType: REQUEST
            FunctionArn: !GetAtt AuthFunction.Arn
            Identity:
              Headers:
                - username
                - password
              ReauthorizeEvery: 0
```

[...]

PR #2105, opened 2021-07-26
fix(bug): Check if Identity.ReauthorizeEvery equals zero

jfuss:
*Issue #, if available:*
#1508

*Description of changes:*
Was originally introduced #1577 but reverted #2038 due to service impact. Specifically, the originally implementation did not consider intrinsics as the value (a dict) and therefore threw a `TypeError` and caused SAM to unexpectedly crash.

This PR reverts the revert (re-introducing the originally changes) but updated the implementation to support handling intrinsics. Also adding more unit tests to cover a wider range of inputs and new functional tests to test a "live" template with intrinsics in it.

Ran both the functional tests and unit tests on develop and this branch. Which pass. 

Here is the output from the unit tests:
```
(venv-python38) ➜  serverless-application-model git:(develop) ✗ pytest -vv tests/ -k TestApiGatewayAuthorizer
=========================================================================================================================================================================== test session starts ============================================================================================================================================================================
platform darwin -- Python 3.8.0, pytest-6.1.2, py-1.9.0, pluggy-0.13.1 -- /Users/jfuss/sam-github/serverless-application-model/venv-python38/bin/python3.8
cachedir: .pytest_cache
rootdir: /Users/jfuss/sam-github/serverless-application-model, configfile: pytest.ini
plugins: cov-2.10.1
collected 1937 items / 1926 deselected / 11 selected

tests/model/test_api.py::TestApiGatewayAuthorizer::test_create_authorizer_fails_with_empty_identity PASSED                                                                                                                                                                                                                                                           [  9%]
tests/model/test_api.py::TestApiGatewayAuthorizer::test_create_authorizer_fails_with_missing_identity_values_and_not_cached PASSED                                                                                                                                                                                                                                   [ 18%]
tests/model/test_api.py::TestApiGatewayAuthorizer::test_create_authorizer_fails_with_string_authorization_scopes PASSED                                                                                                                                                                                                                                              [ 27%]
tests/model/test_api.py::TestApiGatewayAuthorizer::test_create_authorizer_with_identity_ReauthorizeEvery_asNone_valid_with_query_strings PASSED                                                                                                                                                                                                                      [ 36%]
tests/model/test_api.py::TestApiGatewayAuthorizer::test_create_authorizer_with_identity_intrinsic_is_invalid_if_no_querystring_stagevariables_context_headers PASSED                                                                                                                                                                                                 [ 45%]
tests/model/test_api.py::TestApiGatewayAuthorizer::test_create_authorizer_with_identity_intrinsic_is_valid_with_context PASSED                                                                                                                                                                                                                                       [ 54%]
tests/model/test_api.py::TestApiGatewayAuthorizer::test_create_authorizer_with_identity_intrinsic_is_valid_with_headers PASSED                                                                                                                                                                                                                                       [ 63%]
tests/model/test_api.py::TestApiGatewayAuthorizer::test_create_authorizer_with_identity_intrinsic_is_valid_with_query_strings PASSED                                                                                                                                                                                                                                 [ 72%]
tests/model/test_api.py::TestApiGatewayAuthorizer::test_create_authorizer_with_identity_intrinsic_is_valid_with_stage_variables PASSED                                                                                                                                                                                                                               [ 81%]
tests/model/test_api.py::TestApiGatewayAuthorizer::test_create_authorizer_with_non_integer_identity PASSED                                                                                                                                                                                                                                                           [ 90%]
tests/model/test_api.py::TestApiGatewayAuthorizer::test_create_oauth2_auth PASSED
```

*Checklist:*

- [X] Write/update tests
- [X] `make pr` passes
- [X] Update documentation

*Examples?*

### DIFF
--- a/samtranslator/model/apigateway.py
+++ b/samtranslator/model/apigateway.py
@@ -270,8 +270,9 @@ class ApiGatewayAuthorizer(object):
         query_strings = identity.get("QueryStrings")
         stage_variables = identity.get("StageVariables")
         context = identity.get("Context")
+        ttl = identity.get("ReauthorizeEvery")
 
-        if not headers and not query_strings and not stage_variables and not context:
+        if (ttl is None or int(ttl) > 0) and not headers and not query_strings and not stage_variables and not context:
             return True
 
         return False
@@ -314,7 +315,9 @@ class ApiGatewayAuthorizer(object):
                 swagger[APIGATEWAY_AUTHORIZER_KEY]["authorizerCredentials"] = function_invoke_role
 
             if self._get_function_payload_type() == "REQUEST":
-                swagger[APIGATEWAY_AUTHORIZER_KEY]["identitySource"] = self._get_identity_source()
+                identity_source = self._get_identity_source()
+                if identity_source:
+                    swagger[APIGATEWAY_AUTHORIZER_KEY]["identitySource"] = self._get_identity_source()
 
         # Authorizer Validation Expression is only allowed on COGNITO_USER_POOLS and LAMBDA_TOKEN
         is_lambda_token_authorizer = authorizer_type == "LAMBDA" and self._get_function_payload_type() == "TOKEN"
--- a/samtranslator/model/apigateway.py
+++ b/samtranslator/model/apigateway.py
@@ -275,20 +275,14 @@ class ApiGatewayAuthorizer(object):
         required_properties_missing = not headers and not query_strings and not stage_variables and not context
 
         try:
-            int(ttl)
+            ttl_int = int(ttl)
         # this will catch if ttl is None and not convertable to an int
         except TypeError:
             # previous behavior before trying to read ttl
-            if required_properties_missing:
-                return True
-
-            return False
+            return required_properties_missing
 
         # If we can resolve ttl, attempt to see if things are valid
-        if (ttl is None or int(ttl) > 0) and required_properties_missing:
-            return True
-
-        return False
+        return ttl_int > 0 and required_properties_missing
 
     def generate_swagger(self):
         authorizer_type = self._get_type()
--- a/samtranslator/model/apigateway.py
+++ b/samtranslator/model/apigateway.py
@@ -280,7 +280,7 @@ class ApiGatewayAuthorizer(object):
         except TypeError:
             # previous behavior before trying to read ttl
             if required_properties_missing:
-               return True
+                return True
 
             return False
 
--- a/samtranslator/model/apigateway.py
+++ b/samtranslator/model/apigateway.py
@@ -272,7 +272,20 @@ class ApiGatewayAuthorizer(object):
         context = identity.get("Context")
         ttl = identity.get("ReauthorizeEvery")
 
-        if (ttl is None or int(ttl) > 0) and not headers and not query_strings and not stage_variables and not context:
+        required_properties_missing = not headers and not query_strings and not stage_variables and not context
+
+        try:
+            int(ttl)
+        # this will catch if ttl is None and not convertable to an int
+        except TypeError:
+            # previous behavior before trying to read ttl
+            if required_properties_missing:
+               return True
+
+            return False
+
+        # If we can resolve ttl, attempt to see if things are valid
+        if (ttl is None or int(ttl) > 0) and required_properties_missing:
             return True
 
         return False
--- a/tests/model/test_api.py
+++ b/tests/model/test_api.py
@@ -17,3 +17,18 @@ class TestApiGatewayAuthorizer(TestCase):
             auth = ApiGatewayAuthorizer(
                 api_logical_id="logicalId", name="authName", authorization_scopes="invalid_scope"
             )
+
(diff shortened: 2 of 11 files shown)


samtranslator/model/apigateway.py
```diff
@@ -270,8 +270,22 @@ def _is_missing_identity_source(self, identity):
         query_strings = identity.get("QueryStrings")
         stage_variables = identity.get("StageVariables")
         context = identity.get("Context")
+        ttl = identity.get("ReauthorizeEvery")
 
-        if not headers and not query_strings and not stage_variables and not context:
+        required_properties_missing = not headers and not query_strings and not stage_variables and not context
+
+        try:
```
mgrandis, line 277: I think that the reason we reverted was because of an intrinsic function in `ttl`.
If `ttl` is an intrinsic, I think we should return `False` (no access to the value, assume it's valid, pass it to CFN).
We can use the `is_intrinsic` function to test this.

Then we can use `six`, `integer_types`:
`from six import integer_types`
`return (!isinstance(ttl, integer_types) or ttl == 0) and required_properties_missing`

mgrandis (changes requested):

samtranslator/model/apigateway.py
```diff
@@ -270,8 +270,22 @@ def _is_missing_identity_source(self, identity):
...
+        # this will catch if ttl is None and not convertable to an int
+        except TypeError:
+            # previous behavior before trying to read ttl
+            if required_properties_missing:
+                return True
+
+            return False
+
+        # If we can resolve ttl, attempt to see if things are valid
+        if (ttl is None or int(ttl) > 0) and required_properties_missing:
```
mndeveci, line 288: We already checked that `ttl` has an int value, checking `None` is unnecessary here. I think if we go with @mgrandis's suggestion, it will simplify some of the things here.

tests/model/test_api.py
```diff
@@ -14,6 +14,105 @@ def test_create_oauth2_auth(self):
...
+            )
+
+    def test_create_authorizer_fails_with_missing_identity_values_and_not_cached(self):
+        with pytest.raises(InvalidResourceException):
+            ApiGatewayAuthorizer(
+                api_logical_id="logicalId",
+                name="authName",
+                identity={"ReauthorizeEvery": 10},
+                function_payload_type="REQUEST",
+            )
```
mndeveci, line 41: This test is same as the one on line 19, right?

tests/model/test_api.py
```diff
@@ -14,6 +14,105 @@ def test_create_oauth2_auth(self):
...
+
+    def test_create_authorizer_fails_with_empty_identity(self):
+        with pytest.raises(InvalidResourceException):
+            ApiGatewayAuthorizer(
+                api_logical_id="logicalId", name="authName", identity={}, function_payload_type="REQUEST"
+            )
```
mndeveci, line 47: And this is same as the one on line 28?

mndeveci (changes requested):

samtranslator/model/apigateway.py
jfuss: Thanks for the feedback. `try: except:` is pretty normal in python and I prefer this over checking instances types. "Pythonic" style tends to be do the thing and then handle errors. `try: except:` is not slower, which tends to be a reason to avoid it in other languages, https://stackoverflow.com/questions/7604636/better-to-try-something-and-catch-the-exception-or-test-if-its-possible-first has some good explanations. 

With that I prefer the `try: except:` implementation over the checking instance.

I am generally cautious around doing multiple things within a `try` block. If something is changed in the future, you run the risk of the `except` catching on unknown things. Or even if something else unexpected happens, the `except` can mask things. So well scoped `try` tends to be a better practice. I will see how I can simplify this further, but pretty strongly against over-scoping `try` blocks.

samtranslator/model/apigateway.py
jfuss: See above comment about the above suggestion.

tests/model/test_api.py
jfuss: Correct. This was removed in the original PR, so looks like I messed something up when adding them back. I remove this.

tests/model/test_api.py
jfuss: Correct. This was removed in the original PR, so looks like I messed something up when adding them back. I remove this.

mndeveci (approved):

tests/model/test_api.py
```diff
@@ -14,6 +14,100 @@ def test_create_oauth2_auth(self):
...
+            function_payload_type="REQUEST",
+        )
+
+        self.assertIsNotNone(auth)
+
+    def test_create_authorizer_with_identity_intrinsic_is_valid_with_headers(self):
+        auth = ApiGatewayAuthorizer(
+            api_logical_id="logicalId",
+            name="authName",
+            identity={"ReauthorizeEvery": {"FN:If": ["isProd", 10, 0]}, "Headers": ["Accept"]},
```
mgrandis, line 58: nit: the intrinsic key is actually `Fn::If`

mgrandis (approved):

PR #2107, opened 2021-07-29
Serverless canary creates IAM role resource if Role property is not provided + Policies property added

aaythapa:
*Issue #, if available:*

*Description of changes:*

- Implementing the making of AWS::IAM::Role resource if user doesn't provide the Role property in AWS::Serverless::Canary. 
- Added the a Policies property to Serverless::Canary so that user can add policies to the AWS::IAM::Role resource. Similar to the Policies property in AWS::Serverless::Function (https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-resource-function.html#sam-function-policies) 

- Unit tests to make sure an IAM::Role resource was made whenever the Role property wasn't filled and the right roles were attached to the policy. 
- Unit tests to make sure whenever Policies was defined it was appended to the IAM::Role resource
- Integration tests to make sure Serverless::Canary and IAM::Role deployed correctly.

*Checklist:*

- [x] Write/update tests
- [x] `make pr` passes
- [ ] Update documentation

*Examples?*


samtranslator/model/iam.py
```diff
@@ -129,3 +129,45 @@ def lambda_invoke_function_role_policy(cls, function_arn, logical_id):
             },
         }
         return document
+
+    @classmethod
+    def execute_canary_policy(cls, logical_id, result_bucket, canary_name):
+        document = {
+            "PolicyName": logical_id + "CanaryPolicy",
```
c2tarun, line 136: Can we split this into 3 methods, one for each statement?

c2tarun (changes requested):
PR description says that unit tests were added but I didn't find them. Did you forgot to check them in?

samtranslator/model/sam_resources.py
```diff
@@ -845,18 +846,77 @@ def to_cloudformation(self, **kwargs):
...
+        managed_policy_map = kwargs.get("managed_policy_map", {})
         synthetics_canary = self._construct_synthetics_canary()
         resources.append(synthetics_canary)
 
         # A S3 Bucket resource will be added to the transformed template if the user doesn't provide an artifact
         # bucket to store canary results
-        if synthetics_canary.ArtifactS3Location is None:
+
+        artifact_bucket_name = ""
+        if self.ArtifactS3Location is None:
```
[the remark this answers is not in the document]
aaythapa: to keep it consistent with the other times that I check whether or not a property is None. Like for example later in the code is check for self.Tracing and self.VpcConfig

samtranslator/model/sam_resources.py
```diff
@@ -845,18 +846,77 @@ def to_cloudformation(self, **kwargs):
...
+                ArnGenerator.generate_aws_managed_policy_arn("service-role/AWSLambdaVPCAccessExecutionRole")
+            )
+        if self.Tracing:
+            managed_policy_name = get_xray_managed_policy_name()
+            managed_policy_arns.append(ArnGenerator.generate_aws_managed_policy_arn(managed_policy_name))
+
+        # if user has defined Policies property, those policies will be appended to this role
+        function_policies = ResourcePolicies(
+            {"Policies": self.Policies},
+            # No support for policy templates in the "core"
```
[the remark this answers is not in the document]
aaythapa: the `No support for policy templates in the "core"`? Not too sure, it was included in the other places where `ResourcePolicies` is used so I included it here just in case

samtranslator/model/sam_resources.py
```diff
@@ -845,18 +846,76 @@ def to_cloudformation(self, **kwargs):
...
+            policy_template_processor=None,
+        )
+
+        # The policy to execute the canary is only added to the role if the user hasn't defined ArtifactS3Location
+        # this emulates CloudWatch Synthetics Canary dashboard's behavior
+        policy_documents = []
+        if self.ArtifactS3Location is None:
```
aahung, line 900: nit: We didn't use `if self.Tracing is not None` above, to keep consistent, can we use `if not self.ArtifactS3Location:` here?

samtranslator/model/iam.py
```diff
@@ -129,3 +129,45 @@ def lambda_invoke_function_role_policy(cls, function_arn, logical_id):
...
+                                    ],
+                                ]
+                            }
+                        ],
+                    },
+                    {"Effect": "Allow", "Action": ["s3:ListAllMyBuckets"], "Resource": ["*"]},
+                    {
+                        "Effect": "Allow",
+                        "Action": "cloudwatch:PutMetricData",
+                        "Resource": "*",
```
aahung, line 167: Is there a way to be more specific about the resource? (or the condition below is enough)

aaythapa: the condition should be enough, its the same one used by cloudwatch dashboard when you generate a role policy through there

samtranslator/model/iam.py
```diff
@@ -129,3 +129,45 @@ def lambda_invoke_function_role_policy(cls, function_arn, logical_id):
...
+                                    "",
+                                    [
+                                        {
+                                            "Fn::Sub": "arn:${AWS::Partition}:logs:${AWS::Region}:${"
+                                            "AWS::AccountId}:log-group:/${AWS::Partition}/lambda/cwsyn-"
+                                        },
+                                        canary_name,
+                                        "-*",
+                                    ],
+                                ]
```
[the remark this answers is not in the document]
aaythapa: did some more testing on the logs and turns out the canary name is not even needed. Will take it out and take out the `Fn::Join` as well

samtranslator/model/sam_resources.py
```diff
@@ -854,28 +855,36 @@ def to_cloudformation(self, **kwargs):
...
+        -   If the VpcConfig property is specified then the policy to execute VPC will be added
+        -   If the Policies property is specified then the that will be appended to the IAM::Role's Policies property
 
         :returns: the generated IAM Role
         :rtype: model.iam.IAMRole
         """
-
-        assume_role_policy_document = IAMRolePolicies.lambda_assume_role_policy()
         role_attributes = self.get_passthrough_resource_attributes()
+        if self.AssumeRolePolicyDocument is not None:
```
aahung, line 885: nit: Use `if self.AssumeRolePolicyDocument:`

tests/translator/output/aws-cn/canary_with_assume_role_policy.json
```diff
@@ -0,0 +1,61 @@
...
+        "Tags": [
+          {
+            "Key": "lambda:createdBy",
+            "Value": "SAM"
+          }
+        ]
+      }
+    }
+  }
+}
```
aahung, line 61: nit: Add an extra line at the end of the file

aahung (approved):
LGTM, thanks for addressing the comments

samtranslator/model/iam.py
c2tarun: I don't think that is Sam meant about Resources. To answer Sam's question we allow Canary Lambda to putmetricdata so that the integration to publishing metrics is not very brittle. Another approach could be that metric data is created and only permission to those will be granted. But every time we require to publish new metric we have to update it. Plus I couldn't find any way to create metric data using CFN, so it might have to be done by hand. This is similar to `LambdaBasicExecution` permission for logs.

c2tarun (approved):
LGTM!

PR #2108, opened 2021-07-30
fix: handle string RestApiId in Api event source swagger integration

torresxb1:
*Issue #, if available:*

*Description of changes:*
`RestApiId` can be either a string or a dict with `Ref`, but in the swagger integration for the Api event source it assumes it's a dict. This change allows handling a `RestApiId` that's either a string or dict.

Added a test for handling a string `RestApiId` and another test for a dict. The test for the string version fails without the changes and passes with the changes.

EDIT: this PR went through some changes and in the end I realized the part where it assumed the rest api id was a dict didn't end up using the api id. So this PR ended up being more of a refactoring to remove parts where the api id was not used, encapsulate the logic to get the string value and refactor the parts where it's actually used, and add end-to-end tests for rest api id with ref, string, and other intrinsics.

*Checklist:*

- [x] Add/update tests using:
    - [x] Correct values
- [x] `make pr` passes
- [ ] Update documentation

*Examples?*


samtranslator/model/eventsources/push.py
```diff
@@ -747,8 +747,9 @@ def _add_swagger_integration(self, api, function, intrinsics_resolver):
 
             if self.Auth.get("ResourcePolicy"):
                 resource_policy = self.Auth.get("ResourcePolicy")
+                api_id = self.RestApiId if isinstance(self.RestApiId, string_types) else self.RestApiId.get("Ref")
```
jfuss, line 750: Is this to support specifying `RestApiId` as a pure string? If so why should we over failing? This pattern is not in line with CloudFormation, so struggling why we think this is the right thing to do.

jfuss (changes requested):

samtranslator/model/eventsources/push.py
torresxb1: It seems like `RestApiId` can already be either string or dict (correct me if I'm wrong), but this part of the code assumed it to be a dict. So I just updated it to handle strings. If you look in the test file below, in my other changes, the `setUp` already had a string api id. Also, `samtranslator/validator/sam_schema/schema.json` seems to indicate it can be string or object (https://github.com/aws/serverless-application-model/blob/5511198db567a3bae496c35fb44de8994597e5d7/samtranslator/validator/sam_schema/schema.json#L324)

jfuss: Thought it was odd but looks like we have tests for that already: https://github.com/aws/serverless-application-model/blob/develop/tests/translator/input/api_cache.yaml#L12

So I yield this point. :)

jfuss: So instead, is there a way we could encapsulate this logic so that callers getting the RestApiId get it in a standard way? If we have places we already handle this, adding more places just keeps that pattern going and risk when we add something new, it might be missed in other spots.

jfuss (changes requested):
Can we add more end to end or functional (as I like to call them)? This would follow the following pattern.

1) Adding an input template to https://github.com/aws/serverless-application-model/tree/develop/tests/translator/input
2) add outputs files for each partition here: https://github.com/aws/serverless-application-model/tree/develop/tests/translator/output
3) add the file name here: https://github.com/aws/serverless-application-model/blob/develop/tests/translator/test_translator.py#L256 so it will run.

samtranslator/utils/rest_api_id.py
```diff
@@ -0,0 +1,9 @@
+def get_rest_api_id_string(rest_api_id):
+    """
+    rest_api_id can be either a string or a dictionary where the actual api id is the value at key "Ref".
+    If rest_api_id is a dictionary, returns value at key "Ref". Otherwise, return rest_api_id.
+
+    :param rest_api_id: a string or dictionary that contain the rest_api_id
+    :return: string value of rest_api_id
+    """
+    return rest_api_id.get("Ref") if isinstance(rest_api_id, dict) else rest_api_id
```
torresxb1, line 9: Not sure if I should use `.get("Ref")` or `["Ref"]`. If for some reason there was no `"Ref"` key, I believe the former would return `None` while the latter would raise `KeyError`. A couple of the uses above (`translator.py` and second use in `push.py`) were using `.get`, while the other (first use in `push.py`) was using brackets.

torresxb1, line 9: not sure if this is the right place for this

tests/translator/input/api_swagger_integration_with_string_api_id.yaml
```diff
@@ -0,0 +1,23 @@
...
+    Type: AWS::Serverless::Function
+    Properties:
+      CodeUri: s3://sam-demo-bucket/member_portal.zip
+      Handler: index.gethtml
+      Runtime: nodejs12.x
+      Events:
+        GetHtml:
+          Type: Api
+          Properties:
+            RestApiId: HtmlApi
```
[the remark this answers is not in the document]
torresxb1: Every usage I've seen and some documentation (e.g. https://github.com/aws/serverless-application-model/blob/develop/docs/cloudformation_compatibility.rst#api) seems to indicate that it only expects either a `Ref` intrinsic or a `string`. When I try running a local invoke with a `Sub`, I get: `samtranslator.model.exceptions.InvalidDocumentException: [InvalidResourceException('HtmlFunction', 'Event with id [GetHtml] is invalid. Api Event must reference an Api in the same template.')]`. 
I think this would be the case for all other intrinsics as well (https://github.com/aws/serverless-application-model/blob/5511198db567a3bae496c35fb44de8994597e5d7/samtranslator/plugins/api/implicit_rest_api_plugin.py#L88).

samtranslator/utils/rest_api_id.py
```diff
@@ -0,0 +1,9 @@
+def get_rest_api_id_string(rest_api_id):
```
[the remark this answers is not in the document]
torresxb1: I could do that but I think I'd only be able to use that within `push.py`. So, we'd have some similar/duplicate code in `translator.py` and some more in `implicit_api_plugin.py`, which I just realized basically does the same thing. Would it be better for each of these files to handle this logic (through a method/function/@property) or for all of them to reference the util?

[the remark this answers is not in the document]
hawflau: would that be better organized if `get_rest_api_id_string` is a @staticmethod under the `Api` class in push event source?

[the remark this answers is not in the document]
jfuss: I don't fully understand why we need any specific logic in `translator.py` but if everything was serialized into an object, it wouldn't matter. If there isn't a way to refactor what we currently have to do this through objects, you can keep the single method in a file, but let's not call it `utils`. Its just a dumping ground for all the "utilities" or "helpers". What @hawflau suggested meets in the middle.

tests/translator/input/api_swagger_integration_with_string_api_id.yaml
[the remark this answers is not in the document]
jfuss: That's fine but do we have specific tests for this? You are adding/touching a property and before we accept, we need to make sure we have tests that cover different intrinsic cases.

samtranslator/swagger/swagger.py
```diff
@@ -1019,7 +1019,7 @@ def _add_iam_resource_policy_for_method(self, policy_list, effect, resource_list
             statement.extend([policy_statement])
             self.resource_policy["Statement"] = statement
 
-    def _get_method_path_uri_list(self, path, api_id, stage):
```
torresxb1, line 1022: unused parameter, so removed + removed from all calls

samtranslator/swagger/swagger.py
```diff
@@ -902,7 +902,7 @@ def add_models(self, models):
 
             self.definitions[model_name.lower()] = schema
 
-    def add_resource_policy(self, resource_policy, path, api_id, stage):
```
torresxb1, line 905: unused parameter (after removing the unused api_id parameter from `_get_method_path_uri_list`), so removed + removed from all calls

tests/model/eventsources/test_api_event_source.py
```diff
@@ -118,31 +118,6 @@ def test_get_permission_with_just_slash(self):
...
-        self.api_event_source.Auth = {"Authorizer": "AWS_IAM", "ResourcePolicy": {"AwsAccountWhitelist": ["123456"]}}
-        self.api_event_source.to_cloudformation(
-            function=self.func,
-            explicit_api={
-                "__MANAGE_SWAGGER": True,
-                "DefinitionBody": {"openapi": "3.0.1", "paths": {"/foo": {}, "/bar": {}}},
-            },
-            intrinsics_resolver=IntrinsicsResolver({}),
-        )
-
```
torresxb1: The point of these was to test the api id, but it turned out that the api id wasn’t even used in `to_cloudformation`, after I went deeper into the functions where the api id was passed into. I removed this unused api id parameter (the other changes in this commit), and removed the unit tests because I don’t think it made sense to test the api id if it wasn’t used, and less so after removing it from the calls in `to_cloudformation`

samtranslator/model/stepfunctions/events.py
```diff
@@ -259,6 +256,7 @@ def resources_to_link(self, resources):
         permitted_stage = "*"
         stage_suffix = "AllStages"
         explicit_api = None
+        rest_api_id = PushApi.get_rest_api_id_string(self.RestApiId)
```
torresxb1, line 259: Not sure if it's weird calling a push api method from a state machine api method

jfuss (changes requested):

samtranslator/model/eventsources/push.py
```diff
@@ -877,6 +872,17 @@ def _add_swagger_integration(self, api, function, intrinsics_resolver):
 
         api["DefinitionBody"] = editor.swagger
 
+    @staticmethod
+    def get_rest_api_id_string(rest_api_id):
```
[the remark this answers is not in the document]
torresxb1: This is still in the API class in this file. Or do you mean `samtranslator/model/api`? I don't think it was there before

tests/model/eventsources/test_api_event_source.py
```diff
@@ -3,6 +3,7 @@
 
 from samtranslator.model.eventsources.push import Api
 from samtranslator.model.lambda_ import LambdaFunction, LambdaPermission
+from samtranslator.intrinsics.resolver import IntrinsicsResolver
```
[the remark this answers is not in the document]
torresxb1: Thanks for pointing it out 🙏

samtranslator/model/eventsources/push.py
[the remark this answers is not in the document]
jfuss: We discussed this offline. SAM's modeling is weird (in my opinion) but where you have it now seems reasonable enough.

jfuss (approved):
Thanks for keeping up with this and working through the feedback!

PR #2149, opened 2021-09-15
feat: add metrics per resource transformation

mndeveci:
This commit enables metric collection per resource, if a `Metrics` instance is been defined and configured to use CloudWatch.

*Checklist:*

- [x] Add/update tests using:
    - [x] Correct values
- [x] `make pr` passes
- [] Update documentation

### DIFF
--- a/samtranslator/metrics/method_decorator.py
+++ b/samtranslator/metrics/method_decorator.py
@@ -16,8 +16,8 @@ class MetricsMethodWrapperSingleton:
     This singleton will be alive until lambda receives shutdown event
     """
 
-    _METRICS_INSTANCE = None
     _DUMMY_INSTANCE = Metrics("ServerlessTransform", DummyMetricsPublisher())
+    _METRICS_INSTANCE = _DUMMY_INSTANCE
 
     @staticmethod
     def set_instance(metrics):
@@ -28,10 +28,7 @@ class MetricsMethodWrapperSingleton:
         """
         Return the instance, if nothing is set return a dummy one
         """
-        if MetricsMethodWrapperSingleton._METRICS_INSTANCE:
-            return MetricsMethodWrapperSingleton._METRICS_INSTANCE
-
-        return MetricsMethodWrapperSingleton._DUMMY_INSTANCE
+        return MetricsMethodWrapperSingleton._METRICS_INSTANCE
 
 
 def _get_metric_name(prefix, name, func, args):
--- a/samtranslator/metrics/method_decorator.py
+++ b/samtranslator/metrics/method_decorator.py
@@ -1,0 +1,107 @@
+import functools
+from datetime import datetime
+from samtranslator.metrics.metrics import DummyMetricsPublisher, Metrics
+from samtranslator.model import Resource
+import logging
+
+LOG = logging.getLogger(__name__)
+
+
+class MetricsMethodWrapperSingleton:
+    """
+    Keeps the instance of Metrics object.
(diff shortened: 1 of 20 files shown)


samtranslator/metrics/method_decorator.py
```diff
@@ -0,0 +1,110 @@
...
+        The function that is decorated. This will be used as metric name if name is not provided and caller is not an
+        instance of Resource object.
+    args : args
+        Arguments that is originally passed to the caller. This function will check if first element in this function
+        is a Resource then it reads the 'resource_type' property out of it to generate the metric name.
+    """
+    if name:
+        metric_name = name
+    else:
+        if args > 0 and isinstance(args[0], Resource):
```
mgrandis, line 57: Should this be `len(args) > 0`?

samtranslator/metrics/method_decorator.py
```diff
@@ -0,0 +1,110 @@
...
+LOG = logging.getLogger(__name__)
+
+
+class MetricsMethodWrapperSingleton:
+    """
+    Keeps the instance of Metrics object.
+    This singleton will be alive until lambda receives shutdown event
+    """
+
+    _METRICS_INSTANCE = None
```
mgrandis, line 19: I think capital letter are reserved for constants.
This should maybe be lowercase.

samtranslator/model/eventsources/cloudwatchlogs.py
```diff
@@ -4,6 +4,7 @@
 from samtranslator.model.types import is_str
 from samtranslator.translator.arn_generator import ArnGenerator
 from .push import PushEventSource
+from ...metrics.method_decorator import cw_timer
```
mgrandis, line 7: I think absolute paths may be better

mgrandis (changes requested):
A few comments

samtranslator/metrics/method_decorator.py
mndeveci: Shoot, I was converting this from `if len(args) > 0` to `if args`. Let me fix this.

samtranslator/metrics/method_decorator.py
mndeveci: Yes, but static instances are also using uppercase names (`LOG` for instance). I didn't want this to be accepted as instance variable that is why I went with uppercase.

CoshUS (commented):
PR itself looks good to me, but while I was digging into Metrics class, I found some usage of list as default values. They should be fine with the current usages, but could lead to static variable issues in the future. Can we retrospectively change them in this PR?
https://github.com/aws/serverless-application-model/blob/develop/samtranslator/metrics/metrics.py#L122

samtranslator/metrics/method_decorator.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+def _send_cw_metric(prefix, name, execution_time_ms, func, args):
+    """
+    Gets metric name from 'prefix', 'name', 'func' and 'args' parameters, then calls metrics instance from its
+    singleton object to record the latency.
+    """
+    try:
+        metric_name = _get_metric_name(prefix, name, func, args)
+        LOG.info("Execution took %sms for %s", execution_time_ms, metric_name)
```
c2tarun, line 75: Should this be converted to `debug` to reduce some noise?

mgrandis (approved):

c2tarun (approved):

CoshUS (approved):
LGTM! Thanks for addressing the issue.

PR #2155, opened 2021-09-20
Change SAR check using boto3

xazhao:
*Issue #, if available:*

*Description of changes:*
We want to use boto3 to check if SAR is available on the current region instead of hardcoding region name.

*Checklist:*

- [ ] Add/update tests using:
    - [ ] Correct values
- [ ] `make pr` passes
- [ ] Update documentation

*Examples?*

### DIFF
--- a/samtranslator/region_configuration.py
+++ b/samtranslator/region_configuration.py
@@ -28,11 +28,9 @@ class RegionConfiguration(object):
     @classmethod
     def is_sar_supported(cls):
         """
-        SAR is not supported in af-south-1 at the moment.
+        SAR is not supported in some regions.
         https://aws.amazon.com/about-aws/global-infrastructure/regional-product-services/
 
         :return: True, if SAR is supported in current region.
         """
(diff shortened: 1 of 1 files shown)


samtranslator/region_configuration.py
```diff
@@ -28,11 +28,21 @@ def is_apigw_edge_configuration_supported(cls):
...
 
         :return: True, if SAR is supported in current region.
         """
-        return boto3.Session().region_name not in [
-            "af-south-1",
-        ]
+
+        session = boto3.Session()
+        # get the current region
+        region = session.region_name
```
aahung, line 40: is it possible this is `None` by any chance?

CoshUS (approved):

samtranslator/region_configuration.py
aahung: ok, seems `region` is possible to be `None` according the code you referenced, then `region.startswith("cn")` could crash in this case.

PR #2156, opened 2021-09-20
fix: Check GatewayResponse value type.

mildaniel:
Thanks to Ruperto for helping with this PR.

*Issue #, if available:*

*Description of changes:*

*Checklist:*

- [ ] Add/update tests using:
    - [ ] Correct values
- [ ] `make pr` passes
- [ ] Update documentation

*Examples?*

### DIFF
--- a/tests/translator/input/error_gateway_response_invalid_type_int.yaml
+++ b/tests/translator/input/error_gateway_response_invalid_type_int.yaml
@@ -1,0 +1,19 @@
+Resources:
+  ApiResource:
+    Type: AWS::Serverless::Api
+    Properties:
+      StageName: Prod
+      GatewayResponses:
+        DEFAULT_5XX:
+          ResponseParameters:
+            Headers:
+              Access-Control-Allow-Origin: '''*'''
+              Access-Control-Allow-Methods: '''GET,OPTIONS'''
+              Access-Control-Allow-Headers: '''content-type'''
+        DEFAULT_4XX:
+          ResponseParameters:
+            Headers:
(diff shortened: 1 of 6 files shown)


tests/translator/input/error_gateway_response_invalid_type_int.yaml
```diff
@@ -0,0 +1,19 @@
+Resources:
+  ApiResource:
+    Type: AWS::Serverless::Api
+    Properties:
+      StageName: Prod
+      GatewayResponses:
```
jfuss, line 6: I am not following how this results in a "type int" failure. Can you walk me through it so I can understand?

jfuss (changes requested):

tests/translator/input/error_gateway_response_invalid_type_int.yaml
mildaniel: `GatewayResponses` expects a `GatewayResponse` object. Here, you can see that there is a provided `StatusCode` field which is an `int`. Maybe a bit unclear in this test now that you mention it, but this is a very similar case to some the customer template issues we've been seeing.

marekaiv (approved):

tests/translator/input/error_gateway_response_invalid_type_int.yaml
jfuss: I see. Error is a little confusing to me but I don't have a better recommendation at the moment. Raising is also better then SAM silently failing so let's start here and see if the community has a better messaging.

jfuss (approved):

PR #2159, opened 2021-09-22
fix: Throw InvalidEventException when API Event Refs HttpAPI

jfuss:
If a customer miss configures their template by defining an Event of Type
API but Ref an AWS::Serverless::HttpApi, SAM just fails. This results in
a "service error" to customers but SAM should be throwing something helpful
back to customers to help them through the error.

In addition, I updated the test_translator.py tests for error cases to read
all the "error_" templates to test. This allows us not to manually add files
every time we add a new error case. This caught 3 files that were not being run
and had some miss configured output. The commit addresses those as well.

*Issue #, if available:*
n/a

*Description of changes:*
see above

Ran the new tests with and without the change. Fails before but passes after.

*Checklist:*

- [x] Add/update tests using:
    - [x] Correct values
- [x] `make pr` passes
- [ ] Update documentation

*Examples?*


marekaiv (approved):

samtranslator/plugins/api/implicit_api_plugin.py
```diff
@@ -173,8 +173,19 @@ def _add_api_to_swagger(self, event_id, event_properties, template):
         # Need to grab the AWS::Serverless::Api resource for this API event and update its Swagger definition
         api_id = self._get_api_id(event_properties)
 
+        # As of right now, this is for backwards compatability. SAM fails if you have an event type "Api" but that
+        # references "AWS::Serverless::HttpApi". If you do the opposite, SAM still outputs a valid template. Example of that
+        # can be found https://github.com/aws/serverless-application-model/blob/develop/tests/translator/output/api_with_any_method_in_swagger.json.
+        # One would argue that, this is unexpected and should actually fail. Instead of suddenly breaking customers in this
+        # position, we added a check to make sure the Plugin run (Http or Rest) is referencing an api of the same type.
```
mildaniel, line 180: How is the HttpApi event referencing an Api scenario being addressed? As far as I understand, this condition only checks for Api->HttpApi.

tests/translator/test_translator.py
```diff
@@ -32,6 +32,7 @@
 
 BASE_PATH = os.path.dirname(__file__)
 INPUT_FOLDER = os.path.join(BASE_PATH, "input")
+ERROR_FILES_NAMES_FOR_TESTING = [os.path.splitext(f)[0] for f in os.listdir(INPUT_FOLDER) if f.startswith("error_")]
```
mildaniel, line 35: Nice. This is much cleaner.

samtranslator/plugins/api/implicit_api_plugin.py
jfuss: Through the condition: 

```
is_referencing_http_from_api_event = (
            not template.get(api_id)
            or template.get(api_id).type == "AWS::Serverless::HttpApi"
            and not template.get(api_id).type == self.api_type
        )
```

So this `_add_api_to_swagger` is called when the `AWS::Serverless::Function.Event.Type` is `Api` or `HttpApi`, so we can generate the OpenApi (old name swagger). We are checking to see if the referenced Api is or type `AWS::Serverless::HttpApi` and the current API Plugin type matches the one it references.

mildaniel (approved):

PR #2161, opened 2021-09-27
fix: Check DependsOn field type.

mildaniel:
*Issue #, if available:*

*Description of changes:*

*Checklist:*

- [ ] Add/update tests using:
    - [ ] Correct values
- [x] `make pr` passes
- [ ] Update documentation

*Examples?*

### DIFF
--- a/samtranslator/model/eventsources/push.py
+++ b/samtranslator/model/eventsources/push.py
@@ -324,22 +324,18 @@ class S3(PushEventSource):
 
         depends_on = bucket.get("DependsOn", [])
 
-        # Validate that DependsOn is either a list or a string.
-        # If DependsOn is a list, also validate that all elements are also strings
-        check_invalid_type = not (isinstance(depends_on, string_types) or isinstance(depends_on, list)) or (
-            isinstance(depends_on, list) and not all(isinstance(el, str) for el in depends_on)
-        )
-        if check_invalid_type:
+        # DependsOn can be either a list of strings or a scalar string
+        if isinstance(depends_on, string_types):
+            depends_on = [depends_on]
+
+        try:
+            depends_on_set = set(depends_on)
+        except TypeError:
             raise InvalidResourceException(
                 self.logical_id,
                 "Invalid type for field 'DependsOn'. Expected a string or list of strings.",
             )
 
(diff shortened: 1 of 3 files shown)


samtranslator/model/eventsources/push.py
```diff
@@ -324,6 +324,17 @@ def _depend_on_lambda_permissions(self, bucket, permission):
 
         depends_on = bucket.get("DependsOn", [])
 
+        # Validate that DependsOn is either a list or a string.
+        # If DependsOn is a list, also validate that all elements are also strings
+        check_invalid_type = not (isinstance(depends_on, string_types) or isinstance(depends_on, list)) or (
+            isinstance(depends_on, list) and not all(isinstance(el, str) for el in depends_on)
+        )
+        if check_invalid_type:
```
hoffa, line 332: The check below are checks that `depends_on` is a string. Should probably be consolidated.

samtranslator/model/eventsources/push.py
```diff
@@ -324,6 +324,17 @@ def _depend_on_lambda_permissions(self, bucket, permission):
 
         depends_on = bucket.get("DependsOn", [])
 
+        # Validate that DependsOn is either a list or a string.
+        # If DependsOn is a list, also validate that all elements are also strings
+        check_invalid_type = not (isinstance(depends_on, string_types) or isinstance(depends_on, list)) or (
```
marekaiv, line 329: (minor) you can simplify this a bit and make it more readable by moving it a few lines down, after the `if isinstance(depends_on, string_types):` block -- at that point `depends_on` must be a list of strings.

hoffa (approved):

samtranslator/model/eventsources/push.py
```diff
@@ -328,7 +328,14 @@ def _depend_on_lambda_permissions(self, bucket, permission):
         if isinstance(depends_on, string_types):
             depends_on = [depends_on]
 
-        depends_on_set = set(depends_on)
+        try:
+            depends_on_set = set(depends_on)
+        except TypeError:
```
hoffa, line 333: Nice, much cleaner. 🚢

marekaiv (approved):

PR #2162, opened 2021-09-29
Add method to retrieve previously-recorded metrics

marekaiv:
*Issue #, if available:* n/a

*Description of changes:*
Add a method to the metrics object to retrieve previously recorded metrics.  This can be used to derive per-request computed metrics.

Existing unit tests for the class pass.  Also added a new unit test for the new method.

*Checklist:*

- [x] Add/update tests using:
    - [x] Correct values
    - [ ] ~[Intrinsic Functions](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference.html)~ n/a
- [x] `make pr` passes
- [ ] ~Update documentation~ n/a

### DIFF
--- a/samtranslator/metrics/metrics.py
+++ b/samtranslator/metrics/metrics.py
@@ -118,7 +118,7 @@ class Metrics:
         :param metrics_publisher: publisher to publish all metrics
         """
         self.metrics_publisher = metrics_publisher if metrics_publisher else DummyMetricsPublisher()
-        self.metrics_cache = []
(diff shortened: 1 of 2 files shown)


hawflau (approved):

mndeveci (approved):
LGTM! One small thing is once we switched to embedded metric (EMF), we won't need any caching for metrics because they will be just printed into logs. So at that time we can think about moving the caching one layer below, which is `MetricsPublisher` instances.

PR #2171, opened 2021-10-04
Added Exclusion for Non-Method Path Sections

CoshUS:
*Issue #, if available:*
OAS 3 added `summary` and `description` fields to path. SAM tries to resolve these fields as valid method sections.
https://swagger.io/docs/specification/paths-and-operations/
Example:
```
paths:
  /abc:
    summary: dummy summary
    get:
      x-amazon-apigateway-integration:
      contentHandling: CONVERT_TO_TEXT
```
```
method_definition.get(self._X_APIGW_INTEGRATION): -> "dummy summary".get(self._X_APIGW_INTEGRATION)
```

*Description of changes:*

*Checklist:*

- [ ] Add/update tests using:
    - [ ] Correct values
- [ ] `make pr` passes
- [ ] Update documentation

*Examples?*

### DIFF
--- a/samtranslator/swagger/swagger.py
+++ b/samtranslator/swagger/swagger.py
@@ -28,6 +28,7 @@ class SwaggerEditor(object):
     _CACHE_KEY_PARAMETERS = "cacheKeyParameters"
     # https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html
     _ALL_HTTP_METHODS = ["OPTIONS", "GET", "HEAD", "POST", "PUT", "DELETE", "PATCH"]
+    _EXCLUDED_PATHS_FIELDS = ["summary", "description", "parameters"]
     _POLICY_TYPE_IAM = "Iam"
     _POLICY_TYPE_IP = "Ip"
     _POLICY_TYPE_VPC = "Vpc"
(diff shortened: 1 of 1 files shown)


samtranslator/swagger/swagger.py
```diff
@@ -28,6 +28,7 @@ class SwaggerEditor(object):
     _CACHE_KEY_PARAMETERS = "cacheKeyParameters"
     # https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html
     _ALL_HTTP_METHODS = ["OPTIONS", "GET", "HEAD", "POST", "PUT", "DELETE", "PATCH"]
+    _EXCLUDED_PATHS_FIELDS = ["summary", "description", "parameters"]
```
marekaiv, line 31: Looking at https://swagger.io/specification/#path-item-object, should `servers` be in this list as well?  As well, should we add `trace` to `_ALL_HTTP_METHODS`?

marekaiv: That's true.  The thinking was that SAM could support the full standard and then SAM doesn't need to make adjustments if API GW adds support for `trace` (or other parts of the OpenAPI spec.)  In any case, it's a minor point.

marekaiv (approved):

PR #2172, opened 2021-10-04
fix: don't try to resolve the reference id if get attribute fields ar…

mndeveci:
…e not string

*Issue #, if available:*
When trying to resolve resource id's for `GetAtt` intrinsic function, if parameters are not string, it will fail to resolve it.

*Description of changes:*
We don't try to resolve them if parameters are not string, we will just return the given dict, and let CFN to resolve it.

Test with failure example and added unit tests

*Checklist:*

- [x] Add/update tests using:
    - [ ] Correct values
- [x] `make pr` passes
- [ ] Update documentation

*Examples?*

### DIFF
--- a/samtranslator/intrinsics/actions.py
+++ b/samtranslator/intrinsics/actions.py
@@ -490,6 +490,12 @@ class GetAttAction(Action):
         if not isinstance(value, list) or len(value) < 2:
             return input_dict
 
+        # If items in value array is not a string, then following join line will fail. So if any element is not a string
+        # we just pass along the input to CFN for doing the validation
+        for item in value:
+            if not isinstance(item, string_types):
+                return input_dict
+
         value_str = self._resource_ref_separator.join(value)
         splits = value_str.split(self._resource_ref_separator)
         logical_id = splits[0]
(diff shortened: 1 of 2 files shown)


marekaiv (approved):

samtranslator/intrinsics/actions.py
```diff
@@ -490,6 +490,12 @@ def resolve_resource_id_refs(self, input_dict, supported_resource_id_refs):
         if not isinstance(value, list) or len(value) < 2:
             return input_dict
 
+        # If items in value array is not a string, then following join line will fail. So if any element is not a string
+        # we just pass along the input to CFN for doing the validation
+        for item in value:
+            if not isinstance(item, string_types):
+                return input_dict
```
mgrandis, line 497: We may also want to update line 430.
There, we error out if one of the values is not a string.

Based on the [docs](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-getatt.html) (Supported Function, at the bottom of the page), it seems that the resource name must be a string but the attribute name can be a `Ref`.

We may also want to add a test in `tests/translator/input/intrinsic_functions.yaml` with this new case we allow.

PR #2174, opened 2021-10-05
Ensure that QueueLogicalId resolves to a string in case if it exist

moelasmar:
…ule events for lambda function and state machine resources

*Issue #, if available:*

*Description of changes:*

*Checklist:*

- [ ] Add/update tests using:
    - [x] Correct values
- [ ] `make pr` passes
- [ ] Update documentation

*Examples?*

### DIFF
--- a/samtranslator/model/eventbridge_utils.py
+++ b/samtranslator/model/eventbridge_utils.py
@@ -1,3 +1,5 @@
+from six import string_types
+
 from samtranslator.model.sqs import SQSQueue, SQSQueuePolicy, SQSQueuePolicies
 from samtranslator.model.exceptions import InvalidEventException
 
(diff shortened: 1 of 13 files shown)


tests/translator/input/function_with_event_bridge_rule_dlq_intrinsic_function.yaml
```diff
@@ -0,0 +1,40 @@
...
+              detail:
+                values:
+                  status:
+                    - PENDING_REVIEW
+              detail-type:
+                - 'UPDATED'
+            DeadLetterConfig:
+              Type: SQS
+              QueueLogicalId:
+                'Fn::Sub': 'testLambdaFunctionEBRuleDLQ${Env}'
```
jfuss, line 40: What if this is a `Ref`?

samtranslator/model/eventbridge_utils.py
```diff
@@ -41,12 +41,18 @@ def validate_dlq_config(source_logical_id, dead_letter_config):
...
         dlq_queue_arn = cw_event_source.DeadLetterConfig.get("Arn")
         if dlq_queue_arn is not None:
             return dlq_queue_arn, []
         queue_logical_id = cw_event_source.DeadLetterConfig.get("QueueLogicalId")
+        if (
+            isinstance(queue_logical_id, dict)
+            and list(queue_logical_id.keys())[0] in intrinsics_resolver.supported_intrinsics
+        ):
+            queue_logical_id = intrinsics_resolver.resolve_parameter_refs(queue_logical_id)
+            queue_logical_id = list(queue_logical_id.values())[0]
```
jfuss, line 55: Why resolve within this function instead of somewhere more central? The more we adhoc resolving in a bunch of places, the harder this will be to maintain over the long run.

jfuss (changes requested):

jfuss (approved):
Thanks for adding the negative tests!

PR #2180, opened 2021-10-11
Fix no allowed origin

c2tarun:
*Issue #, if available:*

*Description of changes:*
If a template with empty AllowOrigins is passed to SAM we detect and throw `ValueError`, this error is then left uncaught and causes service error. This PR Is converting that `ValueError` to a SAM specific error which is then handled properly and proper error message is returned as part of transform failure.

Added functional test.

*Checklist:*

- [x] Add/update tests using:
    - [ ] Correct values (already exist)
- [x] `make pr` passes
- [ ] Update documentation

*Examples?*


jfuss (approved):
LGTM

wchengru (approved):

PR #2199, opened 2021-11-01
Raise InvalidResourceException when UsagePlan type is not correct

aahung:
*Issue #, if available:*

*Description of changes:*

*Checklist:*

- [x] Add/update tests using:
    - [ ] Correct values
- [x] `make pr` passes
- [ ] Update documentation

*Examples?*

### DIFF
--- a/samtranslator/model/api/api_generator.py
+++ b/samtranslator/model/api/api_generator.py
@@ -697,6 +697,9 @@ class ApiGenerator(object):
         if auth_properties.UsagePlan is None:
             return []
         usage_plan_properties = auth_properties.UsagePlan
+        # throws error if UsagePlan is not a dict
+        if not isinstance(usage_plan_properties, dict):
+            raise InvalidResourceException(self.logical_id, "'UsagePlan' must be a dictionary")
         # throws error if the property invalid/ unsupported for UsagePlan
         if not all(key in UsagePlanProperties._fields for key in usage_plan_properties.keys()):
             raise InvalidResourceException(self.logical_id, "Invalid property for 'UsagePlan'")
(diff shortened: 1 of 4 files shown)


tests/model/api/test_api_generator.py
```diff
@@ -0,0 +1,24 @@
...
+
+from parameterized import parameterized
+
+from samtranslator.model import InvalidResourceException
+from samtranslator.model.api.api_generator import ApiGenerator
+
+
+class TestApiGenerator(TestCase):
+    @parameterized.expand([("this should be a dict",), ("123",), ([{}],)])
+    @patch("samtranslator.model.api.api_generator.AuthProperties")
```
hoffa, line 12: Don't we have some way (functional tests?) to pass a template and ensure it fails some expected manner? Instead of requiring all this mocking.

hoffa (approved):

tests/model/api/test_api_generator.py
aahung: Good call. Added one for the invalid template I intended to fix.

samtranslator/model/api/api_generator.py
```diff
@@ -697,6 +697,9 @@ def _construct_usage_plan(self, rest_api_stage=None):
         if auth_properties.UsagePlan is None:
             return []
         usage_plan_properties = auth_properties.UsagePlan
+        # throws error if UsagePlan is not a dict
+        if not isinstance(usage_plan_properties, dict):
+            raise InvalidResourceException(self.logical_id, "Invalid type for field 'UsagePlan'")
```
marekaiv, line 702: (minor) it might be more user-friendly to tell them the expected type `"UsagePlan" must be a dictionary` rather than just saying it's invalid.

marekaiv (approved):

samtranslator/model/api/api_generator.py
aahung: good one, updated

PR #2207, opened 2021-11-04
[FOSS] hot fix and improvement

mingkun2020:
1. Companion stack for MSK, MQ tests
2. Throttling handle mechanism
3. Option for adding pipeline name as testing stack prefix
4. Update template runtime version


integration/combination/test_function_with_alias.py
```diff
@@ -71,12 +71,16 @@ def test_function_with_alias_with_intrinsics(self):
...
         # This should create a new version and leave existing version intact
         parameters[1]["ParameterValue"] = "code2.zip"
-        self.deploy_stack(parameters)
+        # self.deploy_stack(parameters)
+        self.update_stack("combination/function_with_alias_intrinsics", parameters)
         version_ids = get_function_versions(function_name, self.client_provider.lambda_client)
-        self.assertEqual(["1", "2"], version_ids)
-
+        # only show 1 version, need to investigate why
+        # self.assertEqual(["1", "2"], version_ids)
```
mndeveci, line 78: Why we have commented out tests?

integration/helpers/stack.py
```diff
@@ -0,0 +1,75 @@
...
+    def _get_stack_description(self):
+        try:
+            self.stack_description = self.cfn_client.describe_stacks(StackName=self.stack_name)
+        except botocore.exceptions.ClientError as ex:
+            if "Throttling" in str(ex):
+                raise ThrottlingError(stack_name=self.stack_name, msg=str(ex))
+            raise
+
+    @staticmethod
+    def _generate_out_put_file_path(file_path, output_dir):
```
mndeveci, line 70: Nit: `output`

integration/combination/test_function_with_sns.py
```diff
@@ -1,6 +1,10 @@
+from unittest.case import skipIf
+
 from integration.helpers.base_test import BaseTest
+from integration.helpers.resource import current_region_does_not_support
 
 
+@skipIf(current_region_does_not_support(["SQS"]), "SQS is not supported in this testing region")
```
mndeveci, line 7: Test is SNS but we are checking SQS, is that correct?

mndeveci (dismissed):

integration/combination/test_api_with_authorizers.py
```diff
@@ -1,10 +1,14 @@
...
+
 import requests
 
 from integration.helpers.base_test import BaseTest
 from integration.helpers.deployer.utils.retry import retry
 from integration.helpers.exception import StatusCodeError
+from integration.helpers.resource import current_region_does_not_support
 
 
+@skipIf(current_region_does_not_support(["Cognito"]), "Cognito is not supported in this testing region")
```
CoshUS, line 11: I wonder whether we should put service strings like "Cognito" into constants so that there will be less likely to have typos in the future.

integration/resources/templates/combination/function_with_all_event_types.yaml
```diff
@@ -151,7 +151,7 @@ Outputs:
         - TestSchedule${__StackName__}
         - __StackName__: 
             Fn::Select:
-              - 3
+              - 6
```
CoshUS, line 154: This is changed to 6 because we are adding prefixes to the stack name?
Seems really prone to errors if we ever change the stack names and miss updating this index for all the tests. 
Can we use a different method like passing in parameters?

integration/single/test_basic_api.py
```diff
@@ -41,6 +47,14 @@ def test_basic_api_with_mode(self):
         self.update_and_verify_stack("single/basic_api_with_mode_update")
         response = requests.get(f"{api_endpoint}/get")
```
CoshUS, line 48: nit: This line is not needed anymore.

integration/helpers/deployer/utils/retry.py
```diff
@@ -38,3 +39,38 @@ def wrapper(*args, **kwargs):
...
+        @wraps(func)
+        def wrapper(*args, **kwargs):
+            remaining_attempts = attempts
+            retry_attempt = 1
+
+            while remaining_attempts >= 1:
+                try:
+                    return func(*args, **kwargs)
+                except exc:
+                    sleep_time = random.uniform(0, math.pow(2, retry_attempt) * delay)
```
CoshUS, line 68: What's the reason for adding jitter?
Shouldn't sleep time be calculated by `delay +/- jitter` instead of a random int between 0 to max delay?
The current approach on a 10th retry can range from 0 to 512 seconds. This seems to be a large range and I'm not sure what that will accomplish.

CoshUS (changes requested):

integration/combination/test_function_with_sns.py
mingkun2020: Just copy from the java tests, will change it to SNS.

integration/helpers/deployer/utils/retry.py
mingkun2020: The Cloudformation's`describe_stack` api that we use to check the stack status has hard limit and cannot be changed. Since we run the tests in parallel(for example 10 thread in concurrent), we will have 10 calls in the same time(imagine a big call cluster in a small time frame). So we need to add jitter(add some randomness) to spread the calls over a large time frame to avoid the throttling. If we just use exponential back off, the call cluster will appear once again after exponential time and won't solve the issue, therefore we need the jitter.

CoshUS: Why is the final delay generated with a random between 0 to max delay instead of plus minus a percentage of the max delay? 
In the example of 10th retry, it could have only 1 second delay which defeats the purpose of exponential backoff.

CoshUS: Green as minimal, Red as maximum, and Purple as average.
This is the current implementation:
chrome_f5Cv4MNo4W

With a percentage based jitter:
chrome_SX71SQbQRJ

mingkun2020: Our goal is trying to spread the api call among the test time frame as even as possible(If we use delay + jitter we might have some time with no call). Thinking now how to do some simulation about this.

CoshUS: Tracking:

Current implementation is following the blog post: https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/

One assumption the blog post makes is that the calls will continue to happen until the job is complete. However, in our case, we will fail the test if it fails after an x amount of retries. Although the total amount of "Work" is less, there could be a case where a job rolls low numbers for 5 times and fails the execution.
I think there is a better solution for our use case than either full jitter or equal jitter.

@mingkun2020 will work on a simulator for our use case specifically to test out different approaches.

This comment is not blocking.

integration/combination/test_function_with_alias.py
mingkun2020: leftover from testing, will deleted it.

integration/combination/test_api_with_authorizers.py
mingkun2020: Updated with using constant as service name

integration/single/test_basic_api.py
mingkun2020: deleted

integration/resources/templates/combination/function_with_all_event_types.yaml
mingkun2020: Updated with passing parameter.

CoshUS (approved):
LGTM! Thanks for addressing the comments!

integration/combination/test_api_settings.py
```diff
@@ -1,5 +1,7 @@
 import hashlib
 
+import pytest
```
qingchm, line 3: Is this import used?

mingkun2020: Unused import, removed it.

integration/combination/test_function_with_deployment_preference.py
```diff
@@ -97,7 +102,9 @@ def _get_deployment_groups(self, application_name):
         ]
 
     def _get_deployments(self, application_name, deployment_group):
-        deployments = self.client_provider.code_deploy_client.list_deployments()["deployments"]
+        deployments = self.client_provider.code_deploy_client.list_deployments(
+            applicationName=application_name, deploymentGroupName=deployment_group
+        )["deployments"]
```
qingchm, line 107: For client responses like this are we catching any possible client errors? Also if the client returned a None response due to some error ["deployments"] would also be error prone? (Maybe get("deployments") would be a better option

mingkun2020: If it returns a None, it will break and the test will failed as expected(it shouldn't be none).

qingchm (approved):

PR #2215, opened 2021-11-08
Fix DestinationConfig in streaming event sources

hawflau:
*Issue #, if available:*

*Description of changes:*
Fix `DestinationConfig` of streaming event sources by ensuring `OnFailure` is defined before it's used.
If `OnFailure` is missing, an error will be thrown.

*Checklist:*

- [x] Add/update tests using:
    - [x] Correct values
- [x] `make pr` passes
- [ ] Update documentation

*Examples?*

### DIFF
--- a/samtranslator/model/eventsources/pull.py
+++ b/samtranslator/model/eventsources/pull.py
@@ -110,6 +110,9 @@ class PullEventSource(ResourceMacro):
 
         destination_config_policy = None
         if self.DestinationConfig:
+            if self.DestinationConfig.get("OnFailure") is None:
+                raise InvalidEventException(self.logical_id, "'OnFailure' is a required field for 'DestinationConfig'")
+
             # `Type` property is for sam to attach the right policies
             destination_type = self.DestinationConfig.get("OnFailure").get("Type")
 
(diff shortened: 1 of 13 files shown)


samtranslator/model/eventsources/pull.py
```diff
@@ -110,6 +110,9 @@ def to_cloudformation(self, **kwargs):
 
         destination_config_policy = None
         if self.DestinationConfig:
+            if self.DestinationConfig.get("OnFailure") is None:
+                raise InvalidEventException(self.logical_id, "'OnFailure' is a required field for 'DestinationConfig'")
```
marekaiv, line 114: The [docs](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-lambda-eventsourcemapping-destinationconfig.html) for `DestinationConfig` show that `OnFailure` is not a required field.  I'm not sure why someone would create `DestinationConfig` with its only possible field so I think the docs are incorrect.  Either way, we should make the code and the docs consistent.

marekaiv (approved):

samtranslator/model/eventsources/pull.py
hawflau: There's a discrepancy between doc and code. In addition to the one you mentioned above, the doc is also missing to show `Type` as required under `OnFailure`. I'll request a doc update to address these inconsistencies.

moelasmar (approved):

PR #2216, opened 2021-11-09
fix: handle non-dict DefinitionBody path item in _openapi_postprocess

torresxb1:
*Issue #, if available:*

*Description of changes:*
`_openapi_postprocess` expects the value at a path to be a dict. The following template would cause an unhandled error: 
```
Resources:
  ApiWithInvalidPath:
    Type: AWS::Serverless::Api
    Properties:
      StageName: Prod
      OpenApiVersion: 3.0.1
      DefinitionBody:
        openapi: 3.0.1
        info:
          title: test invalid paths Api
        paths:
          /foo: null
```

This PR checks that it is a dict and throws an error if it's not. Also did some refactoring while I updated this part of the code.

Added end-to-end tests

*Checklist:*

- [x] Add/update tests using:
    - [ ] Correct values
- [x] `make pr` passes
- [x] Update documentation

*Examples?*

### DIFF
--- a/samtranslator/swagger/swagger.py
+++ b/samtranslator/swagger/swagger.py
@@ -139,16 +139,7 @@ class SwaggerEditor(object):
         method = self._normalize_method_name(method)
 
         path_dict = self.paths.setdefault(path, {})
-
-        if not isinstance(path_dict, dict):
-            # Either customers has provided us an invalid Swagger, or this class has messed it somehow
-            raise InvalidDocumentException(
-                [
-                    InvalidTemplateException(
-                        "Value of '{}' path must be a dictionary according to Swagger spec.".format(path)
-                    )
-                ]
-            )
+        SwaggerEditor.validate_path_item_is_dict(path_dict, path)
 
         if self._CONDITIONAL_IF in path_dict:
             path_dict = path_dict[self._CONDITIONAL_IF][1]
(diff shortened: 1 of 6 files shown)


samtranslator/swagger/swagger.py
```diff
@@ -1244,6 +1224,31 @@ def is_valid(data):
...
+        Throws exception if path_item is not a dict
+
+        :param path_item: path_item (value at the path) being validated
+        :param path: path name
+        """
+
+        SwaggerEditor.validate_is_dict(
+            path_item, "Value of '{}' path must be a dictionary according to Swagger spec.".format(path)
+        )
+
```
torresxb1, line 1251: not sure where's the best place for this, but thought this was good for now since some validation is done in this class. Also am I supposed to also add this to `open_api.py`?

torresxb1: because this same action would be done in swagger.py and api_generator.py. I mainly just didn't want to repeat the error message and make sure that it was consistent in every place if it's changed in the future. Not sure if there's a better way of doing it, where I just generate the error message.

moelasmar: ok

moelasmar (approved):

hoffa (approved):

PR #2222, opened 2021-11-15
Run unit tests in parallel

aahung:
*Issue #, if available:*

*Description of changes:*

Run unit tests in parallel, on my 16-inch macbook pro, the time of `make test` is reduced from 227.18s to 64.78s, which is a 72% reduction.

*Checklist:*

- [ ] Add/update tests using:
    - [ ] Correct values
- [x] `make pr` passes
- [ ] Update documentation

*Examples?*

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ init:
 	pip install -e '.[dev]'
 
 test:
-	pytest --cov samtranslator --cov-report term-missing --cov-fail-under 95 tests/*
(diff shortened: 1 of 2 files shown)


hoffa (approved):

mndeveci (approved):
Nice! Can we use the same approach in SAM CLI as well?

hawflau (approved):

aahung:
> Nice! Can we use the same approach in SAM CLI as well?

yep we can

PR #2224, opened 2021-11-15
Conditionally adding metadata to translated resources

andrew-glenn:
Exposing an additional `translated_resource_mapping` property in the `Translator` class allows downstream consumers to apply logic to translated resources. 

An example use case: copying cfn-lint metadata to new resources so that rule exceptions are honored.

See: https://github.com/aws-cloudformation/cfn-lint/pull/2173


moelasmar (changes requested):
could you please add some unit testing cases for this new change?

moelasmar (approved):

jfuss:
@andrew-glenn It isn't clear what the purpose of this is for. Is it only for local tooling support? Why do you want to mutate underlying resources instead of the SAM Resources itself?

jfuss (changes requested):
"Request changes" to indicate not to merge until the use-case is more understood.

andrew-glenn:
@jfuss As an example, `cfn-lint` supports metadata per-resource to define specific configuration options. A byproduct of the SAM Translator is that a single resource is split off into multiple, without the original metadata being copied to the new resources.

Rather than mess with the translator logic - which I'm not familiar with, and has a massive blast radius - I opted to surface a mapping of original SAM resource names --> Generated CFN Resources.

Providing this mapping allows me (downstream) to iterate over the map and perform whatever actions I wish based on this data. 

Admittedly, this property wouldn't be used directly via the SAM CLI, but rather those that import SAM as a python library.

moelasmar:
@jfuss .. my understanding is that change will not affect the transformation logic, but will add some value for the sam translator library.

jfuss:
@moelasmar 

@andrew-glenn and I discussed over slack yesterday afternoon. It is true this wouldn't impact the service but does mean we have to support this "overriding" locally. In the ideal world, I would like `cfn-lint` to lint the SAM Resources not the underlying ones but that requires a spec from SAM (this also decouples `cfn-lint` and SAM, which is an added benefit). I want to understand more of the intention behind the change and why we should support this. Not trying to say this isn't something we should do but there might be other ways to do the same thing. For example, maybe SAM should be taking the Metadata and applying it to all the underlying resources. Then we don't have to this local only "patching" mechanism, making support easier on us. 

I scoped some time for myself on Thursday to think a little deeper and reach back out to @andrew-glenn to understand what he is trying to achieve with this and `cfn-lint`. That should give me a better understanding on what we may want to do here. My initially thinking is either 1) apply Metadata to generated resources or 2) Apply Metadata to generated resources so but make it a flag in the transform call. 

Open to other ideas here, just general cautious of features that only apply locally.

andrew-glenn:
@jfuss @moelasmar 

I've updated the PR to reflect the new approach (adding metadata in translation). I've set it to default disabled for backwards compatibility.

tests/translator/output/translate_convert_metadata.json
```diff
@@ -0,0 +1,93 @@
+{
+  "Resources": {
+    "Images": {
+      "Type": "AWS::S3::Bucket",
```
jfuss, line 4: The bucket is defined in the template above, I don't think we want SAM to mutate that resource. I know we do today, but that is in very specific use-cases (adding events triggers).

samtranslator/translator/translator.py
```diff
@@ -152,7 +153,10 @@ def translate(self, sam_template, parameter_values, feature_toggle=None):
                 del template["Resources"][logical_id]
                 for resource in translated:
                     if verify_unique_logical_id(resource, sam_template["Resources"]):
-                        template["Resources"].update(resource.to_dict())
+                        _r = resource.to_dict()
+                        if resource_dict.get("Metadata") and convert_metadata:
```
jfuss, line 157: I think a comment here could go a long away in understanding what is happening. It took me a while but I think what is happening is the following:
For each SAM resource in the source template, we translate to the underlying CloudFormation Resources. After we translate, we append any metadata in the SAM Resource to the translated CloudFormation Resources.

They generally works but I think there might be a couple edge cases outstanding:
1) S3 buckets are updated (I made a comment below about this)
2) I am not sure how generated resources work in this case. If you have an API Event on an `AWS::Serverless::Function.Event` what happens? Does the Metadata get pass through correctly? That is the `Metadata` on `AWS::Serverless::Function` gets added to all the API Gateway resources? This would be a good test case to add regardless of it it works in the current implementation or not.

jfuss (changes requested):

samtranslator/translator/translator.py
```diff
@@ -152,7 +153,10 @@ def translate(self, sam_template, parameter_values, feature_toggle=None):
                 del template["Resources"][logical_id]
                 for resource in translated:
                     if verify_unique_logical_id(resource, sam_template["Resources"]):
-                        template["Resources"].update(resource.to_dict())
+                        _r = resource.to_dict()
```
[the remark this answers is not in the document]
andrew-glenn: The current  implementation (Line 155) converts it to a dictionary. I assumed this was done for a Very Good(tm) reason, so I simply followed the existing pattern.

tests/translator/output/translate_convert_metadata.json
andrew-glenn: I'm not 100% seeing how this resource is being mutated, since it exists previously. What am I overlooking?

andrew-glenn: Note, to save back and forth:

If we want to exclude certain resource types from the metadata passthrough, that's fine. I'll need a list of resource types that would apply here. I'm a little bit in the dark.

samtranslator/translator/transform.py
```diff
@@ -2,7 +2,7 @@
 from samtranslator.parser.parser import Parser
 
 
-def transform(input_fragment, parameter_values, managed_policy_loader, feature_toggle=None):
+def transform(input_fragment, parameter_values, managed_policy_loader, feature_toggle=None, convert_metadata=False):
```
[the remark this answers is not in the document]
andrew-glenn: Sure, no problem.

samtranslator/translator/translator.py
andrew-glenn: 1. see above

2. Yes. A single SAM resource translates to multiple CFN resources. The metadata on the single SAM resource is passed through to the generated CFN resources, regardless of what the resource type is.

jfuss (approved):

jfuss:
@moelasmar Can you re-review this? Things have changed since you last review

moelasmar (approved):

PR #2234, opened 2021-11-22
Handle when function http event auth is not a string

aahung:
*Issue #, if available:*

*Description of changes:*

*Checklist:*

- [x] Add/update tests using:
    - [ ] Correct values
- [x] `make pr` passes
- [ ] Update documentation

*Examples?*


hoffa (approved):

samtranslator/model/eventsources/push.py
```diff
@@ -1219,6 +1219,11 @@ def _add_auth_to_openapi_integration(self, api, editor):
                     "an explicit 'Authorizer' set or a 'DefaultAuthorizer' "
                     "configured on the HttpApi.",
                 )
+        elif not isinstance(method_authorizer, string_types):
```
marekaiv, line 1222: `method_authorizer` might have been updated in the preceding `if` ... do we know that `DefaultAuthorizer` is a string (was that previously checked?). If not, we could consider turning this into an `if` rather than `elif`.

marekaiv (approved):

samtranslator/model/eventsources/push.py
aahung: Good call, `DefaultAuthorizer` being a dict should not produce this error message (instead, it has its own message). Fixed. I also added an extra unit test to ensure only `DefaultAuthorizer` related error message is generated when it happens.

marekaiv (approved):

PR #2240, opened 2021-11-23
Do not abort SAR loop on throttling

marekaiv:
*Issue #, if available:*
n/a

*Description of changes:*
Currently, if an API call to Serverless Application Repository (SAR) is throttled while querying for the status of an application, the transform fails.  This change allows the transform to sleep and then retry, up to the time limit configured in code.

Unit tests were added.

*Checklist:*

- [x] Add/update tests using:
    - [x] Correct values
    - [ ] ~[Intrinsic Functions](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference.html)~ - n/a
- [x] `make pr` passes
- [ ] ~Update documentation~ - n/a
- [ ] ~Verify transformed template deploys and application functions as expected~ no changes to transformation logic

*Examples?*

### DIFF
--- a/samtranslator/plugins/application/serverless_app_plugin.py
+++ b/samtranslator/plugins/application/serverless_app_plugin.py
@@ -180,9 +180,6 @@ class ServerlessAppPlugin(BasePlugin):
             warning_message = "{}. Unable to verify access to {}/{}.".format(e, app_id, semver)
             LOG.warning(warning_message)
             self._applications[key] = {"Unable to verify"}
-        except ClientError as e:
-            LOG.exception(e)
-            raise e
 
     def _handle_create_cfn_template_request(self, app_id, semver, key, logical_id):
         """
@@ -197,11 +194,7 @@ class ServerlessAppPlugin(BasePlugin):
         create_cfn_template = lambda app_id, semver: self._sar_client.create_cloud_formation_template(
             ApplicationId=self._sanitize_sar_str_param(app_id), SemanticVersion=self._sanitize_sar_str_param(semver)
         )
-        try:
-            response = self._sar_service_call(create_cfn_template, logical_id, app_id, semver)
-        except ClientError as e:
-            LOG.exception(e)
-            raise e
+        response = self._sar_service_call(create_cfn_template, logical_id, app_id, semver)
 
         LOG.info("Requested to create CFN template {}/{} in serverless application repo.".format(app_id, semver))
         self._applications[key] = response[self.TEMPLATE_URL_KEY]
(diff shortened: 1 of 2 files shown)


samtranslator/plugins/application/serverless_app_plugin.py
```diff
@@ -180,6 +180,9 @@ def _handle_get_application_request(self, app_id, semver, key, logical_id):
             warning_message = "{}. Unable to verify access to {}/{}.".format(e, app_id, semver)
             LOG.warning(warning_message)
             self._applications[key] = {"Unable to verify"}
+        except ClientError as e:
+            LOG.exception(e)
+            raise e
```
jfuss, line 185: Doesn't the exception get logged if this is raised and not caught? Any specific reason we added this here?

samtranslator/plugins/application/serverless_app_plugin.py
```diff
@@ -318,8 +327,26 @@ def on_after_transform_template(self, template):
...
+                    response = {"Status": "PREPARING"}  # default response if we can't reach SAR
+
+                    try:
+                        if not throttled:
+                            response = self._sar_service_call(
+                                get_cfn_template, application_id, application_id, template_id
+                            )
+                    except ClientError as e:
+                        error_code = e.response["Error"]["Code"]
+                        if error_code == "TooManyRequestsException":
```
jfuss, line 339: Why only catch this here? Can't this happen in the `create_cloud_formation_template`?

samtranslator/plugins/application/serverless_app_plugin.py
```diff
@@ -372,9 +399,6 @@ def _sar_service_call(self, service_call_lambda, logical_id, *args):
             error_code = e.response["Error"]["Code"]
             if error_code in ("AccessDeniedException", "NotFoundException"):
                 raise InvalidResourceException(logical_id, e.response["Error"]["Message"])
-
```
jfuss, line 375: Is there any reason we shouldn't do the Throttle handling within this (`_sar_service_call`) function? Seems odd to do the checking in multiple places.

samtranslator/plugins/application/serverless_app_plugin.py
```diff
@@ -308,6 +316,7 @@ def on_after_transform_template(self, template):
             while (time() - start_time) < self.TEMPLATE_WAIT_TIMEOUT_SECONDS:
                 temp = self._in_progress_templates
                 self._in_progress_templates = []
+                throttled = False
```
jfuss, line 319: Should we introduce some jitter? 

Should we update our Boto config to be [standard](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/retries.html#standard-retry-mode) (default is legacy)? Seems like that provides a better default retry behavior and I think standard is the newer recommendation.

samtranslator/plugins/application/serverless_app_plugin.py
```diff
@@ -308,6 +316,7 @@ def on_after_transform_template(self, template):
             while (time() - start_time) < self.TEMPLATE_WAIT_TIMEOUT_SECONDS:
```
jfuss, line 316: This resets the `throttle` flag. Is the idea that the 2s sleep will give us enough break from calling that we can safely call SAR again? Might be worth commenting about line 319 to make it clear. This method is pretty hard to parse (lots of nesting).

tests/plugins/application/test_serverless_app_plugin.py
```diff
@@ -253,9 +255,53 @@ def __init__(self, app_id="app_id", semver="1.3.5"):
 
 #     self.plugin.on_before_transform_resource(app_resources[0][0], 'AWS::Serverless::Application', app_resources[0][1].properties)
 
-# class TestServerlessAppPlugin_on_after_transform_template(TestCase):
 
-#     def setUp(self):
-#         self.plugin = SeverlessAppPlugin()
-
-# # TODO: test this lifecycle event
+class TestServerlessAppPlugin_on_after_transform_template(TestCase):
```
jfuss, line 259: Interesting we didn't have test for this block before :(

Thanks for adding them.

tests/plugins/application/test_serverless_app_plugin.py
```diff
@@ -253,9 +255,53 @@ def __init__(self, app_id="app_id", semver="1.3.5"):
...
+    def setUp(self):
+        pass
```
jfuss, line 261: I think you can remove this, since it is only a `pass`.

jfuss (dismissed):

samtranslator/plugins/application/serverless_app_plugin.py
marekaiv: I wondered the same thing -- I added this here to preserve the existing behavior.  I removed `LOG.exception` from `_sar_service_call` as throttles are not really exceptions for `GetCloudFormation`.  I can do a bit more digging to see if we need that extra log.

samtranslator/plugins/application/serverless_app_plugin.py
marekaiv: It probably can -- we just haven't seen it in any availability dips.  There's no loop around the create at the moment so we'd either have to add a loop (to `_sar_service_call` if we want it to be common) or consider throttling an error.

samtranslator/plugins/application/serverless_app_plugin.py
marekaiv: Yes, we should -- I'd suggest doing that in a future PR so we're introducing one change at a time.  I might have misread the botocore code but I thought even the legacy retry had some [jitter](https://github.com/boto/botocore/blob/23ee17f5446c78167ff442302471f9928c3b4b7c/botocore/data/_retry.json#L96) built into it.

samtranslator/plugins/application/serverless_app_plugin.py
```diff
@@ -318,8 +329,26 @@ def on_after_transform_template(self, template):
...
                         )
                     )
-                    response = self._sar_service_call(get_cfn_template, application_id, application_id, template_id)
+                    response = {"Status": "PREPARING"}  # default response if we can't reach SAR
+
+                    try:
+                        if not throttled:
+                            response = self._sar_service_call(
+                                get_cfn_template, application_id, application_id, template_id
+                            )
```
aahung, line 338: since now we manually skip boto3 calls when throttled, it looks like it will go to next iteration immediately, which is likely to result in another throttle. Do you think we should add `sleep` at the end of the iteration if `throttled` is `True`?

aahung (commented):
Question: is it possible we can tune the boto3 sar client's retry strategy instead of having a retrying loop. 

What's the difference between these two?
1. checking SAR A -> B -> C -> A -> B -> C -> A -> B -> C
2. checking SAR A -> A -> A -> B -> B -> B -> C -> C -> C

samtranslator/plugins/application/serverless_app_plugin.py
marekaiv: I should (will?) refactor this code to make it easier to read.  Once `throttled` is `True`, we will continue looping but will not make any more SAR calls.  We continue looping to in order to call `_handle_get_cfn_template_response`.  Once the for loop is done, we will `sleep(self.SLEEP_TIME_SECONDS)` before going through the `while` again, which resets `Throttled`

aahung: Oh I see, just found the line.

samtranslator/plugins/application/serverless_app_plugin.py
marekaiv: The check for throttling and the retry is only in one place (where we're trying to retrieve the CloudFormation template).  The SAR Create call currently does not have a retry and we have not yet seen it get throttled.  If we start seeing it, we could implement a retry in `_sar_service_call`.

samtranslator/plugins/application/serverless_app_plugin.py
```diff
@@ -303,57 +304,73 @@ def on_after_transform_template(self, template):
...
-        status = response["Status"]
-        if status != "ACTIVE":
-            # Other options are PREPARING and EXPIRED.
-            if status == "EXPIRED":
-                message = (
-                    "Template for {} with id {} returned status: {}. Cannot access an expired "
-                    "template.".format(application_id, template_id, status)
-                )
-                raise InvalidResourceException(application_id, message)
-            self._in_progress_templates.append((application_id, template_id))
```
aahung, line 356: this one was hidden here, I like the new change

aahung (approved):

hawflau (approved):

PR #2250, opened 2021-11-29
Validate swagger method value is a valid dict before processing

aahung:
*Issue #, if available:*

*Description of changes:*

*Checklist:*

- [ ] Add/update tests using:
    - [ ] Correct values
- [x] `make pr` passes
- [ ] Update documentation

*Examples?*

### DIFF
--- a/samtranslator/swagger/swagger.py
+++ b/samtranslator/swagger/swagger.py
@@ -537,12 +537,18 @@ class SwaggerEditor(object):
             if normalized_method_name in SwaggerEditor._EXCLUDED_PATHS_FIELDS:
                 continue
             if add_default_auth_to_preflight or normalized_method_name != "options":
-                normalized_method_name = self._normalize_method_name(method_name)
+                SwaggerEditor.validate_is_dict(
+                    method,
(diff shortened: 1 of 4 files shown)


marekaiv (approved):

hoffa (approved):

samtranslator/swagger/swagger.py
```diff
@@ -537,12 +537,18 @@ def set_path_default_authorizer(
...
+                )
                 # It is possible that the method could have two definitions in a Fn::If block.
                 for method_definition in self.get_method_contents(method):
 
                     SwaggerEditor.validate_is_dict(
-                        method_definition, "{} for path {} is not a valid dictionary.".format(method_definition, path)
+                        method_definition,
+                        'Value of "{}" ({}) for path {} is not a valid dictionary.'.format(
+                            method_name, method_definition, path
+                        ),
```
torresxb1, line 551: what's the best way we can avoid having duplicate error messages (like this and the one couple lines above)? To handle this with path items, I created `validate_path_item_is_dict`, but I'm not sure if there's a better way of handling this (maybe a constant for the string, or a function to get the formatted string, or is it good as is?).

PR #2252, opened 2021-11-29
fix: Check type of resource type is a string

mildaniel:
We should verify that the type of resource type is a string before trying to hash it. It should then work similarly to the current behaviour. If the type of a resource is invalid, we should skip that resource and continue processing the other resources.

*Issue #, if available:*

*Description of changes:*

*Checklist:*

- [x] Add/update tests using:
    - [ ] Correct values
- [x] `make pr` passes
- [ ] Update documentation

*Examples?*

### DIFF
--- a/samtranslator/model/__init__.py
+++ b/samtranslator/model/__init__.py
@@ -489,10 +489,7 @@ class ResourceTypeResolver(object):
                 self.resource_types[resource_class.resource_type] = resource_class
 
     def can_resolve(self, resource_dict):
-        if (
-            not isinstance(resource_dict, dict)
-            or not isinstance(resource_dict.get('Type'), string_types)
-        ):
+        if not isinstance(resource_dict, dict) or not isinstance(resource_dict.get("Type"), string_types):
             return False
 
         return resource_dict["Type"] in self.resource_types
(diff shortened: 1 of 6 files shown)


samtranslator/model/__init__.py
```diff
@@ -489,7 +489,11 @@ def __init__(self, *modules):
                 self.resource_types[resource_class.resource_type] = resource_class
 
     def can_resolve(self, resource_dict):
-        if not isinstance(resource_dict, dict) or "Type" not in resource_dict:
+        if (
+            not isinstance(resource_dict, dict)
+            or "Type" not in resource_dict
+            or not isinstance(resource_dict["Type"], string_types)
```
aahung, line 495: nit, these two lines can be combined as `not isinstance(resource_dict.get('Type'), string_types)`

aahung (approved):
LGTM.
on question: Do we want to skip them then pass it to CFN or quickly fail with a helpful error messages? Since customers cannot get the template with one of the resource being a dict, rejecting as an invalid template won't break backward compatibility

mildaniel:
Thanks @aahung. 

> Do we want to skip them then pass it to CFN or quickly fail with a helpful error messages?

I was debating this myself, but went with our current approach which is to just skip over the unreadable resources since it seems less likely to impact existing customers. This check is also done before we start aggregating our customer transform exceptions so it would require a little more work to correctly fail at this point.

moelasmar (approved):

PR #2263, opened 2021-12-14
handle 'Invalid Swagger Document' and refactor some validation into Swagger Editor constructor

torresxb1:
*Issue #, if available:*

*Description of changes:*
Mostly refactoring to make validation a bit less messy. 
Also handles `Invalid Swagger Document` ValueError. 

*Checklist:*

- [x] Add/update tests using:
    - [ ] Correct values
- [x] `make pr` passes
- [x] Update documentation

*Examples?*


samtranslator/model/api/api_generator.py
```diff
@@ -237,6 +237,8 @@ def __init__(
         self.template_conditions = template_conditions
         self.mode = mode
 
+        self.swagger_editor = SwaggerEditor(self.definition_body) if self.definition_body else None
```
torresxb1, line 240: A lot of the validation would happen in the `SwaggerEditor` constructor after these changes, so instantiating it in the `ApiGenerator` constructor will make the validation happen early. 
Also, this file used to instantiate multiple `SwaggerEditors` because it kept modifying `self.definition_body` and then would use the new value in each new instance of SwaggerEditor. Now, there's only one SwaggerEditor used and only at the end we use all of the changes made to update `self.definition_body` once.

samtranslator/model/api/api_generator.py
```diff
@@ -280,7 +282,7 @@ def _construct_rest_api(self):
             rest_api.BodyS3Location = self._construct_body_s3_dict()
         elif self.definition_body:
             # # Post Process OpenApi Auth Settings
-            self.definition_body = self._openapi_postprocess(self.definition_body)
+            self.definition_body = self._openapi_postprocess(self.swagger_editor.swagger)
```
torresxb1, line 285: This is now the only time we update self.definition_body with the result of all of the changes we made with the SwaggerEditor

samtranslator/model/api/api_generator.py
```diff
@@ -610,13 +612,6 @@ def _add_cors(self):
         else:
             raise InvalidResourceException(self.logical_id, INVALID_ERROR)
 
-        if not SwaggerEditor.is_valid(self.definition_body):
-            raise InvalidResourceException(
-                self.logical_id,
-                "Unable to add Cors configuration because "
-                "'DefinitionBody' does not contain a valid Swagger definition.",
-            )
-
```
torresxb1, line 619: removing all of these kind of validations, since they're now done when SwaggerEditor is instantiated

samtranslator/swagger/swagger.py
```diff
@@ -40,26 +40,16 @@ def __init__(self, doc):
...
         self.resource_policy = self._doc.get(self._X_APIGW_POLICY, {})
         self.definitions = self._doc.get("definitions", {})
 
-        # https://swagger.io/specification/#path-item-object
-        # According to swagger spec,
-        # each path item object must be a dict (even it is empty).
-        # We can do an early path validation on path item objects,
-        # so we don't need to validate wherever we use them.
-        for path in self.iter_on_path():
-            SwaggerEditor.validate_path_item_is_dict(self.get_path(path), path)
```
torresxb1, line 62: moved this into `validate_definition_body`

hawflau (changes requested):

samtranslator/swagger/swagger.py
```diff
@@ -1311,6 +1308,32 @@ def is_valid(data):
                 )
         return False
 
+    def validate_definition_body(self, definition_body):
+        """
+        Checks if definition_body is a valid Swagger document
+
+        :param dict definition_body: Data to be validated
+        :return: True, if definition_body is a valid Swagger document
```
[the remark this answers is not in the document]
torresxb1: oops, the docstring shouldn't say it returns `True` either. Thanks for catching it!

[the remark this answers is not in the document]
torresxb1: fixed

samtranslator/swagger/swagger.py
```diff
@@ -47,19 +47,12 @@ def __init__(self, doc):
...
-
         self._doc = copy.deepcopy(doc)
-        self.paths = self._doc["paths"]
-        self.security_definitions = self._doc.get("securityDefinitions", Py27Dict())
-        self.gateway_responses = self._doc.get(self._X_APIGW_GATEWAY_RESPONSES, Py27Dict())
-        self.resource_policy = self._doc.get(self._X_APIGW_POLICY, Py27Dict())
-        self.definitions = self._doc.get("definitions", Py27Dict())
+        self.validate_definition_body(doc)
 
+        self.paths = self._doc.get("paths")
```
hoffa, line 56: Out of curiosity, was was `_doc["paths"]` changed to `_doc.get("paths")`?

torresxb1: I think at some point I had the `self.validate_definition_body(doc)` line after `self.paths = self._doc["paths"]`. So if `paths` wasn't there, it would throw a runtime error (vs. with `.get` it returns `None`). However, shouldn't make a difference now that validation happens before.

moelasmar (approved):

hawflau (approved):

PR #2265, opened 2021-12-15
Update jsonschema dependency version.

JohnPreston:
Update dependency for jsonschema

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -1,6 +1,6 @@
 pyrsistent~=0.16.0; python_version<"3"
 boto3~=1.5
 enum34~=1.1; python_version<"3.4"
-jsonschema~=3.2
+jsonschema>=3.2,<5.0
 six~=1.15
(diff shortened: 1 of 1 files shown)


hawflau:
@JohnPreston Thanks for the contribution. Why is this change necessary?

JohnPreston:
> @JohnPreston Thanks for the contribution. Why is this change necessary?

Keep up with latest release of dependencies, that is all.

hawflau:
@JohnPreston 
unit tests only passed in py3.6 but not in py3.7 and py3.8. `jsonschema>=4.0` got installed in py3.7 and py3.8 but doesn't support py3.6. That seems to be the cause of the test failures.

Because we still support Python3.6, I think it's better to keep `jsonschema~=3.2` for now. 

I'd suggest you close this PR and open a new issue for us to investigate supporting `jsonschema>=4.0`

PR #2273, opened 2021-12-20
fix: Raise exception if provided Auth.Authorizers.Identity isn't a dict

mildaniel:
*Issue #, if available:*

*Description of changes:*

*Checklist:*

- [ ] Add/update tests using:
    - [ ] Correct values
- [x] `make pr` passes
- [ ] Update documentation

*Examples?*

### DIFF
--- a/samtranslator/model/apigateway.py
+++ b/samtranslator/model/apigateway.py
@@ -395,15 +395,15 @@ class ApiGatewayAuthorizer(object):
         return "LAMBDA"
 
     def _get_identity_header(self):
-        try:
-            if not self.identity or not self.identity.get("Header"):
-                return "Authorization"
-        except AttributeError:
+        if self.identity and not isinstance(self.identity, dict):
             raise InvalidResourceException(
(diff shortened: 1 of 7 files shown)


samtranslator/model/apigateway.py
```diff
@@ -395,8 +395,15 @@ def _get_type(self):
...
-            return "Authorization"
+        try:
+            if not self.identity or not self.identity.get("Header"):
+                return "Authorization"
+        except AttributeError:
+            raise InvalidResourceException(
+                self.api_logical_id,
+                "Auth.Authorizers.<Authorizer>.Identity must be a dict (LambdaTokenAuthorizationIdentity "
+                "or LambdaRequestAuthorizationIdentity).",
+            )
```
mndeveci, line 406: Instead of `try/except` block, shouldn't we check if the property is an expected type (dict in this case)?

hawflau (approved):

torresxb1 (approved):

PR #2285, opened 2022-01-10
fix: Raise Invalid Resource When DisableExecuteApiEndpoint: False And Has No DefinitionBody

CoshUS:
*Issue #, if available:*
When DisableExecuteApiEndpoint is False, http api and api generator will treat it as None which means the default, True.

*Description of changes:*

*Checklist:*

- [ ] Add/update tests using:
    - [ ] Correct values
- [ ] `make pr` passes
- [ ] Update documentation

*Examples?*


bin/sam-translate.py
```diff
@@ -28,6 +26,9 @@
 import boto3
 from docopt import docopt
 
+if sys.version_info.major == 3:
```
hawflau, line 29: We can drop this version check and just import the module as we are already py3 only on the `develop` branch

bin/sam-translate.py
```diff
@@ -1,5 +1,3 @@
-#!/usr/bin/env python2
```
hawflau, line 1: We still want to keep this and update from `python2` to `python` as this file is executable.

hawflau (changes requested):

bin/sam-translate.py
CoshUS: Removed.

bin/sam-translate.py
CoshUS: Added back as `python`.

hoffa (approved):

hawflau (approved):

Issue #2288, opened 2022-01-13
Support for PartiQL in DynamoDB policy templates


adriaandejonge:
### Describe your idea/feature/enhancement

AWS SAM policy templates `DynamoDBCrudPolicy`, `DynamoDBReadPolicy` and `DynamoDBWritePolicy` make it easy to assign policies to Lambda functions that interact with DynamoDB. Since November 2020, DynamoDB supports an SQL-like query language called PartiQL as an alternative way of creating/updating/reading/deleting items in a DynamoDB table. 

I wish that the AWS SAM policy templates allowed PartiQL queries without requiring additional custom policies/permissions.

### Proposal

To the policy `DynamoDBCrudPolicy`, add:
 * `dynamodb:PartiQLSelect`
 * `dynamodb:PartiQLUpdate`
 * `dynamodb:PartiQLInsert`
 * `dynamodb:PartiQLDelete`

To the policy `DynamoDBReadPolicy`, add:
 * `dynamodb:PartiQLSelect`

To the policy `DynamoDBWritePolicy`, add:
 * `dynamodb:PartiQLUpdate`
 * `dynamodb:PartiQLInsert`

Things to consider:
[X] The [SAM documentation with the policy template list](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/serverless-policy-template-list.html) will need to be updated --> See PR awsdocs/aws-sam-developer-guide#140

mildaniel:
Thank you for your contribution! We really appreciate it. Once a PR is submitted we weigh it against our roadmap, do regression testing, make sure it’s backwards compatible and maintainable. Please watch this channel from more updates, and feel free to reach out.

hawflau:
@adriaandejonge Thanks for raising the issue. Feel free to raise a PR for this. Contributions are welcomed!

adriaandejonge:
> @adriaandejonge Thanks for raising the issue. Feel free to raise a PR for this. Contributions are welcomed!

Thanks @hawflau - I have already raised PR #2289 for this.

mndeveci:
@adriaandejonge thanks for raising the PR, really appreciated!

I saw that there were some change requests made for it, would you able to address them?

adriaandejonge:
> @adriaandejonge thanks for raising the PR, really appreciated!
> 
> I saw that there were some change requests made for it, would you able to address them?

Thanks @mndeveci Yes, I'll work on this and get back.

hoffa:
[...]

PR #2289, opened 2022-01-13
Add PartiQL support to DynamoDB policy templates

adriaandejonge:
Adds support for PartiQL queries to DynamoDBCrudPolicy, DynamoDBReadPolicy and DynamoDBWritePolicy policy templates.

*Issue #, if available:* #2288

*Description of changes:*

To the policy `DynamoDBCrudPolicy`, add:
 * `dynamodb:PartiQLSelect`
 * `dynamodb:PartiQLUpdate`
 * `dynamodb:PartiQLInsert`
 * `dynamodb:PartiQLDelete`

To the policy `DynamoDBReadPolicy`, add:
 * `dynamodb:PartiQLSelect`

To the policy `DynamoDBWritePolicy`, add:
 * `dynamodb:PartiQLUpdate`
 * `dynamodb:PartiQLInsert`

Validated manually that the proposed permissions are the correct ones for supporting PartiQL queries in a Python based Lambda function using Boto3.

*Checklist:*

- [ ] Add/update tests using:
    - [x] Correct values
    - [ ] ~~Bad/wrong values (None, empty, wrong type, length, etc.)~~ only extended existing test cases with changes
    - [ ] ~~[Intrinsic Functions](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference.html)~~ n/a
- [x] `make pr` passes
- [x] Update documentation See PR awsdocs/aws-sam-developer-guide#140
- [ ] ~~Verify transformed template deploys and application functions as expected~~ Transformation logic unchanged

*Examples?*

### DIFF
--- a/samtranslator/policy_templates_data/policy_templates.json
+++ b/samtranslator/policy_templates_data/policy_templates.json
@@ -133,7 +133,11 @@
               "dynamodb:BatchWriteItem",
               "dynamodb:BatchGetItem",
               "dynamodb:DescribeTable",
-              "dynamodb:ConditionCheckItem"
+              "dynamodb:ConditionCheckItem",
+              "dynamodb:PartiQLSelect",
+              "dynamodb:PartiQLUpdate",
+              "dynamodb:PartiQLInsert",
+              "dynamodb:PartiQLDelete"
             ],
             "Resource": [
               {
@@ -177,7 +181,8 @@
               "dynamodb:Scan",
               "dynamodb:Query",
               "dynamodb:BatchGetItem",
-              "dynamodb:DescribeTable"
+              "dynamodb:DescribeTable",
+              "dynamodb:PartiQLSelect"
             ],
             "Resource": [
               {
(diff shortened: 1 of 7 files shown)


hawflau (changes requested):
Thanks for your contribution. I'd suggest you to create new policies for the PartiQL permissions. For example, create a `DynamoDBPartiQLCrudPolicy` to contain the select, update, insert and delete permissions, rather than add them into the existing `DynamoDBCrudPolicy`. The main reason is that some customers do not expect new permissions being added to existing policy templates that they are already using. If customers want to use both sets of permissions, they can specify both in their function's `Policies`. These policy templates got translated into inline policies in a IAM role, and IAM role can include as many inline policies as you want as long as they don't exceed [certain size limits](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_iam-quotas.html#reference_iam-quotas-entity-length).

hawflau:
Closing due to inactivity. Please open a new PR with the suggested change when it's ready. Thanks!

PR #2294, opened 2022-01-17
fix: For FindInMap intrinsics, prevent dictionary iteration on None objects

qingchm:
*Issue #, if available:*
N.A.

*Description of changes:*
If the map provided for FindInMap does not contain content for keys needed, iteration over None would raise unexpected errors.

Added unit test for none-existent mapping keys. 

*Checklist:*

- [x] Add/update tests using:
    - [ ] Correct values
- [x] `make pr` passes
- [ ] Update documentation

*Examples?*


hoffa (approved):

samtranslator/intrinsics/actions.py
```diff
@@ -561,9 +561,9 @@ def resolve_parameter_refs(self, input_dict, parameters):
             return input_dict
 
         if (
-            map_name not in parameters
-            or top_level_key not in parameters[map_name]
-            or second_level_key not in parameters[map_name][top_level_key]
+            not parameters.get(map_name)
+            or not parameters.get(map_name, {}).get(top_level_key)
+            or not parameters.get(map_name, {}).get(top_level_key, {}).get(second_level_key)
```
moelasmar, line 566: I think it is better to validate that these values are instance of Dict, as for invalid template, these values can exist as strings or None values instead of dictionary

tests/plugins/application/test_serverless_app_plugin.py
```diff
@@ -220,6 +220,15 @@ def test_resolve_intrinsics(self):
 
         self.assertEqual("value1", output)
 
+    def test_resolve_invalid_mapping(self):
```
moelasmar, line 223: could you please add some unit test cases for invalid templates like:
- mappings = {"MapA": "value"}
- mappings = {"MapA": None}

moelasmar (approved):

jfuss:
Closing given how old these are. I chatted with @qingchm and he will re-open and handle feedback once he picks these back up.

PR #2295, opened 2022-01-17
fix: Add handling for OpenApi definition that has tags with invalid format

qingchm:
*Issue #, if available:*
N.A.

*Description of changes:*
If the tags in the definition body is not a list, the tags parsing will cause iteration errors.

Added a check for the validity of the definition body that has tags specified.

*Checklist:*

- [x] Add/update tests using:
    - [ ] Correct values
- [x] `make pr` passes
- [ ] Update documentation

*Examples?*

### DIFF
--- a/samtranslator/open_api/open_api.py
+++ b/samtranslator/open_api/open_api.py
@@ -455,8 +455,8 @@ class OpenApiEditor(object):
                 raise InvalidDocumentException(
                     [
                         InvalidTemplateException(
-                            "Tags in open api definition needs to be a list. {} is a {} not a list.".format(
-                                self.tags, type(self.tags)
+                            "Tags in OpenApi DefinitionBody needs to be a list. {} is a {} not a list.".format(
+                                self.tags, type(self.tags).__name__
                             )
                         )
                     ]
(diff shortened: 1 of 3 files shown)


samtranslator/open_api/open_api.py
```diff
@@ -450,6 +450,17 @@ def add_tags(self, tags):
         :param dict tags: dictionary of tagName:tagValue pairs.
         """
         for name, value in tags.items():
+            # verify the tags definition is in the right format
+            if not isinstance(self.tags, list):
+                raise InvalidDocumentException(
+                    [
+                        InvalidTemplateException(
+                            "Tags in open api definition needs to be a list. {} is a {} not a list.".format(
+                                self.tags, type(self.tags)
```
hoffa, line 459: Perhaps better to get rid of the weird Pythonism (`<class 'dict'>`)?

```suggestion
                                self.tags, type(self.tags).__name__
```

qingchm: Sure yes that makes the message more readable! Will add in this change

hawflau (approved):

hoffa (approved):

PR #2296, opened 2022-01-17
fix: accept empty components property in Open Api Definition

moelasmar:
Use a default dict in case if a template contains an empty components property in the Open Api definition body instead of failing, as Components property is an optional property in the OpenApi definition.

*Issue #, if available:*

*Description of changes:*

*Checklist:*

- [x] Add/update tests using:
- [x] `make pr` passes

*Examples?*


samtranslator/model/api/api_generator.py
```diff
@@ -999,6 +999,7 @@ def _openapi_postprocess(self, definition_body):
         ):
             if definition_body.get("securityDefinitions"):
                 components = definition_body.get("components", Py27Dict())
+                components = components if components else Py27Dict()
```
mndeveci, line 1002: Isn't second parameter of `get` is the default value if the item is `None`? If so, the line 1002 would be redundant?

moelasmar: no, get will only return default value if the `components` is not the dict at all, but if it is in the dict and its value is None, it will return None and not the default value

mndeveci: Got it, can we add a comment next to it so it won't be removed accidentally in the future.

mndeveci (approved):

mildaniel (approved):

samtranslator/model/api/api_generator.py
moelasmar: done

Issue #1274, opened 2019-11-28
Pproperty ApiKeySourceType not defined for resource of type AWS::Serverless::Api


imjustd:
**Description:**
I wanted to specify  the Api Key Source to `AUTHORIZER` in the SAM `template.yaml` file since I use custom Lambda authorizer but it didn't succeed.

**Steps to reproduce the issue:**
1. Create new SAM project and try to add a new resource that uses `AWS::Serverless::Api` similar to:
```
Resources:
  MyApi:
    Type: AWS::Serverless::Api
    Properties:
      ApiKeySourceType: AUTHORIZER
      StageName: Prod
```
2. run `sam validate`

**Observed result:**
When running `sam validate` I got the error:
```
Error: [InvalidResourceException('MyApi', 'property ApiKeySourceType not defined for resource of type AWS::Serverless::Api')] ('MyApi', 'property ApiKeySourceType not defined for resource of type AWS::Serverless::Api')
```

**Expected result:**
- I expect `AWS::Serverless::Api` to accept this parameter `ApiKeySourceType` because it's supported in the `AWS::ApiGateway::RestApi` https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-apigateway-restapi.html#cfn-apigateway-restapi-apikeysourcetype

**Note:** I tried to use `AWS::ApiGateway::RestApi` to define my API in the SAM template but it didn't work out.

imjustd:
Any update on this?

pedrofurlanetto90:
Is there any update on this? Without this parameter, I needed to use swagger definitions on API Gateway

imjustd:
Almost a year has passed, do you have some update?

pedrofurlanetto90:
We really need this feature. Do you have any update? Any information on why AWS::Serverless::Api does not support it?

@imjustd did you find some workaround?

imjustd:
> @imjustd did you find some workaround?

Currenty I use this commands after deploy:
```
aws apigateway update-rest-api --rest-api-id <api_id> --patch-operations op=replace,path=/apiKeySource,value=AUTHORIZER && \
aws apigateway create-deployment --rest-api-id <api_id> --stage-name Prod
```
I would actually try to implement this myself if I know what needs to be changed in the SAM code.

nayaranunes:
The property 'ApiKeySourceType' isn't defined for resource of type AWS::Serverless::Api, try this:
```Resources:
   MyApi:
[...]

PR #2298, opened 2022-01-19
Adding ApiKeySourceType.

lee-11:
#1274 

This PR attempts to make the changes needed to support the feature of API Gateway that allows a Lambda Custom Authorizer to take on the responsibility of identifying the API Key for authorized requests (versus having the client specify the key via the `x-api-key` header).  This PR adds the optional string property `ApiKeySourceType` to the `AWS:Serverless::Api` and adds validation that it holds one of the two acceptable values: `"HEADER"` or `"AUTHORIZER"`. 

To validate this work (I'm not a Python developer, so this is a bit of a learning experience for me) I ran the tests as described in the development guide (i.e. `make pr`).  I didn't see any failed tests and the code coverage remained over `95%` (this is a very small change, so the impact was minimal), so the "optional" nature of the new property seems correct (backward compatible).

I also added a new test for the `AUTHORIZER` use case.  The test makes use of correct value since the default is already tested in others.

Finally, I added an integration test to verify it all works as intended.

I haven't implemented intrinsic functions since I don't believe they are useful for enumerated values like this.  That said, I'm happy to add that in if it makes sense.

*Checklist:*

- [x] Add/update tests using:
    - [x] Correct values
    - [n/a] [Intrinsic Functions](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference.html)
- [X] `make pr` passes
- [X] Update documentation

I have not added examples for this property since it is a fairly complicated topic to explain (requires an Api, Authorizer, Plan and Key at minimum). Also, it looks like the examples have been moved to another repo.


tests/translator/output/api_with_apikey_source.json
```diff
@@ -0,0 +1,318 @@
...
+                  ]
+                }
+              }
+            ]
+          }
+        }
+      }, 
+      "MyApiWithAuthSource": {
+        "Type": "AWS::ApiGateway::RestApi", 
+        "Properties": {
```
hawflau, line 249: This expected output doesn't match with the changes you made. It seems to be the reason for the failed tests.

hawflau (changes requested):
Thanks for your contributions. 

There are some failed tests. Also, you should add the expected output under `tests/translator/output/aws-us-gov` and `tests/translator/output/aws-cn/`.

hawflau:
@lee-11 I didn't notice the PR was still in draft. I also saw the issue discussing about the behavior of `ApiKeySourceType` getting reset. I guess you are still working on it. Please let us know if you need any support!

lee-11:
> Please let us know if you need any support!

Oh, yes.  So far so good, but I'm winging it a bit here and could use some guidance.  I'm not sure how to test a transformed template can actually be deployed. Is there a convenient way to transform a known good SAM template I have on hand?

lee-11:
Is there documentation that should be updated with this PR? I've poked around the repo and don't see anything.  Sorry for asking; I'm totally unfamiliar with both Python and this repo.

hawflau:
> Oh, yes. So far so good, but I'm winging it a bit here and could use some guidance. I'm not sure how to test a transformed template can actually be deployed. Is there a convenient way to transform a known good SAM template I have on hand?

One way is, for example, in the test template you provided, you can replace the dummy s3 bucket/key with actual s3 bucket/key that exists. Then you can deploy it in CloudFormation (via AWS CLI or in AWS Console).
Alternatively you can add an integration test case under the `integration/` directory. Our integration tests are end-to-end tests which are supposed to be deployable when they run.

> Is there documentation that should be updated with this PR? I've poked around the repo and don't see anything. Sorry for asking; I'm totally unfamiliar with both Python and this repo.

[Here](https://github.com/aws/serverless-application-model/blob/develop/DEVELOPMENT_GUIDE.md#verifying-transforms) it is. But I agree that the dev guide is not very up-to-date. We will work on improving the doc.

NickHeap2:
@lee-11 Did you actually get an example to deploy with this change? I did the same type of changes and whilst I could get sam validate to work it failed during sam deploy.

lee-11:
> @lee-11 Did you actually get an example to deploy with this change? I did the same type of changes and whilst I could get sam validate to work it failed during sam deploy.

I did, yes.  However my deploy wasn't through the sam cli, but using the `bin/sam-translate.py` tool.  My usual toolset is to use the cloudformation API directly to create change sets and deploy (which isn't really an option here).

Can you provide the details of how you used the SAM cli to test this?

lee-11:
@hawflau if I may ask, what's the process from here for an external PR? Anything needed from me or that I should be expecting as next steps?

integration/combination/test_api_with_authorizer_apikey.py
```diff
@@ -0,0 +1,128 @@
...
+        base_url = stack_outputs["ApiUrl"]
+
+        self.verify_authorized_request(base_url + "none", 200)
+        self.verify_authorized_request(base_url + "lambda-token-api-key", 401)
+        self.verify_authorized_request(base_url + "lambda-token-api-key", 200, "Authorization", "allow")
+
+        api_key_id = stack_outputs["ApiKeyId"]
+        key = apigw_client.get_api_key(apiKey=api_key_id, includeValue=True)
+
+        self.verify_authorized_request(base_url + "lambda-token-api-key", 403)
```
hawflau, line 71: This line is verifying the same path as L65. Why expected a different status code?

integration/combination/test_api_with_authorizer_apikey.py
```diff
@@ -0,0 +1,128 @@
...
+        self.verify_authorized_request(base_url + "llambda-token-api-key", 403, "x-api-key", key["value"])
```
hawflau, line 72: I believe this should be 401 instead of 403

integration/resources/templates/combination/api_with_authorizer_apikey.yaml
```diff
@@ -0,0 +1,111 @@
...
+  MyUsagePlanKey:
+    Type: AWS::ApiGateway::UsagePlanKey
+    Properties:
+      KeyId:
+        Ref: MyFirstApiKey
+      UsagePlanId:
+        Ref: MyUsagePlan
+      KeyType: API_KEY
+
+Outputs:
```
hawflau, line 102: You should include `ApiKeyId` as an output for the test case to use

hawflau (changes requested):
lee-11 Thanks for updating the PR and adding the integration test case. I tried to run it but encountered some failed validations. Could you please check and update accordingly? Thanks

sawilde:
@lee-11 thanks for this - I have crashed into this issue recently and looked into the myriad of workarounds that seem to involve a lot of make-work, the reason we use AWS::Serverless::APi (all I can say is CORS/Swagger).

@hawflau @lee-11 what do we need to do to progress this PR forward?

integration/resources/templates/combination/api_with_authorizer_apikey.yaml
qingchm: @lee-11 Can we try to address this? Thanks!

hawflau: I've added it in the latest commit

hawflau (approved):

moelasmar (approved):

sawilde:
awesome work folks - in what release will we see this fix emerge?

jfuss:
@sawilde We don't track specific releases, since things outside our control can happen (security patches, non public features, etc). We tend to not combine PRs with some of those processes so we don't accidentally affect others timelines. 

With that said, we are taking about cutting a release in the coming weeks to get Http IAM auth and some other bug fixes merged out. This PR (more specifically anything in `develop` not yet released) would be all released together but have not timeline to share at this point.

sawilde:
Thanks for the update, I'll keep on eye out. 

Is there a releases thread I could follow?

imjustd:
@sawilde This is already released:
- https://github.com/aws/serverless-application-model/issues/1274
- https://github.com/aws/serverless-application-model/releases/tag/v1.46.0

sawilde:
@imjustd thanks for the update

PR #2299, opened 2022-01-19
Clean up Py2 dependencies and update to use native unittest.mock

hawflau:
*Issue #, if available:*

*Description of changes:*

- Remove all Py2 dependencies
- Update to use native `unittest.mock` instead of external `mock` library

*Checklist:*

- [ ] Add/update tests using:
    - [ ] Correct values
- [x] `make pr` passes
- [ ] Update documentation

*Examples?*

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -1,6 +1,4 @@
-pyrsistent~=0.16.0; python_version<"3"
 boto3~=1.5
-enum34~=1.1; python_version<"3.4"
 jsonschema~=3.2
 six~=1.15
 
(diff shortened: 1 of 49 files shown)


requirements/base.txt
```diff
@@ -1,6 +1,4 @@
-pyrsistent~=0.16.0; python_version<"3"
 boto3~=1.5
-enum34~=1.1; python_version<"3.4"
 jsonschema~=3.2
 six~=1.15
```
jfuss, line 3: Do we still need six if we are on py2?

jfuss: Let's do another PR to keep things scoped.

jfuss (approved):

mildaniel (approved):

PR #2305, opened 2022-01-24
fix: ValueError is not caught by int()

wchengru:
*Issue #, if available:*

*Description of changes:*

*Checklist:*

- [ ] Add/update tests using:
    - [ ] Correct values
- [ ] `make pr` passes
- [ ] Update documentation

*Examples?*

### DIFF
--- a/samtranslator/model/apigateway.py
+++ b/samtranslator/model/apigateway.py
@@ -285,7 +285,7 @@ class ApiGatewayAuthorizer(object):
         try:
             ttl_int = int(ttl)
         # this will catch if ttl is None and not convertable to an int
-        except TypeError:
+        except (TypeError, ValueError):
             # previous behavior before trying to read ttl
             return required_properties_missing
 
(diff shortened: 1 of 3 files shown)


samtranslator/model/apigateway.py
```diff
@@ -285,7 +285,7 @@ def _is_missing_identity_source(self, identity):
         try:
             ttl_int = int(ttl)
         # this will catch if ttl is None and not convertable to an int
-        except TypeError:
+        except (TypeError, ValueError):
```
moelasmar, line 288: can you add a unit test case to show when exactly this exception will be thrown?
My understanding is ValueError will be thrown if ttl value contains a non numeric value, and I think in this case we should throw a different exception

wchengru: Thanks, test cases added.

moelasmar (approved):

jfuss (approved):

PR #2308, opened 2022-01-25
fix: Switched to JSON Copy for Performance Optimization

CoshUS:
*Issue #, if available:*

*Description of changes:*
Original time: 139 seconds
With Hash Caching Only: 113 seconds
With Fast Copy Only: 123 seconds
With Fast Copy and Hash Caching: 53 seconds

Lambda authorizer strings are in Py27UniStr. Deepcopy keeps the type but json loads/dumps will resolve it to string.

*Checklist:*

    - [ ] Correct values
- [ ] `make pr` passes
- [ ] Update documentation

*Examples?*


CoshUS:
Tests and docs will be added once implementation has no issue.

hawflau (changes requested):
LGTM overall! 
Can you also add some unit tests? Thanks!

Issue #2291, opened 2022-01-14
[Announcement] Deprecating AWS Serverless Application Model on Python3.6


mndeveci:
As of now, AWS Serverless Application Model can be used as a library within python3.6, python3.7 and python3.8 versions. However, python3.6 has reached its end of life by December 23th, 2021.
### What is deprecation schedule?

AWS Serverless Application Model will not work on python3.6 environments by February 28th, 2022.

### Will this impact my AWS Lambda functions using python3.6?

No, AWS Lambda functions which is using python3.6 as runtime will not be affected by this change. For more information about supported runtimes, please check https://docs.aws.amazon.com/lambda/latest/dg/runtime-support-policy.html

### Am I impacted by this deprecation?

Following use cases will be impacted by this change;

- If you are using python3.6 and installing AWS Serverless Application Model through `pip install`
- If you are using `aws-sam-translator` as a library and you are building your application with python3.6

PatMyron:
Think should stay pinned for a couple quarters since over 5% of downloads are still using python 3.6:
https://pypistats.org/packages/aws-sam-translator

PR #2311, opened 2022-02-01
fix: remove py3.6 support and add py3.9 and py3.10

mndeveci:
*Issue #, if available:*
https://github.com/aws/serverless-application-model/issues/2291

*Description of changes:*
Python3.6 have reached its EOL, this PR removes support for using serverless-application-model as a library in a python3.6 environment.

*Checklist:*

    - [ ] Correct values
- [ ] `make pr` passes
- [ ] Update documentation

*Examples?*


jfuss (approved):

setup.py
```diff
@@ -66,6 +66,7 @@ def read_requirements(req="base.txt"):
         "NOTICE",
         "THIRD_PARTY_LICENSES",
     ),
+    python_requires=">=3.7, <=4.0, !=4.0",
```
jfuss, line 69: Should we restrict this to only supported versions instead of auto enabling new python versions under 4.0?

jfuss (changes requested):

setup.py
jfuss: Rethinking this. Since this repo is a library, leaving it open like this is likely fine. We may run into problems if we use newer lang features but we have to support the min version anyways.

jfuss (approved):

hoffa (approved):

torresxb1 (approved):
There are still some references to python 3.6 in `DEVELOPMENT_GUIDE.md` but otherwise LGTM

jfuss:
> There are still some references to python 3.6 in DEVELOPMENT_GUIDE.md but otherwise LGTM

I can make a follow up PR to address those.

setup.py
michael-k: Why `<=4.0` and `!=4.0` instead of just `<4.0`?

(Or better yet only `>=3.7` https://iscinumpy.dev/post/bound-version-constraints/)

PR #2315, opened 2022-02-07
fix: Add check for event and invoke event destination config type

qingchm:
*Issue #, if available:*
N.A.

*Description of changes:*
Adds a check on destination config format, to ensure on success and on failure events are inputted with correct format

Added integration tests

*Checklist:*

    - [ ] Correct values
- [x] `make pr` passes
- [ ] Update documentation

*Examples?*

### DIFF
--- a/samtranslator/model/eventsources/pull.py
+++ b/samtranslator/model/eventsources/pull.py
@@ -117,6 +117,8 @@ class PullEventSource(ResourceMacro):
         if self.DestinationConfig:
             if self.DestinationConfig.get("OnFailure") is None:
                 raise InvalidEventException(self.logical_id, "'OnFailure' is a required field for 'DestinationConfig'")
+            if not isinstance(self.DestinationConfig.get("OnFailure"), dict):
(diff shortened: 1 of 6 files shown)


samtranslator/model/eventsources/pull.py
```diff
@@ -117,6 +117,8 @@ def to_cloudformation(self, **kwargs):
         if self.DestinationConfig:
             if self.DestinationConfig.get("OnFailure") is None:
                 raise InvalidEventException(self.logical_id, "'OnFailure' is a required field for 'DestinationConfig'")
+            if not isinstance(self.DestinationConfig.get("OnFailure"), dict):
+                raise InvalidEventException(self.logical_id, "'OnFailure' has invalid destination configuration")
```
jfuss, line 121: Can we state what it should be? If a customer sees this error, do they know how to fix it with this error message?

jfuss (changes requested):
Can we updated unit tests with different cases to cover this and make sure we hit the correct permutations?

jfuss:
Closing given how old these are. I chatted with @qingchm and he will re-open and handle feedback once he picks these back up.

PR #2317, opened 2022-02-08
Update doc: Add missing attribute in RequestModel

kavindasr:
*Issue #, if available:*
In doc ReuqestModel has only two attributes. But RequestModel has an attribute call `ValidateBody`
*Description of changes:*

*Checklist:*

    - [ ] Correct values
- [ ] `make pr` passes
- [x] Update documentation

*Examples?*


jfuss:
@kavindasr Thank you for the PR however, we are going to remove this file from the repo. Reason is stated here: https://github.com/aws/serverless-application-model/pull/2334 but we have docs we maintain with each release and deployed to the AWS Doc site. Instead of keeping the repo and AWS Docs updated, we will keep the AWS Docs updated. 

Again thank you for taking the time to submit this.

PR #2321, opened 2022-02-12
Add method for determining service availability in a region

marekaiv:
*Issue #, if available:*
n/a

*Description of changes:*
Generalize existing `RegionConfiguration.is_sar_supported` method into a `is_service_supported` method as we'll need to check availability of other services as well.

Unit tests.

*Checklist:*

    - [x] Correct values
    - [ ] ~[Intrinsic Functions](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference.html)~ n/a
- [ ] Add/update [integration tests](https://github.com/aws/serverless-application-model/blob/develop/INTEGRATION_TESTS.md) -- covered by existing tests that use SAR
- [x] `make pr` passes
- [ ] ~Update documentation~ n/a
- [ ] ~Verify transformed template deploys and application functions as expected~ n/a, low-level change

### DIFF
--- a/samtranslator/plugins/application/serverless_app_plugin.py
+++ b/samtranslator/plugins/application/serverless_app_plugin.py
@@ -111,7 +111,7 @@ class ServerlessAppPlugin(BasePlugin):
 
             if key not in self._applications:
                 try:
-                    if not RegionConfiguration.is_sar_supported():
+                    if not RegionConfiguration.is_service_supported("serverlessrepo"):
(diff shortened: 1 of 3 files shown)


mndeveci (approved):

hawflau (approved):
Approved. But we should be careful that `get_available_regions` might give false negative for some services even though they are actually available, as it depends on the data file in botocore.

PR #2323, opened 2022-02-14
Add validation for SecretsManagerKmsKeyId

jonife:
*Description of changes:*
Added validation to SecretsManagerKmsKeyId 

validation will ensure that SecretsManagerKmsKeyId is of type string

*Checklist:*

    - [x] Correct values
- [x] `make pr` passes
- [ ] Update documentation

*Examples?*


samtranslator/model/eventsources/pull.py
```diff
@@ -188,6 +188,13 @@ def _validate_filter_criteria(self):
         if list(self.FilterCriteria.keys()) not in [[], ["Filters"]]:
             raise InvalidEventException(self.relative_id, "FilterCriteria field has a wrong format")
 
+    def validate_secrets_manager_kms_key_id(self):
```
moelasmar, line 191: can you make it private method, and do the validation only if the SecretsManagerKmsKeyId is not None.

tests/translator/input/error_function_with_mq_kms_invalid_type.yaml
```diff
@@ -0,0 +1,24 @@
+Parameter:
```
moelasmar, line 1: can you add another test case to cover the self managed Kafka event.

moelasmar (changes requested):

samtranslator/model/eventsources/pull.py
jonife: the validation is called only when `SecretsManagerKmsKeyId` is provided. It is not a required property for an eventsource type

moelasmar (approved):

jfuss (commented):
Can we add unit tests as well? The end to end tests are great but we can create a better coverage if we add unit tests and validate different input (lists, dict, int, etc).

tests/model/eventsources/test_mq_event_source.py
```diff
@@ -40,3 +41,49 @@ def test_get_policy_statements(self):
             }
         ]
         self.assertEqual(policy_statements, expected_policy_document)
+
+    def test_must_validate_secrets_manager_kms_key_id(self):
+        self.mq_event_source.SourceAccessConfigurations = [{"Type": "BASIC_AUTH", "URI": "SECRET_URI"}]
+        self.mq_event_source.Broker = "BROKER_ARN"
+        self.mq_event_source.SecretsManagerKmsKeyId = ["1abc23d4-567f-8ab9-cde0-1fab234c5d67"]
```
jfuss, line 48: Could we parameterize this to include anything that is not a string? We covered list here but would be good to add int, bool and dict to add move coverage.

jfuss (changes requested):

jfuss (approved):
Thank you for going through additional iterations with adding more unit tests.

PR #2324, opened 2022-02-15
update validation for dead letter queue

jonife:
*Issue #, if available:*

*Description of changes:*
Update validation for dlq 

*Checklist:*

    - [x] Correct values
- [x] `make pr` passes
- [ ] Update documentation

*Examples?*

### DIFF
--- a/samtranslator/model/sam_resources.py
+++ b/samtranslator/model/sam_resources.py
@@ -593,6 +593,12 @@ class SamFunction(SamResourceMacro):
                 "'DeadLetterQueue' requires Type and TargetArn properties to be specified.".format(valid_dlq_types),
             )
 
+        if not (isinstance(self.DeadLetterQueue.get("Type"), str)):
+            raise InvalidResourceException(
+                self.logical_id,
+                "'DeadLetterQueue' property 'Type' should be of type str.",
+            )
+
         # Validate required Types
         if not self.DeadLetterQueue["Type"] in self.dead_letter_queue_policy_actions:
             raise InvalidResourceException(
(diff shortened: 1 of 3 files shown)


samtranslator/model/sam_resources.py
```diff
@@ -593,6 +593,12 @@ def _validate_dlq(self):
                 "'DeadLetterQueue' requires Type and TargetArn properties to be specified.".format(valid_dlq_types),
             )
 
+        if not (isinstance(self.DeadLetterQueue.get("Type"), str)):
```
moelasmar, line 596: can you update this method to validate that the `DeadLetterQueue` is of type Dict, and if not raise a validation error.

moelasmar (changes requested):

samtranslator/model/sam_resources.py
jonife: I don't know this is necessary as this is already handled by [sam](https://github.com/aws/serverless-application-model/blob/develop/samtranslator/model/sam_resources.py/#L68). And this error is thrown when `DeadLetterQueue` is not dict `Type of property 'DeadLetterQueue' is invalid.`

moelasmar: not sure where I can find this error, but I found this usage where we just check that the `self.DeadLetterQueue` is not None 
https://github.com/aws/serverless-application-model/blob/cae272936f714999fb962d97fa6a8b68507b14b3/samtranslator/model/sam_resources.py#L139-L140

moelasmar (approved):

PR #2343, opened 2022-03-08
test: Add retries on flaky integration test

mildaniel:
*Issue #, if available:*

*Description of changes:*

*Checklist:*

    - [ ] Correct values
- [ ] `make pr` passes
- [ ] Update documentation

*Examples?*

### DIFF
--- a/integration/combination/test_api_with_gateway_responses.py
+++ b/integration/combination/test_api_with_gateway_responses.py
@@ -1,6 +1,7 @@
 from unittest.case import skipIf
 
 from integration.helpers.base_test import BaseTest
+from integration.helpers.deployer.utils.retry import retry
 from integration.helpers.resource import current_region_does_not_support
 from integration.config.service_names import GATEWAY_RESPONSES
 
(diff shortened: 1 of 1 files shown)


integration/combination/test_api_with_gateway_responses.py
```diff
@@ -30,8 +31,12 @@ def test_gateway_responses(self):
         self.assertEqual(gateway_response.get("statusCode"), None, "gatewayResponse: status code must be none")
 
         base_url = stack_outputs["ApiUrl"]
-        response = self.verify_get_request_response(base_url + "iam", 403)
-        access_control_allow_origin = response.headers["Access-Control-Allow-Origin"]
+        self._verify_request_response_and_cors(base_url + "iam", 403)
```
jfuss, line 34: I assume this is due to us calling to quickly after a deployment? If so, should we make our tests wait for some amount of time before trying to verify? This might be true across our integration tests. 

Retrying is fine but I wonder if there is a wider fix we could/should be applying.

jfuss (approved):

integration/combination/test_api_with_gateway_responses.py
mildaniel: Yeah, I do wonder if there is a generic "Wait after stack creation" mechanism we should be employing. It's hard to say exactly why these are failing sometimes. I was able to reproduce the error a few times, but nothing really stood out as an issue in the deployment. My best guess is either some transient bug in APIGW/CFN, or we're trying to assert before the stack is fully ready.

It's also impossible to say for sure if the fix works. I tried to reproduce after making these changes and wasn't able to, but it could just be that the probability of the failure event is low and I didn't hit it. The retry decorator does employ a timeout before retries, so we can monitor these changes and if we don't see the same error again, we can add some global waiting mechanism.

jfuss: For inconsistent failures it is hard to tell if the fix does work, as it could be that we just didn't hit the right situation. It is good that you were able to repro at some rate and not at all after though. 

My guess in what is happening is some kind of network latency or some kind of eventually consistency in that the stack is deployed but when we call too soon after, we hit some state that is outdated (or not there). Completely just guessing here, but I do think having a wider solution here that will be helpful. It will make our testing run a little longer but I think that is ok if the result is we can rely on our integration tests. Maybe worth throwing something in the backlog that we can handle sooner rather than later.

hawflau (approved):

PR #2349, opened 2022-03-15
 Manage Policy Loader error handling

jonife:
A spike of invocations caused higher concurrent executions and thus more cold starts as Lambda spun up more sandboxes to handle the higher concurrency. More cold starts made more ListPolicies to IAM and therefore led to a higher chance of being throttled.

*Description of changes:*

*Checklist:*

    - [ ] Correct values
- [ ] `make pr` passes
- [ ] Update documentation

*Examples?*

### DIFF
--- a/samtranslator/translator/managed_policy_translator.py
+++ b/samtranslator/translator/managed_policy_translator.py
@@ -1,4 +1,5 @@
 import logging
+import botocore
 
 from samtranslator.metrics.method_decorator import cw_timer
 
@@ -32,5 +33,8 @@ class ManagedPolicyLoader(object):
 
     def load(self):
         if self._policy_map is None:
-            self._load_policies_from_iam()
+            try:
+                self._load_policies_from_iam()
+            except botocore.exceptions.ClientError as error:
+                raise error
         return self._policy_map


samtranslator/translator/managed_policy_translator.py
```diff
@@ -32,5 +33,8 @@ def _load_policies_from_iam(self):
 
     def load(self):
         if self._policy_map is None:
-            self._load_policies_from_iam()
+            try:
+                self._load_policies_from_iam()
+            except botocore.exceptions.ClientError as error:
```
jfuss, line 38: Should we be catching all Errors here? Can we log the error?

samtranslator/translator/managed_policy_translator.py
```diff
@@ -32,5 +33,8 @@ def _load_policies_from_iam(self):
...
+                raise error
```
jfuss, line 39: Not sure I understand the purpose of just catching the error and then immediately raising it. If that is the case, why do we need the `try: except:` at all?

jfuss (changes requested):

jfuss:
Closing as this has not been touched by the author in 2 months with open comments. If you would like to continue, feel free to reopen or create a new PR.

PR #2352, opened 2022-03-16
fix: Add retries to SAR service calls on throttle

mildaniel:
*Issue #, if available:*

*Description of changes:*

*Checklist:*

    - [ ] Correct values
- [x] `make pr` passes
- [ ] Update documentation

*Examples?*


samtranslator/plugins/__init__.py
```diff
@@ -1,152 +1,10 @@
 import logging
 
-from samtranslator.model.exceptions import InvalidResourceException, InvalidDocumentException
 from enum import Enum
 
 LOG = logging.getLogger(__name__)
 
 
-class SamPlugins(object):
-    """
```
mildaniel, line 10: Move this to a separate file to resolve a circular dependency.
`samtranslator.plugins (__init__.py)` -> `samtranslator.model.exceptions…` -> `samtranslator.plugins.LifeCycleEvents`

samtranslator/plugins/application/serverless_app_plugin.py
```diff
@@ -118,11 +118,30 @@ def on_before_transform_template(self, template_dict):
...
-                    service_call(app_id, semver, key, logical_id)
+                    self._make_service_call_with_retry(service_call, app_id, semver, key, logical_id)
                 except InvalidResourceException as e:
                     # Catch all InvalidResourceExceptions, raise those in the before_resource_transform target.
                     self._applications[key] = e
 
+    def _make_service_call_with_retry(self, service_call, app_id, semver, key, logical_id):
+        call_succeeded = False
+        start_time = time()
+        while (time() - start_time) < self.TEMPLATE_WAIT_TIMEOUT_SECONDS:
```
hawflau, line 129: There's also a similar retry pattern used in `on_after_transform_template`. I can see things are done a bit differently within this retry and that retry. But can we try to refactor the retry part into a decorator that can be used on both?

hawflau (changes requested):

moelasmar (approved):

hawflau (approved):

PR #2354, opened 2022-03-17
add iam retry emf metrics

mingkun2020:
*Issue #, if available:*

*Description of changes:*
- enable samtranslator to publish EMF metrics
- collect IAM EMF metrics

- validate in beta pipeline

*Checklist:*

    - [ ] Correct values
- [ ] `make pr` passes
- [ ] Update documentation

*Examples?*

### DIFF
--- a/samtranslator/metrics/metrics.py
+++ b/samtranslator/metrics/metrics.py
@@ -3,6 +3,7 @@ Helper classes to publish metrics
 """
 import logging
 from datetime import datetime
+from aws_embedded_metrics import metric_scope
 
 LOG = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


samtranslator/metrics/metrics.py
```diff
@@ -3,6 +3,7 @@
 """
 import logging
 from datetime import datetime
+from aws_embedded_metrics import metric_scope
```
jfuss, line 6: Were does this module come from? I don't think this is anything in SAM and therefore will cause an import error and crash.

samtranslator/translator/managed_policy_translator.py
```diff
@@ -11,6 +13,8 @@ def __init__(self, iam_client):
         self._policy_map = None
         self.max_items = 1000
 
+        self.iam_metrics = Metrics(namespace="Downstream-retry", metrics_publisher=EMFMetricsPublisher())
```
jfuss, line 16: Should we hardcode `Downstream-retry` or should we pull this from an env var we set on the service?

samtranslator/translator/managed_policy_translator.py
```diff
@@ -27,9 +31,17 @@ def _load_policies_from_iam(self):
...
             name_to_arn_map.update(map(lambda x: (x["PolicyName"], x["Arn"]), page["Policies"]))
 
+            retry_count = page.get("ResponseMetadata", {}).get("RetryAttempts")
+            self.iam_metrics.record_count("Retry", retry_count, {"downstream_service": "iam"})
+
         LOG.info("Finished loading policies from IAM.")
         self._policy_map = name_to_arn_map
 
+        try:
+            self.iam_metrics.publish()
```
jfuss, line 41: Doesn't this will always print information. We should disable this if not running in the service (like the other metrics we have).

samtranslator/translator/managed_policy_translator.py
```diff
@@ -27,9 +31,17 @@ def _load_policies_from_iam(self):
...
+        except Exception as e:
+            LOG.warning("Failed to publish iam emf metrics: %s", e)
```
jfuss, line 43: Why is this a warning and not `exception`?

jfuss (changes requested):

samtranslator/metrics/metrics.py
mingkun2020: @jfuss Thanks for the comments. We already have another mechanism to collect the IAM metrics, so close this pr.

PR #2359, opened 2022-03-23
sync foss changes

mingkun2020:
*Description of changes:*
Sync foss public repo with recent changes.

Tested in local and success

- [x] `make pr` passes
- [ ] Update documentation

*Examples?*


mndeveci (approved):

jfuss:
@mingkun2020 builds are failing due to improper formatting. Can you run `make black`?

mingkun2020:
@jfuss Just wanna do the reformatting after second reviewer's approvals. If I do reformat now it may hide the actual changes which might confuse the reviewer.

jfuss:
@mingkun2020 Please do the reformatted and I can review. Reformatting through black shouldn't hide changes, or I am miss understanding your concern. 

I am not comfortable approving or reviewing while the builds do not work.

integration/helpers/client_provider.py
```diff
@@ -115,6 +116,16 @@ def cloudwatch_event_client(self):
                 self._cloudwatch_event_client = boto3.client("events")
         return self._cloudwatch_event_client
 
+    @property
+    def cloudwatch_client(self):
+        """
+        CloudWatch Client
+        """
+        with self._lock:
```
jfuss, line 124: Why do we have a lock? First reaction is this seems odd but don't understand why we would need to do this.

integration/resources/expected/single/basic_api_with_mode_update.json
```diff
@@ -1,10 +1,10 @@
 [
         {"LogicalResourceId": "MyApi", "ResourceType": "AWS::ApiGateway::RestApi"},
-        {"LogicalResourceId": "MyApiDeploymentb30ecf9df1", "ResourceType": "AWS::ApiGateway::Deployment"},
+        {"LogicalResourceId": "MyApiDeployment", "ResourceType": "AWS::ApiGateway::Deployment"},
```
jfuss, line 3: Why are the hashes removed here?

integration/helpers/client_provider.py
mingkun2020: This client provider may call by multi threads, so we add lock here to prevent concurrency issues.

jfuss: I wonder if there is a different way but not critical for this. Thanks for explaining.

integration/resources/expected/single/basic_api_with_mode_update.json
jfuss: So the hashes are their to help guard against changes that accidentally change how SAM hashes those values. When the hashes change, this causes new resources in the stacks and can cause impact to. We had an event where we changed how Layers hashing was done and this caused throttling for customers.

Can we add these back?

jfuss:
Approving with the condition we add back the hashing, otherwise our tests are not validating we are not spontaneously changing how the logical id's are generated. This causes unknowing deployments on customers, so the logical ids should always be deterministic.

jfuss (approved):

PR #2368, opened 2022-04-01
reduce CloudFormation template size

PatMyron:
also: https://github.com/aws/aws-cdk/pull/19656, https://github.com/cloudtools/troposphere/pull/2028, https://github.com/awslabs/goformation/pull/445

[CloudFormation templates can currently only be 1MB](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cloudformation-limits.html), and [template size limits apply even after transformation processing](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/template-macros.html#template-macros-use)

Simply reducing indentation from 2 to 1 should remove ~1/4 of the template file size for everyone by default while still preserving indentation formatting. Beyond improving the default, those wishing to reduce readability for further reduced file size could opt into using something like [`jq`](https://stedolan.github.io/jq/) on their own for now

*Checklist:*

    - [ ] Correct values
- [ ] `make pr` passes
- [ ] Update documentation

### DIFF
--- a/bin/sam-translate.py
+++ b/bin/sam-translate.py
@@ -107,7 +107,7 @@ def transform_template(input_file_path, output_file_path):
             region=None,
         )
         cloud_formation_template = transform(sam_template, {}, ManagedPolicyLoader(iam_client), feature_toggle)
-        cloud_formation_template_prettified = json.dumps(cloud_formation_template, indent=2)
+        cloud_formation_template_prettified = json.dumps(cloud_formation_template, indent=1)
 
         with open(output_file_path, "w") as f:
             f.write(cloud_formation_template_prettified)


jfuss:
This seems like a fine change but this script is/should only be used for testing changes to the SAM repo. If you are using this to deploy a SAM template, you are using something not maintained for that and should be using SAM CLI or AWS CLI.

jfuss (approved):

PatMyron:
@jfuss not sure I understand the relationship between this repo and https://github.com/aws/aws-sam-cli if you're saying these changes would be more meaningful there, mind linking to where these changes could be made more meaningfully?

jfuss:
@PatMyron This repo just holds the SAM Spec to CloudFormation conversion. SAM CLI's repo uses this to read the spec but holds the `sam *` commands.

There are a couple places you will need to update:
1. https://github.com/aws/aws-sam-cli/blob/587ad2d22c3066b472bf3c8029e2e84eb643dda7/samcli/yamlhelper.py#L96
2. https://github.com/aws/aws-sam-cli/blob/587ad2d22c3066b472bf3c8029e2e84eb643dda7/samcli/commands/package/package_context.py#L163

For 1, this covers `sam build` and `sam package` if the original template is in yaml. 2 covers the json case within `sam package`. I think that hits them all but 100% not sure.

moelasmar (approved):

PR #2376, opened 2022-04-12
fix: use InvalidDocumentException instead of ValueError

mndeveci:
*Issue #, if available:*
N/A

*Description of changes:*
Replaced `ValueError` exceptions with `InvalidDocumentException` which should help SAM Transform to gracefully exit execution for invalid Swagger/OpenAPI inputs.

*Checklist:*

    - [ ] Correct values
- [x] `make pr` passes
- [ ] Update documentation

### DIFF
--- a/samtranslator/open_api/open_api.py
+++ b/samtranslator/open_api/open_api.py
@@ -37,10 +37,10 @@ class OpenApiEditor(object):
         modifications on this copy.
 
         :param dict doc: OpenApi document as a dictionary
-        :raises ValueError: If the input OpenApi document does not meet the basic OpenApi requirements.
+        :raises InvalidDocumentException: If the input OpenApi document does not meet the basic OpenApi requirements.
         """
         if not OpenApiEditor.is_valid(doc):
-            raise ValueError(
+            raise InvalidDocumentException(
                 "Invalid OpenApi document. "
                 "Invalid values or missing keys for 'openapi' or 'paths' in 'DefinitionBody'."
             )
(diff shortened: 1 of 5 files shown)


samtranslator/open_api/open_api.py
```diff
@@ -37,10 +37,10 @@ def __init__(self, doc):
         modifications on this copy.
 
         :param dict doc: OpenApi document as a dictionary
-        :raises ValueError: If the input OpenApi document does not meet the basic OpenApi requirements.
+        :raises InvalidDocumentException: If the input OpenApi document does not meet the basic OpenApi requirements.
         """
         if not OpenApiEditor.is_valid(doc):
-            raise ValueError(
+            raise InvalidDocumentException(
```
jfuss, line 43: Thank you! We should probably do another pass and stop throwing generic python errors/exceptions.

samtranslator/open_api/open_api.py
```diff
@@ -37,10 +37,10 @@ def __init__(self, doc):
         modifications on this copy.
 
         :param dict doc: OpenApi document as a dictionary
-        :raises ValueError: If the input OpenApi document does not meet the basic OpenApi requirements.
+        :raises InvalidDocumentException: If the input OpenApi document does not meet the basic OpenApi requirements.
```
jfuss, line 40: Should this be InvalidDocument or InvalidResource?

jfuss (approved):

samtranslator/open_api/open_api.py
mndeveci: We can throw `InvalidResource` but we need logical id of the resource.

jfuss: Got it. That gets me thinking but nothing specific for this PR.

mildaniel (approved):

Issue #2373, opened 2022-04-07
Function.FunctionUrlConfig generates incorrect policy for aliased function


aidansteele:
### Description:

I have created a function that has an alias and URL configuration with the following template:

```yaml
Transform: AWS::Serverless-2016-10-31

Resources:
  Function:
    Type: AWS::Serverless::Function
    Properties:
      CodeUri: ./bootstrap
      Architectures: [arm64]
      Runtime: provided.al2
      Handler: unused
      AutoPublishAlias: live # notice defined alias
      FunctionUrlConfig:
        AuthType: NONE
```

This is the template after the serverless transformation. Note that the 7th line is incorrectly referencing the _function_ and not the _function's alias_.

```yaml
Resources:
  FunctionUrlPublicPermissions:
    Type: AWS::Lambda::Permission
    Properties:
      Action: lambda:InvokeFunctionUrl
      FunctionName:
        Ref: Function # this should be Ref: FunctionAliaslive
      Principal: "*"
      FunctionUrlAuthType: NONE
  FunctionUrl:
    Type: AWS::Lambda::Url
    Properties:
      TargetFunctionArn:
        Ref: FunctionAliaslive
      AuthType: NONE
  Function:
    Type: AWS::Lambda::Function
    Properties:
      Code:
        S3Bucket: stackit-ap-southeast-2-0123465690
        S3Key: urlshenanigans/bootstrap.zip/b15b257b4b1d56015ffb6160f1564d2e
        S3ObjectVersion: ThuT.NfqNqrs7apqjJT7nb9P4ox8TpiT
      Handler: unused
      Role:
        Fn::GetAtt:
          - FunctionRole
          - Arn
      Runtime: provided.al2
      Timeout: 30
      Tags:
        - Key: lambda:createdBy
          Value: SAM
      Architectures:
        - arm64
  FunctionVersionbe0cebe936:
    Type: AWS::Lambda::Version
    DeletionPolicy: Retain
    Properties:
      FunctionName:
        Ref: Function
  FunctionAliaslive:
    Type: AWS::Lambda::Alias
    Properties:
      Name: live
      FunctionName:
        Ref: Function
      FunctionVersion:
        Fn::GetAtt:
          - FunctionVersionbe0cebe936
          - Version
  FunctionRole:
    Type: AWS::IAM::Role
    Properties:
      AssumeRolePolicyDocument:
        Version: '2012-10-17'
        Statement:
          - Action:
              - sts:AssumeRole
[...]

PR #2378, opened 2022-04-13
Furl AutopublishAlias bug fix 

jonife:
*Issue #, if available:*
https://github.com/aws/serverless-application-model/issues/2373

*Description of changes:*
Grant public access to function alias

*Checklist:*

    - [x] Correct values
- [x] `make pr` passes
- [ ] Update documentation

*Examples?*

### DIFF
--- a/samtranslator/model/sam_resources.py
+++ b/samtranslator/model/sam_resources.py
@@ -175,7 +175,7 @@ class SamFunction(SamResourceMacro):
         if self.FunctionUrlConfig:
             lambda_url = self._construct_function_url(lambda_function, lambda_alias)
             resources.append(lambda_url)
-            url_permission = self._construct_url_permission(lambda_function)
+            url_permission = self._construct_url_permission(lambda_function, lambda_alias)
             if url_permission:
                 resources.append(url_permission)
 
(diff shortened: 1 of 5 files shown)


mndeveci (changes requested):
Thanks for the PR @jonife .

Can we check the following failure?
```
FAILED tests/model/api/test_http_api_generator.py::TestHttpApiGenerator::test_enabling_auth_iam_does_not_clobber_conflicting_custom_authorizer - samtranslator.model.exceptions.InvalidResourceException: ('HttpApiId', "Specify either 'Defi...
```

And also, will this require any documentation change?

jfuss (approved):
LGTM, only question is are there integration tests we need to update with these? We recently adding python integration tests in the repo and want to make sure we have the correct coverage.

jonife:
> LGTM, only question is are there integration tests we need to update with these? We recently adding python integration tests in the repo and want to make sure we have the correct coverage.

can't seem to think of any one of the top of my head, i think the integration test covers everything

mndeveci (approved):

hawflau:
> > LGTM, only question is are there integration tests we need to update with these? We recently adding python integration tests in the repo and want to make sure we have the correct coverage.
> 
> can't seem to think of any one of the top of my head, i think the integration test covers everything

Should we test calling the Function URL to confirm if the Lambda Function Alias is invoked without any error? From what I understand, the integration test for FURL does not make any call to the created FURL.

hawflau:
note: we will submit another PR to address the integration tests

PR #2380, opened 2022-04-20
chore: remove sys.path override

jfuss:
*Issue #, if available:*
#1911 

*Description of changes:*
Remove `sys.path` overrides that were done to support a testing script.

Remove `sys.path` overriding.

*Checklist:*

    - [ ] Correct values
- [ ] `make pr` passes
- [ ] Update documentation

*Examples?*

### DIFF
--- a/samtranslator/feature_toggle/feature_toggle.py
+++ b/samtranslator/feature_toggle/feature_toggle.py
@@ -13,9 +13,6 @@ from samtranslator.feature_toggle.dialup import (
 )
 from samtranslator.metrics.method_decorator import cw_timer
 
-my_path = os.path.dirname(os.path.abspath(__file__))
-sys.path.insert(0, my_path + "/..")
-
 LOG = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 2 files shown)


samtranslator/feature_toggle/feature_toggle.py
```diff
@@ -13,9 +13,6 @@
 )
 from samtranslator.metrics.method_decorator import cw_timer
 
-my_path = os.path.dirname(os.path.abspath(__file__))
```
mndeveci, line 16: Why we needed this before?

mndeveci (approved):

samtranslator/feature_toggle/feature_toggle.py
jfuss: @mndeveci See discussion here: https://github.com/aws/serverless-application-model/pull/2076#discussion_r796798622

TLDR: Was due to the test script. Since this is clobbering things on customers machines, I removed feature_toggle from the test script. If we really need it there (which I don't think we do), we can find a different way to add it back.

hawflau (approved):

PR #2401, opened 2022-05-19
Update FURL integ test to test invoke

hawflau:
*Issue #, if available:* 
N/A

*Description of changes:*
Update FURL integration test to test invoke the created Lambda Function

Running the updated tests from my laptop

*Checklist:*

    - [ ] Correct values
- [x] `make pr` passes
- [ ] Update documentation

*Examples?*

### DIFF
--- a/requirements/dev.txt
+++ b/requirements/dev.txt
@@ -14,7 +14,7 @@ parameterized~=0.7.4
 # Integration tests
 click~=7.1
 dateparser~=0.7
-boto3~=1.17
+boto3>=1.23,<2
 
 # Requirements for examples
 requests~=2.24.0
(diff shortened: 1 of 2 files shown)


requirements/dev.txt
```diff
@@ -14,7 +14,7 @@ parameterized~=0.7.4
 # Integration tests
 click~=7.1
 dateparser~=0.7
-boto3~=1.17
+boto3>=1.23,<2
```
mndeveci, line 17: Why we needed boto version bump here? In order to support a new parameter to `lambda_client.invoke`?

mndeveci (approved):

requirements/dev.txt
hawflau: It was a miss from the FURL PR actually, as the FURL integration test is using `lambda_client.get_function_url_config`, which is only available in boto3 1.22.15 or above

jfuss (approved):

PR #2409, opened 2022-06-06
fix: check null method for open api

mndeveci:
*Issue #, if available:*
N/A

*Description of changes:*
Checks if method is not `None` before proceeding to work on it.

Adding a new test case and verifying the older ones.

*Checklist:*

    - [ ] Correct values
- [x] `make pr` passes
- [ ] Update documentation

*Examples?*

### DIFF
--- a/samtranslator/open_api/open_api.py
+++ b/samtranslator/open_api/open_api.py
@@ -397,6 +397,11 @@ class OpenApiEditor(object):
                             ]
                         )
                     for method_definition in self.get_conditional_contents(method):
+                        # check if there is any method_definition given by customer
+                        if not method_definition:
+                            raise InvalidDocumentException(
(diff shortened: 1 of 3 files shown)


jfuss (approved):

torresxb1 (changes requested):
I see this `for method_definition in self.get_conditional_contents(method)` is used a few times in the file. In order to catch this early without having to run this validation each time, could we add this check in the constructor?
We do something like this in the SwaggerEditor constructor: https://github.com/aws/serverless-application-model/blob/develop/samtranslator/swagger/swagger.py#L65

By the way, could we also add this check on the SwaggerEditor 🙏

torresxb1:
oh, I was too late

mndeveci:
> oh, I was too late

No worries, let me raise a new PR to address your comment.

PR #2413, opened 2022-06-07
fix: Allow Implicit HTTP APIs to write when value of method is None

jfuss:
SAM would crash if a the customer was using Implicit APIs that had a method
whose value was None/Null. SAM assumed the value would be a dictonary and therefore
would crash with a NoneType error when adding a key/value pair.

*Issue #, if available:*
N/A

*Description of changes:*
SAM would crash if a the customer was using Implicit APIs that had a method
whose value was None/Null. SAM assumed the value would be a dictonary and therefore
would crash with a NoneType error when adding a key/value pair.

Ran the example in the tests before the change, which produced a crash. The tests passed with the patch applied.

*Checklist:*

    - [x] Correct values
- [x] `make pr` passes
- [ ] Update documentation

*Examples?*


tests/openapi/test_openapi.py
```diff
@@ -203,11 +203,32 @@ def setUp(self):
...
                 "/foo": {"post": {"a": [1, 2, "b"], "responses": {"something": "is already here"}}},
                 "/bar": {"get": {_X_INTEGRATION: {"a": "b"}}},
+                "/nullmethod": {"get": None},
             },
         }
 
         self.editor = OpenApiEditor(self.original_openapi)
 
+    def test_must_override_null_path(self):
+        path = "/newpath"
```
moelasmar, line 213: could you explain what you are trying to test here. my understanding is you are testing that the `/nullmethod` path in the `original_openapi` defined in the `setUp` method  will be override. But I am not able to understand why you used `/newpath`

tests/translator/output/aws-us-gov/function_with_alias_and_code_sha256.json
```diff
@@ -0,0 +1,89 @@
+{
```
moelasmar, line 1: why did this file got changed? I am not able to understand how it is related to this change.

tests/openapi/test_openapi.py
jfuss: I did change it locally but somehow got a conflict and didn't update. Not sure how this happened but will be updated in the next diff.

moelasmar (approved):

PR #2425, opened 2022-06-27
fix: Fail gracefully when AutoPublishAlias is an empty string

jfuss:
*Issue #, if available:*
n/a

*Description of changes:*
Check if AutoPublishAlias is "truthy", which matches what we do elsewhere in the code. 

Added the test first, which failed and after patching the test passed. 

*Checklist:*

    - [x] Correct values
- [x] `make pr` passes
- [ ] Update documentation

*Examples?*


tests/translator/output/error_auto_publish_alias_empty_string.json
```diff
@@ -0,0 +1,3 @@
+{
+    "errorMessage": "Invalid Serverless Application Specification document. Number of errors found: 1. Resource with id [MinimalFunction] is invalid. 'DeploymentPreference' requires AutoPublishAlias property to be specified."
```
torresxb1, line 2: if we wanna be even more specific we could say ```... requires a non-empty AutoPublishAlias ...```

PR #2450, opened 2022-07-20
feat - Creating ability to add validators to api gw method without needing a model

Rondineli:
*Issue #, if available:* #1403 - Feature improvement

*Description of changes:*  Add ability to add validators if no Model is present. In some cases, when request Parameters is present but no model we still want to add validators to the path. This PR aim add a second way of add a validator to the swagger specs, even if a model is not defined.
In case both are defined (model and RequestValidators) it will add the unique validator to the path, keeping model and requestParameters as part of the swagger spec definitions. 

*Description of how you validated changes:* Create a path with RequestParameters and add a RequestValidators to it. It will update the validators to the final spec, validating the requests as intended.

*Checklist:*

    - [x] Correct values
    - [-] [Intrinsic Functions](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference.html)
- [x] `make pr` passes
- [ ] Update documentation
- [*] Verify transformed template deploys and application functions as expected
- [*] Do these changes include any template validations?
    - [-] Did the newly validated properties support intrinsics prior to adding the validations? (If unsure, please review [Intrinsic Functions](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference.html) before proceeding).
        - [*] Does the pull request ensure that intrinsics remain functional with the new validations?

*Examples?*


samtranslator/model/eventsources/push.py
```diff
@@ -800,31 +828,17 @@ def _add_swagger_integration(self, api, function, intrinsics_resolver):
...
-                        path=self.Path,
-                        method_name=self.Method,
-                        validate_body=validate_body,
-                        validate_parameters=validate_parameters,
-                    )
+                self._add_validators(editor, validate_body, validate_parameters)
+
+        if self.RequestValidators:
+            # default is no Validators
+            validators = {"ValidateBody": None, "ValidateParameters": None}
```
aahung, line 835: Seems this dictionary is not necessary and we can directly use

```
self._add_validators(editor, self.RequestValidators.get("ValidateBody"), self.RequestValidators.get("ValidateParameters"))
```

tests/translator/output/error_api_request_model_with_strings_validator.json
```diff
@@ -1,3 +1,3 @@
 {
-  "errorMessage": "Invalid Serverless Application Specification document. Number of errors found: 1. Resource with id [HtmlFunction] is invalid. Event with id [GetHtml] is invalid. Unable to set Validator to RequestModel [User] on API method [get] for path [/] ValidateBody and ValidateParameters must be a boolean type, strings or intrinsics are not supported."
+  "errorMessage": "Invalid Serverless Application Specification document. Number of errors found: 1. Resource with id [HtmlFunction] is invalid. Event with id [GetHtml] is invalid. Unable to set Validator on API method [get] for path [/] ValidateBody and ValidateParameters must be a boolean type, strings or intrinsics are not supported."
```
aahung, line 2: I guess the error message is still actionable without mentioning the model. Fine with it

aahung (commented):
thanks for the PR!

Rondineli:
@aahung fixed, thanks for the review!!

aahung (approved):
lgtm!

samtranslator/model/eventsources/push.py
```diff
@@ -800,31 +828,16 @@ def _add_swagger_integration(self, api, function, intrinsics_resolver):
...
-                        method_name=self.Method,
-                        validate_body=validate_body,
-                        validate_parameters=validate_parameters,
-                    )
+        if self.RequestValidators:
+            # default is no Validators
+            # if we have validator configured to a request parameter
+            # we want to add it to the editor but remove it from original dict before it iterates
+            # If validators is present or not, this method will handle the check
+            self._add_validators(
```
moelasmar, line 838: I am not sure if this should be the correct behaviour. I found it confusing, if the customer defined set the validation to `True` in the `RequestModel` but to `False` in the `RequestValidators`, why should we disable the validations. I think it is not a good customer experience.

moelasmar:
Thanks @Rondineli  for your contribution. I just have a concern about the customer experience of the new property `RequestValidators`. My issue is if the customer enables the validations in the requestModel, and disables it in the RequestValidators, I found it confusing for the customers to use the request validators to overwrite the request model.

My understanding of the purpose of this PR is to allow the customers to set validations for a service Path even if there is no model. Let me share my thoughts, and I will also validate it with the team. Is it possible to allow the customer to do that by making the `Model` attribute to be optional in the `RequestModel` property (https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-property-function-requestmodel.html). So the customer can enable the validations even if there is no model.

Rondineli:
@moelasmar thanks for replying. That makes sense and it is a valid concern. My thoughts were get a way of not change the actual behaviour as it can break existent config. Let me know the best way to proceed then I can change my PR to meet the requirements. The idea of changing the model to no be required also sounds confusing as it still having to have a "RequestModel" on the template.

We could maybe validate If the template has a model with a validator and if both are present, raise an exception saying that it must have either a validator on  its model block to the method or a validator on its request parameters to the method.

In any case I would be happy to change the PR to meet the desired behaviour.

Issue #2452, opened 2022-07-22
Removing the Alarms property from the serverless function DeploymentPreference does not remove the alarms from the created deployment group


moelasmar:
### Description:
When in a SAM template, Alarms are removed from DeploymentPreference it doesn't not remove Alarm value on CodeDeploy Deployment Group console.

### Steps to reproduce:
1- deploy the following SAM Template
```
AWSTemplateFormatVersion: '2010-09-09'
Transform: 'AWS::Serverless-2016-10-31'

Description: Simple Hello World Example

Resources:
  HelloWorldFunction:
    Type: AWS::Serverless::Function
    Properties:
       Handler: HelloWorld.handler_aws
       CodeUri: ./
       Runtime: python3.9
       AutoPublishAlias: test
       DeploymentPreference:
          Enabled: true
          Type: Canary10Percent10Minutes
          Alarms:
            - Ref: AliasErrorMetricGreaterThanZeroAlarm
  AliasErrorMetricGreaterThanZeroAlarm:
    Type: AWS::CloudWatch::Alarm
    Properties:
      AlarmName: AliasErrorMetricGreaterThanZeroAlarm
      ComparisonOperator: LessThanLowerOrGreaterThanUpperThreshold
      EvaluationPeriods: 1
      Metrics:
      - Expression: ANOMALY_DETECTION_BAND(m1, 2)
        Id: ad1
      - Id: m1
        MetricStat:
          Metric:
            MetricName: Invocations
            Namespace: AWS/Lambda
          Period: !!int 86400
          Stat: Sum
      ThresholdMetricId: ad1
      TreatMissingData: breaching
```
2- check the created Deployment Group instance, and make sure that the alarms configuration is there.
3- update the created stack using the following SAM template:
```
AWSTemplateFormatVersion: '2010-09-09'
Transform: 'AWS::Serverless-2016-10-31'

Description: Simple Hello World Example

Resources:
  HelloWorldFunction:
    Type: AWS::Serverless::Function
    Properties:
       Handler: HelloWorld.handler_aws
       CodeUri: ./
       Runtime: python3.9
       AutoPublishAlias: test
       DeploymentPreference:
          Enabled: true
          Type: Canary10Percent10Minutes
```
4- check the deployment group again. You will find the alarms is still there, although the expectation is it should be removed.

The issue is the generated CFN template after removing the Alarm configuration is as following:
```
HelloWorldFunctionDeploymentGroup:
    Properties:
      ApplicationName:
[...]

PR #2453, opened 2022-07-22
fix: clean the DeploymentGroup alarms configuration in case of updating the DeploymentPreference and removing the alarms attribute.

moelasmar:
*Issue #, if available:*
https://github.com/aws/serverless-application-model/issues/2452

*Description of changes:*
If the Alarms property in Deployment preference is not provided, update the generated DeploymentGroup resource to contain an empty AlarmsConfiguration. So in case of updating the DeploymentGroup resource, the empty AlarmsConfiguration will force cleaning the alarms value from the DeploymentGroup resource.

*Checklist:*

- [X] Add/update [unit tests](https://github.com/aws/serverless-application-model/blob/develop/DEVELOPMENT_GUIDE.md#unit-testing-with-multiple-python-versions) using:
    - [X] Correct values
- [ ] `make pr` passes
- [ ] Update documentation

*Examples?*

### DIFF
--- a/samtranslator/model/preferences/deployment_preference_collection.py
+++ b/samtranslator/model/preferences/deployment_preference_collection.py
@@ -226,7 +226,13 @@ class DeploymentPreferenceCollection(object):
             If Alarms is in the wrong format
         """
         if not preference_alarms or is_intrinsic_no_value(preference_alarms):
-            return None
+            # return the empty alarms configuration in case if there is no alarms at all defined in the
+            # DeploymentPreference to force CFN to update the deployment group and remove the alarms values from the
+            # created DeploymentGroup. Check issue https://github.com/aws/serverless-application-model/issues/2452
+            return {
+                "Enabled": False,
+                "Alarms": [],
+            }
 
         if is_intrinsic_if(preference_alarms):
             processed_alarms = copy.deepcopy(preference_alarms)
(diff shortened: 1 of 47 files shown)


samtranslator/model/preferences/deployment_preference_collection.py
```diff
@@ -226,7 +226,13 @@ def _convert_alarms(self, preference_alarms):
             If Alarms is in the wrong format
         """
         if not preference_alarms or is_intrinsic_no_value(preference_alarms):
-            return None
+            # return the empty alarms configuration in case if there is no alarms at all defined in the
+            # DeploymentPreference to force CFN to update the deployment group and remove the alarms values from the
+            # created DeploymentGroup. Check issue https://github.com/aws/serverless-application-model/issues/2452
+            return {
```
hawflau, line 232: Is there any concern that it would be a breaking change? Or put it other way - is there any concern that there could be some customers relying on this incorrect behaviour?

moelasmar: The only thing that I can think about, if the customers did not remove the actual alarm, but removed the alarm property from the deployment preference. The customer will find that some of the deployments will be rolled back because of this alarm, and they want this behaviour. But I do not think that this is a real scenario, I always expect that when the customers remove the alarm configuration, they should expect no rollbacks

hawflau:
Closing the PR as the issue should be fixed on the CodeDeploy CFN resource.

PR #2455, opened 2022-07-25
chore: Add region skip condition and retry for http test

qingchm:
*Issue #, if available:*
N.A.

*Description of changes:*
To exclude regions for http tests so it does not get tested in regions not supporting http api

This change uses the same decorator used in other http api tests

*Checklist:*

    - [ ] Correct values
- [x] `make pr` passes
- [ ] Update documentation

*Examples?*

### DIFF
--- a/integration/single/test_function_with_http_api_and_auth.py
+++ b/integration/single/test_function_with_http_api_and_auth.py
@@ -14,6 +14,7 @@ class TestFunctionWithHttpApiAndAuth(BaseTest):
     """
     AWS::Lambda::Function tests with http api events and auth
     """
+
     @retry(
         stop=stop_after_attempt(5),
         wait=wait_exponential(multiplier=1, min=4, max=10) + wait_random(0, 1),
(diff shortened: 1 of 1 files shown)


integration/single/test_function_with_http_api_and_auth.py
```diff
@@ -1,11 +1,27 @@
...
+LOG = logging.getLogger(__name__)
 
 
+@skipIf(current_region_does_not_support(["HttpApi"]), "HttpApi is not supported in this testing region")
 class TestFunctionWithHttpApiAndAuth(BaseTest):
     """
     AWS::Lambda::Function tests with http api events and auth
     """
 
+    @retry(
```
sriram-mv, line 18: Could there be issues surrounding the fact that the stacks were not cleaned up correctly, is that a possibility here?

mildaniel (approved):

integration/single/test_function_with_http_api_and_auth.py
qingchm: Nope for each stack the stack name is generated with a unique hash, random.choice(string.ascii_lowercase) for i in range(RANDOM_SUFFIX_LENGTH), so a failed clean up will never affect stack creation

Issue #2301, opened 2022-01-20
SAM Validator Does Not Support SSM Parameter Types


CoshUS:
### Description:

SSM parameter types:
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/parameters-section-structure.html#aws-ssm-parameter-types

SAM validate does not support these parameter types and will error out in the validator.
https://github.com/aws/serverless-application-model/blob/develop/samtranslator/validator/sam_schema/definitions/parameter.json

### Observed result:

```
"Template schema validation reported the following errors: Parameters.SecurityGroupSSMEntry.Type]'AWS::SSM::Parameter::Value<List<AWS::EC2::SecurityGroup::Id>>'isnotoneof['String''Number''List<Number>''CommaDelimitedList''AWS::EC2::AvailabilityZone::Name''AWS::EC2::Image::Id''AWS::EC2::Instance::Id''AWS::EC2::KeyPair::KeyName''AWS::EC2::SecurityGroup::GroupName''AWS::EC2::SecurityGroup::Id''AWS::EC2::Subnet::Id''AWS::EC2::Volume::Id''AWS::EC2::VPC::Id''AWS::Route53::HostedZone::Id''List<AWS::EC2::AvailabilityZone::Name>''List<AWS::EC2::Image::Id>''List<AWS::EC2::Instance::Id>''List<AWS::EC2::SecurityGroup::GroupName>''List<AWS::EC2::SecurityGroup::Id>''List<AWS::EC2::Subnet::Id>''List<AWS::EC2::Volume::Id>''List<AWS::EC2::VPC::Id>''List<AWS::Route53::HostedZone::Id>''List<String>']".
```

### Expected result:

SSM Parameter types should be valid.

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

1. OS:
2. If using SAM CLI, `sam --version`:
3. AWS region:

`Add --debug flag to any SAM CLI commands you are running`

ljayatRabbit:
The same happened to us :(

ljayatRabbit:
@CoshUS
I went through the docs and, correct me if I'm wrong, I think the problem lies here: https://github.com/aws/serverless-application-model/blob/develop/samtranslator/validator/sam_schema/definitions/parameter.json#L37

Could adding the new type the enum solve this problem? 
Something like:
```json
        "Type": {
          "enum": [
            "String",
            "Number",
            "List<Number>",
            "CommaDelimitedList",
[...]

PR #2469, opened 2022-08-10
feat: adding support for SSM Parameters (string) as Type

LautaroJayat:
*Issue #, if available:*
https://github.com/aws/serverless-application-model/issues/2301

*Description of changes:*
* added new entry on parameters.json > Types > enum
* added one input to test all correct types pass
* added one input and output to test that it will fail if a Parameter is not included in the enum from parameters.json
* updated test_validator_root to include these new cases

* Added missing tests to validate Parameter types are included in the parameter types enum.

*Checklist:*

    - [x] Correct values
- [ ] `make pr` passes
- [ ] Update documentation

*Examples?*


hawflau (approved):
Thanks for the PR! LGTM!

LautaroJayat:
If someone could give me some info on what are the actual SSM Parameter Types that SAM support, I would really love to update the PR to include all of them.

moelasmar (approved):

hawflau:
> If someone could give me some info on what are the actual SSM Parameter Types that SAM support, I would really love to update the PR to include all of them.

That's a good question. In fact, SAM works as a [CloudFormation Macro](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/template-macros.html#template-macros-author). As the doc suggests, it supplies `templateParameterValues` to SAM. So I think SAM would just receive whatever value (string or list) the SSM parameter resolves to. Here's the list of supported SSM parameter types in CloudFormation - https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/parameters-section-structure.html#aws-ssm-parameter-types.

LautaroJayat:
@hawflau,
For what I've read in the source, I understand SAM the same way as you are commenting above.
Nonetheless, I would like to only add the parameters that actually will end up working fine after the cloudformation resources are created. That is why I'm trying to find that list, just to let `sam validate` be consistent with the things that actually work after `sam build` and `sam deploy`.

hawflau:
@LautaroJayat 

Yep, the inaccurate part in my comment was " I think SAM would just receive whatever value (string or list) the SSM parameter resolves to". But that's not the point in this PR. I agree that the parity between `sam validate` and other commands like `sam deploy` should be fixed. 

Just want to confirm - do other supported SSM Parameter types also work in `sam deploy`? If so, I think we can include them in this PR.

LautaroJayat:
@hawflau 
I think we could use this list: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/parameters-section-structure.html#aws-specific-parameter-types-supported
Is the official documentation, I guess it is up to date.
If you agree, I will update the PR to include all of them.

PR #2473, opened 2022-08-22
Add validation for S3 NotificationConfiguration in Serverless Functin…

hawflau:
*Issue #, if available:*
N/A/

*Description of changes:*
Add Validation for `NotifcationConfiguration` in a S3 Bucket if it is used in a Serverless Function as an event source.

Updated test to catch this error.

*Checklist:*

    - [ ] Correct values
- [x] `make pr` passes
- [ ] Update documentation

*Examples?*

### DIFF
--- a/samtranslator/model/eventsources/push.py
+++ b/samtranslator/model/eventsources/push.py
@@ -313,7 +313,7 @@ class S3(PushEventSource):
         #   merging is literally "last one wins", which works fine because we linearly loop through the template once.
         #   The de-dupe happens inside `samtranslator.translator.Translator.translate` method when merging results of
         #   to_cloudformation() to output template.
-        self._inject_notification_configuration(function, bucket)
+        self._inject_notification_configuration(function, bucket, bucket_id)
         resources.append(S3Bucket.from_dict(bucket_id, bucket))
 
         return resources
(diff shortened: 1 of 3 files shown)


mildaniel (approved):

samtranslator/model/eventsources/push.py
```diff
@@ -407,13 +407,16 @@ def _inject_notification_configuration(self, function, bucket):
             notification_config = {}
             properties["NotificationConfiguration"] = notification_config
 
+        if not isinstance(notification_config, dict):
+            raise InvalidResourceException(bucket_id, "Invalid type for NotificationConfiguration. Must be a dict.")
```
jfuss, line 411: Is `dict` the right wording here? I only ask because `dict` is pretty Python specific. How does CloudFormation communicate things like this?

jfuss (changes requested):

samtranslator/model/eventsources/push.py
hawflau: CloudFormation doesn't seem have a generic name for an object/dict (e.g. it's called [`NotificationConfiguration`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-s3-bucket-notificationconfig.html) type for this case).

Instead of "dict", I think we can just say "Invalid type for NotificationConfiguration". I think it's informative enough for customer to find out how to fix it.

jfuss (approved):

PR #2487, opened 2022-09-01
chore: Integration test improvements with ApiKey skips and retry time

qingchm:
*Issue #, if available:*
N.A.

*Description of changes:*
Change the integration tests to add skip for ApiKey, also modifying retry time to be longer wait according to ApiGateway recommendation

Ran local testing

*Checklist:*

    - [ ] Correct values
- [x] `make pr` passes
- [ ] Update documentation

*Examples?*


integration/helpers/base_test.py
```diff
@@ -452,7 +452,7 @@ def verify_stack(self, end_state="CREATE_COMPLETE"):
 
     @retry(
         stop=stop_after_attempt(3),
```
hawflau, line 454: should we increase the number of attempts as well?

integration/helpers/base_test.py
```diff
@@ -452,7 +452,7 @@ def verify_stack(self, end_state="CREATE_COMPLETE"):
...
-        wait=wait_exponential(multiplier=1, min=4, max=10) + wait_random(0, 1),
+        wait=wait_exponential(multiplier=1, min=20, max=30) + wait_random(0, 1),
```
hawflau, line 455: I'm wondering if this config makes sense at all.
Considering we have max 3 attempts -> the sequence of wait times we will get is 1, 2, 4. With min set as 20, that means we will always wait for 20s + some random time between 0s and 1s. Then why do we exponential backoff at all? 

And in total we have 3 attempts in 60s, which is still less than the 2min suggested time from API GW?

hawflau (changes requested):

integration/helpers/base_test.py
qingchm: Sure we can do that as well to increase to 5

integration/helpers/base_test.py
qingchm: So the wait time is going to increase exponentially starting from 20s, but not exceeding 30, so the total wait time is around 90s, we can either increase the max wait time or the total retry attempts to meet the 2 minute mark

hawflau: I was looking at https://github.com/jd/tenacity/blob/main/tenacity/wait.py#L136 so that's why the sequence is "1, 2, 4", and then bounded by 20 and 30. 

But I think adding more attempts should do the work.

hawflau (approved):

moelasmar (approved):

PR #2497, opened 2022-09-11
feat: Add State Property to EventBridgeRule EventSource in Function

nallenscott:
*Issue #, if available:*

N/A

*Description of changes:*

Add state property to EventBridgeRule EventSource for AWS::Serverless::Function.

Added invalid input and transform tests.

*Checklist:*

    - [x] Correct values
- [x] `make pr` passes
- [ ] Update documentation

*Examples?*

### DIFF
--- a/samtranslator/model/eventsources/push.py
+++ b/samtranslator/model/eventsources/push.py
@@ -186,6 +186,8 @@ class CloudWatchEvent(PushEventSource):
         "Input": PropertyType(False, is_str()),
         "InputPath": PropertyType(False, is_str()),
         "Target": PropertyType(False, is_type(dict)),
+        "Enabled": PropertyType(False, is_type(bool)),
+        "State": PropertyType(False, is_str()),
     }
 
     @cw_timer(prefix=FUNCTION_EVETSOURCE_METRIC_PREFIX)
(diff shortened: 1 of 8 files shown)


samtranslator/model/eventsources/push.py
```diff
@@ -218,6 +220,15 @@ def to_cloudformation(self, **kwargs):
             )
             resources.extend(dlq_resources)
 
+        if self.State and self.Enabled is not None:
```
lucashuy, line 223: Quick question, is there a particular reason why we also have the option of providing `Enabled: boolean` instead of just passing in `State: string`?

lucashuy (commented):
Hi, thanks for opening this PR to add `State` to EventSource and EventBridgeRule! Just had question regarding one of the behaviours

samtranslator/model/eventsources/push.py
nallenscott: Consistency with other event sources: https://github.com/aws/serverless-application-model/blob/eda646d88a8c8499c5ac2630a253e2e62d1b62be/samtranslator/model/stepfunctions/events.py#L85-L86

lucashuy (approved):
Thanks for contributing, this looks good to me!

moelasmar (approved):

PR #2507, opened 2022-09-21
chore: Use pre-created SAR app during integration tests

mndeveci:
Instead of using a SAR App from another account, create one before tests begin and use that one.

### DIFF
--- a/integration/resources/templates/single/basic_application_sar_location_with_intrinsics.yaml
+++ b/integration/resources/templates/single/basic_application_sar_location_with_intrinsics.yaml
@@ -1,40 +1,39 @@
 Parameters:
-  SemanticVersion:
+  SarApplicationId:
     Type: String
-    Default: 1.0.2
 
 Mappings:
   SARApplication:
     us-east-1:
-      ApplicationId: arn:aws:serverlessrepo:us-east-1:077246666028:applications/hello-world-python3
+      SemanticVersion: 1.0.3
     us-east-2:
-      ApplicationId: arn:aws:serverlessrepo:us-east-1:077246666028:applications/hello-world-python3
+      SemanticVersion: 1.0.3
     us-west-1:
-      ApplicationId: arn:aws:serverlessrepo:us-east-1:077246666028:applications/hello-world-python3
+      SemanticVersion: 1.0.3
     us-west-2:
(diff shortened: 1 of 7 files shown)


integration/resources/templates/single/basic_application_sar_location_with_intrinsics.yaml
```diff
@@ -1,53 +1,52 @@
...
-  SemanticVersion:
+  SarApplicationId:
     Type: String
-    Default: 1.0.2
 
 Mappings:
   SARApplication:
     us-east-1:
-      ApplicationId: arn:aws:serverlessrepo:us-east-1:077246666028:applications/hello-world-python3
+      SemanticVersion: 1.0.3
```
moelasmar, line 8: is there a way to take rid of this configuration. We are adding the same value for all regions any way.

integration/single/test_basic_application.py
```diff
@@ -9,6 +11,10 @@ class TestBasicApplication(BaseTest):
     Basic AWS::Serverless::Application tests
     """
 
+    @pytest.fixture(autouse=True)
+    def companion_stack_outputs(self, get_serverless_application_repository_app):
+        self.sar_app_id = get_serverless_application_repository_app
```
moelasmar, line 16: are we calling the `get_serverless_application_repository_app` function here or what we are doing?

moelasmar (commented):
left some questions

integration/resources/templates/single/basic_application_sar_location_with_intrinsics.yaml
mndeveci: I know it doesn't make any sense, but this is just to confirm that SAR integration can work with `Fn::FindInMap` and `Fn::Ref`. The previous content was re-using same SAR App ARN for all the regions.

integration/single/test_basic_application.py
mndeveci: `get_serverless_application_repository_app` is a fixture and adding fixture as parameter will call that function and return its result. Its result is the ARN of the SAR App which then be used in the parameters of the template.

moelasmar: thanks Mehmet for the explanation

moelasmar (approved):

hnnasit (approved):
LGTM!

qingchm (approved):

PR #2517, opened 2022-09-29
chore: fix custom domain integration tests for feature toggle

hnnasit:
*Issue #, if available:*
N/A
*Description of changes:*
Custom domain resources are created in the same account for the internal integration tests to run
Tests were run locally with --internal flag
*Checklist:*

    - [ ] Correct values
- [x] `make pr` passes
- [ ] Update documentation

*Examples?*


mndeveci (approved):

integration/helpers/base_test.py
```diff
@@ -45,6 +45,13 @@
 
 STACK_NAME_PREFIX = "sam-integ-stack-"
 S3_BUCKET_PREFIX = "sam-integ-bucket-"
+FEATURE_TOGGLE_JSON_FILES = [
```
qingchm, line 48: Where is the file name of the non-FT json files stored?

hnnasit: The file names for non-FT json files are not stored anywhere. The idea is if the test is related to feature toggle then it would use non-FT + feature_toggle.json file

qingchm (approved):

Issue #2534, opened 2022-10-11
Connector with event source mapping and DLQ requires `DependsOn`


hoffa:
Consider the following template:

```yaml
Transform: AWS::Serverless-2016-10-31
Resources:
  MyTable:
    Type: AWS::DynamoDB::Table
    Properties:
      BillingMode: PAY_PER_REQUEST
      AttributeDefinitions:
        - AttributeName: id
          AttributeType: S
      KeySchema:
        - AttributeName: id
          KeyType: HASH
      StreamSpecification:
        StreamViewType: NEW_IMAGE

  MyRole:
    Type: AWS::IAM::Role
    Properties:
      AssumeRolePolicyDocument:
        Statement:
          - Effect: Allow
            Action: sts:AssumeRole
            Principal:
              Service: lambda.amazonaws.com
      ManagedPolicyArns:
        - arn:aws:iam::aws:policy/service-role/AWSLambdaBasicExecutionRole

  MyFunction:
    Type: AWS::Lambda::Function
    Properties:
      Runtime: nodejs16.x
      Handler: index.handler
      Role: !GetAtt MyRole.Arn
      Code:
        ZipFile: |
          exports.handler = async (event, context) => {
            console.log(JSON.stringify(event));
          };

  MyQueue:
    Type: AWS::SQS::Queue

  MyEventSourceMapping:
    Type: AWS::Lambda::EventSourceMapping
    Properties:
      EventSourceArn: !GetAtt MyTable.StreamArn
      FunctionName: !Ref MyFunction
      StartingPosition: TRIM_HORIZON
      DestinationConfig:
        OnFailure:
          Destination: !GetAtt MyQueue.Arn

  MyConnector:
    Type: AWS::Serverless::Connector
    Properties:
      Source: 
        Id: MyTable
      Destination: 
        Id: MyFunction
      Permissions:
        - Read

  LambdaToQueue:
    Type: AWS::Serverless::Connector
    Properties:
      Source: 
        Id: MyFunction
      Destination: 
        Id: MyQueue
      Permissions:
        - Write
```

Deploying it will fail with:

> MyEventSourceMapping
[...]

PR #2537, opened 2022-10-12
feat: replace `DependsOn` to connector by its generated logical IDs

hoffa:
### Issue #, if available

https://github.com/aws/serverless-application-model/issues/2534

### Description of changes

A connector logical ID in `DependsOn` will be automatically replaced by the logical IDs of its generated resources. This makes situations where `DependsOn` is necessary a little easier (e.g. event source mapping with DLQ), and it's consistent with how currently the abstract resources are referenceable.

For example, the following:

```yaml
DependsOn:
  - SomeConnector
```

Will be replaced by:

```yaml
DependsOn:
  - SomeConnectorWriteLambdaPermission 
```

### Checklist

    - [x] Correct values
- [ ] Update documentation

### Examples?


tests/translator/input/connector_bucket_to_function.yaml
```diff
@@ -12,7 +12,7 @@ Resources:
 
   Bucket:
     # See also https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-s3-bucket-notificationconfig.html
-    DependsOn: [MyConnectorWriteLambdaPermission]
+    DependsOn: MyConnector
```
GavinZZ, line 15: nit: This should still be [MyConnector] right?

GavinZZ (approved):
Nice!

tests/translator/input/connector_bucket_to_function.yaml
hoffa: Not necessarily, both are correct: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-dependson.html

samtranslator/utils/utils.py
```diff
@@ -0,0 +1,23 @@
...
+def insert_unique(xs: Any, vs: Any) -> List[Any]:
+    """
+    Return copy of `xs` extended with values of `vs` that do not exist in `xs`.
+
+    Inputs are converted to lists if they already aren't.
+    """
+    xs = as_array(copy.deepcopy(xs))
+    vs = as_array(copy.deepcopy(vs))
+
+    for v in vs:
```
aaythapa, line 19: Maybe I'm misreading the logic but can't you convert to sets and use the .union function

awood45 (approved):

samtranslator/utils/utils.py
hoffa: Could, but we want it to be always predictable and deterministic (ensures we don't cause changes/redeployments and tests keep working regardless of Python version, and we can determine order without knowing implementation details).

For example:

```python
>>> list(set(['a', 'wat', 'lol']))
['lol', 'wat', 'a']
```

The order changes due to implementation details (the hashing); we don't want to rely on it staying as-is (in fact during the Python 2 transition to Python 3, there was a lot of work needed to keep code that relied on the underlying hashing algorithm backwards compatible).

aaythapa: Ah I see, I thought .union had a guaranteed order. Thanks for the clarification

hawflau (approved):
LGTM. I originally thought this is better done via a plugin. But this seems more precise as each connector handles its own logical ID.

PR #2538, opened 2022-10-12
chore: pylint fix

GavinZZ:
### Description of changes
Enabling more pyling rules. Putting more strict requirement on the code.

Ran integ tests and all passed.
`2739 passed in 56.52s`

### DIFF
--- a/samtranslator/model/stepfunctions/generators.py
+++ b/samtranslator/model/stepfunctions/generators.py
@@ -1,17 +1,13 @@
 import json
-from uuid import uuid4
 from copy import deepcopy
 
-import samtranslator.model.eventsources.push
 from samtranslator.metrics.method_decorator import cw_timer
-from samtranslator.model import ResourceTypeResolver
 from samtranslator.model.exceptions import InvalidEventException, InvalidResourceException
 from samtranslator.model.iam import IAMRolePolicies
 from samtranslator.model.resource_policies import ResourcePolicies
 from samtranslator.model.role_utils import construct_role_for_resource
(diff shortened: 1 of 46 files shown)


samtranslator/plugins/__init__.py
```diff
@@ -56,7 +56,7 @@ def on_before_transform_resource(self, logical_id, resource_type, resource_prope
         # NoOp implementation
         pass
 
-    def on_before_transform_template(self, template):
+    def on_before_transform_template(self, template_dict):
```
hoffa, line 59: Why?

hoffa (approved):
Nice 👍

samtranslator/plugins/__init__.py
GavinZZ: This is changed due to pyling rule W0237. 
```
Used when a method parameter has a different name than in the implemented interface or in an overridden method. 
```
https://pylint.pycqa.org/en/latest/user_guide/messages/warning/arguments-renamed.html

samtranslator/model/stepfunctions/generators.py
```diff
@@ -262,7 +258,7 @@ def _generate_event_resources(self):
                     for name, resource in self.event_resources[logical_id].items():
                         kwargs[name] = resource
                 except (TypeError, AttributeError) as e:
-                    raise InvalidEventException(logical_id, "{}".format(e))
+                    raise InvalidEventException(logical_id, f"{e}")
```
aahung, line 261:
```suggestion
                    raise InvalidEventException(logical_id, str(e))
```

aahung (approved):

PR #2551, opened 2022-10-18
chore: Add transform test creation script

GavinZZ:
### Description of changes
Added a script to allow easier transform tests creation.

Verified by manual trigger. 

Tested with and without `--disable-api-configuration`, work as expected.
Tested with `arn:aws...`, work as expected.

Not affecting SAM T functionality.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,6 +28,8 @@ lint:
 	pylint --rcfile .pylintrc samtranslator
 	# mypy performs type check
 	mypy samtranslator
+	# Add --ignore-missing-imports flag to address Skipping analyzing "docopt": module is installed, but missing library stubs or py.typed marker
+	mypy bin/add_transform_test.py --ignore-missing-imports
 
 prepare-companion-stack:
 	pytest -v --no-cov integration/setup -m setup
--- a/Makefile
+++ b/Makefile
@@ -27,9 +27,7 @@ lint:
 	# Linter performs static analysis to catch latent bugs
 	pylint --rcfile .pylintrc samtranslator
 	# mypy performs type check
-	mypy samtranslator
-	# Add --ignore-missing-imports flag to address Skipping analyzing "docopt": module is installed, but missing library stubs or py.typed marker
-	mypy bin/add_transform_test.py --ignore-missing-imports
+	mypy samtranslator bin/add_transform_test.py
 
 prepare-companion-stack:
 	pytest -v --no-cov integration/setup -m setup
--- a/bin/add_transform_test.py
+++ b/bin/add_transform_test.py
@@ -60,7 +60,6 @@ def replace_arn_partitions(input_file_path: str) -> None:
         yaml.dump(replaced_template, f, default_flow_style=False)
 
 
-
 def read_json_file(file_path: str) -> Dict[str, Any]:
     with open(file_path, "r") as f:
         sam_template: Dict[str, Any] = json.load(f)
--- a/bin/add_transform_test.py
+++ b/bin/add_transform_test.py
@@ -17,7 +17,7 @@ import shutil
 import sys
 import yaml
 import tempfile
-from docopt import docopt
+from docopt import docopt  # type: ignore
 from pathlib import Path
 from typing import Any, Dict
 
--- a/bin/add_transform_test.py
+++ b/bin/add_transform_test.py
@@ -24,7 +24,7 @@ from typing import Any, Dict
 from samtranslator.yaml_helper import yaml_parse
 
 SCRIPT_DIR = os.path.dirname(os.path.realpath(__file__))
-TRANSFORM_TEST_DIR = os.path.dirname(os.path.realpath(__file__)) + "/../tests/translator/"
+TRANSFORM_TEST_DIR = SCRIPT_DIR + "/../tests/translator"
 CLI_OPTIONS = docopt(__doc__)
 
 
--- a/bin/add_transform_test.py
+++ b/bin/add_transform_test.py
@@ -18,17 +18,18 @@ import yaml
 import tempfile
 from docopt import docopt
 from pathlib import Path
-from typing import Any
+from typing import Any, Dict
 
 from samtranslator.yaml_helper import yaml_parse
 
-cwd = os.getcwd()
+SCRIPT_DIR = os.path.dirname(os.path.realpath(__file__))
+TRANSFORM_TEST_DIR = os.path.dirname(os.path.realpath(__file__)) + "/../tests/translator/"
 CLI_OPTIONS = docopt(__doc__)
 
 
 def get_input_file_path() -> str:
     input_file_option = CLI_OPTIONS.get("--template-file")
-    return os.path.join(cwd, input_file_option)
+    return os.path.join(os.getcwd(), input_file_option)
 
 
 def copy_input_file_to_transform_test_dir(input_file_path: str, transform_test_input_path: str) -> None:
@@ -38,21 +39,27 @@ def copy_input_file_to_transform_test_dir(input_file_path: str, transform_test_i
     print(f"Transform Test input file generated {transform_test_input_path}")
 
 
-def add_regional_endpoint_configuration_if_needed(temp_output_file_path: str) -> None:
-    with open(temp_output_file_path, "r") as f:
+def read_json_file(file_path: str) -> Dict[str, Any]:
+    with open(file_path, "r") as f:
         sam_template = json.load(f)
+    return sam_template
 
-    for logical_id, resource in sam_template["Resources"].items():
+
+def write_json_file(obj: Dict[str, Any], file_path: str) -> None:
+    with open(file_path, "w") as f:
+        json.dump(obj, f, indent=2)
+
+
+def add_regional_endpoint_configuration_if_needed(template: Dict[str, Any]) -> None:
+    for _, resource in template["Resources"].items():
         if resource["Type"] == "AWS::ApiGateway::RestApi":
-            if "EndpointConfiguration" not in resource["Properties"]:
-                sam_template["Resources"][logical_id]["Properties"]["EndpointConfiguration"] = {"Types": ["REGIONAL"]}
-            if "Parameters" not in resource["Properties"]:
-                sam_template["Resources"][logical_id]["Properties"]["Parameters"] = {
-                    "endpointConfigurationTypes": "REGIONAL"
-                }
+            properties = resource["Properties"]
+            if "EndpointConfiguration" not in properties:
+                properties["EndpointConfiguration"] = {"Types": ["REGIONAL"]}
+            if "Parameters" not in properties:
+                properties["Parameters"] = {"endpointConfigurationTypes": "REGIONAL"}
 
-    with open(temp_output_file_path, "w") as f:
-        json.dump(sam_template, f, indent=2)
+    return template
 
 
 def generate_transform_test_output_files(input_file_path: str, file_basename: str) -> None:
@@ -63,7 +70,7 @@ def generate_transform_test_output_files(input_file_path: str, file_basename: st
         subprocess.run(
             [
                 "python",
-                os.path.dirname(os.path.realpath(__file__)) + "/sam-translate.py",
+                SCRIPT_DIR + "/sam-translate.py",
                 "--template-file",
                 input_file_path,
                 "--output-template",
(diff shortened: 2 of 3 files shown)


bin/add_transform_test.py
```diff
@@ -0,0 +1,98 @@
...
+
+def get_input_file_path():
+    input_file_option = CLI_OPTIONS.get("--template-file")
+    return os.path.join(cwd, input_file_option)
+
+
+def copy_input_file_to_transform_test_dir(input_file_path, transform_test_input_path):
+    shutil.copyfile(input_file_path, transform_test_input_path)
+
+    replace_arn_partitions(transform_test_input_path)
```
hoffa, line 33: One thing I often encountered was with `AWS::ApiGateway::RestApi` needing (see any of the API transform tests e.g. `api_with_apikey_required`)

```json
        "EndpointConfiguration": {
          "Types": [
            "REGIONAL"
          ] 
        },
        "Parameters": {
          "endpointConfigurationTypes": "REGIONAL"
        }
```

Maybe we could add that by default?

bin/add_transform_test.py
```diff
@@ -0,0 +1,98 @@
...
+    # copy the output files into correct directories
+    transform_test_output_paths = [
+        cwd + path + output_file_option
+        for path in [
+            "/tests/translator/output/",
+            "/tests/translator/output/aws-cn/",
+            "/tests/translator/output/aws-us-gov/",
+        ]
+    ]
+    for output_path in transform_test_output_paths:
```
hoffa, line 52: Would be good at least print out that the file is copied, otherwise there's no output and it could be confusing.

bin/add_transform_test.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+def replace_arn_partitions(input_file_path):
+    with open(input_file_path, "r") as f:
+        sam_template = yaml_parse(f)
+
+    replaced_template = replace_arn_partitions_nested(sam_template, lambda v: replace_arn(v))
+
+    with open(input_file_path, "w") as f:
+        yaml.dump(replaced_template, f, default_flow_style=False)
```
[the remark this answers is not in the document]
GavinZZ: The original file is untouched. This only rewrites the copied file in the transform test input directory.
`shutil.copyfile(input_file_path, transform_test_input_path)`

bin/add_transform_test.py
```diff
@@ -0,0 +1,98 @@
...
+    shutil.copyfile(input_file_path, transform_test_input_path)
+
+    replace_arn_partitions(transform_test_input_path)
+
+
+def generate_transform_test_output_files(input_file_path, file_basename):
+    output_file_option = file_basename + ".json"
+
+    # run sam_translate.py and get the temporary output file
+    temp_output_file_path = os.path.join(cwd, output_file_option)
```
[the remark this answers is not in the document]
GavinZZ: Nice! That's good to know

bin/add_transform_test.py
GavinZZ: Good point!

bin/add_transform_test.py
```diff
@@ -0,0 +1,98 @@
...
+from samtranslator.yaml_helper import yaml_parse
+from sam_translate import transform_template
+
+cwd = os.getcwd()
+CLI_OPTIONS = docopt(__doc__)
+
+
+def get_input_file_path():
+    input_file_option = CLI_OPTIONS.get("--template-file")
+    return os.path.join(cwd, input_file_option)
```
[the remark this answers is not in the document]
hoffa: I don't mind keeping it, but I'd argue the consistency here provides little to no value.

bin/sam_translate.py
```diff
@@ -5,9 +5,9 @@
 Known limitations: cannot transform CodeUri pointing at local directory.
 
 Usage:
-  sam-translate.py --template-file=sam-template.yaml [--verbose] [--output-template=<o>]
-  sam-translate.py package --template-file=sam-template.yaml --s3-bucket=my-bucket [--verbose] [--output-template=<o>]
-  sam-translate.py deploy --template-file=sam-template.yaml --s3-bucket=my-bucket --capabilities=CAPABILITY_NAMED_IAM --stack-name=my-stack [--verbose] [--output-template=<o>]
+  sam_translate.py --template-file=sam-template.yaml [--verbose] [--output-template=<o>] [--disable-api-configuration]
```
GavinZZ, line 8: Had to add this parameter here as well, otherwise docopt will generate Usage warning.

hoffa: What that makes no sense 🤔

bin/add_transform_test.py
```diff
@@ -0,0 +1,133 @@
...
+        subprocess.run(
+            [
+                "python",
+                cwd + "/bin/sam-translate.py",
+                "--template-file",
+                input_file_path,
+                "--output-template",
+                temp_output_file.name,
+            ]
+        )
```
hoffa, line 72: So it fails if command fails:

```suggestion
            ],
            check=True,
        )
```

bin/add_transform_test.py
```diff
@@ -0,0 +1,137 @@
...
+
+def generate_transform_test_output_files(input_file_path: str, file_basename: str) -> None:
+    output_file_option = file_basename + ".json"
+
+    # run sam-translate.py and get the temporary output file
+    with tempfile.NamedTemporaryFile() as temp_output_file:
+        subprocess.run(
+            [
+                "python",
+                cwd + "/bin/sam-translate.py",
```
hoffa, line 66: This should be `<directory-that-script-lives-in> / sam-translate.py` as we're only guaranteed both scripts are in the same directory.

bin/add_transform_test.py
```diff
@@ -0,0 +1,138 @@
...
+        regional_transform_test_output_paths = [
+            cwd + path + output_file_option
+            for path in [
+                "/tests/translator/output/aws-cn/",
+                "/tests/translator/output/aws-us-gov/",
+            ]
+        ]
+
+        if not CLI_OPTIONS.get("--disable-api-configuration"):
+            add_regional_endpoint_configuration_if_needed(temp_output_file.name)
```
hoffa, line 88: This is only for `aws-cn` and `aws-us-gov`.

GavinZZ: Yes, this is.

hoffa: Oh I didn't see it was writing twice!

bin/add_transform_test.py
```diff
@@ -0,0 +1,138 @@
...
+                "--template-file",
+                input_file_path,
+                "--output-template",
+                temp_output_file.name,
+            ],
+            check=True,
+        )
+
+        # copy the output files into correct directories
+        transform_test_output_path = cwd + "/tests/translator/output/" + output_file_option
```
hoffa, line 76: `cwd` has same problem as L66 previously; if the script is not run from the repo root the path is wrong. So instead using `os.path.dirname(os.path.realpath(__file__)) / ".." / "tests" / ...` is better.

bin/add_transform_test.py
```diff
@@ -0,0 +1,138 @@
...
+    print(f"Transform Test input file generated {transform_test_input_path}")
+
+
+def add_regional_endpoint_configuration_if_needed(temp_output_file_path: str) -> None:
+    with open(temp_output_file_path, "r") as f:
+        sam_template = json.load(f)
+
+    for logical_id, resource in sam_template["Resources"].items():
+        if resource["Type"] == "AWS::ApiGateway::RestApi":
+            if "EndpointConfiguration" not in resource["Properties"]:
```
hoffa, line 47: nit: you can make this more legible by adding `properties = resource["Properties"]` and then directly setting `properties["EndpointConfiguration"]` etc.; they're all references so it'll mutate the original object.

bin/add_transform_test.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+def generate_transform_test_output_files(input_file_path: str, file_basename: str) -> None:
+    output_file_option = file_basename + ".json"
+
+    # run sam-translate.py and get the temporary output file
+    with tempfile.NamedTemporaryFile() as temp_output_file:
+        subprocess.run(
+            [
+                "python",
```
hoffa, line 65: I just found this: https://docs.python.org/3/library/sys.html#sys.executable

That should be more reliable as we know we have some interpreter already that's running this script.

bin/add_transform_test.py
```diff
@@ -0,0 +1,148 @@
...
+import yaml
+import tempfile
+from docopt import docopt
+from pathlib import Path
+from typing import Any, Dict
+
+from samtranslator.yaml_helper import yaml_parse
+
+SCRIPT_DIR = os.path.dirname(os.path.realpath(__file__))
+TRANSFORM_TEST_DIR = os.path.dirname(os.path.realpath(__file__)) + "/../tests/translator/"
```
hoffa, line 27:
```suggestion
TRANSFORM_TEST_DIR = SCRIPT_DIR + "/../tests/translator"
```

The trailing slash would've made the script path `//` which I'm not sure is valid.

hoffa (approved):
:shipit:

bin/add_transform_test.py
```diff
@@ -0,0 +1,148 @@
...
+
+    with open(input_file_path, "w") as f:
+        yaml.dump(replaced_template, f, default_flow_style=False)
+
+
+def main() -> None:
+    input_file_path = get_input_file_path()
+    file_basename = Path(input_file_path).stem
+
+    transform_test_input_path = TRANSFORM_TEST_DIR + "/input/" + file_basename + ".yaml"
```
aahung, line 141: how about windows dev who wants to create them

GavinZZ: Good catch. Updated to use `os.path.join`

aahung (approved):

bin/add_transform_test.py
```diff
@@ -0,0 +1,149 @@
...
+
+    replace_arn_partitions(transform_test_input_path)
+    print(f"Transform Test input file generated {transform_test_input_path}")
+
+
+def main() -> None:
+    input_file_path = get_input_file_path()
+    file_basename = Path(input_file_path).stem
+
+    transform_test_input_path = os.path.join(TRANSFORM_TEST_DIR, "input/", file_basename + ".yaml")
```
hoffa, line 142:
```suggestion
    transform_test_input_path = os.path.join(TRANSFORM_TEST_DIR, "input", file_basename + ".yaml")
```

bin/add_transform_test.py
```diff
@@ -0,0 +1,149 @@
...
+                "--template-file",
+                input_file_path,
+                "--output-template",
+                temp_output_file.name,
+            ],
+            check=True,
+        )
+
+        # copy the output files into correct directories
+        transform_test_output_path = os.path.join(TRANSFORM_TEST_DIR, "output/", output_file_option)
```
hoffa, line 105:
```suggestion
        transform_test_output_path = os.path.join(TRANSFORM_TEST_DIR, "output", output_file_option)
```

Makefile
```diff
@@ -28,6 +28,8 @@ lint:
 	pylint --rcfile .pylintrc samtranslator
 	# mypy performs type check
 	mypy samtranslator
+	# Add --ignore-missing-imports flag to address Skipping analyzing "docopt": module is installed, but missing library stubs or py.typed marker
+	mypy bin/add_transform_test.py --ignore-missing-imports
```
hoffa, line 32: Doesn't `# type: ignore` work?

xazhao:
A general question: If we use the script to create transform tests, we use translated templates to compare with translated templates which will for sure succeed. Are we supposed to write output templates on ourselves so we can make sure it's expected?

On the other hand, if it is to test if Sam-T functionality has changed, I think this method makes sense.

hoffa:
> A general question: If we use the script to create transform tests, we use translated templates to compare with translated templates which will for sure succeed. Are we supposed to write output templates on ourselves so we can make sure it's expected?
> 
> On the other hand, if it is to test if Sam-T functionality has changed, I think this method makes sense.

We should always verify the output is as _we_ expect. The tool only makes the step of generating it easier.

GavinZZ:
To answer Xia's question, I think this script just makes our live easier. We will still need to manually verify the output to make sure it's what we expect.

xazhao:
Oh makes sense to me then! Could we add this to development guide too?

GavinZZ:
Will add a message at the end to let users to check output manually.

bin/add_transform_test.py
```diff
@@ -0,0 +1,148 @@
...
+import yaml
+import tempfile
+from docopt import docopt  # type: ignore
+from pathlib import Path
+from typing import Any, Dict
+
+from samtranslator.yaml_helper import yaml_parse
+
+SCRIPT_DIR = os.path.dirname(os.path.realpath(__file__))
+TRANSFORM_TEST_DIR = os.path.join(SCRIPT_DIR, "../tests/translator")
```
hoffa, line 27:
```suggestion
TRANSFORM_TEST_DIR = os.path.join(SCRIPT_DIR, "..", "tests", "translator")
```

xazhao (approved):

PR #2253, opened 2021-11-30
fix: state machine input escaping

hoffa:
### Issue #, if available

#1895

### Description of changes

Unescapes singles quotes in state machine input escaped by `escapeJavaScript` (which isn't valid JSON).

The final mapping template isn't JSON by design, so can't use `json.dumps()`. For example:

```json
{"input": "$util.escapeJavaScript($input.json('$')).replaceAll("\\'","'")", "stateMachineArn": "arn:aws:states:us-east-1:792766875239:stateMachine:Post-n7ZKNljUJsZZ"}
```

### Description of how you validated changes

```bash
pytest integration --no-cov -k test_state_machine_with_api_single_quotes_input -s -vvv
make test
```

### Checklist

- [ ] Add/update tests using:
    - [ ] Correct values
    - [ ] Bad/wrong values (None, empty, wrong type, length, etc.)
    - [ ] [Intrinsic Functions](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference.html)
- [x] `make pr` passes
- [ ] Update documentation
- [ ] Verify transformed template deploys and application functions as expected

### Examples?

Please reach out in the comments, if you want to add an example. Examples will be 
added to `sam init` through https://github.com/awslabs/aws-sam-cli-app-templates/

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

Issue #1895, opened 2021-01-17
escapeJavaScript ecaping in request templates


russau:
https://github.com/aws/serverless-application-model/blob/16fa8522d596a153965d6801943249276b8d974b/samtranslator/model/stepfunctions/events.py#L437

Using `escapeJavaScript` here can cause problems with apostrophes.  More details in the forum posts:

https://forums.aws.amazon.com/thread.jspa?threadID=228917
https://forums.aws.amazon.com/thread.jspa?messageID=710397

hoffa:
Thanks for reporting this!

Do you have a template that would help reproducing any issue you've had?

russau:
Hello @hoffa, I have a template below that will reproduce what I'm seeing.

I see the mapping template generated looks like

``` json
{"input": "$util.escapeJavaScript($input.json('$'))", "stateMachineArn": "arn:aws:states:us-west-2:..."}
```

A post to my `echo` API without any single quotes works:

```
curl --location --request POST 'https://....execute-api.us-west-2.amazonaws.com/Prod/echo' \
--header 'Content-Type: application/json' \
--data-raw '{"hello": "world"}'
```

A post that contains quotes fails with `{"__type":"com.amazon.coral.service#SerializationException"}`.

```
curl --location --request POST 'https://....execute-api.us-west-2.amazonaws.com/Prod/echo' \
--header 'Content-Type: application/json' \
--data-raw '{"hello": "'\''world'\''"}'
```

The template..

``` yaml
AWSTemplateFormatVersion: '2010-09-09'
Transform: AWS::Serverless-2016-10-31
Description: Repro

Resources:
  MyApi:
    Type: AWS::Serverless::Api
    Properties:
      StageName: Prod

  HelloWorldFunction:
    Type: AWS::Serverless::Function
    Properties:
      InlineCode: |
        def handler(event, context):
            print(event)
            return "do nothing"
      Handler: index.handler
      Runtime: python3.8


  Post:
    Type: AWS::Serverless::StateMachine
    Properties:
      Role: null
      Policies:
        - arn:aws:iam::aws:policy/AWSLambdaFullAccess
      Definition:
        StartAt: One
        States:
          One:
            Type: Task
            Resource: !GetAtt HelloWorldFunction.Arn
            End: true
      Events:
        PostEcho:
          Type: Api
          Properties:
            RestApiId: !Ref MyApi
            Path: /echo
            Method: POST
```

hoffa:
[...]

PR #2591, opened 2022-11-15
feat: add `UnescapeMappingTemplate` to state machine `Api` event

hoffa:
### Issue #, if available

https://github.com/aws/serverless-application-model/issues/1895

### Description of changes

Adds `UnescapeMappingTemplate` boolean to [state machine `Api` event](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-property-statemachine-statemachineapi.html), which unescapes singles quotes in state machine input escaped by `escapeJavaScript` (which isn't valid JSON).

The final mapping template isn't JSON by design, so can't use `json.dumps()`. For example:

```json
{"input": "$util.escapeJavaScript($input.json('$')).replaceAll("\\'","'")", "stateMachineArn": "arn:aws:states:us-east-1:792766875239:stateMachine:Post-n7ZKNljUJsZZ"}
```

Originally done in https://github.com/aws/serverless-application-model/pull/2253, but using an opt-in property instead to avoid large-scale redeployments.

```bash
pytest integration --no-cov -k test_state_machine_with_api_single_quotes_input -s
pytest --no-cov samtranslator -n auto tests/translator/test_translator.py -k state_machine_with_api
```

### Checklist

    - [ ] Correct values
- [ ] Update documentation

### Examples?


GavinZZ (approved):

xazhao:
Do we have doc update for this?

integration/single/test_basic_api.py
```diff
@@ -113,3 +115,21 @@ def test_basic_api_with_tags(self):
         self.assertIsNotNone(stage)
         self.assertEqual(stage["tags"]["TagKey1"], "TagValue1")
         self.assertEqual(stage["tags"]["TagKey2"], "")
+
+    def test_state_machine_with_api_single_quotes_input(self):
```
xazhao, line 119: Another test without `UnescapeMappingTemplate` to show it's not desired?

integration/helpers/base_test.py
```diff
@@ -560,6 +560,17 @@ def verify_options_request(self, url, expected_status_code, headers=None):
...
         return response
 
+    def verify_post_request(self, url: str, body_obj, expected_status_code: int):
+        """Return response to POST request and verify matches expected status code."""
+        response = self.do_post_request(url, body_obj)
+        if response.status_code != expected_status_code:
+            raise StatusCodeError(
+                "Request to {} failed with status: {}, expected status: {}".format(
+                    url, response.status_code, expected_status_code
+                )
```
aahung, line 570: nit: we should switch to use f-string now like we did for new code. Fine for testing code

aahung (approved):

integration/single/test_basic_api.py
hoffa: It's always desired AFAICS, but have to do as opt-in property to avoid large-scale redeployments.

xazhao (approved):

PR #2603, opened 2022-11-17
fix: Raise correct exception when Api authorizer identity is not a dict

aahung:
### Checklist

    - [ ] Correct values
- [ ] Update documentation

### Examples?


samtranslator/model/apigateway.py
```diff
@@ -291,9 +299,11 @@ def _is_missing_identity_source(self, identity):  # type: ignore[no-untyped-def]
 
         required_properties_missing = not headers and not query_strings and not stage_variables and not context
 
+        if ttl is None:
+            return required_properties_missing
         try:
             ttl_int = int(ttl)
-        # this will catch if ttl is None and not convertable to an int
+        # this will catch if and not convertable to an int
```
aahung, line 306: As the type can be derived by `mypy`, it contains about `int(ttl)`'s `ttl` could be `None`.

GavinZZ (approved):

hoffa (approved):

ssenchenko (approved):

PR #2604, opened 2022-11-17
fix: KeyError: u'AWS_IAM' and add typing

GavinZZ:
### Description of changes
Fixing Availability Dip `KeyError: u'AWS_IAM'`.

Unit and integration tests passed locally.

### Checklist

    - [ ] Correct values
- [ ] Update documentation

### Examples?


samtranslator/model/api/api_generator.py
```diff
@@ -1172,7 +1175,6 @@ def _set_default_authorizer(  # type: ignore[no-untyped-def]
                 default_authorizer,
                 authorizers=authorizers,
                 add_default_auth_to_preflight=add_default_auth_to_preflight,
-                api_authorizers=api_authorizers,
```
GavinZZ, line 1175: No need to pass `api_authorizers`. `authorizers` already contains everything from `api_authorizers`. 
https://github.com/aws/serverless-application-model/blob/develop/samtranslator/model/api/api_generator.py#L723

samtranslator/model/api/api_generator.py
```diff
@@ -724,12 +724,11 @@ def _add_auth(self):  # type: ignore[no-untyped-def]
 
         if authorizers:
             swagger_editor.add_authorizers_security_definitions(authorizers)  # type: ignore[no-untyped-call]
-            self._set_default_authorizer(  # type: ignore[no-untyped-call]
+            self._set_default_authorizer(
                 swagger_editor,
                 authorizers,
                 auth_properties.DefaultAuthorizer,
                 auth_properties.AddDefaultAuthorizerToCorsPreflight,
-                auth_properties.Authorizers,
```
aaythapa, line 732: Why do we take out `auth_properties.Authorizers` here? Don't we need it for the method

ssenchenko (commented):
Somehow we missed this case in our tests, which ld to a runtime error.
Let's add some tests, which can reproduce the situation when the key was missing and make sure it's not gonna happen again. Even unit test with mocking dependencies would be better then nothing

hoffa (approved):
Any tests to ensure the issue doesn't get introduced again later?

samtranslator/model/api/api_generator.py
GavinZZ: I have removed all use case of `auth_properties.Authorizers` inside the function. This is actually a duplicate of `authorizers` as `authorizers` build from `auth_properties.Authorizers`

aaythapa (approved):

aahung (approved):

tests/model/api/test_http_api_generator.py
```diff
@@ -95,6 +95,48 @@ def test_auth_iam_enabled(self):
             },
         )
 
+    def test_auth_iam_enabled_with_default_iam_authorizer(self):
```
hoffa, line 98: Why not a regular transform test? These are hard to read.

ssenchenko: Aren't these tests more stable (not flakey)?

GavinZZ: Transform tests aren't flaky either. Only some of the integ tests are flaky.

ssenchenko (commented):
How come these are easier to read? A lot of copy/pasting, if we change anything, need to change in 4 files?

hoffa:
> How come these are easier to read? A lot of copy/pasting, if we change anything, need to change in 4 files?

Transforms should nearly always be deterministic, so changes to transform outputs.

ssenchenko:
> > How come these are easier to read? A lot of copy/pasting, if we change anything, need to change in 4 files?
> 
> Transforms should nearly always be deterministic, so no changes to transform outputs.
> 
> Edit: neither is the epitome of readability to be fair. There's a few issues with unit tests for transforms, since it only tests small sections of the code. e.g. it could be possible to have a plugin that runs after the transform that breaks the transform, but the unit test would still pass.

I believe it's possible, but shouldn't it be tested in that plugin then? I'm just worried that we keep adding tests which are flakey...

hoffa:
> > > How come these are easier to read? A lot of copy/pasting, if we change anything, need to change in 4 files?
> > 
> > 
> > Transforms should nearly always be deterministic, so no changes to transform outputs.
> > Edit: neither is the epitome of readability to be fair. There's a few issues with unit tests for transforms, since it only tests small sections of the code. e.g. it could be possible to have a plugin that runs after the transform that breaks the transform, but the unit test would still pass.
> 
> I believe it's possible, but shouldn't it be tested in that plugin then? I'm just worried that we keep adding tests which are flakey...

These are not integration tests, they're never flaky. They run the full transform code locally and compare the outputs.

ssenchenko:
> gration tests, they're never flaky. They run the full transform code locally and

gotcha, my bad :)

GavinZZ:
Transform tests are not flaky. It's simply comparing the output against the expected json files. 

I do agree with Chris that unit tests are highly couple to implementation details, so if a small implementation detail changes we need to spend effort fixing them.

ssenchenko (approved):

xazhao (approved):

PR #2605, opened 2022-11-17
adding support for all SSM Parameters types

aaythapa:
### Issue #, if available

https://github.com/aws/serverless-application-model/issues/2301

### Description of changes

SAM currently doesn't support all the parameter types defined [here](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/parameters-section-structure.html#aws-ssm-parameter-types). We added support for the SSM parameter String type in [ this PR](https://github.com/aws/serverless-application-model/pull/2469)\, adding support for the rest of the types.

- added new entry on parameters.json > Types > enum
- added one input to test all correct types pass
- added to error message if Parameter is not included in parameter.json

### Checklist

    - [x] Correct values
- [ ] Update documentation

### Examples?


ssenchenko (commented):
Overall approving.

Just an idea. It feels weird that we have to maintain that list of properties manually. What if there appear a new value? Should we consider generating this schema? (Not in this PR) Can we get the list of those enum values from any other place or service?

aaythapa:
> Overall approving.
> 
> Just an idea. It feels weird that we have to maintain that list of properties manually. What if there appear a new value? Should we consider generating this schema? (Not in this PR) Can we get the list of those enum values from any other place or service?

Had the same thought and I agree. I'm not sure if there is a way to get the list of enums from another place but if there is then we should do that. Manually maintaining parameters is what caused the original issue.

hoffa (approved):

ssenchenko (approved):

PR #2606, opened 2022-11-17
Create SAM StateMachine without providing a role or policy

ConnorRobertson:
### Description of changes
Allows a SAM StateMachine to be created without a policy or role by generating a role when no role or policy is given.

Added a test that uses Connectors without giving the SAM StateMachine a policy and changed the no policy given to test to see if a Role exists.

### Checklist

    - [x] Correct values
- [ ] Update documentation

### Examples?


samtranslator/model/stepfunctions/generators.py
```diff
@@ -129,20 +129,28 @@ def to_cloudformation(self):  # type: ignore[no-untyped-def]
...
+                        "Version": "2012-10-17",
+                        "Statement": [
+                            {
+                                "Action": ["Lambda:InvokeFunction"],
+                                "Effect": "Allow",
+                                "Resource": "arn:aws:lambda:us-west-2:534568764167:function:SomethingDoesNotMatter*",
+                            }
+                        ],
+                    }
+                ]
```
GavinZZ, line 150: Why are we adding an arbitrary policy?

tests/model/stepfunctions/test_state_machine_generator.py
```diff
@@ -56,12 +56,8 @@ def test_state_machine_no_role_or_policies(self):
         self.kwargs["definition_uri"] = "s3://my-demo-bucket/my_asl_file.asl.json"
         self.kwargs["role"] = None
         self.kwargs["policies"] = None
-        with self.assertRaises(InvalidResourceException) as error:
-            StateMachineGenerator(**self.kwargs).to_cloudformation()
-        self.assertEqual(
-            error.exception.message,
-            "Resource with id [StateMachineId] is invalid. Either 'Role' or 'Policies' property must be specified.",
-        )
+        generated_event_resources = StateMachineGenerator(**self.kwargs).to_cloudformation()
```
GavinZZ, line 59: nit: `generated resources`

samtranslator/model/stepfunctions/generators.py
```diff
@@ -129,20 +129,28 @@ def to_cloudformation(self):  # type: ignore[no-untyped-def]
...
-
+            if not self.policies:
+                self.policies = [
+                    {
+                        "Version": "2012-10-17",
+                        "Statement": [
+                            {
+                                "Action": ["Lambda:InvokeFunction"],
+                                "Effect": "Allow",
+                                "Resource": "arn:aws:lambda:us-west-2:534568764167:function:SomethingDoesNotMatter*",
```
hoffa, line 146: 🤨

samtranslator/model/stepfunctions/generators.py
ConnorRobertson: The connector doesn't work if you just leave it empty.

integration/resources/expected/combination/connector_sfn_to_function_without_policy.json
```diff
@@ -0,0 +1,22 @@
...
+    },
+    {
+      "LogicalResourceId": "MyFunction",
+      "ResourceType": "AWS::Lambda::Function"
+    },
+    {
+      "LogicalResourceId": "MyConnectorPolicy",
+      "ResourceType": "AWS::IAM::ManagedPolicy"
+    }
+  ]
```
aahung, line 22: you can run `make pr` to catch some formatting issues

samtranslator/model/stepfunctions/generators.py
aahung: 🤔

tests/model/stepfunctions/test_state_machine_generator.py
```diff
@@ -73,7 +69,7 @@ def test_state_machine_both_role_and_policies(self):
             StateMachineGenerator(**self.kwargs).to_cloudformation()
         self.assertEqual(
             error.exception.message,
-            "Resource with id [StateMachineId] is invalid. Specify either 'Role' or 'Policies' property and not both.",
+            "Resource with id [StateMachineId] is invalid. Specify 'Role' or 'Policies' or neither property.",
```
ssenchenko, line 72: I'd suggest either:
- assert only error class if it's enough
- make an error message a constant where you raise it ie `SFN_INVALID_PROPERTY_BLAH_BLAH = "Our error message"` and import it here
Otherwise, it's additional friction to maintain unit tests up-to-date without no value in this case

samtranslator/model/stepfunctions/generators.py
```diff
@@ -129,20 +129,28 @@ def to_cloudformation(self):  # type: ignore[no-untyped-def]
             )
 
         if self.role and self.policies:
-            raise InvalidResourceException(
-                self.logical_id, "Specify either 'Role' or 'Policies' property and not both."
-            )
+            raise InvalidResourceException(self.logical_id, "Specify 'Role' or 'Policies' or neither property.")
```
ssenchenko, line 132: So does it support now both `Role` and `Policy` specified at the same time? I can't see a code change which guarantees it. However the error message is not that clear anymore, that both at the same moment are not allowed.

ssenchenko (changes requested):

samtranslator/model/stepfunctions/generators.py
ConnorRobertson: It does not support both Role and Policy at the same time. I am not sure how to word that error message. Saying "Specify 'Role' or 'Policies' or neither property, and not both." Seems wordy.

ssenchenko: How about "Specify either 'Role' or 'Policies' (but not both at the same time) or neither of them"?

aahung (commented):
generally look good to me! Thanks! Can we add a transform test so that we can take a look on what is generated?

integration/resources/templates/combination/connector_sfn_to_function_without_policy.yaml
```diff
@@ -0,0 +1,36 @@
+Resources:
```
hoffa, line 1: Can you add this as a transform test as well?

Should be something like:

```
python bin/add_transform_test.py --template-file integration/resources/templates/combination/connector_sfn_to_function_without_policy.yaml
```

GavinZZ (approved):
LGTM. Thanks for addressing the comments!

samtranslator/model/stepfunctions/generators.py
```diff
@@ -129,20 +132,17 @@ def to_cloudformation(self):  # type: ignore[no-untyped-def]
...
         if self.role:
             self.state_machine.RoleArn = self.role
-        elif self.policies:
-            if not self.managed_policy_map:
+        else:
+            if self.policies and not self.managed_policy_map:
                 raise Exception("Managed policy map is empty, but should not be.")
-
+            if not self.policies:
+                self.policies = []
```
ssenchenko, line 142: Is it it? Where do we actually create a role or a policy?

aahung (approved):

samtranslator/model/stepfunctions/generators.py
aahung: in `_construct_role()` in generator.py

ConnorRobertson: Yeah, in _constructor_role() a role is generator based on a policy.

ssenchenko (approved):

PR #2615, opened 2022-11-18
fix: Handle when openapi/swagger security property is null

aahung:
### Issue #, if available

For template like this, it won't fail:
```
         /api/payment-gateway/servipag/payment-confirmation:
            post:
              responses: {}
              security: null
```

According to the swagger doc, template should uses `security: []` https://swagger.io/docs/specification/authentication/
I haven't tested whether `security: null` can be deployed or not. This PR make SAM-T to hand over the validation to CFN so in case swagger allows security to be `null` in the future, we don't need to do anything.

### Checklist

    - [ ] Correct values
- [ ] Update documentation

### Examples?


samtranslator/open_api/open_api.py
```diff
@@ -461,7 +461,9 @@ def _set_method_authorizer(self, path, method_name, authorizer_name, authorizers
             authorization_scopes = []
 
         for method_definition in self.iter_on_method_definitions_for_path_at_method(path, method_name):  # type: ignore[no-untyped-call]
-            existing_security = method_definition.get("security", [])
+            existing_security = method_definition.get("security")
+            if not existing_security:
+                existing_security = []
```
GavinZZ, line 466: why? Aren't they doing the same thing?

aahung: see the description above, if `security: null`, it won't return the default value. The default value is for the key doesn't exist

samtranslator/open_api/open_api.py
```diff
@@ -413,7 +413,7 @@ def set_path_default_authorizer(
                         # If no integration given, then we don't need to process this definition (could be AWS::NoValue)
                         if not self.method_definition_has_integration(method_definition):  # type: ignore[no-untyped-call]
                             continue
-                        existing_security = method_definition.get("security", [])
+                        existing_security = method_definition.get("security")
```
ssenchenko, line 416: Did you remove `method_definition.get("security", [])` just because `[]` is excessive?

samtranslator/open_api/open_api.py
ssenchenko: maybe add a comment. it's not clear if one doesn't know/remember that `security: null` can be an actual pair

ssenchenko: And also, should we add a test for that?

GavinZZ (approved):

samtranslator/open_api/open_api.py
aahung: yep, as it only check if it is falsy

samtranslator/open_api/open_api.py
aahung: I changed the implementation and it now raises an exception

aahung: one test added

GavinZZ (approved):

ssenchenko (approved):

PR #2621, opened 2022-11-18
chore: Experiment on unitify property value validation messages

aahung:
### Issue #, if available

After yesterday's various fixes, it was a pain that each validation has its own error message and they are not consistent, e.g.:
- XXX must be XXX
- Invalid type of 
- should be a **map** vs should be a **dictionary**
- Unexpected type "Py27UniStr" **completely expose our its implementation, very confusing**

This PR is a probe to unify the error messaging and provide customers a consistent way to raise invalid template errors. It starts with types but we will extend it to other validations. 

### Checklist

    - [ ] Correct values
- [ ] Update documentation

### Examples?


ssenchenko (approved):
Looks like a good idea!

GavinZZ (approved):

Issue #1133, opened 2019-09-08
SAM JSON schema


jlhood:
**Description:**

Today, SAM's spec is a document meant for human consumption only. SAM should also have an authoritative, machine-readable spec (schema). This would be beneficial in multiple ways:

1. This would allow cfn-lint to remove its dependency on the SAM translator code, which can cause dependency conflicts with other modules if they depend on a different version of SAM translator. cfn-lint could keep a copy the SAM schema similar to what it does for [CloudFormation schemas](https://github.com/aws-cloudformation/cfn-python-lint/tree/master/src/cfnlint/data/CloudSpecs). Then there would no longer be a dependency conflict.
    1. Somewhat tangential: I'd prefer for SAM CLI's `sam validate` command to just run cfn-lint against the template. I think this would be far more valuable than what the current `sam validate` command offers. Removing cfn-lint's dependency on SAM translator would unblock this.
1. If we could embed the documentation of the current spec into the schema, this could lead to the human-readable version of the spec being auto-generated. This would allow us to auto-generate the SAM spec in the official [SAM documentation](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/what-is-sam.html) after each SAM release.

kddejong:
From a cfn lint perspective I like this idea. sam  translator is currently essential to us being able to lint serverless templates. Having a shema could remove us from that requirement and allow us to directly lint the original file.

nija-at:
+1 to this. 

[AWS CDK](https://github.com/aws/aws-cdk) is currently using [the spec for awslabs/goformation](https://github.com/awslabs/goformation/blob/master/generate/sam-2016-10-31.json) to auto-generate CDK L1 resources (i.e. language bindings) and we're seeing a couple of cases where the handcrafted spec deviates from SAM's spec, specifically in terms of required and optional attributes, that blocks our customers from doing the right thing. 

Referencing https://github.com/aws/aws-cdk/issues/3841 and https://github.com/awslabs/goformation/issues/87.

stilvoid:
Another +1. I'd like to add SAM resource types to https://github.com/aws-cloudformation/rain :)

tj:
There's https://github.com/aws/serverless-application-model/blob/develop/samtranslator/validator/sam_schema/schema.json , but it doesn't seem up-to-date, it's missing HttpApi.

jfuss:
[...]

PR #2623, opened 2022-11-19
Add Schema Type to AWS::Serverless::Lambda

GavinZZ:
### Issue #, if available

#1133 

### Checklist

    - [ ] Correct values
- [ ] Update documentation

### Examples?

### DIFF
--- a/samtranslator/schema/schema.py
+++ b/samtranslator/schema/schema.py
@@ -79,7 +79,7 @@ class FunctionDeadLetterQueue(BaseModel):
     Type: Literal["SNS", "SQS"]
 
 
-class FunctionEventInvokeConfiguration(BaseModel):
+class FunctionEventInvokeConfig(BaseModel):
     DestinationConfig: Optional[PassThrough]
     MaximumEventAgeInSeconds: Optional[int]
     MaximumRetryAttempts: Optional[int]
@@ -107,7 +107,7 @@ class FunctionProperties(BaseModel):
     Description: Optional[PassThrough]
     Environment: Optional[PassThrough]
     EphemeralStorage: Optional[PassThrough]
-    EventInvokeConfig: Optional[FunctionEventInvokeConfiguration]
+    EventInvokeConfig: Optional[FunctionEventInvokeConfig]
     Events: Optional[Dict[str, FunctionEvent]]
     FileSystemConfigs: Optional[PassThrough]
     FunctionName: Optional[PassThrough]
--- a/samtranslator/schema/schema.py
+++ b/samtranslator/schema/schema.py
@@ -3,7 +3,8 @@ from __future__ import annotations
 from typing_extensions import Literal
 from typing import Any, Dict, List, Optional, Union
 
-from pydantic import Extra, constr
+import pydantic
+from pydantic import Extra
 
 # TODO: Get rid of this in favor of proper types
 Unknown = Optional[Any]
--- a/samtranslator/schema/schema.py
+++ b/samtranslator/schema/schema.py
@@ -3,9 +3,7 @@ from __future__ import annotations
 from typing_extensions import Literal
 from typing import Any, Dict, List, Optional, Union
 
-import pydantic
-from pydantic import Extra
-
+from pydantic import Extra, constr
 
 # TODO: Get rid of this in favor of proper types
 Unknown = Optional[Any]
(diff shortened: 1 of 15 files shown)


tests/translator/input/api_with_method_aws_iam_auth.yaml
```diff
@@ -45,21 +45,21 @@ Resources:
...
+              InvokeRole: !Sub arn:${AWS::Partition}:iam::123:role/AUTH_AWS_IAM
+        MyApiWithAwsIamAuthAndDefaultInvokeRoleAnyMethod:
+          Type: Api
+          Properties:
+            RestApiId: !Ref MyApiWithoutAuth
+            Path: /any/three
+            Method: any
+            Auth:
+              Authorizer: AWS_IAM
+              InvokeRole: CALLER_CREDENTIALS
```
aahung, line 65: unbelievable!

tests/translator/input/eventbridgerule_with_retry_policy.yaml
```diff
@@ -10,8 +10,8 @@ Resources:
           Type: Schedule
           Properties:
             Schedule: rate(1 minute)
-          RetryPolicy:
-            MaximumRetryAttempts: 3
+            RetryPolicy:
+              MaximumRetryAttempts: 3
```
aahung, line 14: :o

tests/translator/input/function_with_deployment_preference_from_parameters.yaml
```diff
@@ -40,6 +40,7 @@ Resources:
       CodeUri: s3://my-bucket/mySanityTestFunction.zip
       DeploymentPreference:
         Enabled: !Ref MyLowerFalseParameter
+        Type: Linear10PercentEvery1Minute
```
aahung, line 44: what are these changes?

tests/translator/input/function_with_deployment_preference_all_parameters.yaml
```diff
@@ -38,6 +39,7 @@ Resources:
       CodeUri: s3://my-bucket/myValidationTestFunction.zip
       DeploymentPreference:
         Enabled: false
+        Type: Linear10PercentEvery1Minute
```
hoffa, line 42: Why are you changing these? We should be first of all testing against the status quo, with no changes.

tests/translator/input/function_with_deployment_preference_from_parameters.yaml
GavinZZ: It's a required field, but it's missing. Json schema won't let it pass.

samtranslator/schema/schema.py
```diff
@@ -45,42 +45,84 @@ class AwsServerlessConnector(BaseModel):
...
+    Hooks: Optional[FunctionDeploymentPreferenceHooks]
+    PassthroughCondition: Optional[Union[bool, SamIntrinsic]]
+    Role: Optional[Union[str, SamIntrinsic]]
+    TriggerConfigurations: Optional[PassThrough]
+    Type: Union[str, SamIntrinsic]
+
+
+class FunctionDeadLetterQueue(BaseModel):
+    TargetArn: str
+    Type: Literal["SNS", "SQS"]
```
hoffa, line 71: Let's separate these so the `Type` is coupled to whatever properties it has. Otherwise makes it a pain to expand later.

tests/translator/input/function_with_deployment_preference_all_parameters.yaml
GavinZZ: It's a required field, but missing from the template. This results json schema check failure. 
https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-property-function-deploymentpreference.html#sam-function-deploymentpreference-type

samtranslator/schema/schema.py
```diff
@@ -45,42 +45,84 @@ class AwsServerlessConnector(BaseModel):
...
+
+
+class FunctionEventInvokeConfiguration(BaseModel):
+    DestinationConfig: Optional[PassThrough]
+    MaximumEventAgeInSeconds: Optional[int]
+    MaximumRetryAttempts: Optional[int]
+
+
+class FunctionEvent(BaseModel):
+    Properties: Any
```
hoffa, line 81: Missing types. Also same as comment above; should couple both `Properties` and `Type` so that each model only has a single `Type`.

tests/translator/input/function_with_deployment_preference_all_parameters.yaml
hoffa: Let's skip it. Let's not change any templates currently. There is no formal schema at the moment. First step is to document what our tests consider as valid, as that is the effective source of truth (since that's what deal with the logic). We should document the discrepancies and later figure out whether docs need to change, or tests need to change.

samtranslator/schema/schema.py
hoffa: Ignore, this one is fine. Still applies to `Events`.

Makefile
```diff
@@ -34,8 +34,6 @@ lint:
 	mypy --strict samtranslator bin
 	# Linter performs static analysis to catch latent bugs
 	pylint --rcfile .pylintrc samtranslator
-	# Ensure templates adhere to JSON schema
-	bin/validate.sh
```
hoffa, line 38: Should not be removed. We need to ensure all templates going forward adhere to the schema.

samtranslator/schema/schema.py
GavinZZ: Will do this in a separate PR. It's a huge change.

tests/translator/output/aws-us-gov/api_with_method_aws_iam_auth.json
```diff
@@ -100,9 +140,9 @@
       },
       "Type": "AWS::ApiGateway::RestApi"
     },
-    "MyApiWithoutAuthDeployment60e94c7d4c": {
+    "MyApiWithoutAuthDeployment54f5d55d46": {
```
hoffa, line 143: Something's off; this shouldn't change or it'll cause redeployments.

tests/translator/input/api_with_method_aws_iam_auth.yaml
```diff
@@ -27,7 +27,7 @@ Resources:
             Method: post
             Auth:
               Authorizer: AWS_IAM
-              InvokeRole: rn:aws:iam::123:role/AUTH_AWS_IAM
+              InvokeRole: !Sub arn:${AWS::Partition}:iam::123:role/AUTH_AWS_IAM
```
hoffa, line 30: Why this change?

bin/validate.sh
```diff
@@ -3,6 +3,6 @@ set -euxo pipefail
 
 # TODO: Enable the valid templates
 # TODO: Switch to way-faster Python script
-find tests/translator/input -type f -name '*.yaml' | grep -v error_ | grep -v unsupported_resources | grep -v resource_with_invalid_type | shuf | while read -r template; do
+find tests/translator/input -type f -name '*.yaml' | grep -v error_ | grep -v null_events | grep -v unsupported_resources | grep -v resource_with_invalid_type |  while read -r template; do
```
hoffa, line 6: Was `null_events` supposed to skip two tests? If yes, clearer:

```suggestion
find tests/translator/input -type f -name '*.yaml' | grep -v error_ | grep -v function_with_null_events | grep -v state_machine_with_null_events | grep -v unsupported_resources | grep -v resource_with_invalid_type |  while read -r template; do
```

GavinZZ: Yes, it's supposed to skip two tests.

tests/translator/input/api_with_method_aws_iam_auth.yaml
GavinZZ: The arn was incorrect `rn:aws:iam...`.

hoffa: Okay, so I'm assuming it transformed correctly. Transform tests should ideally never change, so we can skip making these changes as well (maybe make a note for later cleanup) for easier review (e.g. I'm not sure whether the logical ID change is due to this or the indentation change).

samtranslator/schema/schema.py
```diff
@@ -52,42 +51,86 @@ class AwsServerlessConnector(BaseModel):
...
+    AutoPublishAlias: Optional[Union[str, SamIntrinsic]]
+    AutoPublishCodeSha256: Optional[Union[str, SamIntrinsic]]
+    CodeSigningConfigArn: Optional[Union[str, SamIntrinsic]]
+    CodeUri: Optional[Union[str, FunctionCodeUri]]
+    DeadLetterQueue: Optional[Union[SamIntrinsic, FunctionDeadLetterQueue]]
+    DeploymentPreference: Optional[FunctionDeploymentPreference]
+    Description: Optional[PassThrough]
+    Environment: Optional[PassThrough]
+    EphemeralStorage: Optional[PassThrough]
+    EventInvokeConfig: Optional[FunctionEventInvokeConfiguration]
```
hoffa, line 110: nit: we'll have a lot of names, so keeping it consistent throughout will make it easier to reason (i.e. use `Config` if that's what the property uses)

tests/translator/input/api_with_method_aws_iam_auth.yaml
GavinZZ: The transform is correct, and the id change is due to indentation change I believe. It's when I make the indentation change, I found the arn seems off. And there are two occurrence of wrong arn format.

hoffa (approved):

aahung (approved):

PR #2635, opened 2022-11-21
Add detailed type to AWS::Serverless::Lambda Schema

GavinZZ:
### Issue #, if available
https://github.com/aws/serverless-application-model/issues/1133

### Checklist

    - [ ] Correct values
- [ ] Update documentation

### Examples?


acristin (approved):

samtranslator/schema/schema.py
```diff
@@ -176,65 +223,61 @@ class FunctionSqsEvent(BaseModel):
...
 
-class FunctionApiProperties(BaseModel):
-    Auth: Unknown
-    Method: Unknown
-    Path: Unknown
-    RequestModel: Unknown
-    RequestParameters: Unknown
-    RestApiId: Unknown
+class FunctionResourcePolicy(BaseModel):
+    AwsAccountBlacklist: Optional[List[ResourcePolicy]]
```
hoffa, line 227: Are you sure this is correct for all? It passed with `Optional[List[str]]` for the state machine.

samtranslator/schema/schema.py
```diff
@@ -12,6 +12,8 @@
 # Value passed directly to CloudFormation; not used by SAM
 PassThrough = Any  # TODO: Make it behave like typescript's unknown
 
+ResourcePolicy = Union[str, Dict[str, Any]]
```
hoffa, line 15: nit: `ResourcePolicy` sounds more abstract than it is (a dict or string) and can't really be reused

samtranslator/schema/schema.py
GavinZZ: Screen Shot 2022-11-21 at 2 59 32 PM

hoffa: That's for `CustomStatements`, which by it's name I assume stands for policy statement, so a dict makes sense. Is that the case for all others?

GavinZZ: I'll change the statemachine one to match function. I checked document for SFN, it's also a dict or string.

samtranslator/schema/schema.py
hoffa: Actually I don't like this name, since now all the ones in `ResourcePolicy` have this type, but it's not obvious that they all should share the same type (whatever it is).

samtranslator/schema/schema.py
```diff
@@ -51,6 +51,65 @@ class AwsServerlessConnector(BaseModel):
...
+    StartDate: Optional[PassThrough]
+    State: Optional[PassThrough]
+
+
+class EventsScheduleV2(BaseModel):
+    Type: Literal["ScheduleV2"]
+    Properties: EventsScheduleV2Properties
+
+
+class ResourcePolicy(BaseModel):
```
hoffa, line 99: nit: there's always a little risk of sharing models unless we're sure they share it by design (and do not just currently _happen_ to have the same properties), as new properties will be added to both and tests will pass even if it was meant for only one of them

hoffa (approved):

PR #2643, opened 2022-11-22
Validate schema as unit test

GavinZZ:
### Issue #, if available

#1133 

### Checklist

    - [ ] Correct values
- [ ] Update documentation

### Examples?


ssenchenko (approved):

Makefile
```diff
@@ -38,8 +38,6 @@ lint:
 	mypy --strict samtranslator bin
 	# Linter performs static analysis to catch latent bugs
 	pylint --rcfile .pylintrc samtranslator
-	# Ensure templates adhere to JSON schema
-	bin/validate_schema.py
```
hoffa, line 42: The file should be deleted as well.

tests/schema/test_validate_schema.py
```diff
@@ -0,0 +1,86 @@
...
+        list(Path("tests/translator/input").glob("**/*.yaml"))
+        + list(Path("tests/translator/input").glob("**/*.yml"))
+        + list(Path("tests/validator/input").glob("**/*.yaml"))
+        + list(Path("tests/validator/input").glob("**/*.yml"))
+        + list(Path("integration/resources/templates").glob("**/*.yaml"))
+        + list(Path("integration/resources/templates").glob("**/*.yml"))
+    )
+
+
+SCHEMA_VALIDATION_TESTS = [os.path.splitext(f)[0] for f in get_all_test_templates() if not should_skip_test(str(f))]
```
hoffa, line 72: Why the split?

acristin (approved):
This seems fine, but it would be useful to document the reasoning behind moving from a lint task to a unit test.

tests/translator/input/error_schema_validation.yaml
```diff
@@ -0,0 +1,25 @@
...
+
+  MyBucket:
+    Type: AWS::S3::Bucket
+
+  MyConnector:
+    Type: AWS::Serverless::Connector
+    Properties:
+      Source:
+        Id: MyFunction
+      Desitation: # wrong spelling
```
hoffa, line 21: Another test for wrong type would be great as well! 🙇 

Maybe:

```yaml
Resources:
  MyFunction:
    Properties:
      Runtime: nodejs14.x
      Handler: index.handler
      InlineCode: |
        const AWS = require('aws-sdk');
        exports.handler = async (event) => {
          console.log(JSON.stringify(event));
        };
  MyBucket:
    Type: AWS::S3::Bucket

  MyConnector:
    Type: AWS::Serverless::Connector
    Properties:
      Source:
        Id: MyFunction
      Destination:
        Id: 42 # wrong type
      Permissions:
      - Read
      - Write
```

tests/schema/test_validate_schema.py
```diff
@@ -0,0 +1,86 @@
...
+            return True
+    return False
+
+
+def get_all_test_templates():
+    return (
+        list(Path("tests/translator/input").glob("**/*.yaml"))
+        + list(Path("tests/translator/input").glob("**/*.yml"))
+        + list(Path("tests/validator/input").glob("**/*.yaml"))
+        + list(Path("tests/validator/input").glob("**/*.yml"))
```
xazhao, line 66: Didn't see any `yml` file in this folder and translator/input folder. Maybe I missed some files?

hoffa: Just in case. Both are valid extensions for YAML.

tests/schema/test_validate_schema.py
GavinZZ: This is for showing the exact template name when running pytest. Otherwise, it will only have numerics.
Screen Shot 2022-11-22 at 1 00 17 PM

Makefile
GavinZZ: Oh I thought I deleted it. My force push must have reverted this change...

tests/schema/test_validate_schema.py
```diff
@@ -0,0 +1,92 @@
...
+    )
+
+
+SCHEMA_VALIDATION_TESTS = [os.path.splitext(f)[0] for f in get_all_test_templates() if not should_skip_test(str(f))]
+
+
+class TestValidateSchema(TestCase):
+    @parameterized.expand(itertools.product(SCHEMA_VALIDATION_TESTS))
+    def test_validate_schema(self, testcase):
+        file_name = testcase + ".yaml"
```
hoffa, line 78: What if my file is `test.yml`?

tests/translator/output/error_schema_validation_wrong_type.json
```diff
@@ -0,0 +1,8 @@
+{
+  "errorMessage": "Invalid Serverless Application Specification document. Number of errors found: 1. Resource with id [MyConnector] is invalid. 'Id' is missing or not a string.",
+  "errors": [
+    {
+      "errorMessage": "Resource with id [MyConnector] is invalid. 'Id' is missing or not a string."
+    }
+  ]
```
hoffa, line 7: I think the `errors` is unused?

```suggestion
  "errorMessage": "Invalid Serverless Application Specification document. Number of errors found: 1. Resource with id [MyConnector] is invalid. 'Id' is missing or not a string."
```

tests/schema/test_validate_schema.py
```diff
@@ -0,0 +1,90 @@
...
+        list(Path("tests/translator/input").glob("**/*.yaml"))
+        + list(Path("tests/translator/input").glob("**/*.yml"))
+        + list(Path("tests/validator/input").glob("**/*.yaml"))
+        + list(Path("tests/validator/input").glob("**/*.yml"))
+        + list(Path("integration/resources/templates").glob("**/*.yaml"))
+        + list(Path("integration/resources/templates").glob("**/*.yml"))
+    )
+
+
+SCHEMA_VALIDATION_TESTS = [str(f) for f in get_all_test_templates() if not should_skip_test(str(f))]
```
hoffa, line 72: nit: maybe add the skip to `test_validate_schema`? so it's obvious when running that the files were skipped

e.g.

```python
if ...:
    self.skipTest()
```

GavinZZ: This method will add some noise to the pytest output. I don't think I care about this when I am running `make test`
Screen Shot 2022-11-22 at 2 39 31 PM

hoffa: We should be resolving all of those, so getting a constant reminder doesn't hurt either tbh.

hoffa (approved):

PR #2647, opened 2022-11-23
Multi dest connector

xazhao:
### Checklist

    - [ ] Correct values
- [ ] Update documentation

### Examples?


samtranslator/model/sam_resources.py
```diff
@@ -1785,88 +1787,106 @@ def to_cloudformation(self, **kwargs: Any) -> List[Resource]:  # type: ignore
...
         replace_depends_on_logical_id(self.logical_id, generated_logical_ids, resource_resolver)
 
-        self._add_connector_metadata(generated_resources, original_template, source, destination)
-        if generated_resources:
-            return generated_resources
+        if list_generated_resources:
+            return list_generated_resources
 
         # Should support all profile types
         raise TypeError(f"Unknown profile policy type '{profile_type}'")
```
aahung, line 1892: since this variable `profile_type` was created in the `for` loop, it doesn't make sense to use it outside the loop anymore

samtranslator/model/sam_resources.py
```diff
@@ -1785,88 +1787,106 @@ def to_cloudformation(self, **kwargs: Any) -> List[Resource]:  # type: ignore
...
+                )
+            if profile_type == "AWS_SNS_TOPIC_POLICY":
+                generated_resources.append(
+                    self._construct_sns_topic_policy(source, destination, profile_properties, index, multi_dest)
+                )
+            if profile_type == "AWS_LAMBDA_PERMISSION":
+                generated_resources.extend(
+                    self._construct_lambda_permission_policy(source, destination, profile_properties, index, multi_dest)
+                )
+
```
aahung, line 1881: here we should do a `raise TypeError(f"Unknown profile policy type '{profile_type}'")`

samtranslator/model/sam_resources.py
aahung: this should change to "the connector destination is empty"

aahung (commented):
Overall it looks good! Just some minor error handling issues. Can we make sure those errors are correct using some transform tests?

samtranslator/model/sam_resources.py
```diff
@@ -1785,88 +1787,106 @@ def to_cloudformation(self, **kwargs: Any) -> List[Resource]:  # type: ignore
...
 
-        # removing duplicate permissions
-        self.Permissions = list(set(self.Permissions))
-        profile_type, profile_properties = profile["Type"], profile["Properties"]
-        profile_permissions = profile_properties["AccessCategories"]
-        valid_permissions_combinations = profile_properties.get("ValidAccessCategories")
+        list_generated_resources: List[Resource] = []
 
-        valid_permissions_str = ", ".join(profile_permissions)
+        for index, dest in enumerate(self.Destination):
```
hoffa, line 1797: A large nested function is hard to read. Should be possible to separate the concerns into logical parts; adding a 1:1 connector is one thing, and multi-destination is a generalization of that.

samtranslator/model/sam_resources.py
```diff
@@ -1923,6 +1947,8 @@ def _construct_lambda_permission_policy(
         source: ConnectorResourceReference,
         destination: ConnectorResourceReference,
         profile: ConnectorProfile,
+        index: int,
+        multi_dest: bool,
```
hoffa, line 1951: Could have a single variable by e.g. using `index: Optional[int]` instead.

samtranslator/model/sam_resources.py
xazhao: Synced offline. Will move the majority of for loop into a function `generate_resource`.

samtranslator/model/sam_resources.py
xazhao: the index is from the loop. So I think we still need `multi_dest` flag to decide it's a single destination or a list that only includes 1 destination.

samtranslator/model/sam_resources.py
xazhao: Good catch! Will update it.

samtranslator/model/sam_resources.py
xazhao: agree

samtranslator/model/sam_resources.py
hoffa: `index` is only used for multi-resource connectors right? An "index" doesn't make sense for a non-list structure anyway, so I don't see why it should be relevant when not using multi-resource.

hoffa: (Also might want to make that `dest_index` for clarity. There's a good change we'll do multi-source as well.)

xazhao:
@aahung Actually those errors can not be approached by transform test. They will be errors before.

samtranslator/model/sam_resources.py
```diff
@@ -1770,42 +1770,41 @@ class SamConnector(SamResourceMacro):
...
             raise InvalidResourceException(
                 self.logical_id,
                 f"Unable to create connector from {source.resource_type} to {destination.resource_type}; it's not supported or the template is invalid.",
             )
 
+        profile_type, profile_properties = profile["Type"], profile["Properties"]
+
         # removing duplicate permissions
         self.Permissions = list(set(self.Permissions))
-        profile_type, profile_properties = profile["Type"], profile["Properties"]
```
hoffa, line 1803: nit: should avoid unrelated changes (especially for cognitively high-load PRs like this), makes reviewing harder

samtranslator/model/sam_resources.py
```diff
@@ -1816,7 +1815,7 @@ def to_cloudformation(self, **kwargs: Any) -> List[Resource]:  # type: ignore
             if permission not in profile_permissions:
                 raise InvalidResourceException(
                     self.logical_id,
-                    f"Unsupported 'Permissions' provided; valid values are: {valid_permissions_str}.",
+                    f"Unsupported 'Permissions' provided for connector from {source.resource_type} to {destination.resource_type}; valid values are: {valid_permissions_str}.",
```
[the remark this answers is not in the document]
GavinZZ: > Why the change?
> 
> The error message already includes the logical ID which allows looking up any information necessary. Also `resource_type` is the CloudFormation type, so it doesn't necessarily match what the customer uses.

If users are using multi dest connector, isn't it clearer to tell which source-destination pair results in this error message?

samtranslator/model/sam_resources.py
```diff
@@ -1848,28 +1848,62 @@ def to_cloudformation(self, **kwargs: Any) -> List[Resource]:  # type: ignore
...
 
-        self._add_connector_metadata(generated_resources, original_template, source, destination)
-        if generated_resources:
-            return generated_resources
+        if list_generated_resources:
+            return list_generated_resources
 
-        # Should support all profile types
-        raise TypeError(f"Unknown profile policy type '{profile_type}'")
+        raise TypeError("The destination is empty")
```
GavinZZ, line 1906: nit: This error message doesn't seem quite right. There are other scenarios where `list_generated_resources` can be empty list.

samtranslator/model/sam_resources.py
```diff
@@ -1888,6 +1922,8 @@ def _construct_iam_policy(
         destination: ConnectorResourceReference,
         profile: ConnectorProfile,
         resource_resolver: ResourceResolver,
+        dest_index: int,
+        multi_dest: bool,
```
GavinZZ, line 1926: Instead of these two variables, can we make `dest_index` as a optional? The checks below can become 
`policy_name = f"{self.logical_id}PolicyDestination{dest_index}" if dest_index != None else f"{self.logical_id}Policy"`

samtranslator/model/sam_resources.py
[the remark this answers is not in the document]
xazhao: Synced with Chris offline. It's clearer to have customer know which resource pair doesn't support the permission.

samtranslator/model/sam_resources.py
```diff
@@ -1848,28 +1848,62 @@ def to_cloudformation(self, **kwargs: Any) -> List[Resource]:  # type: ignore
...
+        list_generated_resources: List[Resource] = []
+
+        for dest_index, dest in enumerate(self.Destination):
+            try:
+                destination = get_resource_reference(dest, resource_resolver, self.Source)
+                source = get_resource_reference(self.Source, resource_resolver, dest)
+            except ConnectorResourceError as e:
+                raise InvalidResourceException(self.logical_id, str(e))
+
+            generated_resource = self.generate_resource(source, destination, dest_index, multi_dest, resource_resolver)
```
hoffa, line 1895: This should be _resources_, right?

samtranslator/model/sam_resources.py
xazhao: Do you know what scenarios `list_generated_resources` can be empty list? Trying to figure out the right message here.

samtranslator/model/sam_resources.py
xazhao: will update it.

GavinZZ (commented):
Generally LGTM, just some minor comments.

samtranslator/model/sam_resources.py
xazhao: yes.

GavinZZ: `generated_resource` seems right here. It's passing only this specific source and destination pair.

samtranslator/model/sam_resources.py
aahung: I don't think this will ever happen if our profile is right. So if it happens, it should be an availability dip. We can raise something like unexpected error

samtranslator/model/sam_resources.py
xazhao: For lambda permission with `READ + WRITE`, we creates 2 lambda permissions. In other cases, we only create 1 policy

samtranslator/model/sam_resources.py
xazhao: We would still need a flag to identify it's a single destination or a list only contains one destination.

samtranslator/model/sam_resources.py
xazhao: updates: an empty destination list will trigger this error. So I updated to `InvalidResourceException` because it's a template error.

GavinZZ (approved):

samtranslator/model/sam_resources.py
```diff
@@ -1850,26 +1848,61 @@ def to_cloudformation(self, **kwargs: Any) -> List[Resource]:  # type: ignore
...
+                self._construct_lambda_permission_policy(
+                    source, destination, profile_properties, dest_index, multi_dest
+                )
+            )
+        else:
+            raise TypeError(f"Profile type {profile_type} is not supported")
+        return generated_resources
+
+    @cw_timer
+    def to_cloudformation(self, **kwargs: Any) -> List[Resource]:  # type: ignore
```
hoffa, line 1874: Why `# type: ignore`?

samtranslator/model/sam_resources.py
```diff
@@ -1850,26 +1848,61 @@ def to_cloudformation(self, **kwargs: Any) -> List[Resource]:  # type: ignore
...
 
-        self._add_connector_metadata(generated_resources, original_template, source, destination)
-        if generated_resources:
-            return generated_resources
+        if list_generated_resources:
+            return list_generated_resources
 
-        # Should support all profile types
-        raise TypeError(f"Unknown profile policy type '{profile_type}'")
+        raise InvalidResourceException(self.logical_id, "The destination is an empty list")
```
hoffa, line 1905: To match the wording so far:

```suggestion
        raise InvalidResourceException(self.logical_id, "'Destination' is an empty list")
```

hoffa (approved):

aahung (approved):
:shipit:

PR #2648, opened 2022-11-23
chore: Refactor some chained dict.get() with dict_deep_get()

aahung:
### Issue #, if available

There aren't a lot of them: 
image

### Checklist

    - [ ] Correct values
- [ ] Update documentation

### Examples?


samtranslator/utils/utils.py
```diff
@@ -21,3 +21,27 @@ def insert_unique(xs: Any, vs: Any) -> List[Any]:
...
+        return path_prefix + "." + self.relative_path
+
+
+def dict_deep_get(d: Any, path: str) -> Optional[Any]:
+    relative_path = ""
+    _path_nodes = path.split(".")
+    while _path_nodes:
+        if d is None:
+            return None
+        if not isinstance(d, dict):
```
hoffa, line 42: Does this enforce the return type? If yes it would be clearer explicitly specified by caller.

hoffa: Also unit tests for this would be great if it's to become the generic way to get stuff out of dicts. It's not clear reading the code how it fails.

ssenchenko (approved):

samtranslator/utils/utils.py
aahung: > Does this enforce the return type? If yes it would be clearer explicitly specified by caller.
similar to `dict.get()` it doesn't enforce the return value. 

> Also unit tests for this would be great if it's to become the generic way to get stuff out of dicts. It's not clear reading the code how it fails.

yeah that's good idea, will add soon

aahung: it's validating the parents

aahung:
@ssenchenko @hoffa updated.

I removed a test where `responses = null`. It won't cause availability dips anymore and now `dict_deep_get` will simply return `None` and skip it.

hoffa (approved):

ssenchenko (approved):

PR #2659, opened 2022-11-24
feat: Role Path for AWS::Serverless::Function

GavinZZ:
### Description of changes
Added Role Path support for SAM Function.

### Checklist

    - [ ] Correct values
- [ ] Update documentation

### Examples?


ssenchenko (approved):

docs/globals.rst
```diff
@@ -67,6 +67,7 @@ Currently, the following resources and properties are being supported:
       Layers:
       AutoPublishAlias:
       DeploymentPreference:
+      RolePath:
```
hoffa, line 70: We should remove this whole section and instead refer to the schema directly, as that’s the source of truth.

samtranslator/schema/aws_serverless_function.py
```diff
@@ -489,6 +490,7 @@ class Properties(BaseModel):
     Layers: Optional[Layers] = prop("Layers")
     MemorySize: Optional[MemorySize] = prop("MemorySize")
     PackageType: Optional[PassThrough] = prop("PackageType")
+    RolePath: Optional[RolePath]
```
hoffa, line 493: nit: TODO to remember adding docs when live

samtranslator/model/sam_resources.py
```diff
@@ -141,6 +142,7 @@ class SamFunction(SamResourceMacro):
     Role: Optional[Intrinsicable[str]]
     AssumeRolePolicyDocument: Optional[Dict[str, Any]]
     Policies: Optional[List[Any]]
+    RolePath: Optional[Intrinsicable[str]]
```
hoffa, line 145: Why str? We just pass the value to CloudFormation, and we don’t care what it is.

docs/globals.rst
hoffa: Let's do it in a separate PR. Might have properties not in schema, or other interesting information.

samtranslator/model/sam_resources.py
GavinZZ: IAM::Role in `iam.py` expects this property as `str`. I think it makes more sense to keep it as `str` than `Any` here.

tests/translator/input/function_with_role_path.yaml
```diff
@@ -0,0 +1,8 @@
+Resources:
```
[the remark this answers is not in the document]
GavinZZ: Good catch. Let me add it.

samtranslator/model/sam_resources.py
hoffa: What is `iam.py`? Why does it care?

GavinZZ:
```
class IAMRole(Resource):
    resource_type = "AWS::IAM::Role"
    property_types = {
        "AssumeRolePolicyDocument": PropertyType(True, is_type(dict)),
        "ManagedPolicyArns": PropertyType(False, is_type(list)),
        "Path": PropertyType(False, is_str()),
        "Policies": PropertyType(False, is_type(list)),
        "PermissionsBoundary": PropertyType(False, is_str()),
        "Tags": PropertyType(False, list_of(is_type(dict))),
    }

    runtime_attrs = {"name": lambda self: ref(self.logical_id), "arn": lambda self: fnGetAtt(self.logical_id, "Arn")}
```
I'm talking about this class definition.

GavinZZ: Keeping `RolePath` as str will also allow early type check since we know it's going to fail if customers input an array as input.

hoffa: Hmm okay. Regardless, I still think `Optional[Intrinsicable[str]]` is misleading. It's a pass-through value and don't need to make any assumptions about the underlying resource we don't control, so `Optional[PassThrough]` is a lot clearer (and easier; one less thing to figure out).

GavinZZ: Discussed offline. Will update it to use `PassThrough`.

hoffa (approved):

PR #2660, opened 2022-11-24
feat: update transform test script to update ManagedPolicyArn with partion

xazhao:
### Issue #, if available

The partition generated by the script is always `aws`
Now adding a new option to set partition to `aws-cn` or `aws-us-gov` based on the output path

### Checklist

    - [ ] Correct values
- [ ] Update documentation

### Examples?


GavinZZ (commented):
I actually think it would be nice to have this by default.

xazhao:
@GavinZZ There are some tests case which partition is hardcoded. In this case I don't want to have the script to update those arns.

bin/add_transform_test.py
```diff
@@ -45,6 +47,20 @@ def add_regional_endpoint_configuration_if_needed(template: Dict[str, Any]) -> D
 
     return template
 
+def update_partition(region: str, template: Dict[str, Any]) -> Dict[str, Any]:
```
ssenchenko, line 50: Since we actually doing smth here and in `generate_transform_test_output_files` it'd be great to have some tests

GavinZZ (approved):

aaythapa (approved):
Thank you for this 🙏

bin/add_transform_test.py
xazhao: hmm it's a script to run the translator given a file and move those files to related folder. Unit tests seems useless and integration test is a overkill because to really see the desired file, we have to set region to cn or us-gov. So I don't find a meaningful testing solution here.

hoffa:
> @GavinZZ There are some tests case which partition is hardcoded. In this case I don't want to have the script to update those arns.

I agree with @GavinZZ, why not have this by default? This script is only used for new tests, so I don't see why the existing tests matter.

bin/add_transform_test.py
```diff
@@ -46,6 +48,21 @@ def add_regional_endpoint_configuration_if_needed(template: Dict[str, Any]) -> D
     return template
 
 
+def update_partition(region: str, template: Dict[str, Any]) -> Dict[str, Any]:
```
hoffa, line 51: Can we do simpler and more generic:

```python
return s.replace("aws:arn", partition)
```

bin/add_transform_test.py
ssenchenko: Right, that's the script!

ssenchenko (approved):

bin/add_transform_test.py
xazhao: sure we can do that

bin/add_transform_test.py
```diff
@@ -67,21 +81,21 @@ def generate_transform_test_output_files(input_file_path: str, file_basename: st
...
         if not CLI_OPTIONS.get("--disable-api-configuration"):
             template = read_json_file(temp_output_file.name)
             template = add_regional_endpoint_configuration_if_needed(template)
             write_json_file(template, temp_output_file.name)
 
-        for output_path in regional_transform_test_output_paths:
+        for region, output_path in regional_transform_test_output_paths.items():
             shutil.copyfile(temp_output_file.name, output_path)
+            template = read_json_file(output_path)
+            template = update_partition(region, template)
```
hoffa, line 97: Can we have an option to opt out, like `--disable-api-configuration`?

DEVELOPMENT_GUIDE.md
```diff
@@ -143,6 +143,11 @@ To disable this feature, run the following command instead.
 python3 bin/add_transform_test.py --template-file template.yaml --disable-api-configuration
 ```
 
+The script will update the partition of arn based on the output path. To disable updating the partition of arn
```
hoffa, line 146:
```suggestion
The script automatically updates hardcoded ARN partitions to match the output partition. To disable this, use:
```

hoffa (approved):

Issue #2279, opened 2022-01-07
Hyphens are stripped from API mapping paths


AndrewKvalheim:
### Description:

The template transform incorrectly [strips hyphens](https://github.com/aws/serverless-application-model/blob/ab6943a340a3f489af62b8c70c1366242b2887fe/samtranslator/model/api/api_generator.py#L503) from API mapping paths.

This was previously reported [in a comment](https://github.com/aws/serverless-application-model/issues/1456#issuecomment-652779157) on a related bug but was never addressed.

### Steps to reproduce:

Specify an API mapping path containing a hyphen:

```yaml
Globals:
  Api:
    Domain:
      BasePath: foo-bar
```

### Observed result:

The transformed path is missing the hyphen:

```json
{
  "Type": "AWS::ApiGateway::BasePathMapping",
  "Properties": {
    "BasePath": "foobar"
  }
}
```

### Expected result:

Hyphens are not removed, e.g. `foo-bar`.

### Additional environment details

- SAM CLI 1.35.0
- NixOS 21.11

AndrewKvalheim:
Workaround:

```yaml
Resources:
  Issue2279BasePathMapping:
    Type: AWS::ApiGateway::BasePathMapping
    Properties:
      RestApiId: !Ref ServerlessRestApi
      Stage: !Ref ServerlessRestApiProdStage
      DomainName: !Ref ServerlessRestApi.DomainName
      BasePath: foo-bar
```

mildaniel:
Thanks for reporting the issue and providing a workaround, we will be investigating it further. Please watch this channel for more updates, and feel free to reach out.

GavinZZ:
Hi there, this issue has been addressed and it's currently in pending release state.

GavinZZ:
The latest release [1.56.0](https://github.com/aws/serverless-application-model/releases/tag/v1.56.0) has included the fix.

PR #2680, opened 2022-11-29
Don't strip hyphen from BasePathMapping

GavinZZ:
### Issue #, if available
https://github.com/aws/serverless-application-model/issues/2279

### Description of changes
Input template 
```
Globals:
  Api:
    Domain:
      BasePath: foo-bar
```

Observed output template
```
"MyApifoobarBasePathMapping": {
  "Type": "AWS::ApiGateway::BasePathMapping",
  "Properties": {
    "BasePath": "foobar"
  }
}
```

Expected output template
```
"MyApifoobarBasePathMapping": {
  "Type": "AWS::ApiGateway::BasePathMapping",
  "Properties": {
    "BasePath": "foo-bar"
  }
}
```

### Checklist

    - [ ] Correct values
- [ ] Update documentation

### Examples?


samtranslator/model/api/api_generator.py
```diff
@@ -513,16 +513,16 @@ def _construct_api_domain(self, rest_api, route53_record_set_groups):  # type: i
             basepath_mapping.Stage = ref(rest_api.logical_id + ".Stage")
             basepath_resource_list.extend([basepath_mapping])
         else:
-            for path in basepaths:
-                path = "".join(e for e in path if e.isalnum())
+            for basepath in basepaths:
+                path = "".join(e for e in basepath if e.isalnum())
```
hoffa, line 517: What changed? Any tests?

GavinZZ: oh wait, the test file changes didn't got committed. Let me add that

tests/translator/output/api_with_basic_custom_domain.json
```diff
@@ -165,7 +165,7 @@
     },
     "MyApifetchBasePathMapping": {
       "Properties": {
-        "BasePath": "fetch",
+        "BasePath": "/fetch",
```
hoffa, line 168: Test outputs should typically never change. Why are these changing?

GavinZZ: removed. Changed the code to match `Http API` where we remove trailing and leading `/`.

tests/translator/input/api_with_custom_base_path.yaml
```diff
@@ -0,0 +1,20 @@
+Globals:
+  Api:
+    Domain:
+      BasePath: foo-bar
```
hoffa, line 4: I don't see this in the output files. 🤔

hoffa: And how is this relevant to https://github.com/aws/serverless-application-model/issues/2279#issuecomment-1007055489?

Does something change with `Globals` versus `Properties`?

samtranslator/model/api/api_generator.py
```diff
@@ -513,16 +513,18 @@ def _construct_api_domain(self, rest_api, route53_record_set_groups):  # type: i
             basepath_mapping.Stage = ref(rest_api.logical_id + ".Stage")
             basepath_resource_list.extend([basepath_mapping])
         else:
-            for path in basepaths:
-                path = "".join(e for e in path if e.isalnum())
+            for basepath in basepaths:
+                # ignore leading and trailing `/` in the path name
```
hoffa, line 517: Why?

samtranslator/model/api/api_generator.py
```diff
@@ -513,16 +513,18 @@ def _construct_api_domain(self, rest_api, route53_record_set_groups):  # type: i
...
+                basepath = basepath.strip("/")
+                path = "".join(e for e in basepath if e.isalnum())
```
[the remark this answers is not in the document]
GavinZZ: path is for the logical id.  basepath is used for `BasePath` property.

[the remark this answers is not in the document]
aaythapa: ah i missed that 👍

aaythapa (approved):

samtranslator/model/api/api_generator.py
GavinZZ: `A base path may only contain letters, numbers, and one of "$-_.+!*'()"`

hoffa: That doesn't explain why it's only stripping slashes.

samtranslator/model/api/api_generator.py
```diff
@@ -502,6 +502,9 @@ def _construct_api_domain(self, rest_api, route53_record_set_groups):  # type: i
         else:
             basepaths = None
 
+        # Boolean to allow/disallow hyphens in BasePath property
+        strip_hyphens = self.domain.get("StripHyphens", True)
```
hoffa, line 506: For future, we should have some way of specifying the default in the props definition, so we don't need to sprinkle custom logic like this.

samtranslator/model/api/api_generator.py
```diff
@@ -513,16 +516,19 @@ def _construct_api_domain(self, rest_api, route53_record_set_groups):  # type: i
             basepath_mapping.Stage = ref(rest_api.logical_id + ".Stage")
             basepath_resource_list.extend([basepath_mapping])
         else:
-            for path in basepaths:
-                path = "".join(e for e in path if e.isalnum())
+            for basepath in basepaths:
+                # Remove possible leading and trailing '/' because a base path may only
+                # contain letters, numbers, and one of "$-_.+!*'()"
+                basepath = basepath.strip("/")
```
hoffa, line 522: The `strip_hyphens` ternary would be much clearer here, as this is dead code if it's `True`.

tests/translator/output/api_with_custom_base_path.json
```diff
@@ -99,6 +99,57 @@
...
+        "RestApiId": {
+          "Ref": "MyApiWithStripHyphens"
+        },
+        "StageName": "Prod"
+      },
+      "Type": "AWS::ApiGateway::Stage"
+    },
+    "MyApiWithStripHyphensfoobarBasePathMapping": {
+      "Properties": {
+        "BasePath": "foo-bar",
```
hoffa, line 140: `WithStripHyphens` but still has hyphens?

xazhao (approved):

samtranslator/model/api/api_generator.py
GavinZZ: Agree.

tests/translator/input/api_with_custom_base_path.yaml
```diff
@@ -32,3 +32,18 @@ Resources:
...
+  MyApiWithoutStripHyphens:
+    Type: AWS::Serverless::Api
+    Properties:
+      OpenApiVersion: 3.0.1
+      StageName: Prod
+      Domain:
+        DomainName: MyEdgeDomainName
+        CertificateArn: MyEdgeDomainCert
+        EndpointConfiguration: EDGE
+        StripHyphens: false
```
hoffa, line 45: 1. Is `StripHyphens` clear enough from context? An uglier but clearer alternative could be `BasePathStripHyphens`.
2. Is `StripHyphens` correct? I don't see code specifically stripping hyphens, but anything non-alphanumeric.

GavinZZ: Good call. `BasePathStripSymbols` is probably better in this context.

hoffa: Symbol is quite a fancy word, and makes me think of symbol tables in binaries. Plus any character can really be a "symbol".

Some bad alternatives:

- `BasePathSanitize`/`SanitizeBasePath`
- `BasePathAlphanumeric`

aahung (approved):

tests/translator/input/api_with_custom_base_path.yaml
```diff
@@ -32,3 +32,18 @@ Resources:
...
+  MyApiWithoutNormalizedBasePath:
+    Type: AWS::Serverless::Api
+    Properties:
+      OpenApiVersion: 3.0.1
+      StageName: Prod
+      Domain:
+        DomainName: MyEdgeDomainName
+        CertificateArn: MyEdgeDomainCert
+        EndpointConfiguration: EDGE
+        NormalizeBasePath: false
```
hoffa, line 45: Would be clearer if it used `BasePath` with slashes as well, to make sure we don't revert this later and tests keep passing.

hoffa (approved):

PR #2690, opened 2022-11-30
Added support for mTLS auth for MSK and Kafka

mbfreder:
### Description of changes

Support [mTLS authentication](https://aws.amazon.com/about-aws/whats-new/2021/11/aws-lambda-supports-mtls-authentication-amazon-msk-event-source/) for MSK and Kafka.

Ran the template tests

### Checklist

    - [x] Correct values
- [ ] Update documentation

### Examples?


tests/model/eventsources/test_msk_event_source.py
```diff
@@ -0,0 +1,47 @@
+from unittest import TestCase
+from samtranslator.model.eventsources.pull import MSK
+from samtranslator.model.exceptions import InvalidEventException
```
msharris1, line 3: Is this import used anywhere? If not it can be removed

samtranslator/model/eventsources/pull.py
```diff
@@ -287,7 +287,26 @@ def get_policy_arn(self):  # type: ignore[no-untyped-def]
         return ArnGenerator.generate_aws_managed_policy_arn("service-role/AWSLambdaMSKExecutionRole")  # type: ignore[no-untyped-call]
 
     def get_policy_statements(self):  # type: ignore[no-untyped-def]
-        return None
+        document = None
+        if self.SourceAccessConfigurations:
+            for conf in self.SourceAccessConfigurations:
+                if isinstance(conf, dict) and conf.get("Type") == "CLIENT_CERTIFICATE_TLS_AUTH":
```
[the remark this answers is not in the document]
mbfreder: Right.

msharris1 (approved):

tests/model/eventsources/test_msk_event_source.py
```diff
@@ -0,0 +1,46 @@
+from unittest import TestCase
+from samtranslator.model.eventsources.pull import MSK
+
+
+class MSKEventSource(TestCase):
```
hoffa, line 5: Just curious, any specific value to this test compared to a transform test? We typically prefer transform tests, as they mimic customer templates much closer, and are validated against [the JSON schema](https://github.com/aws/serverless-application-model/discussions/2645).

samtranslator/model/eventsources/pull.py
```diff
@@ -287,7 +287,26 @@ def get_policy_arn(self):  # type: ignore[no-untyped-def]
         return ArnGenerator.generate_aws_managed_policy_arn("service-role/AWSLambdaMSKExecutionRole")  # type: ignore[no-untyped-call]
 
     def get_policy_statements(self):  # type: ignore[no-untyped-def]
-        return None
+        document = None
+        if self.SourceAccessConfigurations:
+            for conf in self.SourceAccessConfigurations:
+                if isinstance(conf, dict) and conf.get("Type") == "CLIENT_CERTIFICATE_TLS_AUTH" and conf.get("URI"):
+                    document = {
```
hoffa, line 294: Shouldn't it automatically `return [document]` here?

Would also be good to include a comment explaining why we only pick one; is that a constraint of the underlying resource?

hoffa (approved):

samtranslator/model/eventsources/pull.py
```diff
@@ -287,7 +287,26 @@ def get_policy_arn(self):  # type: ignore[no-untyped-def]
...
+                        "PolicyName": "SamAutoGeneratedMSKPolicy",
```
hoffa, line 295: Any documentation for reference? Why this specific policy document?

tests/translator/output/aws-cn/msk_with_mtls_auth.json
```diff
@@ -0,0 +1,104 @@
...
+                    "secretsmanager:GetSecretValue"
+                  ],
+                  "Effect": "Allow",
+                  "Resource": {
+                    "Fn::Sub": "arn:${AWS::Partition}:secretsmanager:us-west-2:123456789012:secret:my-path/my-secret-name-1a2b3c"
+                  }
+                }
+              ]
+            },
+            "PolicyName": "SamAutoGeneratedMSKPolicy"
```
hoffa, line 91: Is this analogous to the `SelfManagedKafkaExecutionRolePolicy` for `SelfManagedKafka` below? If yes, could this be `MSKExecutionRolePolicy` so the symmetry is becomes obvious?

samtranslator/model/eventsources/pull.py
```diff
@@ -287,7 +287,26 @@ def get_policy_arn(self):  # type: ignore[no-untyped-def]
         return ArnGenerator.generate_aws_managed_policy_arn("service-role/AWSLambdaMSKExecutionRole")  # type: ignore[no-untyped-call]
 
     def get_policy_statements(self):  # type: ignore[no-untyped-def]
```
hoffa, line 289: I see this changes `MSK` events; which part adds in this PR adds support for `SelfManagedKafka`?

mbfreder: SelfManagedKafka is the last class in the file.

hoffa: Ah I see, it adds a new `AUTH_MECHANISM`, and the code checks whether `Type` is one of them. Then the question becomes why is `MSK` different? Asked in another comment.

samtranslator/model/eventsources/pull.py
hoffa: For my latter question, I see on https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-lambda-eventsourcemapping-sourceaccessconfiguration.html that Amazon MSK only has two types; `SASL_SCRAM_512_AUTH` looks unrelated to this PR. So the change makes sense.

And for client TLS it says:
> The Secrets Manager ARN of your secret key

samtranslator/model/eventsources/pull.py
```diff
@@ -287,7 +287,26 @@ def get_policy_arn(self):  # type: ignore[no-untyped-def]
...
-        return None
+        document = None
+        if self.SourceAccessConfigurations:
+            for conf in self.SourceAccessConfigurations:
+                if isinstance(conf, dict) and conf.get("Type") == "CLIENT_CERTIFICATE_TLS_AUTH" and conf.get("URI"):
```
hoffa, line 293: So it's quickly clear why this specific policy:

```suggestion
                # https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-lambda-eventsourcemapping-sourceaccessconfiguration.html
                if isinstance(conf, dict) and conf.get("Type") == "CLIENT_CERTIFICATE_TLS_AUTH" and conf.get("URI"):
```

samtranslator/model/eventsources/pull.py
mbfreder: Sorry, I'm not sure I understand your question. What do you mean by "explaining why we only pick one"? are you referring to the source access type? According to the [docs](https://docs.aws.amazon.com/lambda/latest/dg/API_SourceAccessConfiguration.html), CLIENT_CERTIFICATE_TLS_AUTH is the only one needed for mTLS auth.

tests/model/eventsources/test_msk_event_source.py
mbfreder: This is just a unit test that I added to test my logic. I found a bunch of tests in that same folder for other event sources so I thought I should add one as well for the logic I added in MSK. I also added transform tests.

samtranslator/model/eventsources/pull.py
hoffa: So if I have multiple `CLIENT_CERTIFICATE_TLS_AUTH` configurations, the `document` will only refer to the last `URI`, correct? As it loops over all of them, but each time it matches the condition, it overwrites `document`.

tests/model/eventsources/test_msk_event_source.py
hoffa: I see, makes sense. 👍  In general if there are transform tests covering the transform logic, there's no need for separate unit tests.

samtranslator/model/eventsources/pull.py
mbfreder: Correct.

mbfreder: Do you think we should end the loop after the first match?

hoffa: My question is: why? Why pick only one? Is it a constraint of the underlying resource?

mbfreder: We only need one ARN of the secret key containing the client certificate. Can't think of why we would need more.

hoffa: It's not obvious to me reading the code why that assumption is made (without spending time reading through the internals and how it works). A "why" comment will help clarify for future readers.

hoffa: Perhaps another way to put it, if Lambda supports multiple `CLIENT_CERTIFICATE_TLS_AUTH`s in the same configuration, perhaps there's a valid use case that we shouldn't break. If Lambda doesn't support it, that is the "why".

mbfreder: Ah I get what you mean now. I have not seen mentioned anywhere in the docs that Lambda supports multiple `CLIENT_CERTIFICATE_TLS_AUTH`. You're right. I can add a comment to make it clear for future readers.

tests/translator/output/aws-cn/msk_with_mtls_auth.json
mbfreder: Yep. Makes sense.

samtranslator/model/eventsources/pull.py
mbfreder: Added the comment.

tests/translator/output/aws-cn/msk_with_mtls_auth.json
mbfreder: Changed the policy name.

samtranslator/model/eventsources/pull.py
```diff
@@ -287,7 +287,28 @@ def get_policy_arn(self):  # type: ignore[no-untyped-def]
...
+                                        ],
+                                        "Effect": "Allow",
+                                        "Resource": conf.get("URI"),
+                                    }
+                                ]
+                            },
+                        }
+                    ]
+
+            return None
```
hoffa, line 311: Technically identical, but a little clearer and mypy/pylint won't complain when the `# type: ignore` is removed:

```suggestion
        return None
```

hoffa (approved):
🚢

samtranslator/model/eventsources/pull.py
```diff
@@ -287,7 +287,28 @@ def get_policy_arn(self):  # type: ignore[no-untyped-def]
         return ArnGenerator.generate_aws_managed_policy_arn("service-role/AWSLambdaMSKExecutionRole")  # type: ignore[no-untyped-call]
 
     def get_policy_statements(self):  # type: ignore[no-untyped-def]
-        return None
+        if self.SourceAccessConfigurations:
```
aahung, line 290: nit: Wanna make sure we are not introducing any potential internal server error here. Can we add a test when `SourceAccessConfigurations` is not a list and see what will the error be like?

mbfreder: According to the [docs](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-lambda-eventsourcemapping-sourceaccessconfiguration.html) `SourceAccessConfigurations` has to be a list.

mbfreder: when we add a test where sourceAccessConfigurations is not a list, we get the following error: 
`ERROR:__main__:Invalid Serverless Application Specification document. Number of errors found: 1. Resource with id [MyMskStreamProcessorMyMskEvent] is invalid. Type of property 'SourceAccessConfigurations' is invalid.`

hoffa: What @aahung is referring to, is adding a test to ensure it stays that way (and so we can verify the failure behavior).

See e.g. https://github.com/aws/serverless-application-model/pull/2742 for an example. Basically add a template with `error_` prefix, then a corresponding `error_*.json` in the outputs with a `errorMessage` containing the expected error message.

aahung (approved):
thanks for addressing the comments!

Issue #2227, opened 2021-11-16
AWS SAM generates Event Rules names longer than 64 characters


danemauland:
### Description:
When creating an EventSource of type EventBridgeRule in a StateMachine resource definition, SAM generates a name for it by taking the StateMachine ID, and appending to it the EventSource ID and the string "StepFunctionsTarget". If this resulting name is over 64 characters, the stack update will fail. 

### Steps to reproduce:

Deploy a stack with a StateMachine resource that also defines an EventBridgeRule event source, where the length of the state machine ID and event bridge rule ID sum to at least 45, such as the following:

`"StateMachineIdWith30Characters": {
  "Type": "AWS::Serverless::StateMachine",
  "Properties": {
    "Events": {
      "EventIdWith23Characters": {
        "Type": "EventBridgeRule",
        "Properties": {
          "EventBusName": {
            "Fn::ImportValue": {
              "Fn::Sub": "${Environment}-ImproovyEventBusName"
            }
          },
          "InputPath": "$.detail",
          "Pattern": {
            "source": [
              "ImproovyWebHook"
            ],
            "detail-type": [
              "JustcallSms"
            ]
          }
        }
      }
    },
  },
},`

### Observed result:
1 validation error detected: Value 'StateMachineIdWith30CharactersEventIdWith23CharactersStepFunctionsTarget' at 'targets.1.member.id' failed  to satisfy constraint: Member must have length less than or equal to 64 (Service: AmazonCloudWatchEvents; Status Code: 400; Error Code: ValidationException; Request ID: ...; Proxy: null)

### Expected result:
SAM to either truncate the generated name or provide a way to override the generated name.

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

1. OS: Amazon Linux
2. `sam --version`: SAM CLI, version 1.31.0
3. AWS region: US-East-1

ScottKelly:
I just ran into this issue as well. In case anyone else has this same problem, I'm going to try to use the normal CloudFormation Events Rule to work around it.

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-events-rule.html

farski:
Probably this should have the same functionality available as Functions [got a while back](https://github.com/aws/serverless-application-model/pull/1747/files)

xazhao:
[...]

PR #2692, opened 2022-12-01
feat: Add target properties to state machine event source

xazhao:
### Issue #, if available
https://github.com/aws/serverless-application-model/issues/2227

### Checklist

    - [ ] Correct values
- [x] Update documentation

### Examples?


samtranslator/model/stepfunctions/events.py
```diff
@@ -28,6 +28,8 @@ class EventSource(ResourceMacro):
     # TODO: Make `EventSource` an abstract class and not giving `principal` initial value.
     principal: str = None  # type: ignore
 
+    Target: Optional[Dict[str, str]]
```
GavinZZ, line 31: What's the benefit of using a dict over a str? It seems that we only want a `target_id`.

samtranslator/schema/docs.json
```diff
@@ -482,7 +482,8 @@
...
-      "State": "The state of the rule\\.  \n*Accepted values:* `DISABLED | ENABLED`  \nSpecify either the `Enabled` or `State` property, but not both\\.\n*Type*: String  \n*Required*: No  \n*AWS CloudFormation compatibility*: This property is passed directly to the [`State`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-events-rule.html#cfn-events-rule-state) property of an `AWS::Events::Rule` resource\\."
+      "State": "The state of the rule\\.  \n*Accepted values:* `DISABLED | ENABLED`  \nSpecify either the `Enabled` or `State` property, but not both\\.\n*Type*: String  \n*Required*: No  \n*AWS CloudFormation compatibility*: This property is passed directly to the [`State`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-events-rule.html#cfn-events-rule-state) property of an `AWS::Events::Rule` resource\\.",
+      "Target": "Configures the AWS resource that EventBridge invokes when a rule is triggered. Currently it only supports `Id` property\\."
```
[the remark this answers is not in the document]
xazhao: Yeah I think you're right. Looks like our schema is bound tight with docs. I will discard changes in docs/schema.

samtranslator/model/stepfunctions/events.py
xazhao: It's to match what we have here: https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-property-function-target.html

GavinZZ (approved):

samtranslator/model/stepfunctions/events.py
```diff
@@ -173,6 +179,7 @@ class CloudWatchEvent(EventSource):
         "DeadLetterConfig": PropertyType(False, is_type(dict)),
         "RetryPolicy": PropertyType(False, is_type(dict)),
         "State": PropertyType(False, is_str()),
+        "Target": PropertyType(False, is_type(dict)),
```
aaythapa, line 182: In this CloudWatchEvent [docs](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-property-statemachine-statemachinecloudwatchevent.html) it says that "AWS SAM will add support for new features only to EventBridgeRule". Does that apply here?

xazhao: Hmm I didn't notice this comment. `CloudWatchEvent` and `EventBridgeRule` basically share the same code and it doesn't match what doc says... And this is more like a bug fix instead of a new feature. So I think it's fine to keep it here.

aaythapa: If we're pushing customers to use `EventBridgeRule` over `CloudWatchEvent` then I think not adding Target to `CloudWatchEvent` might be a good thing as it gives them another reason to not use it.

aaythapa (approved):

samtranslator/model/stepfunctions/events.py
hoffa, line 182: Should be `Property` or `PassThroughProperty`, see https://github.com/aws/serverless-application-model/blob/develop/DEVELOPMENT_GUIDE.md#development-guidelines

xazhao: Oh yes! Will update it

samtranslator/schema/aws_serverless_statemachine.py
```diff
@@ -97,6 +98,7 @@ class EventBridgeRuleEventProperties(BaseModel):
     InputPath: Optional[PassThrough] = eventbridgeruleeventproperties("InputPath")
     Pattern: Optional[PassThrough] = eventbridgeruleeventproperties("Pattern")
     RetryPolicy: Optional[PassThrough] = eventbridgeruleeventproperties("RetryPolicy")
+    Target: Optional[PassThrough]  # TODO: Add docs
```
hoffa, line 101: Property doesn't say it's pass-through.

samtranslator/model/stepfunctions/events.py
hoffa, line 31: We need to figure out some way so we could use the classes from the schema directly throughout the code. Not needed in this PR.

hoffa (approved):

samtranslator/model/stepfunctions/events.py
xazhao: Yeah that sounds like a good idea.

Issue #1112, opened 2019-08-30
Handle ParameterName value with or without '/' when creaing generating SSMParameterReadPolicy policy


whimzyLive:
**Description:**
When using `SSMParameterReadPolicy` policy template with `ParameterName` !Ref..ed to SSM resource `SSMParameterReadPolicy` will generate incorrect arn mapping.
For example:
```yml
  MyCredParameter:
    Type: AWS::SSM::Parameter
    Properties:
      Name: /prefix/environment/applications/mycred
      Type: String
      Value: !Sub |
        {
          "accessKeyId": "<Your accessKeyId>",
          "accessKeySecret": "<Your accessKeySecret>"
        }

  MyFunction:
    Type: AWS::Serverless::Function
    Properties:
      Handler: file-name.js
      Runtime: nodejs8.10
      CodeUri: dir
      Policies:
        - SSMParameterReadPolicy:
            ParameterName: !Ref MyCredParameter
      Events:
        SomeEvent:
          Type: SNS
          Properties:
            Topic: !Ref MyTopic
      Environment:
        Variables:
          MY_CRED: !Ref MyCredParameter
``` 

Here I am creating SSM resource and lambda function that will use this SSM parameter, as well as I am using SAM policy templates to allow `MyFunction` to allow read access to `MyCredParameter`.

This will deploy successfully, However, when Lambda tries to get value from SSM params with name process.env.MY_CRED, it will fail.

After some debugging I found out that, My lambda function role had permission to call `arn:aws:ssm:my-region:000000000:parameter//prefix/environment/applications/mycred` SSM parameter instead of `arn:aws:ssm:my-region:000000000:parameter/prefix/environment/applications/mycred`.
Notice that extra '/' after parameter in first arn.

**Observed result:**
This was because SAM policy templates uses `ParameterName` as is to create arn without doing any sort of validation. 
```json
{
"SSMParameterReadPolicy": {
      "Description": "Gives access to a parameter to load secrets in this account. If not using default key, KMSDecryptPolicy will also be needed.",
      "Parameters": {
        "ParameterName": {
          "Description":"The name of the secret stored in SSM in your account."
        }
      },
      "Definition": {
        "Statement": [
          {
            "Effect": "Allow",
            "Action": [
              "ssm:DescribeParameters"
            ],
            "Resource": "*"
          },
          {
            "Effect": "Allow",
            "Action": [
              "ssm:GetParameters",
              "ssm:GetParameter",
[...]

PR #2693, opened 2022-12-01
feat: Ssm paramater policy fix

aaythapa:
### Issue #, if available

#1112 

### Description of changes

The SSM parameter name can have a leading slash and not have one (as per the [docs](https://docs.aws.amazon.com/systems-manager/latest/userguide/sysman-paramstore-su-create.html#sysman-parameter-name-constraints)). Right now the SSMParamaeterReadPolicy doesn't account for names that have a leading slash. The resource arn is `arn:${AWS::Partition}:ssm:${AWS::Region}:${AWS::AccountId}:parameter/${parameterName}` which adds an extra slash if the `${parameterName}` has a leading slash. Can't change the policy directly as that would break compatibility. Added a new policy that customers can use when their parameter name has a leading slash. 

### Checklist

    - [ ] Correct values
- [ ] Update documentation

### Examples?


aaythapa:
Opened security consultation, waiting for approval

aaythapa:
Got security approval

hoffa (approved):

ConnorRobertson (approved):

xazhao:
Can we remove the extra slash if it is invalid?

aaythapa:
> Can we remove the extra slash if it is invalid?

Having a name with a slash and and without a slash are both valid so we need both policies

PR #2714, opened 2022-12-09
chore: retry 2 times with 30s wait if the stack deletion fails

xazhao:
### Checklist

    - [ ] Correct values
- [ ] Update documentation

### Examples?

### DIFF
--- a/integration/helpers/base_test.py
+++ b/integration/helpers/base_test.py
@@ -125,7 +125,7 @@ class BaseTest(TestCase):
         if self.stack_name:
             client = self.client_provider.cfn_client
             client.delete_stack(StackName=self.stack_name)
-            waiter = client.get_waiter('stack_delete_complete')
+            waiter = client.get_waiter("stack_delete_complete")
             waiter.wait(StackName=self.stack_name)
         if self.output_file_path and os.path.exists(self.output_file_path):
             os.remove(self.output_file_path)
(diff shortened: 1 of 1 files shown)


integration/helpers/base_test.py
```diff
@@ -115,9 +116,17 @@ def setUp(self):
         self.output_file_path = None
         self.sub_input_file_path = None
 
+    @retry(
+        stop=stop_after_attempt(3),
+        wait=wait_fixed(30),
```
aahung, line 121: If there is multiple tests running at the same time (assuming so) and they all throttle at the same time, will they throttle again when they wait for 30 seconds? Do we need to add some randomness or exponential to it?

aahung (approved):

integration/helpers/base_test.py
```diff
@@ -115,9 +116,17 @@ def setUp(self):
         self.output_file_path = None
         self.sub_input_file_path = None
 
+    @retry(
+        stop=stop_after_attempt(3),
```
hoffa, line 120:
```suggestion
        stop=stop_after_attempt(5),
```

xazhao: I pushed the same change at the same time lol

integration/helpers/base_test.py
xazhao: I assume the deployment time is different for different tests (not sure if the difference is big enough though).

hoffa (approved):

PR #2723, opened 2022-12-14
docs: add clause to backward compatibility guidelines

hoffa:
### Description of changes

- A template that has successfully deployed so far should continue to do so. If a change causes a template to no longer be deployable, it isn't backward compatible.
- Removed the "As rule of thumb"; it's all "guidelines" and the section ends with "There are some exceptions" anyway.

### Checklist

    - [ ] Using correct values
    - [ ] Using wrong values

### Examples?


GavinZZ (approved):

acristin (approved):
Good change. I might word it the opposite direction though, something along the lines of `No change may cause a previously valid template to become undeployable`, but that's a personal preference. As is, perfectly fine!

PR #2724, opened 2022-12-14
Add a cache on top of python27 hashing

acristin:
Profiling indicated that we spend a lot of time in our custom deepcopy, mostly due to maintaining compatibility with python2.7 hashes and their ordering of keys. If we change the ordering of keys, we will change logical IDs of some generated resources, breaking backwards compatibility.

We found that during the hashing code, we made some 13MM calls to Hash.hash, but with caching this is reduced to 1040 calls for a large test template.

On my laptop, this reduces wall clock time from ~15s to ~13s, or ~34s to ~27s with cProfiler.

### Checklist

    - [ ] Using correct values
    - [ ] Using wrong values

### Examples?

### DIFF
--- a/samtranslator/third_party/py27hash/hash.py
+++ b/samtranslator/third_party/py27hash/hash.py
@@ -7,19 +7,7 @@ This is designed for compatibility not performance.
 
 import ctypes
 import math
-
-# See https://github.com/python/mypy/issues/5107
-from typing import TYPE_CHECKING, TypeVar, Callable, Any
-
-if TYPE_CHECKING:
-    F = TypeVar("F", bound=Callable[[Any], Any])
-
-    def lru_cache(maxsize: int = 128, typed: bool = False) -> Callable[[F], F]:
-        pass
-
-
-else:
-    from functools import lru_cache
+from functools import lru_cache
 
 
 def hash27(value):  # type: ignore[no-untyped-def]
@@ -33,7 +21,7 @@ def hash27(value):  # type: ignore[no-untyped-def]
         Python 2.7 hash
     """
 
-    return Hash.hash(value)  # type: ignore[no-untyped-call]
+    return Hash.hash(value)
 
 
 class Hash(object):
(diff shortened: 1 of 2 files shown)


aahung (approved):

aahung:
on 2019 16-inch macbook pro, 42.8s -> 34.8s: 19% less time

samtranslator/third_party/py27hash/hash.py
```diff
@@ -28,6 +42,7 @@ class Hash(object):
     """
 
     @staticmethod
+    @lru_cache(maxsize=2048)
```
aahung, line 45: we should move the "str" check to the top as that's what hits the most

samtranslator/third_party/py27hash/hash.py
```diff
@@ -2,11 +2,25 @@
...
+from typing import TYPE_CHECKING, TypeVar, Callable, Any
+
+if TYPE_CHECKING:
+    F = TypeVar("F", bound=Callable[[Any], Any])
+
+    def lru_cache(maxsize: int = 128, typed: bool = False) -> Callable[[F], F]:
+        pass
+
+
+else:
```
hoffa, line 21: Any benefit versus `# type: ignoring` with a comment?

acristin: I was probably doing it wrong, but I was unable to make `# type:ignoring` work; I still got complaints about unused type annotations as in original commit.

samtranslator/third_party/py27hash/hash.py
```diff
@@ -2,11 +2,25 @@
 Compatibility methods to support Python 2.7 style hashing in Python 3.X+
 
 This is designed for compatibility not performance.
```
hoffa, line 4: 😏

samtranslator/third_party/py27hash/hash.py
hoffa: Ah ok :( Even when you put `# type: ignore` on the exact line that it complains about?

samtranslator/third_party/py27hash/hash.py
acristin: Yeah :/

hoffa (approved):

PR #2737, opened 2022-12-15
Make jwtConfiguration fields lower case

ssenchenko:
### Description of changes
`JwtConfiguration` fields from `Auth` from `Aws::Serverless::HttpApi` is attached to OpenApi body and therefore should be lower case. Currently, we do not make sure those fields are lowercase and create quite a confusion for users. 

### Checklist

    - [x] Using correct values
    - [x] Using wrong values
- [ ] Add/update [integration tests](https://github.com/aws/serverless-application-model/blob/develop/INTEGRATION_TESTS.md) - not required here

### Examples?


tests/translator/input/http_api_multiple_authorizers.yaml
```diff
@@ -62,9 +62,9 @@ Resources:
             - scope
             IdentitySource: $request.header.Authorization
             JwtConfiguration:
-              audience:
+              Audience:
```
hoffa, line 65: Why is the input file changed? It’s hard to gauge what about the transform is being changed, and whether existing templates transform the same.

samtranslator/model/apigatewayv2.py
```diff
@@ -299,3 +304,9 @@ def _get_reauthorize_every(self):  # type: ignore[no-untyped-def]
             return None
 
         return self.identity.get("ReauthorizeEvery")
+
+    @staticmethod
+    def _get_jwt_configuration(props: Optional[Dict[str, Union[str, List[str]]]]) -> Optional[JwtConfiguration]:
+        if not props:
+            return None
+        return {k.lower(): v for k, v in props.items()}
```
hoffa, line 312: Any reference that explains why this is needed?

ssenchenko: The issue in PR? Isn't it enough?

hoffa: So why is this specific to `JwtConfiguration`?

tests/translator/input/http_api_multiple_authorizers.yaml
ssenchenko: The issue is about removal the confusion created by the fact that `issuer` and `audience` should be lower case when everything else is capitalized. So the point is to change input, run the tests and see that the output has not changed. There is no change in output files in the PR, this is the proof that output stays the same with capitalized `Issuer` and `Audience`

hoffa: Ah okay I see, and just checked we have a bunch of non-changing tests with `audience`, so this is fine.

samtranslator/model/apigatewayv2.py
hoffa: If this only applies to `JwtConfiguration` because it's inconsistent with `AWS::ApiGatewayV2::Authorizer`, we should explain that. Maybe just a comment above saying e.g. "AWS::ApiGatewayV2::Authorizer uses uppercase, OpenAPI lowercase; make lowercase to avoid confusion" or something.

GavinZZ (approved):

samtranslator/model/apigatewayv2.py
xazhao: Actually I see this comment: 
Use lowercase issuer and audience properties with `AWS::Serverless::HttpApi` and uppercase Issuer and Audience properties with `AWS::ApiGatewayV2::Authorizer` here: https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-property-httpapi-oauth2authorizer.html#sam-httpapi-oauth2authorizer-jwtconfiguration

I'm just wondering if there was a specific reason to do this.

ssenchenko: The reason for the comment in the documentation? Because SAM-T provided not the best UX. Because of the way an authoriser is created. I know about the comment.

ssenchenko: @hoffa I've explained why we need to do that in the docstring

hoffa: Ok that's clearer. And there's no use case where the capitalized version has any purpose in `AWS::Serverless::HttpApi`?

hoffa (approved):

PR #2749, opened 2022-12-16
chore: adding types part 2

aaythapa:
### Checklist

    - [ ] Using correct values
    - [ ] Using wrong values

### Examples?


samtranslator/plugins/sam_plugins.py
```diff
@@ -45,13 +46,13 @@ class SamPlugins(object):
     set by the plugin. SAM translator will convert this into a nice error message and display to the user.
     """
 
-    def __init__(self, initial_plugins=None):  # type: ignore[no-untyped-def]
+    def __init__(self, initial_plugins: Optional[Any] = None) -> None:
```
hoffa, line 49: No better type than `Any`? Looking at where it's used, I see `isinstance(plugin, ServerlessAppPlugin)`.

samtranslator/intrinsics/resolver.py
```diff
@@ -46,7 +46,7 @@ def resolve_parameter_refs(self, _input: Any) -> Any:
         """
         return self._traverse(_input, self.parameters, self._try_resolve_parameter_refs)  # type: ignore[no-untyped-call]
 
-    def resolve_sam_resource_refs(self, input, supported_resource_refs):  # type: ignore[no-untyped-def]
+    def resolve_sam_resource_refs(self, input: Dict[str, Any], supported_resource_refs: Any) -> Any:
```
hoffa, line 49: What's the `SupportedResourceReferences` that it mentions just below?

The return value can be anything?

aaythapa: `SupportedResourceReferences` is a class in intrinsics. It wasn't imported in the file, is it worth importing for this type hint? 

Also the return value for this will be the return value from the self._traverse function. The self._traverse function modifies an input variable and returns it and the input can be "Any primitive type  (dict, array, string etc) whose value might contain an intrinsic function". Would that not be Any?

GavinZZ (approved):

samtranslator/plugins/sam_plugins.py
aaythapa: You're right it should be `Optional[List[BasePlugin]]`

samtranslator/intrinsics/resolver.py
hoffa: > SupportedResourceReferences is a class in intrinsics. It wasn't imported in the file, is it worth importing for this type hint?

I'd say yes. It's already imported in `translator.py`, so there's shouldn't be any practical impact.

> Would that not be Any?

Looks fine.

hoffa (approved):

aahung (approved):

PR #2750, opened 2022-12-16
Add typing and remove 'type: ignore' Part Three

GavinZZ:
### Checklist

    - [ ] Using correct values
    - [ ] Using wrong values

### Examples?


samtranslator/model/api/api_generator.py
```diff
@@ -326,7 +327,7 @@ def _add_endpoint_extension(self):  # type: ignore[no-untyped-def]
         editor.add_disable_execute_api_endpoint_extension(self.disable_execute_api_endpoint)  # type: ignore[no-untyped-call]
         self.definition_body = editor.swagger
 
-    def _construct_body_s3_dict(self):  # type: ignore[no-untyped-def]
+    def _construct_body_s3_dict(self) -> Dict[str, Any]:
```
aaythapa, line 330: Wouldn't it be `Dict[str, str]`?

samtranslator/model/api/api_generator.py
```diff
@@ -494,24 +500,26 @@ def _construct_api_domain(self, rest_api, route53_record_set_groups):  # type: i
...
+        basepath_value = self.domain.get("BasePath")
         # Create BasepathMappings
-        if self.domain.get("BasePath") and isinstance(self.domain.get("BasePath"), str):
-            basepaths = [self.domain.get("BasePath")]
-        elif self.domain.get("BasePath") and isinstance(self.domain.get("BasePath"), list):
-            basepaths = self.domain.get("BasePath")
+        if self.domain.get("BasePath") and isinstance(basepath_value, str):
+            basepaths = [basepath_value]
+        elif self.domain.get("BasePath") and isinstance(basepath_value, list):
+            basepaths = cast(Optional[List[str]], basepath_value)
```
aahung, line 509: nit: it is not necessarily list of str. It can be list of something else...

samtranslator/model/api/api_generator.py
```diff
@@ -826,22 +836,25 @@ def _construct_usage_plan(self, rest_api_stage=None):  # type: ignore[no-untyped
...
+            if usage_plan_properties.get("UsagePlanName"):
+                usage_plan.UsagePlanName = usage_plan_properties.get("UsagePlanName")
+            if usage_plan_properties.get("Description"):
+                usage_plan.Description = usage_plan_properties.get("Description")
+            if usage_plan_properties.get("Quota"):
+                usage_plan.Quota = usage_plan_properties.get("Quota")
+            if usage_plan_properties.get("Tags"):
+                usage_plan.Tags = usage_plan_properties.get("Tags")
+            if usage_plan_properties.get("Throttle"):
+                usage_plan.Throttle = usage_plan_properties.get("Throttle")
```
aahung, line 852: can we have a for loop to remove this mess...

aahung (approved):

samtranslator/model/api/api_generator.py
GavinZZ: s3_pointer is of type `Dict[str, Any]`. I can't guarantee the output of `s3_pointer[...]` is str right?
```
# Construct body_s3 as py27 dict
        body_s3 = Py27Dict()
        body_s3["Bucket"] = s3_pointer["Bucket"]
        body_s3["Key"] = s3_pointer["Key"]
        if "Version" in s3_pointer:
            body_s3["Version"] = s3_pointer["Version"]
        return body_s3
```

aaythapa: it does `s3_pointer["Key"] = Py27UniStr(s3_pointer["Key"])` in the lines above this code block. Doesn't that guarantee string?

GavinZZ: That's one scenario. There is another path 
```
if isinstance(self.definition_uri, dict):
            if not self.definition_uri.get("Bucket", None) or not self.definition_uri.get("Key", None):
                # DefinitionUri is a dictionary but does not contain Bucket or Key property
                    self.logical_id, "'DefinitionUri' requires Bucket and Key properties to be specified."
                )
            s3_pointer = self.definition_uri
```

aaythapa (approved):

PR #2752, opened 2022-12-16
chore: Make PropertyType.expected_type auto gen for dict and str

aahung:
### Checklist

    - [ ] Using correct values
    - [ ] Using wrong values

### Examples?


hoffa (approved):

samtranslator/model/__init__.py
```diff
@@ -5,16 +5,16 @@
...
 from samtranslator.plugins import LifeCycleEvents
 from samtranslator.model.tags.resource_tagging import get_tag_list
 
 
 class PropertyType(object):
     """Stores validation information for a CloudFormation resource property.
 
-    The argument "expected_type" is only used by InvalidResourcePropertyTypeException
-    to generate an error message. When it is not provided,
+    The arribute "expected_type" is only used by InvalidResourcePropertyTypeException
```
xazhao, line 16: `attribute`

GavinZZ (commented):
What's the benefit of this?

aahung:
> What's the benefit of this?

so we don't have to pass `expected_type = xxx` maually

samtranslator/model/__init__.py
aahung: lol

aahung: fixed

GavinZZ (approved):
Nice

hoffa:
> What's the benefit of this?

Related to previous PR that added `PropertyType.optional_dict()`. Which is confusing since technically a property could be both required and `Optional`; `optional_dict` isn't `Optional[dict]`.

samtranslator/model/__init__.py
```diff
@@ -5,16 +5,16 @@
 
 from samtranslator.intrinsics.resolver import IntrinsicsResolver
 from samtranslator.model.exceptions import ExpectedType, InvalidResourceException, InvalidResourcePropertyTypeException
```
xazhao, line 7: `ExpectedType` can be removed from import?

hoffa: If this is a valid comment, we should instead fix `pylint` so that it complains.

aahung: it is used at line 30

xazhao (approved):

PR #2753, opened 2022-12-19
fix: Raise correct exception when apigateway-endpoint-configuration is not a map

aahung:
…s not a map

### Checklist

    - [ ] Using correct values
    - [ ] Using wrong values

### Examples?


samtranslator/utils/utils.py
```diff
@@ -49,3 +49,21 @@ def dict_deep_get(d: Any, path: str) -> Optional[Any]:
...
+    Update the value deep in the dict.
+
+    If any value along the path doesn't exist, create empty dict along the way.
+    If any parent node exists but is not a dict, raise InvalidValueType.
+    """
+    relative_path = ""
+    _path_nodes = path.split(".")
+    while _path_nodes:
+        relative_path = (relative_path + f".{_path_nodes[0]}").lstrip(".")
+        d = d.setdefault(_path_nodes[0], {})
```
hoffa, line 65: What if `d` isn't a dict?

hoffa: And what if `_path_nodes` is empty?

samtranslator/utils/utils.py
```diff
@@ -49,3 +49,21 @@ def dict_deep_get(d: Any, path: str) -> Optional[Any]:
         relative_path = (relative_path + f".{_path_nodes[0]}").lstrip(".")
         _path_nodes = _path_nodes[1:]
     return d
+
+
+def dict_deep_update(d: Any, path: str, dict_to_merge: Dict[str, Any]) -> None:
+    """
+    Update the value deep in the dict.
+
+    If any value along the path doesn't exist, create empty dict along the way.
```
hoffa, line 58: I'd find it more intuitive if `dict_to_merge` included the "path", so it's always just a merge of two dicts. The `path` just adds confusion IMO.

aahung: interesting

aahung: Good idea, adopted

hoffa: The new version includes the literal `path` in the key?

I meant instead of:

```
merge(d, "foo.bar", e)
```

Do:

```
merge(d, {"foo": {"bar": e}})
```

aahung: this works better

aahung: wait, what you proposed is not flexible. It doesn't allow me to control down to which level I want to do a "replace" instead of "merge".

For example, let's say, I want to set the `a.b.c` to `{"Mode": "Flexible"}. If the original dict already has `a.b.c` set to `{"Disabled": True}`, that old key/value will not be removed

aahung: I am getting rid of this deep update thingy. It is not worth it

aahung: resolving this one as we don't `dict_deep_update` anymore

samtranslator/utils/utils.py
aahung: updated

samtranslator/utils/utils.py
```diff
@@ -49,3 +49,25 @@ def dict_deep_get(d: Any, path: str) -> Optional[Any]:
         relative_path = (relative_path + f".{_path_nodes[0]}").lstrip(".")
         _path_nodes = _path_nodes[1:]
     return d
+
+
+def dict_deep_set(d: Any, path: str, value: Any) -> None:
+    """
+    Get the value deep in the dict.
```
hoffa, line 56:
```suggestion
    Set the value deep in the dict.
```

Also I like this name better, makes much more sense. 👍 Wasn't aware of the merge/override thing that `path` dealt with.

hoffa (approved):

acristin (approved):

xazhao (approved):

PR #2759, opened 2022-12-21
S3ObjectTagging: allow Object Tagging actions in policy template for S3 Read/Write/Crud

mustafa-sadiq:
### Issue #, if available

None I can find

### Description of changes

Using boto3 to write S3 Object Tags using the put_object(Tagging='string') is not currently supported with the S3WritePolicy which does not allow the S3 Object Tagging actions. Users currently will have to write a managed policy to be able to use the S3 Object Tagging API.

I have added the three actions Put/Get/Delete to the respective policies:

S3CrudPolicy: PutObjectTagging, GetObjectTagging and DeleteObjectTagging
S3WritePolicy: PutObjectTagging
S3ReadPolicy: GetObjectTagging

The S3FullAccessPolicy already has all these actions allowed. 

policy_templates.json is still valid JSON `cat samtranslator/policy_templates_data/policy_templates.json | jq .`
verified S3 Object Tagging actions from 
https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutObjectTagging.html
https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetObjectTagging.html
https://docs.aws.amazon.com/AmazonS3/latest/API/API_DeleteObjectTagging.html

### Checklist

    - [ ] Using correct values
    - [ ] Using wrong values

### Examples?

### DIFF
--- a/samtranslator/policy_templates_data/policy_templates.json
+++ b/samtranslator/policy_templates_data/policy_templates.json
@@ -348,7 +348,8 @@
               "s3:ListBucket",
               "s3:GetBucketLocation",
               "s3:GetObjectVersion",
-              "s3:GetLifecycleConfiguration"
+              "s3:GetLifecycleConfiguration",
+              "s3:GetObjectTagging"
             ],
             "Resource": [
               {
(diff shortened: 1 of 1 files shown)


acristin:
Hi Mustafa,

Thank you for the request and PR. We appreciate that it is extra work to define managed policies if you want to use the tagging APIs, but because tags are often used for access control we believe that create, update and delete permissions should be explicitly defined when needed and not provided by default. 

We are open to adding `s3:GetObjectTagging` to the `S3ReadPolicy`, pending some internal discussion around policy templates.

Thanks again for taking the time to submit this PR, but we will not be able to accept it.

PR #2772, opened 2023-01-06
feat: Attach connectors to source

aaythapa:
### Description of changes

We want users to be able to attach their connector definition to the source property.

### Checklist

    - [ ] Using correct values
    - [ ] Using wrong values

### Examples?


samtranslator/translator/translator.py
```diff
@@ -269,6 +272,64 @@ def _get_resources_to_iterate(
...
+
+def add_embedded_connectors(sam_template: Dict[str, Any]) -> Any:
+    resources = sam_template.get("Resources", {})
+    connectors_list = []
+
+    # Loop through the resources in the template and see if any connectors have been attached
+    for source_logicalId, resource in resources.items():
+        if "Connectors" in resource:
+            if not isinstance(resource["Connectors"], dict):
+                raise make_error("'Connectors' must be a dict")
```
aahung, line 287: can we use `sam_expect` for this?

samtranslator/translator/translator.py
```diff
@@ -269,6 +272,64 @@ def _get_resources_to_iterate(
...
+
+
+def get_generated_connector(
+    source_logicalId: str, connector_logicalId: str, connector_dict: Dict[str, Any]
+) -> Tuple[str, Dict[str, Any]]:
+    connector = connector_dict
+    connector["Type"] = "AWS::Serverless::Connector"
+
+    # No need to raise an error for this instance as the error will be caught by the parser
+    if isinstance(connector["Properties"], dict):
```
aahung, line 320: nit: there are a lot of `connector["Properties"]` here. Can we use `properties = connector["Properties"]`?

samtranslator/translator/translator.py
```diff
@@ -269,6 +272,64 @@ def _get_resources_to_iterate(
...
+        resources[connector_logicalId] = connector_dict
+
+    return resources
+
+
+def get_generated_connector(
+    source_logicalId: str, connector_logicalId: str, connector_dict: Dict[str, Any]
+) -> Tuple[str, Dict[str, Any]]:
+    connector = connector_dict
+    connector["Type"] = "AWS::Serverless::Connector"
```
aahung, line 317: nit: find a constant with value `"AWS::Serverless::Connector"` somewhere and use it here.

samtranslator/translator/translator.py
```diff
@@ -269,6 +272,64 @@ def _get_resources_to_iterate(
...
+            )
+        resources[connector_logicalId] = connector_dict
+
+    return resources
+
+
+def get_generated_connector(
+    source_logicalId: str, connector_logicalId: str, connector_dict: Dict[str, Any]
+) -> Tuple[str, Dict[str, Any]]:
+    connector = connector_dict
```
aahung, line 316: this line seems out of place... can we just name the `connector_dict` as `connector` if you don't like the parameter?

samtranslator/translator/translator.py
```diff
@@ -269,6 +272,64 @@ def _get_resources_to_iterate(
...
+def make_error(msg: str) -> Exception:
+    return InvalidDocumentException([InvalidTemplateException(msg)])
+
+
+def add_embedded_connectors(sam_template: Dict[str, Any]) -> Any:
+    resources = sam_template.get("Resources", {})
+    connectors_list = []
+
+    # Loop through the resources in the template and see if any connectors have been attached
+    for source_logicalId, resource in resources.items():
```
aahung, line 284:
```suggestion
    for source_logical_id, resource in resources.items():
```

samtranslator/translator/translator.py
```diff
@@ -124,6 +124,9 @@ def translate(
 
         self.sam_parser.parse(sam_template=sam_template, parameter_values=parameter_values, sam_plugins=sam_plugins)
 
+        updated_resources = add_embedded_connectors(sam_template=sam_template)
```
aahung, line 127: > del resource["Connectors"]

> del connector["Properties"]["SourceReference"]

I am not a fan of mutating the input parameters without making it super clear. 

couple of possible ways to make it clearer:
1. rename the function and no need to return anything
2. return the connector resources only and have a separate function to strip away "Connectors:"

samtranslator/translator/translator.py
```diff
@@ -269,6 +272,64 @@ def _get_resources_to_iterate(
...
+            # delete connectors key, so it doesn't show up in the transformed template
+            del resource["Connectors"]
+
+    # go through the list of generated connectors and add it to resources dict
+    for connector_logicalId, connector_dict in connectors_list:
+        if connector_logicalId in resources:
+            raise make_error(
+                f"A resource with the id {connector_logicalId} already exists in the "
+                f"resource. Please use a different id for that resource.",
+            )
```
aahung, line 307: This check might not be effective: this happen before any serverless transform is done. Not having conflict at this point doesn't mean no conflict at later stage. We might want to consider:
1. generate the connector list right now but merge into the template at a later stage (after serverless transform except Serverless::Connector).
2. no other ideas yet.

samtranslator/translator/translator.py
```diff
@@ -269,6 +272,64 @@ def _get_resources_to_iterate(
...
+                connectors_list.append(data)
+
+            # delete connectors key, so it doesn't show up in the transformed template
+            del resource["Connectors"]
+
+    # go through the list of generated connectors and add it to resources dict
+    for connector_logicalId, connector_dict in connectors_list:
+        if connector_logicalId in resources:
+            raise make_error(
+                f"A resource with the id {connector_logicalId} already exists in the "
```
aahung, line 305: The message could be a bit more specific like "The logical ID of generated connector XXX has conflict with XXX of type XXX, please rename either of them."

samtranslator/translator/translator.py
```diff
@@ -269,6 +272,64 @@ def _get_resources_to_iterate(
...
+    connectors_list = []
+
+    # Loop through the resources in the template and see if any connectors have been attached
+    for source_logicalId, resource in resources.items():
+        if "Connectors" in resource:
+            if not isinstance(resource["Connectors"], dict):
+                raise make_error("'Connectors' must be a dict")
+
+            # Go through each of the connectors that have been attached and create a Serverless Connector resource
+            for connector_logicalId, connector_dict in resource["Connectors"].items():
```
aahung, line 290: Does this `connector_logicalId` needs to be unique in the whole template? It seems a bit hard to come up with different names for customers.

samtranslator/translator/translator.py
```diff
@@ -269,6 +272,64 @@ def _get_resources_to_iterate(
...
+        connector["Properties"]["Source"] = {"Id": source_logicalId}
+
+        if "SourceReference" in connector["Properties"]:
+            if not isinstance(connector["Properties"]["SourceReference"], dict):
+                raise make_error(f"The SourceReference property for {connector_logicalId} must be a dict")
+
+            connector["Properties"]["Source"].update(connector["Properties"]["SourceReference"])
+            del connector["Properties"]["SourceReference"]
+
+    return connector_logicalId, connector
```
xazhao, line 330: `connector_logicalId` is not needed in return?

samtranslator/translator/translator.py
```diff
@@ -269,6 +272,64 @@ def _get_resources_to_iterate(
...
+        if "Connectors" in resource:
+            if not isinstance(resource["Connectors"], dict):
+                raise make_error("'Connectors' must be a dict")
+
+            # Go through each of the connectors that have been attached and create a Serverless Connector resource
+            for connector_logicalId, connector_dict in resource["Connectors"].items():
+                if not isinstance(connector_dict, dict):
+                    raise make_error("Invalid 'Connectors' type. Must be a dict.")
+
+                data = get_generated_connector(source_logicalId, connector_logicalId, connector_dict)
```
xazhao, line 294: I feel using a list of tuple is overkill. A dict can do the same thing?

samtranslator/translator/translator.py
```diff
@@ -269,6 +272,64 @@ def _get_resources_to_iterate(
         return functions + statemachines + apis + others + connectors
 
 
+def make_error(msg: str) -> Exception:
```
xazhao, line 275: `make_error` looks very general. Maybe use a more specific name.

samtranslator/translator/translator.py
aaythapa: Yea it has to be unique. We decided to allow users to set the logical id for the generated connectors to take away any convoluted logical id generation and to make it easier for them to reference the connectors if they needed to.

samtranslator/translator/translator.py
aaythapa: Is it not ok to have a conflict at a later stage? 

For example if user has a serverless function "MyFunction" without the role defined and an attached connector called "MyFunctionRole" then it would conflict with the "MyFunctionRole" thats generated from the serverless function and it will show an error message asking the user to change the id of the connector. 

Is that not the right flow we want?

samtranslator/translator/translator.py
aaythapa: removed the return from the function. Any suggestions on what the function name should be changed to?

samtranslator/translator/translator.py
```diff
@@ -269,6 +272,80 @@ def _get_resources_to_iterate(
...
+    # Loop through the resources in the template and see if any connectors have been attached
+    for source_logical_id, resource in resources.items():
+        if "Connectors" in resource:
+            if not isinstance(resource["Connectors"], dict):
+                raise_error("'Connectors' must be a dict")
+
+            # Go through each of the connectors that have been attached and create a Serverless Connector resource
+            for connector_logical_id, connector_dict in resource["Connectors"].items():
+                if not isinstance(connector_dict, dict):
+                    raise_error(f"Invalid resource type. '{connector_logical_id}' be a dict.")
```
xazhao, line 298: must be a dict?

samtranslator/translator/translator.py
```diff
@@ -269,6 +272,80 @@ def _get_resources_to_iterate(
...
+    connector_class = SamConnector(logical_id="SamConnector")
+    connector["Type"] = connector_class.resource_type
+    properties = connector["Properties"]
+
+    # No need to raise an error for this instance as the error will be caught by the parser
+    if isinstance(properties, dict):
+        properties["Source"] = {"Id": source_logical_id}
+
+        if "SourceReference" in properties:
+            if not isinstance(properties["SourceReference"], dict):
```
hoffa, line 342: These are the properties complementing what's deduced from the source resource? If yes, that's not very clear from the name, which seems to suggest a fully-defined reference to the source resource. I don't have any good suggestions right now that's not mega verbose like `AdditionalSourceProperties`. 🤔

aaythapa: The UX team preferred `SourceReference` (at least they preferred it over using just `Source`) but we can change it to something clearer. Maybe `SourceResourceReference`? Customers should be familiar with the `ResourceReference` if they read the docs and adding Source as a prefix could minimize confusion.

samtranslator/schema/aws_serverless_api.py
```diff
@@ -217,6 +217,7 @@ class Resource(BaseModel):
     Properties: Properties
     Condition: Optional[PassThrough]
     DeletionPolicy: Optional[PassThrough]
+    Connectors: Optional[PassThrough]
```
hoffa, line 220: We support `Connectors` on non-SAM resources as well, so there might be a more generic way of supporting this property. Perhaps allow additional properties (using `LenientBaseModel`), and use another schema to only validate `Connectors` on any resource with `Type` of `AWS::*::*`. No need to do in this PR; from top of my head I'm not entirely sure how to do this. Perhaps the existing [`any_cfn_resource`](https://github.com/aws/serverless-application-model/blob/af76a9e9158dcb30a52df650c6587d1546e550a6/samtranslator/schema/any_cfn_resource.py#L7-L9) gives inspiration.

samtranslator/schema/aws_serverless_function.py
```diff
@@ -542,6 +542,7 @@ class Globals(BaseModel):
 class Resource(BaseModel):
     Type: Literal["AWS::Serverless::Function"]
     Properties: Optional[Properties]
+    Connectors: Optional[PassThrough]
```
hoffa, line 545: It's not `PassThrough`. We know the schema.

samtranslator/schema/aws_serverless_api.py
hoffa: So we should have [schema validation test](https://github.com/aws/serverless-application-model/blob/a4b4b9ac4e720107b4af3635de0a4fc99feb3ad7/tests/schema/test_validate_schema.py#L85-L94) that makes sure a non-SAM resource with `Connectors` and e.g. integer `Id` fails validation.

samtranslator/translator/translator.py
```diff
@@ -269,6 +274,83 @@ def _get_resources_to_iterate(
...
+                    connector_logical_id, connector_logical_id, resources[connector_logical_id]["Type"]
+                )
+            )
+        else:
+            resources[connector_logical_id] = connector_dict
+
+
+def get_generated_connector(
+    source_logical_id: str, connector_logical_id: str, connector: Dict[str, Any], document_errors: List[Any]
+) -> None:
```
hoffa, line 329: IMO would be clearer if this took the connector from the resource attribute as input, and returned the connector resource as output. Currently it mutates an input variable, making it hard to audit.

samtranslator/translator/translator.py
```diff
@@ -269,6 +274,83 @@ def _get_resources_to_iterate(
...
+    Adds those attached connectors as Connector resources to the list of resources
+
+    :param dict sam_template: SAM template
+    """
+    resources = sam_template.get("Resources", {})
+    connectors = {}
+
+    # Loop through the resources in the template and see if any connectors have been attached
+    for source_logical_id, resource in resources.items():
+        if "Connectors" in resource:
```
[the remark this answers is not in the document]
GavinZZ: +1. Less indentation and making it easier to read.
```
if "Connectors" not in resource:
   continue
```

tests/translator/input/embedded_connectors_api_to_function.yaml
```diff
@@ -0,0 +1,83 @@
...
+      Integration:
+        Type: MOCK
+      ResourceId: !GetAtt MyApiGateway.RootResourceId
+      RestApiId: !Ref MyApiGateway
+
+  MyApiGatewayV2:
+    Type: AWS::ApiGatewayV2::Api
+    Connectors:
+      MyConnectorApiV2ToLambda:
+        Properties:
```
hoffa, line 40: Why the extra level of identation? What value does `Properties` give?

samtranslator/translator/translator.py
```diff
@@ -269,6 +274,83 @@ def _get_resources_to_iterate(
...
+                    resource["Connectors"], source_logical_id, "Connectors", is_resource_attribute=True
+                ).to_be_a_map()
+
+                # Go through each of the connectors that have been attached and create a Serverless Connector resource
+                for connector_logical_id, connector_dict in resource["Connectors"].items():
+
+                    # The logical id for the transformed connector will be <source_logical_id> + <embedded_connector_logical_id>
+                    connector_logical_id = source_logical_id + connector_logical_id
+
+                    # can't use sam_expect since this is neither a property nor a resource attribute
```
GavinZZ, line 301: nit: Maybe it's just me, but it's great to have comments. However, I feel like there are too many comments that's kinda make the code readability a bit hard. Plus, these codes are pretty straightforward to understand without comments.

samtranslator/translator/translator.py
```diff
@@ -269,6 +274,83 @@ def _get_resources_to_iterate(
...
+                    if not isinstance(connector_dict, dict):
+                        raise InvalidResourceException(connector_logical_id, f"{connector_logical_id} should be a map.")
+
+                    # mutates the connector_dict to make it a connector resource
+                    get_generated_connector(source_logical_id, connector_logical_id, connector_dict, document_errors)
```
GavinZZ, line 306: nit: `get_generated_connector` implies it's going to return something, but doesn't seem to return anything in this case instead modify some of the input.

samtranslator/translator/translator.py
```diff
@@ -269,6 +274,83 @@ def _get_resources_to_iterate(
...
+    connector_class = SamConnector(logical_id="SamConnector")
+    connector["Type"] = connector_class.resource_type
+
+    # No need to raise an error for this instance as the error will be caught by the parser
+    if "Properties" in connector and isinstance(connector["Properties"], dict):
+        properties = connector["Properties"]
+        properties["Source"] = {"Id": source_logical_id}
+        if "SourceReference" in properties:
+            try:
+                sam_expect(properties["SourceReference"], connector_logical_id, "SourceReference").to_be_a_map()
```
hoffa, line 346: Not sure the error checking is worth it. It's not very thorough, and it'll be checked properly later anyway.

GavinZZ (commented):
Generally LGTM. Some minor comments.

xazhao:
Could you add a few integration tests to the PR? Maybe re-use the template, deploy and verify the permissions. Not necessary to cover all cases though.

samtranslator/translator/translator.py
```diff
@@ -269,6 +274,83 @@ def _get_resources_to_iterate(
...
+
+                properties["Source"].update(properties["SourceReference"])
+                del properties["SourceReference"]
```
awood45, line 349: We should be a bit more strongly typed here - theoretically I could override the Id, that shouldn't be possible.

awood45 (changes requested):

samtranslator/schema/common.py
```diff
@@ -59,3 +60,36 @@ class Config:
...
+    SourceReference: Optional[SourceReference]  #  TODO: add docs for SourceReference
+    Destination: ResourceReference = connector_properties("Destination")
+    Permissions: List[Literal["Read", "Write"]] = connector_properties("Permissions")
+
+
+class EmbeddedConnector(BaseModel):
+    Properties: EmbeddedConnectorProperties
+    DependsOn: Optional[PassThroughProp]
+    DeletionPolicy: Optional[PassThroughProp]
+    MetaData: Optional[PassThroughProp]
```
hoffa, line 94: `Metadata`?

samtranslator/schema/common.py
```diff
@@ -59,3 +60,36 @@ class Config:
 # https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-ref.html
 class Ref(BaseModel):
     Ref: str
+
+
+connector_properties = get_prop("sam-resource-connector")
+resourcereference = get_prop("sam-property-connector-resourcereference")
+
+
+class ResourceReference(BaseModel):
```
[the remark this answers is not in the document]
aaythapa: Would importing it from the connectors file be better?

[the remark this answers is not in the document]
aaythapa: Nvm, will cause errors. Going to put the embedded connectors class in the connectors file rather than `common.py`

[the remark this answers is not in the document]
hoffa: > Would importing it from the connectors file be better?

Yes, if they're the same the schema should reflect that.

samtranslator/schema/aws_serverless_connector.py
```diff
@@ -28,3 +28,21 @@ class Properties(BaseModel):
...
+
+
+class SourceReference(BaseModel):
+    Qualifier: Optional[PassThroughProp] = resourcereference("Qualifier")
+
+
+class EmbeddedConnectorProperties(BaseModel):
+    SourceReference: Optional[SourceReference]  # TODO: add docs for SourceReference
+    Destination: ResourceReference = properties("Destination")
+    Permissions: List[Literal["Read", "Write"]] = properties("Permissions")
```
hoffa, line 40: Should always be same as `Permissions` in regular connector, so better to refactor so both use the same.

tests/translator/input/embedded_connectors_esm_depends_on.yaml
```diff
@@ -0,0 +1,28 @@
...
+      MyConnector:
+        Properties:
+          Destination:
+            Id: MyFunction
+          Permissions:
+          - Read
+          - Write
+
+  MyEventSourceMapping:
+    DependsOn: MyQueueMyConnectorPolicy
```
hoffa, line 24: Should also have a test with `DependsOn` on the connector logical ID (verbatim). We automatically expand to the generated resource logical IDs:

https://github.com/aws/serverless-application-model/blob/b4f9ea1b083469a304ae248e26f3b2afb3a8b6f7/samtranslator/model/sam_resources.py#L1895-L1896

hoffa (approved):

samtranslator/translator/translator.py
```diff
@@ -268,6 +276,101 @@ def _get_resources_to_iterate(
 
         return functions + statemachines + apis + others + connectors
 
+    @staticmethod
+    def _update_resources(resources: Dict[str, Any], connectors_list: List[SamConnector]) -> None:
```
hoffa, line 280: Can make this pure (hence easier to reason and less error-prone) by returning the connectors as dict, and mutating at the caller level.

samtranslator/translator/translator.py
```diff
@@ -268,6 +276,101 @@ def _get_resources_to_iterate(
...
+        for connector in connectors_list:
+            resources.update(connector.to_dict())
+
+    @staticmethod
+    def _delete_connectors_attribute(resources: Dict[str, Any]) -> None:
+        for source_logical_id, resource in resources.items():
+            if "Connectors" not in resource:
+                continue
+            del resources[source_logical_id]["Connectors"]
```
hoffa, line 289: Note sure if this works:

```suggestion
            del resource["Connectors"]
```

samtranslator/translator/verify_logical_id.py
```diff
@@ -1,3 +1,6 @@
+from ..model import Resource
```
hoffa, line 1: nit: we tend to use full path (makes recognition a little quicker and doesn't need changing if moving the file)

samtranslator/translator/translator.py
```diff
@@ -268,6 +276,101 @@ def _get_resources_to_iterate(
...
+                        )
+                    connectors.append(generated_connector)
+                except (InvalidResourceException, DuplicateLogicalIdException) as e:
+                    self.document_errors.append(e)
+
+        return connectors
+
+    def _get_generated_connector(
+        self, source_logical_id: str, connector_logical_id: str, connector_dict: Dict[str, Any]
+    ) -> Any:
```
hoffa, line 342: Can be a `Resource`/`SamResource`.

samtranslator/schema/aws_serverless_function.py
```diff
@@ -547,6 +555,7 @@ class Globals(BaseModel):
 class Resource(BaseModel):
     Type: Literal["AWS::Serverless::Function"]
     Properties: Optional[Properties]
+    Connectors: Optional[Dict[str, EmbeddedConnector]]
```
hoffa, line 558: In a follow-up PR, let's make this part of all resources (currently it's only defined on SAM resources), along with a test that ensures schema validation fails on templates with wrong `Connectors` on non-SAM resource.

We can probably make `Resources` use `allOff` (1) lenient `Connectors` model for any resource, and (2) `anyOf` the current list of resources it validates.

hoffa (approved):

samtranslator/translator/translator.py
```diff
@@ -268,6 +278,103 @@ def _get_resources_to_iterate(
...
+                    properties.get("SourceReference"),
+                    connector_logical_id,
+                    f"{connector_logical_id}.Properties.SourceReference",
+                ).to_be_a_map()
+
+                # can't allow user to override the Id using SourceReference
+                if "Id" in properties["SourceReference"]:
+                    raise InvalidResourceException(connector_logical_id, "Id shouldn't be defined in 'SourceReference'")
+
+                properties["Source"].update(properties["SourceReference"])
```
hoffa, line 373: I thought we only allow `Qualifier`? None of the other properties should work alongside `Id` AFAIK.

awood45 (approved):

samtranslator/translator/translator.py
aaythapa: If they use anything other than `Qualifier` then it'll throw an error from the connectors class https://github.com/aws/serverless-application-model/blob/b4f9ea1b083469a304ae248e26f3b2afb3a8b6f7/samtranslator/model/connector/connector.py#L95

samtranslator/schema/aws_serverless_function.py
aaythapa: Added a TODO

tests/translator/input/embedded_connectors_depends_on_connector.yaml
```diff
@@ -0,0 +1,28 @@
...
+      InlineCode: |
+        exports.handler = async (event) => {
+          console.log(event);
+        };
+
+  MyQueue:
+    Type: AWS::SQS::Queue
+    Connectors:
+      MyConnector:
+        Properties:
```
hoffa, line 16: Missing tests with stuff next to `Properties`?

samtranslator/translator/translator.py
```diff
@@ -268,6 +278,103 @@ def _get_resources_to_iterate(
...
+            if "SourceReference" in properties:
+                sam_expect(
+                    properties.get("SourceReference"),
+                    connector_logical_id,
+                    f"{connector_logical_id}.Properties.SourceReference",
+                ).to_be_a_map()
+
+                # can't allow user to override the Id using SourceReference
+                if "Id" in properties["SourceReference"]:
+                    raise InvalidResourceException(connector_logical_id, "Id shouldn't be defined in 'SourceReference'")
```
hoffa, line 371: Is this tested?

hoffa, line 371:
```suggestion
                    raise InvalidResourceException(connector_logical_id, "'Id' shouldn't be defined in 'SourceReference'")
```

samtranslator/translator/translator.py
```diff
@@ -268,6 +278,103 @@ def _get_resources_to_iterate(
...
+                    generated_connector = self._get_generated_connector(
+                        source_logical_id,
+                        connector_logical_id,
+                        connector_dict,
+                    )
+
+                    if not verify_unique_logical_id(generated_connector, resources):
+                        raise DuplicateLogicalIdException(
+                            source_logical_id, connector_logical_id, generated_connector.resource_type
+                        )
```
hoffa, line 337: Missing test?

tests/translator/input/embedded_connectors_depends_on_connector.yaml
aaythapa: Will be part of another PR where connector can support resource attributes

xazhao (approved):

samtranslator/translator/translator.py
aaythapa: tests added

samtranslator/translator/translator.py
aaythapa: tests added

PR #2780, opened 2023-01-12
Remove not implemented methods

GavinZZ:
### Description of changes
Remove NotImplemented error and turn these classes into abstract classes.

### Checklist

    - [ ] Using correct values
    - [ ] Using wrong values

### Examples?

### DIFF
--- a/samtranslator/model/__init__.py
+++ b/samtranslator/model/__init__.py
@@ -371,7 +371,7 @@ class Resource(object, metaclass=ABCMeta):
         :return: Dictionary that will resolve to value of the attribute when CloudFormation stack update is executed
         """
         if attr_name not in self.runtime_attrs:
-            raise KeyError(f"{attr_name} attribute is not implemented for resource {self.resource_type}")
+            raise KeyError(f"{attr_name} attribute is not supported for resource {self.resource_type}")
 
         return self.runtime_attrs[attr_name](self)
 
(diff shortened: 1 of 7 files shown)


samtranslator/model/__init__.py
```diff
@@ -370,10 +370,10 @@ def get_runtime_attr(self, attr_name: str) -> Any:
 
         :return: Dictionary that will resolve to value of the attribute when CloudFormation stack update is executed
         """
+        if attr_name not in self.runtime_attrs:
+            raise KeyError(f"{attr_name} attribute is not supported for resource {self.resource_type}")
```
acristin, line 374: Why is KeyError correct?

aaythapa (approved):

acristin (approved):

samtranslator/model/__init__.py
acristin: Spoke offline; KeyError is already used in similar places, so this shouldn't cause confusion and maintains consistency.

GavinZZ: Put my comment here for visibility: `I definitely didn’t think NotImplemented is the correct exception type if runtime attr is not found. I was thinking either KeyError or RuntimeError. Since other methods in that class thrown KeyError when attributes are not found, I picked it as well.`

GavinZZ: Open for suggestion for a better alternative.

ssenchenko (approved):

PR #2787, opened 2023-01-14
chore: Adding more type hints

GavinZZ:
### Checklist

    - [ ] Using correct values
    - [ ] Using wrong values

### Examples?

### DIFF
--- a/samtranslator/intrinsics/actions.py
+++ b/samtranslator/intrinsics/actions.py
@@ -1,6 +1,6 @@
 import re
 from abc import ABC
-from typing import Any, Dict, Optional, Tuple, Callable, List
+from typing import Any, Dict, Optional, Tuple, Callable, List, Union
 
 from samtranslator.model.exceptions import InvalidTemplateException, InvalidDocumentException
 
@@ -331,7 +331,7 @@ class SubAction(Action):
 
         return input_dict
 
-    def _handle_sub_value(self, sub_value: Any, handler_method: Callable[[str, str], str]) -> Any:
+    def _handle_sub_value(self, sub_value: Union[str, List[str]], handler_method: Callable[[str, str], str]) -> Any:
         """
         Generic method to handle value to Fn::Sub key. We are interested in parsing the ${} syntaxes inside
         the string portion of the value.
@@ -517,7 +517,7 @@ class GetAttAction(Action):
         return True
 
     def _get_resolved_dictionary(
-        self, input_dict: Optional[Dict[Any, Any]], key: str, resolved_value: Optional[Any], remaining: List[str]
+        self, input_dict: Optional[Dict[str, Any]], key: str, resolved_value: Optional[str], remaining: List[str]
     ) -> Optional[Any]:
         """
         Resolves the function and returns the updated dictionary
--- a/samtranslator/intrinsics/actions.py
+++ b/samtranslator/intrinsics/actions.py
@@ -1,6 +1,6 @@
 import re
 from abc import ABC
-from typing import Any, Dict, Optional, Tuple, Callable, List, Union
+from typing import Any, Dict, Optional, Tuple, Callable, List
 
 from samtranslator.model.exceptions import InvalidTemplateException, InvalidDocumentException
 
@@ -331,7 +331,7 @@ class SubAction(Action):
 
         return input_dict
 
-    def _handle_sub_value(self, sub_value: Union[str, List[str]], handler_method: Callable[[str, str], str]) -> Any:
+    def _handle_sub_value(self, sub_value: Any, handler_method: Callable[[str, str], str]) -> Any:
         """
         Generic method to handle value to Fn::Sub key. We are interested in parsing the ${} syntaxes inside
         the string portion of the value.
--- a/samtranslator/intrinsics/actions.py
+++ b/samtranslator/intrinsics/actions.py
@@ -1,6 +1,6 @@
 import re
 from abc import ABC
-from typing import Any, Dict, Optional, Tuple
+from typing import Any, Dict, Optional, Tuple, Callable, List
 
 from samtranslator.model.exceptions import InvalidTemplateException, InvalidDocumentException
 
@@ -15,15 +15,7 @@ class Action(ABC):
     """
 
     _resource_ref_separator = "."
-
-    # Note(xinhol): `Action` should have been an abstract class. Disabling the type check for the next
-    # line to avoid any potential behavior change.
-    # TODO: Make `Action` an abstract class and not giving `intrinsic_name` initial value.
-    intrinsic_name: str = None  # type: ignore
-
-    def __init__(self) -> None:
-        if not self.intrinsic_name:
-            raise TypeError("Subclass must provide a intrinsic_name")
+    intrinsic_name: str
 
     def resolve_parameter_refs(self, input_dict: Optional[Any], parameters: Dict[str, Any]) -> Optional[Any]:
         """
@@ -185,7 +177,7 @@ class SubAction(Action):
         :return: Resolved
         """
 
-        def do_replacement(full_ref, prop_name):  # type: ignore[no-untyped-def]
+        def do_replacement(full_ref: str, prop_name: str) -> Any:
             """
             Replace parameter references with actual value. Return value of this method is directly replaces the
             reference structure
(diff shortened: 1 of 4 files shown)


samtranslator/intrinsics/resolver.py
```diff
@@ -194,13 +198,13 @@ def _try_resolve_sam_resource_refs(self, input, supported_resource_refs):  # typ
...
         """
-        if not self._is_intrinsic_dict(input):  # type: ignore[no-untyped-call]
+        if not self._is_intrinsic_dict(input):
             return input
 
         function_type = list(input.keys())[0]
         return self.supported_intrinsics[function_type].resolve_resource_refs(input, supported_resource_refs)
 
-    def _try_resolve_sam_resource_id_refs(self, input, supported_resource_id_refs):  # type: ignore[no-untyped-def]
+    def _try_resolve_sam_resource_id_refs(self, input: Any, supported_resource_id_refs: Dict[str, str]) -> Any:
```
aaythapa, line 207: `input` isn't a `dict`?

samtranslator/intrinsics/actions.py
```diff
@@ -315,9 +311,9 @@ def do_replacement(full_ref, ref_value):  # type: ignore[no-untyped-def]
             # syntax and retain other attributes. Ex: ${LogicalId.Property.Arn} => ${SomeOtherLogicalId.Arn}
             return full_ref.replace(logical_id, resolved_value)
 
-        return self._handle_sub_action(input_dict, do_replacement)  # type: ignore[no-untyped-call]
+        return self._handle_sub_action(input_dict, do_replacement)
 
-    def _handle_sub_action(self, input_dict, handler):  # type: ignore[no-untyped-def]
+    def _handle_sub_action(self, input_dict: Optional[Any], handler: Callable[[str, str], str]) -> Optional[Any]:
```
ssenchenko, line 316: Shouldn't it be then `input_dict: Optional[Dict[str, Any]]` rather than `Any`?

samtranslator/intrinsics/actions.py
```diff
@@ -522,7 +518,9 @@ def _check_input_value(self, value):  # type: ignore[no-untyped-def]
 
         return True
 
-    def _get_resolved_dictionary(self, input_dict, key, resolved_value, remaining):  # type: ignore[no-untyped-def]
+    def _get_resolved_dictionary(
+        self, input_dict: Optional[Any], key: str, resolved_value: Optional[Any], remaining: List[str]
```
aaythapa, line 522: `input_dict = Optional[Dict[Any, Any]]`?

samtranslator/intrinsics/actions.py
```diff
@@ -506,9 +502,9 @@ def resolve_resource_id_refs(
         remaining = splits[1:]  # if any
 
         resolved_value = supported_resource_id_refs.get(logical_id)
-        return self._get_resolved_dictionary(input_dict, key, resolved_value, remaining)  # type: ignore[no-untyped-call]
+        return self._get_resolved_dictionary(input_dict, key, resolved_value, remaining)
 
-    def _check_input_value(self, value):  # type: ignore[no-untyped-def]
+    def _check_input_value(self, value: Any) -> bool:
```
aaythapa, line 507: Comment below says `Value` must be an array

samtranslator/intrinsics/actions.py
```diff
@@ -333,11 +329,11 @@ def _handle_sub_action(self, input_dict, handler):  # type: ignore[no-untyped-de
         key = self.intrinsic_name
         sub_value = input_dict[key]
 
-        input_dict[key] = self._handle_sub_value(sub_value, handler)  # type: ignore[no-untyped-call]
+        input_dict[key] = self._handle_sub_value(sub_value, handler)
 
         return input_dict
 
-    def _handle_sub_value(self, sub_value, handler_method):  # type: ignore[no-untyped-def]
+    def _handle_sub_value(self, sub_value: Any, handler_method: Callable[[str, str], str]) -> Any:
```
ssenchenko, line 336: Is `sub_value` really `Any`? Maybe some `Union` will make sense?

samtranslator/intrinsics/actions.py
ssenchenko, line 507: The comment says that `value` must be an array. Why `Any` then?

ssenchenko (commented):
It feels like we're adding too much `Any` just to get rid of `#type: ignore`. I don't think it's what we should do. Some of those `Any` at least should be some Container[Any].
I made a couple of comments and there are more cases like that below.

samtranslator/intrinsics/actions.py
GavinZZ: I think the reason for the comment to say `Value` must be an array is because it does a type check in the function and if it's not `list`, we immediately return false.

GavinZZ: I think the reason for the comment to say Value must be an array is because it does a type check in the function and if it's not list, we immediately return false.

samtranslator/intrinsics/actions.py
```diff
@@ -333,11 +327,11 @@ def _handle_sub_action(self, input_dict, handler):  # type: ignore[no-untyped-de
...
```
ssenchenko, line 334: It seems to me that the function returns `Union[str, List[str]]`. Am I right? I think the purpose of this and similar functions in this PR is to transition from `Any` to 1 or 2 types

samtranslator/intrinsics/actions.py
```diff
@@ -506,9 +500,9 @@ def resolve_resource_id_refs(
...
 
         resolved_value = supported_resource_id_refs.get(logical_id)
-        return self._get_resolved_dictionary(input_dict, key, resolved_value, remaining)  # type: ignore[no-untyped-call]
+        return self._get_resolved_dictionary(input_dict, key, resolved_value, remaining)
 
-    def _check_input_value(self, value):  # type: ignore[no-untyped-def]
+    def _check_input_value(self, value: Any) -> bool:
         # Value must be an array with *at least* two elements. If not, this is invalid GetAtt syntax. We just pass along
         # the input to CFN for it to do the "official" validation.
         if not isinstance(value, list) or len(value) < 2:
```
ssenchenko, line 508: It seems like we can use here `sam_expect` like in this example https://github.com/aws/serverless-application-model/blob/09fcb574c3dbf2ab5e9ab30651fae21a29c91fed/samtranslator/model/apigateway.py#L182

samtranslator/intrinsics/actions.py
```diff
@@ -352,15 +346,15 @@ def _handle_sub_value(self, sub_value, handler_method):  # type: ignore[no-untyp
         # because that's the best we can do here.
         if isinstance(sub_value, str):
```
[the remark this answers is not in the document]
ssenchenko: See example below

samtranslator/intrinsics/actions.py
```diff
@@ -522,7 +516,9 @@ def _check_input_value(self, value):  # type: ignore[no-untyped-def]
 
         return True
 
-    def _get_resolved_dictionary(self, input_dict, key, resolved_value, remaining):  # type: ignore[no-untyped-def]
+    def _get_resolved_dictionary(
+        self, input_dict: Optional[Dict[Any, Any]], key: str, resolved_value: Optional[Any], remaining: List[str]
```
ssenchenko, line 520: Are sure that `input_dict` is not `Dict[str, Any]`? Also can `resolved_value` be `Optional[str]`?

samtranslator/intrinsics/actions.py
GavinZZ: How can you tell it's only `Union[str, List[str]]`? Isn't `Dict` also one of the viable type?

samtranslator/intrinsics/actions.py
GavinZZ: I think `sam_expect` is too strong here. It will throw an exception if type doesn't match. However, we do not want to throw an exception here.

samtranslator/intrinsics/actions.py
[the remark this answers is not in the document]
GavinZZ: I think sam_expect is too strong here. It will throw an exception if type doesn't match. However, we do not want to throw an exception here.

samtranslator/intrinsics/actions.py
GavinZZ: Good catch. It should be Dict[str, Any]

samtranslator/intrinsics/actions.py
aahung: wait, why it is not `Any` here? We expect this to be anything right? See test `test_must_skip_invalid_input_dict()`

GavinZZ: Good catch.

ssenchenko (approved):

aaythapa (approved):

PR #2789, opened 2023-01-16
Update sam_expect to be used for resource attributes

aaythapa:
### Issue #, if available

Add ability to use `sam_expect` on Resource attributes like Properties. Going to be used on the `Connectors` attribute too. Still needs some refactoring for it be used on things like DeletionPolicy and Type.

### Checklist

    - [ ] Using correct values
    - [ ] Using wrong values

### Examples?


samtranslator/model/exceptions.py
```diff
@@ -102,21 +102,40 @@ class InvalidResourcePropertyTypeException(InvalidResourceException):
...
+    ) -> None:
+        message = message or self._default_message(logical_id, key_path, expected_type)
+        super().__init__(logical_id, message)
+
+    @staticmethod
+    def _default_message(logical_id: str, key_path: str, expected_type: Optional[ExpectedType]) -> str:
+        if expected_type:
+            type_description, _ = expected_type.value
+            return f"Attribute '{key_path}' for resource '{logical_id}' should be a {type_description}."
+        return f"Type of attribute '{key_path}' for resource '{logical_id}'is invalid."
```
hoffa, line 138:
```suggestion
        return f"Type of attribute '{key_path}' for resource '{logical_id}' is invalid."
```

hoffa: Where will this be used?

aaythapa: For the `Connectors` attribute and for other attributes like `Type` and `DeletionPolicy`. Those other attributes are in the `resource.py` file which needs some refactoring for it to use `sam_expect` (I left a TODO there). Going to open another PR for that and do it later on.

samtranslator/parser/parser.py
```diff
@@ -41,16 +46,15 @@ def validate_datatypes(sam_template):  # type: ignore[no-untyped-def]
...
-                            "All 'Resources' must be Objects and have a 'Properties' Object. If "
-                            "you're using YAML, this may be an indentation issue.",
-                        )
-                    ]
+            try:
+                sam_expect(
+                    sam_resource.properties, resource_logical_id, "Properties", is_resource_attribute=True
+                ).to_be_a_map(
+                    "All 'Resources' must be Objects and have a 'Properties' Object. If "
+                    "you're using YAML, this may be an indentation issue."
```
aahung, line 54: This is too verbose IMO. Can we remove this custom message and leave it default?

samtranslator/model/exceptions.py
```diff
@@ -102,21 +102,40 @@ class InvalidResourcePropertyTypeException(InvalidResourceException):
...
+    ) -> None:
+        message = message or self._default_message(logical_id, key_path, expected_type)
+        super().__init__(logical_id, message)
+
+    @staticmethod
+    def _default_message(logical_id: str, key_path: str, expected_type: Optional[ExpectedType]) -> str:
+        if expected_type:
+            type_description, _ = expected_type.value
+            return f"Attribute '{key_path}' for resource '{logical_id}' should be a {type_description}."
+        return f"Type of attribute '{key_path}' for resource '{logical_id}' is invalid."
```
hoffa, line 138: Do we have to mention logical ID? In the test example the logical ID is already specified ("Resource with id [Function] is invalid.").

aaythapa: You're right, its not mentioned when resource property type is invalid so it shouldn't be mentioned here.

hoffa (approved):

aahung (approved):

PR #2793, opened 2023-01-16
feat: unified SAM/CloudFormation schema

hoffa:
### Description of changes

Adds new `unified.schema.json` that includes both SAM and CloudFormation resources.

We need control over the CloudFormation schema, so that e.g. we can allow [`additionalProperties`](https://json-schema.org/understanding-json-schema/reference/object.html#additional-properties) such as https://github.com/aws/serverless-application-model/pull/2772 on them.

Other notes:
- Eventually will rename `schema.json` to `sam.schema.json`, and `unified.schema.json` to `schema.json` so that existing users of the schema will automatically get the improvements.
- The CloudFormation schema doesn't support intrinsic functions (see https://github.com/aws/aws-toolkit-vscode/issues/2596).
- We should add some automation to automatically update the schema.

### Checklist

    - [x] Using correct values
    - [x] Using wrong values

### Examples?


ssenchenko (approved):

Makefile
```diff
@@ -44,15 +48,22 @@ lint:
 prepare-companion-stack:
 	pytest -v --no-cov integration/setup -m setup
 
-schema:
-	python samtranslator/schema/schema.py > samtranslator/schema/schema.json
+get-cfn-schema:
+	mkdir -p .tmp
+	test -f $(CFN_SCHEMA_PATH) || curl -o $(CFN_SCHEMA_PATH) https://raw.githubusercontent.com/awslabs/goformation/$(CFN_SCHEMA_VERSION)/schema/cloudformation.schema.json
```
awood45, line 53: This feels like a fragile dependency - is there a better, more persistent way to do this?

awood45 (changes requested):

acristin (approved):

aahung (approved):

Makefile
hoffa: Didn't want to include another 5 MB, but seems fair. And it simplifies upkeep somewhat.

awood45 (approved):

PR #2794, opened 2023-01-17
chore: Rewrite the AppConfig part to make it efficient

aahung:
### Issue #, if available

> This API action has been deprecated. Calls to receive configuration data should use the [StartConfigurationSession](https://docs.aws.amazon.com/appconfig/2019-10-09/APIReference/API_appconfigdata_StartConfigurationSession.html) and [GetLatestConfiguration](https://docs.aws.amazon.com/appconfig/2019-10-09/APIReference/API_appconfigdata_GetLatestConfiguration.html) APIs instead.
    GetConfiguration is a priced call. For more information, see [Pricing](https://aws.amazon.com/systems-manager/pricing/).
    AppConfig uses the value of the ClientConfigurationVersion parameter to identify the configuration version on your clients. If you don’t send ClientConfigurationVersion with each call to GetConfiguration , your clients receive the current configuration. You are charged each time your clients receive a configuration. **To avoid excess charges, we recommend you use the [StartConfigurationSession](https://docs.aws.amazon.com/appconfig/2019-10-09/APIReference/StartConfigurationSession.html) and [GetLatestConfiguration](https://docs.aws.amazon.com/appconfig/2019-10-09/APIReference/GetLatestConfiguration.html) APIs, which track the client configuration version on your behalf.** If you choose to continue using GetConfiguration , we recommend that you include the ClientConfigurationVersion value with every call to GetConfiguration . The value to use for ClientConfigurationVersion comes from the ConfigurationVersion attribute returned by GetConfiguration when there is new or updated data, and should be saved for subsequent calls to GetConfiguration .
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/appconfig.html#AppConfig.Client.get_configuration

### Checklist

    - [ ] Using correct values
    - [ ] Using wrong values

### Examples?

### DIFF
--- a/samtranslator/translator/translator.py
+++ b/samtranslator/translator/translator.py
@@ -111,7 +111,7 @@ class Translator:
         self.feature_toggle = (
             feature_toggle
             if feature_toggle
-            else FeatureToggle(FeatureToggleDefaultConfigProvider(), stage=None, account_id=None, region=None)  # type: ignore[no-untyped-call, no-untyped-call]
+            else FeatureToggle(FeatureToggleDefaultConfigProvider(), stage=None, account_id=None, region=None)
         )
         self.function_names: Dict[Any, Any] = {}
         self.redeploy_restapi_parameters = {}
--- a/samtranslator/feature_toggle/feature_toggle.py
+++ b/samtranslator/feature_toggle/feature_toggle.py
@@ -1,12 +1,17 @@
 import json
+import time
 from abc import ABC, abstractmethod
-from typing import Any, Dict, cast
+from typing import Any, Dict, Optional, Type, cast
 
 import boto3
 import logging
 
+from botocore.client import BaseClient
 from botocore.config import Config
+from botocore.exceptions import ClientError
+
 from samtranslator.feature_toggle.dialup import (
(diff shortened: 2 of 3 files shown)


samtranslator/translator/translator.py
```diff
@@ -111,7 +111,7 @@ def translate(
         self.feature_toggle = (
             feature_toggle
             if feature_toggle
-            else FeatureToggle(FeatureToggleDefaultConfigProvider(), stage=None, account_id=None, region=None)  # type: ignore[no-untyped-call, no-untyped-call]
+            else FeatureToggle(FeatureToggleDefaultConfigProvider(), stage=None, account_id=None, region=None)
```
marekaiv, line 114: Does this actually fix the problem?  Since we construct `FeatureToggleAppConfigConfigProvider` every time `translate` is invoked, and `FeatureToggleAppConfigConfigProvider.__init__` invokes `app_config_data_client.start_configuration_session`, it seems like we'd still be retrieving the config on every request.  Or am I misreading this?

aahung: move discussion to offline

Issue #2292, opened 2022-01-14
Support both verbose and compact resource tag structures


farski:
The tagging structure that SAM tends to use (`{my_key: my_value}`) is significantly less flexible than the format than the format that many CloudFormation templates use (`[{ Key: my_key, Value: my_value }]`), and it would be great if SAM allowed for using the older, more verbose format as well.

I understand and appreciate that SAM generally tries to simplify the deployment of many complex resources, and allows for writing less code in many cases. In this case, though, since tagging in inherently a function designed for user-defined customization, the lack of choice to use the more verbose structure conflicts with that design intent of tagging.

As an example of where this creates an issue, there are times where I create templates and I want some tag key and tag value to be user-defined through stack parameters. Using the verbose structure of tagging, this very easy to achieve (`[{ Key: !Ref TagKey Value: !Ref TagValue }]`. Using the more compact structure, there's no way to use `Ref` to define the tag key.

It's certainly possible in these cases to replace all SAM resources with their CloudFormation counterparts, but it's strange to have to give up all the benefits and time savings associated with SAM, just to achieve something so basic with tagging.

Given that, in nearly all cases, SAM has to translate from the compact structure to the verbose structure to create the underlying resources during transformation, it would be great if it could identify when the `Tags` property is an array rather than a map, and simply pass the tags along without transforming them in that case.

hawflau:
Thanks @farski  for raising the feature request. I think your points about using `!Ref` in tag keys are valid. 

We try to take in popular feature requests based on 👍🏽 from the community. We will be adding this to our intake for prioritization.

ssenchenko:
That linked PR doesn't have anything to do with supporting verbose tag structures. Stating it here so it won't confuse the community.

PR #2803, opened 2023-01-23
feat: create high level AppSync::GraphQLApi properties mapping

pradhapanr:
### Description of changes

Created a new resource AWS::Serverless::GraphQLApi, mapping properties to AWS::AppSync::GraphQLApi CFN resource.

Current properties include:
- Name
- Tags
- AuthenticationType
- XrayEnabled

Created schema changes and ran against new transform tests.

### Checklist

    - [x] Using correct values
    - [x] Using wrong values

### Examples?


samtranslator/model/appsync.py
```diff
@@ -0,0 +1,24 @@
...
+    Type: str
+
+
+class GraphQLApi(Resource):
+    resource_type = "AWS::AppSync::GraphQLApi"
+    property_types = {
+        "Name": PropertyType(True, IS_STR),
+        "Tags": PropertyType(False, list_of(IS_DICT)),
+        "XrayEnabled": PropertyType(False, is_type(bool)),
+        "AuthenticationType": PropertyType(True, IS_STR),
```
GavinZZ, line 18: This doesn't look complete to me. Any reason we aren't adding all the properties? https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-appsync-graphqlapi.html

samtranslator/model/sam_resources.py
```diff
@@ -2064,3 +2065,40 @@ def _add_connector_metadata(
...
+    def to_cloudformation(self, **kwargs: Any) -> List[Resource]:  # type: ignore
+        appsync_api = self._construct_appsync_api()
+        resources: List[Resource] = [appsync_api]
+
+        return resources
+
+    def _construct_appsync_api(self) -> GraphQLApi:
+        api = GraphQLApi(logical_id=self.logical_id, depends_on=self.depends_on, attributes=self.resource_attributes)
+
+        if "Type" not in self.Auth:
```
ssenchenko, line 2096: We need more resilient solution. We don't have mandatory vs optional fields in the TypedDict, do we?

samtranslator/schema/common.py
```diff
@@ -59,3 +59,8 @@ class Config:
 # https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-ref.html
 class Ref(BaseModel):
     Ref: str
+
+
+class Tags(BaseModel):
```
GavinZZ, line 64: Nice, we can use `Tags` in other schema files.

ssenchenko (changes requested):

samtranslator/schema/common.py
ssenchenko: That's the point! I suggested, Prad makes a separate PR for that

samtranslator/model/appsync.py
ssenchenko: It's not. complete obviously. The branch is work-in-progress

pradhapanr: This is just a very high level mapping to start with the most basic necessary properties. I think it's best to move slowly with this as we may refactor other parts of the design, but these core properties will certainly remain.

Also, some of the configuration properties can get a bit complex so I think it's better to take it slow here and add functionality in small pieces.

GavinZZ: Makes sense. Thanks for the clarifying.

samtranslator/model/appsync.py
```diff
@@ -0,0 +1,24 @@
...
+    }
+
+    Name: str
+    AuthenticationType: str
+    Tags: Optional[List[Dict[str, Any]]]
```
hoffa, line 23: Not using the `Tags` class?

samtranslator/schema/common.py
hoffa, line 64: This is a single tag.

samtranslator/model/appsync.py
hoffa: Never mind, thought this was in `schema`.

samtranslator/schema/common.py
hoffa: Where else is this format used? Typically it's a tag key-value map, here it's a list of maps.

pradhapanr: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-appsync-graphqlapi.html

It's listed as a list of tag objects here.

hoffa: I mean, why is this in `common`? Which other resources will benefit from this?

pradhapanr: Both [lambda functions](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-lambda-function.html) and [state machines](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-stepfunctions-statemachine.html) (TagEntrys instead of Tags but they're equivalent shapes) are using lists. I'm not sure why it's split as I can see here that [ApiGateway](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-stepfunctions-statemachine.html ) uses just a dictionary, but I think this will be useful for other resources as we choose to add them or modify typings for existing ones.

samtranslator/model/sam_resources.py
pradhapanr: After some offline discussion, we decided that we will remove these type checks and create a custom validator to have a more robust solution for checking valid properties are inputted by the user. This will be done in an upcoming PR.

samtranslator/schema/common.py
hoffa: Those are for CFN resources. The schema here is for SAM resources. If that format is not used commonly for SAM resources, it shouldn't be in `common.py`.

pradhapanr: Is it not more sensible to have a one to one mapping from SAM to CFN for these cases? It's minor but it sounds like bad UX to me for a customer to change the typing for how they define `Tags` if they are coming from CFN.

pradhapanr: But also, having a lack of consistency for what is essentially the same property across different resources is also bad UX...

hoffa: Perhaps. My point is mainly just that `Tags` should be in `aws_serverless_graphqlapi.py` unless there are good reasons not to. No SAM resource currently uses lists for tags, so I don't see why this is in `common.py`.

As for which format for tags to use, I'd probably lean toward use whatever CFN uses, as this also gives intrinsics support for keys. But it would set a new precedent, which is probably fine.

pradhapanr: That is fair. I'll move the `Tags` class to `aws_serverless_graphqlapi.py` for now. We can discuss if we'd like to change our formatting for `Tags` in our SAM resources to match CFN later and we can make any appropriate adjustments based on the outcome.

samtranslator/schema/aws_serverless_graphqlapi.py
```diff
@@ -0,0 +1,28 @@
+from typing import Optional, List
+
+from typing_extensions import Literal
+
+from samtranslator.schema.common import BaseModel, get_prop
+
+properties = get_prop("sam-resource-graphqlapi")
+
+# TODO: add docs
+class Tags(BaseModel):
```
hoffa, line 10: `Tag`?

samtranslator/schema/common.py
ssenchenko: @pradhapanr I think @hoffa raised a good point. List of dictionaries of 2 fields doesn't seem to be an example of great UX. Besides, it will go against UX with other SAM resources. I think it's better to have tags as a simple Dict for SAM resource and then transform it into `[{"key": "xxx", "value": "yyy"},]` for CFN

pradhapanr: Yeah, so I talked a bit more with @hoffa and it makes sense not to break precedent just for this one resource. There are [reasons](https://github.com/aws/serverless-application-model/issues/2292) to look into using the other way of declaring tags, or possibly supporting both. But that would require more discussion.

I'm going to create another commit with the dictionary typing and adjustments where necessary.

ssenchenko (approved):

GavinZZ (approved):

samtranslator/model/appsync.py
hoffa, line 18: Is any of these pass-through?

samtranslator/model/sam_resources.py
```diff
@@ -2064,3 +2065,39 @@ def _add_connector_metadata(
...
+class SamGraphQLApi(SamResourceMacro):
+    """SAM GraphQL API Macro (WIP)."""
+
+    resource_type = "AWS::Serverless::GraphQLApi"
+    property_types = {
+        "Name": PropertyType(False, IS_STR),
+        "Tags": PropertyType(False, IS_DICT),
+        "XrayEnabled": PropertyType(False, is_type(bool)),
+        "Auth": PropertyType(True, IS_DICT),
+    }
```
hoffa, line 2079: If these are pass-throughs, should use `PassThroughProperty`:

https://github.com/aws/serverless-application-model/blob/96691c1b9e8d7133f72f8c9c95ecc1098e3d96f4/samtranslator/model/__init__.py#L57-L65

Also `PropertyType` is deprecated: https://github.com/aws/serverless-application-model/blob/develop/DEVELOPMENT_GUIDE.md#development-guidelines and:

https://github.com/aws/serverless-application-model/blob/c39c2807bbf327255de8abed8b8150b18c60f053/samtranslator/model/__init__.py#L16

pradhapanr: Thanks! Did not realize that PropertyType was deprecated. Also that reminds me that `Tags` is technically no longer a PassThrough, but i will change `XrayEnabled` to match that.

samtranslator/model/sam_resources.py
```diff
@@ -2064,3 +2066,39 @@ def _add_connector_metadata(
...
+    def to_cloudformation(self, **kwargs: Any) -> List[Resource]:  # type: ignore
+        appsync_api = self._construct_appsync_api()
+        resources: List[Resource] = [appsync_api]
+
+        return resources
+
+    def _construct_appsync_api(self) -> GraphQLApi:
+        api = GraphQLApi(logical_id=self.logical_id, depends_on=self.depends_on, attributes=self.resource_attributes)
+
+        api.AuthenticationType = self.Auth["Type"]
```
hoffa, line 2097: How does it fail when there's no `Type`? Want to make sure it's a customer error, and not a `KeyError`.

pradhapanr: Me and Slava talked about this [here](https://github.com/aws/serverless-application-model/pull/2803#discussion_r1084744398), we're going to make a more resilient solution to catch errors in the `Auth` object in a followup PR.

hoffa (approved):

PR #2816, opened 2023-01-25
feat: CloudFormation docs in schema

hoffa:
### Description of changes

> Full diff here: https://github.com/aws/serverless-application-model/pull/2816.diff

Adds CloudFormation docs to schema:

Screen Shot 2023-01-26 at 13 39 02

Currently uses heuristics to guess the [documentation page](https://github.com/awsdocs/aws-cloudformation-user-guide/tree/main/doc_source) matching `definitions` in the CloudFormation JSON schema.

It works reasonably well (e.g. `AWS::EC2::CapacityReservationFleet.TagSpecification` corresponds to [`aws-properties-ec2-capacityreservation-tagspecification.md`](https://github.com/awsdocs/aws-cloudformation-user-guide/blob/main/doc_source/aws-properties-ec2-capacityreservation-tagspecification.md)), but there are exceptions such as `AWS::EC2::Instance.Ebs` which correspond to [`aws-properties-ec2-blockdev-template.md`](https://github.com/awsdocs/aws-cloudformation-user-guide/blob/main/doc_source/aws-properties-ec2-blockdev-template.md). 🤔 

There are probably more robust ways of cross-referencing, without depending on a specific format [as provided by GoFormation](https://github.com/awslabs/goformation/blob/master/schema/cloudformation.schema.json), and without heuristics (e.g. using the [CloudFormation resource specification](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-resource-specification.html), which includes a `Documentation` URL which has the correct slug).

The decompressed sizes are relatively large (7 to 11 MB), but all compress nicely (3 MB in total gzipped).

Checked diff and tested sample templates in VS Code.

### Checklist

    - [ ] Using correct values
    - [ ] Using wrong values

### Examples?


bin/add_docs_cfn_schema.py
```diff
@@ -0,0 +1,88 @@
...
+    )
+    args = parser.parse_args()
+
+    schema = json.loads(args.schema.read_text())
+    docs = json.loads(args.docs.read_text())["properties"]
+
+    # Assumes schema is from GoFormation and has consistent structure
+    # TODO: Use a more generic walk
+    # TODO: Use a more robust cross-referencing method than guessing slugs, doesn't always work
+    #       e.g. AWS::EC2::Instance.Ebs corresponds to aws-properties-ec2-blockdev-template
```
acristin, line 58: Do we need to special case known exceptions?

Makefile
```diff
@@ -47,10 +47,22 @@ prepare-companion-stack:
 
 update-schema-data:
 	mkdir -p .tmp
+
+	# Update and parse SAM docs
 	rm -rf .tmp/aws-sam-developer-guide
-	git clone https://github.com/awsdocs/aws-sam-developer-guide.git .tmp/aws-sam-developer-guide
+	git clone --depth 1 https://github.com/awsdocs/aws-sam-developer-guide.git .tmp/aws-sam-developer-guide
```
GavinZZ, line 53: Why is the `--depth 1` needed here?

hoffa: Faster pull. No need for whole history.

GavinZZ (approved):

bin/add_docs_cfn_schema.py
hoffa: There are more robust ways of doing the cross-referencing, probably using the [CloudFormation resource specification](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-resource-specification.html) which includes a `Documentation` link (that seemingly matches the filename in https://github.com/awsdocs/aws-cloudformation-user-guide/tree/main/doc_source) from each property that we can use.

This PR is mostly "good enough" to get the initial iteration out. There are improvements to the process I want to look at, e.g. whether the GoFormation schema is adequate (or whether we should use https://github.com/aws-cloudformation/cloudformation-template-schema or the source [resource provider schemas](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/resource-type-schemas.html)), so don't want to overfit to the current assumptions just yet.

acristin (approved):

ssenchenko (approved):

bin/add_docs_cfn_schema.py
```diff
@@ -0,0 +1,88 @@
...
+    for def_name, def_schema in schema["definitions"].items():
+        if not def_name.startswith("AWS::"):
+            log(f"Skipping {def_name}: does not start with AWS::")
+            continue
+        # If e.g. AWS::S3::Bucket, we only look under Properties
+        # TODO: Support resource attributes et al.
```
aaythapa, line 64: et al?

hoffa: e.g. `Properties` itself, or `Type` and whatnot

acristin:
```
Et al. comes from the Latin phrase meaning “and others.” It is usually styled with a period, but you will occasionally see et al as well. 
```
https://www.merriam-webster.com/dictionary/et%20al. for more details :)

aaythapa (approved):

PR #2822, opened 2023-01-26
feat: Allow resource attributes to be used on connectors

aaythapa:
### Description of changes

As of now none of the [resource attributes](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-specification-resource-attributes.html) supported by SAM can be used with connector resources. Also updated the schema for connectors + other SAM resources with the right resource attributes

### Checklist

    - [ ] Using correct values
    - [ ] Using wrong values

### Examples?


samtranslator/schema/aws_serverless_connector.py
```diff
@@ -48,4 +53,5 @@ class EmbeddedConnector(BaseModel):
     DependsOn: Optional[PassThroughProp]
     DeletionPolicy: Optional[PassThroughProp]
     Metadata: Optional[PassThroughProp]
-    UpdatePolicy: Optional[PassThroughProp]
```
acristin, line 51: Forgive me if this is a dumb question, but could renaming/removing this break any existing templates?

samtranslator/schema/aws_serverless_httpapi.py
```diff
@@ -141,6 +141,9 @@ class Globals(BaseModel):
 class Resource(BaseModel):
     Type: Literal["AWS::Serverless::HttpApi"]
     Properties: Optional[Properties]
-    Metadata: Optional[PassThroughProp]
     Condition: Optional[PassThroughProp]
     Connectors: Optional[Dict[str, EmbeddedConnector]]
+    UpdateReplacePolicy: Optional[PassThroughProp]
```
hoffa, line 146: nit: if all of them have the same resource attributes, we could have a common class that online defines the resource attributes, and then `Resource`s inherit from it; less typing/risk

samtranslator/schema/aws_serverless_connector.py
aaythapa: It shouldn't since its the schema

samtranslator/schema/aws_serverless_connector.py
```diff
@@ -30,6 +30,11 @@ class Properties(BaseModel):
 class Resource(BaseModel):
     Type: Literal["AWS::Serverless::Connector"]
     Properties: Properties
+    DependsOn: Optional[PassThroughProp]
+    DeletionPolicy: Optional[PassThroughProp]
+    Metadata: Optional[PassThroughProp]
+    UpdateReplacePolicy: Optional[PassThroughProp]
+    Condition: Optional[PassThroughProp]
```
GavinZZ, line 37: Can we move these properties into a common class so that we don't have to duplicate these?

GavinZZ (approved):
One nit comment, LGTM otherwise.

hoffa (approved):

aaythapa:
Removed the tests using `UpdatePolicy` on a Serverless API resource since according [this](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-updatepolicy.html) you can use `UpdatePolicy` on only select resources and none of the generated resources from Serverless API can use them

PR #2838, opened 2023-01-31
feat: New opt-in property to update Lambda version whenever a property changes

aaythapa:
### Issue #, if available

https://github.com/aws/serverless-application-model/issues/413

### Description of changes

Add a new opt in property that will update a lambda version whenever a property in the transformed lambda function changes. 

Used integrations tests. Pushed a template with `AutoPublishAliasAllProperties` set to `True`, then changed the handler property and made sure a new lambda version was created

### Checklist

    - [ ] Using correct values
    - [ ] Using wrong values

### Examples?


ConnorRobertson (approved):
LGTM

schema_source/aws_serverless_function.py
```diff
@@ -448,6 +448,7 @@ class ScheduleV2Event(BaseModel):
 KmsKeyArn = Optional[PassThroughProp]
 Layers = Optional[PassThroughProp]
 AutoPublishAlias = Optional[SamIntrinsicable[str]]
+AutoPublishAliasAdditionalProperties = Optional[SamIntrinsicable[bool]]
```
hoffa, line 451: Since it's now a boolean, perhaps `AutoPublishAliasAllProperties` or something similar would be clearer? Currently it's not obvious what "additional" refers to.

samtranslator/model/sam_resources.py
```diff
@@ -121,6 +121,7 @@ class SamFunction(SamResourceMacro):
         # Intrinsic functions in value of Alias property are not supported, yet
         "AutoPublishAlias": PropertyType(False, one_of(IS_STR)),
         "AutoPublishCodeSha256": PropertyType(False, one_of(IS_STR)),
+        "AutoPublishAliasAdditionalProperties": PropertyType(False, is_type(bool)),
```
hoffa, line 124: We shouldn't use `PropertyType`: https://github.com/aws/serverless-application-model/blob/develop/DEVELOPMENT_GUIDE.md#development-guidelines

Intrinsic resolution is no longer handled by the transform, and `PropertyType` introduces sneaky bugs like in this case. For example, if a customer passes an `!If` that should resolve to false, it'll behave as if it was true anyway (because any non-empty dict is truthy).

samtranslator/model/sam_resources.py
```diff
@@ -161,6 +162,7 @@ class SamFunction(SamResourceMacro):
     EphemeralStorage: Optional[Dict[str, Any]]
     AutoPublishAlias: Optional[Intrinsicable[str]]
     AutoPublishCodeSha256: Optional[Intrinsicable[str]]
+    AutoPublishAliasAdditionalProperties: Optional[Intrinsicable[bool]]
```
hoffa, line 165: Don't resolve intrinsics on new properties, it's handled by `AWS::LanguageExtensions`. So this isn't `Intrinsicable`.

schema_source/aws_serverless_function.py
hoffa: Also same comment about intrinsics, this shouldn't support them explicitly.

ssenchenko (approved):

schema_source/aws_serverless_function.py
aaythapa: Yea I like `AutoPublishAliasAllProperties` more since they're not defining additional properties

integration/combination/test_function_with_alias.py
```diff
@@ -35,6 +35,42 @@ def test_updating_version_by_changing_property_value(self):
         self.assertEqual(len(alias), 1)
```
hoffa, line 35: Should also include transform test (you can use the existing integration test template) to ensure the the hash doesn't change unexpectedly.

samtranslator/model/sam_resources.py
```diff
@@ -886,6 +889,11 @@ def _construct_version(
             # If SnapStart is enabled we want to publish a new version, to have the corresponding snapshot
             if function.SnapStart and function.SnapStart.get("ApplyOn", "None") != "None":
                 logical_dict.update({"SnapStart": function.SnapStart})
+            # We can't directly change AutoPublishAlias as that would be a breaking change, so we have to add this opt-in
+            # property that when set to true would change the lambda version whenever a property in the lambda function changes
+            if self.AutoPublishAliasAllProperties:
```
hoffa, line 894: Could this interfere with the `ApplyOn` logic above?

hoffa, line 894: Should this be supported in `Globals` as well?

aaythapa: Not too sure, do we add every new property as `Globals` or only certain ones?

aaythapa: Possibly, if they have a function and they add `Snapstart` set to `None` then it would create a new version without the Snapstart being enabled. Should that case be accounted for?

hoffa: It looks like `AutoPublishAlias` is, so probably want for this one as well: https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-specification-template-anatomy-globals.html

hoffa: Not sure, I don't think it matters. It basically checks that SnapStart is enabled. For `AutoPublishAliasAllProperties` we don't care about any of that I assume.

samtranslator/model/sam_resources.py
```diff
@@ -886,6 +889,11 @@ def _construct_version(
...
+                properties = function._generate_resource_dict().get("Properties", {})
+                logical_dict.update(properties)
```
hoffa, line 896: Why not `logical_dict = properties`?

aaythapa: Thought something like that would mess with the `code_dict` thats created and added to the `logical_dict` earlier in the file but doesn't look like it does. Going to restructure this so that `AutoPublishAliasAllProperties ` is checked first

tests/translator/input/function_with_alias_and_all_properties_property.yaml
```diff
@@ -0,0 +1,20 @@
+Resources:
+  MinimalFunction:
+    Type: AWS::Serverless::Function
+    Properties:
+      CodeUri: s3://sam-demo-bucket/hello.zip
+      Handler: hello.handler
+      Runtime: python2.7
+      AutoPublishAlias: live
+      AutoPublishAliasAllProperties: true
```
hoffa, line 9: I'd add another minimal function with this as `true` and any other property not supported by `AutoPublishAlias`, to confirm the hash changes.

ssenchenko (approved):

hoffa (approved):

samtranslator/model/sam_resources.py
```diff
@@ -874,18 +877,25 @@ def _construct_version(
...
-                logical_dict.update(function.Environment)
-            if function.MemorySize:
-                logical_dict.update({"MemorySize": function.MemorySize})
-            # If SnapStart is enabled we want to publish a new version, to have the corresponding snapshot
-            if function.SnapStart and function.SnapStart.get("ApplyOn", "None") != "None":
-                logical_dict.update({"SnapStart": function.SnapStart})
+            try:
+                logical_dict = code_dict.copy()
+            except (AttributeError, UnboundLocalError):
+                pass  # noqa: try-except-pass
```
aahung, line 889: We should use https://docs.python.org/3/library/contextlib.html#contextlib.suppress to be more specific. Also some comment can help to understand why

samtranslator/model/sam_resources.py
```diff
@@ -874,18 +878,22 @@ def _construct_version(
...
+            with suppress(AttributeError, UnboundLocalError):
+                logical_dict = code_dict.copy()
             if function.Environment:
                 logical_dict.update(function.Environment)
             if function.MemorySize:
                 logical_dict.update({"MemorySize": function.MemorySize})
             # If SnapStart is enabled we want to publish a new version, to have the corresponding snapshot
             if function.SnapStart and function.SnapStart.get("ApplyOn", "None") != "None":
                 logical_dict.update({"SnapStart": function.SnapStart})
+
```
hoffa, line 896: nit: surgery

```suggestion
```

aahung (approved):

ConnorRobertson (approved):
LGTM

PR #2839, opened 2023-01-31
feat: load managed policies locally

hoffa:
### Description of changes

Faster and more resilient transforms by bundling the name-to-ARN managed policy map with the code.

The ARN of a managed policy name is determined by `Translator` in the following order:

1. From the `managed_policy_map` argument. Can be `None`.
2. From the bundled managed policy map.
3. From the `get_managed_policy_map` function. Must be efficient to call, as it can be called multiple times within a transform.

Got the managed policies from my personal account using:

```bash
aws iam list-policies --scope AWS | jq '.Policies[] | {(.PolicyName): .Arn}' | jq -s add --sort-keys
```

Originally cached `get_managed_policy_map()` within the `Translator` instance, but anecdotally `Translator` is instantiated in hot paths, so letting the caller do the caching is simpler to integrate.

### Checklist

    - [ ] Using correct values
    - [ ] Using wrong values

### Examples?


aahung:
gogogo

bin/list_managed_policies.py
```diff
@@ -0,0 +1,33 @@
...
+    return re.sub(r"^arn:\w+:", f"arn:{partition}:", arn)
+
+
+def main() -> None:
+    # Any region from `aws` partition; IAM is global
+    iam = boto3.client("iam", region_name="us-west-2")
+
+    policies_aws = ManagedPolicyLoader(iam).load()  # type: ignore
+    policies_aws_cn = {k: replace_partition(v, "aws-cn") for k, v in policies_aws.items()}
+    policies_aws_us_gov = {k: replace_partition(v, "aws-us-gov") for k, v in policies_aws.items()}
```
aahung, line 19: non-aws partitions might have different list of managed policies. Simply replacing them might not work properly

GavinZZ (approved):

aahung (commented):
remember to add it to manifest.in

samtranslator/internal/managed_policies.py
```diff
@@ -0,0 +1,56 @@
...
+    """
+    # Caller-provided policy map (eager)
+    arn = _dict_get(managed_policy_map, name)
+    if arn:
+        return arn
+
+    # Bundled policy map
+    partition = ArnGenerator.get_partition_name()
+    bundled_managed_policies = _BUNDLED_MANAGED_POLICIES.get(partition)
+    arn = _dict_get(bundled_managed_policies, name)
```
aahung, line 45: why we need a method for that? can't it just be `arn = bundled_managed_policies and bundled_managed_policies.get(name)`?

samtranslator/internal/managed_policies.py
```diff
@@ -0,0 +1,56 @@
...
+
+from samtranslator.internal.types import GetManagedPolicyMap
+from samtranslator.translator.arn_generator import ArnGenerator
+
+with Path(os.path.dirname(os.path.abspath(__file__)), "data", "managed_policies.json").open(encoding="utf-8") as f:
+    _BUNDLED_MANAGED_POLICIES: Dict[str, Dict[str, str]] = json.load(f)
+
+
+def _dict_get(d: Optional[Dict[str, str]], k: str) -> Optional[str]:
+    if isinstance(d, dict) and k in d:
```
aahung, line 14: adding a check like this suppress the error what `d` is not a dict, which should be detected before sam-t is shipped.

samtranslator/translator/translator.py
```diff
@@ -137,6 +140,15 @@ def translate(  # noqa: too-many-branches
...
         intrinsics_resolver = IntrinsicsResolver(parameter_values)
 
+        # The available managed policies aren't expected to change during the
+        # run of a transform; cache function return value as it can be called
+        # multiple times
+        @lru_cache(maxsize=None)
+        def cached_get_managed_policy_map() -> Optional[Dict[str, str]]:
+            if not get_managed_policy_map:
+                return None
+            return get_managed_policy_map()
```
aahung, line 150: I am not a fan of line functions, neither sure whether `@lru_cache` will work inside another function, as every time this is called, it creates another cached `cached_get_managed_policy_map`. I don't think caching is a concern of this part of the code, but the caller of `samtranslator` library.

aahung: Another concern is memory leak of `lru_cache`. What if `lru_cache` still keep the cache even when the function is out of scope?

hoffa: Yeah, it's not pretty. Included as forgetting it could mean a lot of IAM calls, and weaving the caching into the instances got messy. I removed for now and will check whether I can put it in the instances more cleanly.

samtranslator/internal/managed_policies.py
hoffa: In case it isn't a dict. But it's true we probably want to hard fail in that case since something's wrong.

aahung (approved):
🎉

samtranslator/internal/data/managed_policies.json
```diff
@@ -0,0 +1,1040 @@
+{
```
aahung, line 1: nit: rename this file to `aws_managed_policies.json`?

samtranslator/translator/translator.py
```diff
@@ -70,8 +70,9 @@ def __init__(  # type: ignore[no-untyped-def] # noqa: PLR0913
             ArnGenerator.BOTO_SESSION_REGION_NAME = self.boto_session.region_name
 
     def _get_managed_policy_map(self) -> Optional[Dict[str, str]]:
-        if not self._fallback_managed_policy_map and self._get_fallback_managed_policy_map:
-            self._fallback_managed_policy_map = self._get_fallback_managed_policy_map()
+        if self._fallback_managed_policy_map is None:
+            if self._get_fallback_managed_policy_map:
```
aahung: does ruff allow this nested if?

samtranslator/model/role_utils/role_constructor.py
```diff
@@ -83,8 +130,12 @@ def construct_role_for_resource(  # type: ignore[no-untyped-def] # noqa: too-man
             #
 
             policy_arn = policy_entry.data
-            if isinstance(policy_entry.data, str) and policy_entry.data in managed_policy_map:
-                policy_arn = managed_policy_map[policy_entry.data]
+            if isinstance(policy_entry.data, str):
```
aahung, line 133: nit: 
```suggestion
            if isinstance(policy_arn, str):
```

samtranslator/model/role_utils/role_constructor.py
```diff
@@ -83,8 +130,12 @@ def construct_role_for_resource(  # type: ignore[no-untyped-def] # noqa: too-man
...
+                policy_arn = _get_managed_policy_arn(
+                    policy_entry.data,
```
aahung, line 135:
```suggestion
                    policy_arn,
```

samtranslator/translator/translator.py
hoffa: No, that's previous commit.

aahung: oh..

aahung (approved):
🎉

samtranslator/model/role_utils/role_constructor.py
```diff
@@ -1,7 +1,57 @@
...
+
+    # Bundled managed policy map
+    partition = ArnGenerator.get_partition_name()
+    bundled_managed_policy_map = get_bundled_managed_policy_map(partition)
+    if bundled_managed_policy_map:
+        arn = bundled_managed_policy_map.get(name)
+        if arn:
+            return arn
+
+    # Caller-provided managed policy map (fallback)
```
acristin, line 46: This isn't caller provided

samtranslator/model/role_utils/role_constructor.py
```diff
@@ -83,8 +134,12 @@ def construct_role_for_resource(  # type: ignore[no-untyped-def] # noqa: too-man
             #
 
             policy_arn = policy_entry.data
-            if isinstance(policy_entry.data, str) and policy_entry.data in managed_policy_map:
-                policy_arn = managed_policy_map[policy_entry.data]
+            if isinstance(policy_arn, str):
```
acristin, line 137: For my own curiosity, what should happen if this is the empty string?

samtranslator/translator/transform.py
```diff
@@ -15,13 +19,19 @@ def transform(input_fragment, parameter_values, managed_policy_loader, feature_t
     sam_parser = Parser()
     to_py27_compatible_template(input_fragment, parameter_values)
     translator = Translator(  # type: ignore[no-untyped-call]
-        managed_policy_loader.load(),
+        None,
         sam_parser,
     )
+
+    @lru_cache(maxsize=None)
```
acristin, line 26: Nit/question: Why `maxsize=None` vs `mazsize=1`?

acristin (approved):
Just a minor complaint and two questions.

samtranslator/model/role_utils/role_constructor.py
hoffa, line 46:
```suggestion
    # Caller-provided function to get managed policy map (fallback)
```

samtranslator/translator/transform.py
hoffa: One fewer things to think about and avoid bugs if we e.g. add an optional parameter later.

`@lru_cache(maxsize=None)` is same as `@cache` in Python 3.9+, i.e. a simple args-to-value cache without the LRU machinery.

samtranslator/model/role_utils/role_constructor.py
hoffa: It falls back to whatever the string is, so empty string. And deployment will fail.

PR #2857, opened 2023-02-02
feat: Add connector type

GavinZZ:
### Description of changes
Added connector type (standard, embedded, source, or embedded + source) to metadata for logging purposes.

### Checklist

    - [ ] Using correct values
    - [ ] Using wrong values

### Examples?

### DIFF
--- a/samtranslator/model/sam_resources.py
+++ b/samtranslator/model/sam_resources.py
@@ -1837,7 +1837,7 @@ class SamConnector(SamResourceMacro):
                 source, destination, dest_index, multi_dest, resource_resolver
             )
 
-            self._add_connector_metadata(generated_resources, original_template, source, destination)
+            self._add_connector_metadata(generated_resources, original_template, source, destination, multi_dest)
             list_generated_resources.extend(generated_resources)
 
         generated_logical_ids = [resource.logical_id for resource in list_generated_resources]
@@ -2096,6 +2096,7 @@ class SamConnector(SamResourceMacro):
         original_template: Dict[str, Any],
         source: ConnectorResourceReference,
         destination: ConnectorResourceReference,
+        is_multi_dest: bool,
     ) -> None:
         """
         Add metadata attribute to generated resources.
@@ -2111,18 +2112,7 @@ class SamConnector(SamResourceMacro):
         original_resources = original_template.get("Resources", {})
         original_source_type = original_resources.get(source.logical_id, {}).get("Type")
         original_dest_type = original_resources.get(destination.logical_id, {}).get("Type")
-        metadata = {
-            "aws:sam:connectors": {
-                self.logical_id: {
-                    # If the source/destination is a serverless resource,
-                    # we prefer to include the original serverless resource type
-                    # over the transformed CFN resource type so it can distinguish
-                    # connector usage between serverless resources and CFN resources.
-                    "Source": {"Type": original_source_type or source.resource_type},
-                    "Destination": {"Type": original_dest_type or destination.resource_type},
-                }
-            }
-        }
+
         for resource in generated_resources:
             # Although as today the generated resources do not have any existing metadata,
             # To make it future proof, we still does a merge to avoid overwriting.
@@ -2130,4 +2120,24 @@ class SamConnector(SamResourceMacro):
                 original_metadata = resource.get_resource_attribute("Metadata")
             except KeyError:
                 original_metadata = {}
+
+            connector_type: Dict[str, Any] = {
+                "multidest:connector": is_multi_dest,
+                "embedded:connector": "aws:sam:embedded:connector" in original_metadata,
+            }
+            original_metadata = {k: v for k, v in original_metadata.items() if k != "aws:sam:embedded:connector"}
+
+            metadata = {
+                "aws:sam:connectors": {
+                    self.logical_id: {
+                        # If the source/destination is a serverless resource,
+                        # we prefer to include the original serverless resource type
+                        # over the transformed CFN resource type so it can distinguish
+                        # connector usage between serverless resources and CFN resources.
+                        "Source": {"Type": original_source_type or source.resource_type},
+                        "Destination": {"Type": original_dest_type or destination.resource_type},
+                        **connector_type,
+                    }
+                }
+            }
             resource.set_resource_attribute("Metadata", {**original_metadata, **metadata})
--- a/samtranslator/model/sam_resources.py
+++ b/samtranslator/model/sam_resources.py
@@ -2122,8 +2122,8 @@ class SamConnector(SamResourceMacro):
                 original_metadata = {}
 
             connector_type: Dict[str, Any] = {
-                "MultiDestConnector": is_multi_dest,
-                "EmbeddedConnector": "aws:sam:embedded:connector" in original_metadata,
+                "MultiDestination": is_multi_dest,
+                "Embedded": "aws:sam:embedded:connector" in original_metadata,
             }
             original_metadata = {k: v for k, v in original_metadata.items() if k != "aws:sam:embedded:connector"}
 
--- a/samtranslator/model/sam_resources.py
+++ b/samtranslator/model/sam_resources.py
@@ -2122,8 +2122,8 @@ class SamConnector(SamResourceMacro):
                 original_metadata = {}
 
             connector_type: Dict[str, Any] = {
-                "multidest:connector": is_multi_dest,
-                "embedded:connector": "aws:sam:embedded:connector" in original_metadata,
+                "MultiDestConnector": is_multi_dest,
+                "EmbeddedConnector": "aws:sam:embedded:connector" in original_metadata,
             }
             original_metadata = {k: v for k, v in original_metadata.items() if k != "aws:sam:embedded:connector"}
 
--- a/samtranslator/translator/translator.py
+++ b/samtranslator/translator/translator.py
@@ -1,4 +1,5 @@
 import copy
+from contextlib import suppress
 from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple
 
 from samtranslator.feature_toggle.feature_toggle import (
@@ -330,6 +331,14 @@ class Translator:
                         source_logical_id,
                         connector_logical_id,
                         connector_dict,
+                    )
+
+                    # In case the generated resource already contains Metadata, we need to append to it
+                    original_metadata = {}
+                    with suppress(KeyError):
+                        original_metadata = generated_connector.get_resource_attribute("Metadata")
+                    generated_connector.set_resource_attribute(
+                        "Metadata", {**original_metadata, "aws:sam:embedded:connector": True}
                     )
 
                     if not verify_unique_logical_id(generated_connector, resources):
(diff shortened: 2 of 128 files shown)


ssenchenko (approved):

samtranslator/translator/translator.py
```diff
@@ -332,6 +332,15 @@ def _get_embedded_connectors(self, resources: Dict[str, Any]) -> List[Resource]:
...
 
+                    # In case the generated resource already contains Metadata, we need to append to it
+                    try:
+                        original_metadata = generated_connector.get_resource_attribute("Metadata")
+                    except KeyError:
+                        original_metadata = {}
+                    generated_connector.set_resource_attribute(
+                        "Metadata", {**original_metadata, "aws:sam:connectors:type": "Embedded"}
+                    )
+
```
aaythapa, line 343: Can we use [suppress](https://docs.python.org/3/library/contextlib.html#contextlib.suppress) here? Something like:
```
original_metadata = {}
suppress(KeyError):
    original_metadata = generated_connector.get_resource_attribute("Metadata")
generated_connector.set_resource_attribute(
    "Metadata", {**original_metadata, "aws:sam:connectors:type": "Embedded"}
)
```

samtranslator/model/sam_resources.py
```diff
@@ -2111,23 +2120,36 @@ def _add_connector_metadata(
...
             # To make it future proof, we still does a merge to avoid overwriting.
             try:
                 original_metadata = resource.get_resource_attribute("Metadata")
             except KeyError:
                 original_metadata = {}
+
+            connector_type = STANDARD_CONNECTOR
+            if "aws:sam:connectors:type" in original_metadata and is_multi_dest:
+                connector_type = EMBEDDED_SOURCE_CONNECTOR
+                del original_metadata["aws:sam:connectors:type"]
```
hoffa, line 2135: Any way to avoid mutations? It's hard to review.

samtranslator/model/connector/connector.py
```diff
@@ -20,6 +21,11 @@
     ],
 )
 
+STANDARD_CONNECTOR = "StandardConnector"
+EMBEDDED_CONNECTOR = "EmbeddedConnector"
+SOURCE_CONNECTOR = "SourceConnector"
```
xazhao, line 25: `SourceConnector` is a little vague. Maybe try something like `MultiDestConnector`?

samtranslator/model/sam_resources.py
```diff
@@ -2081,8 +2088,9 @@ def _construct_sqs_queue_policy(
         queue_policy_name = (
             f"{self.logical_id}QueuePolicyDestination{dest_index}" if multi_dest else f"{self.logical_id}QueuePolicy"
         )
+        # For multi-destination connector, modifying resource attribute will affect others, use deep copy to prevent it
```
hoffa, line 2091: > For multi-destination connector, modifying resource attribute will affect others

Why?

samtranslator/model/connector/connector.py
GavinZZ: Sounds good.

samtranslator/model/connector/connector.py
```diff
@@ -20,6 +20,11 @@
...
+EMBEDDED_SOURCE_CONNECTOR = "EmbeddedSourceConnector"
```
aaythapa, line 26: Should we add this now or when we actually implement it?

samtranslator/translator/translator.py
```diff
@@ -332,6 +332,15 @@ def _get_embedded_connectors(self, resources: Dict[str, Any]) -> List[Resource]:
                         connector_dict,
                     )
 
+                    # In case the generated resource already contains Metadata, we need to append to it
+                    try:
+                        original_metadata = generated_connector.get_resource_attribute("Metadata")
+                    except KeyError:
+                        original_metadata = {}
+                    generated_connector.set_resource_attribute(
+                        "Metadata", {**original_metadata, "aws:sam:connectors:type": "Embedded"}
```
hoffa, line 341: Why is there both `aws:sam:connectors:type` and `ConnectorType`?

samtranslator/model/sam_resources.py
GavinZZ: When we use multi-destination and embedded connector together 
```
MyServerlessFunction:
    Connectors:
      MixedDestConnector:
        Properties:
          Metadata: xxx
          Destination:
          - Id: MyBucket
          - Id: MyQueue
          Permissions:
          - Read
          - Write
         ...
```
This is going to become two IAM Managed Policies. After generating the first IAM Managed Policy but before the second one, if I modify Metadata, it will it to change for the second IAM Managed Policy generation.

samtranslator/model/sam_resources.py
```diff
@@ -2111,23 +2120,36 @@ def _add_connector_metadata(
...
                 original_metadata = resource.get_resource_attribute("Metadata")
             except KeyError:
                 original_metadata = {}
+
+            connector_type = STANDARD_CONNECTOR
+            if "aws:sam:connectors:type" in original_metadata and is_multi_dest:
+                connector_type = EMBEDDED_SOURCE_CONNECTOR
+                del original_metadata["aws:sam:connectors:type"]
+            elif "aws:sam:connectors:type" in original_metadata:
+                connector_type = EMBEDDED_CONNECTOR
```
hoffa, line 2137: This seems weird... because it has that key it's an embedded connector?

samtranslator/model/sam_resources.py
xazhao, line 2091: What do you mean here?

samtranslator/translator/translator.py
GavinZZ: `aws:sam:connectors:type` is used to distinguish it's a embedded connector type and later deleted.

samtranslator/model/connector/connector.py
GavinZZ: I'm not sure if I'm following. These are being used now.

samtranslator/model/sam_resources.py
hoffa: Can it be avoided altogether by refactoring code?

samtranslator/model/connector/connector.py
xazhao: It's already supported after we implement `EMBEDDED_CONNECTOR` and `SOURCE_CONNECTOR`

samtranslator/model/sam_resources.py
GavinZZ: Not easily, I can think of another method that requires passing a parameter around. I like the current one better. Open for suggestions.

samtranslator/model/sam_resources.py
GavinZZ: the value is actually unused, but metadata requires to be key value pair, perhaps renaming the key to `aws:sam:embedded:connector` is better

samtranslator/model/sam_resources.py
GavinZZ: removed. refactored to use another method

samtranslator/model/connector/connector.py
aaythapa: Ah I didn't realize `EMBEDDED_SOURCE_CONNECTOR` was multi dest in embedded connector. Much cleared with the name change

samtranslator/model/connector/connector.py
```diff
@@ -20,6 +20,11 @@
     ],
 )
 
+STANDARD_CONNECTOR = "StandardConnector"
+EMBEDDED_CONNECTOR = "EmbeddedConnector"
+MULTIDEST_CONNECTOR = "MultiDestConnector"
+EMBEDDED_MULTIDEST_CONNECTOR = "EmbeddedMultiDestConnector"
```
hoffa, line 26: Why a separate type? Instead of individual booleans, and then for metrics can check combination. Otherwise with more kinds of connectors this will grow exponentially. Also should keep code simpler.

samtranslator/model/sam_resources.py
```diff
@@ -2111,23 +2116,36 @@ def _add_connector_metadata(
...
         for resource in generated_resources:
             # Although as today the generated resources do not have any existing metadata,
             # To make it future proof, we still does a merge to avoid overwriting.
             try:
                 original_metadata = resource.get_resource_attribute("Metadata")
             except KeyError:
                 original_metadata = {}
+
+            connector_type = STANDARD_CONNECTOR
+            if "aws:sam:embedded:connector" in original_metadata and is_multi_dest:
```
aaythapa, line 2129: Would it be worth making `connector_type` a list and adding `EMBEDDED_CONNECTOR` and `MULTIDEST_CONNECTOR` rather than making a separate `EMBEDDED_MULTIDEST_CONNECTOR` type? It'd be one less type

samtranslator/model/connector/connector.py
GavinZZ: great ideas

samtranslator/model/sam_resources.py
GavinZZ: I'll actually make them a boolean value, like Chris suggested.

samtranslator/model/sam_resources.py
```diff
@@ -2111,23 +2112,32 @@ def _add_connector_metadata(
...
         for resource in generated_resources:
             # Although as today the generated resources do not have any existing metadata,
             # To make it future proof, we still does a merge to avoid overwriting.
             try:
                 original_metadata = resource.get_resource_attribute("Metadata")
             except KeyError:
                 original_metadata = {}
+
+            connector_type: Dict[str, Any] = {
+                "MultiDestConnector": is_multi_dest,
```
hoffa, line 2125: We don't use abbreviations elsewhere in the template, so I'd rather keep it as `Destination`. And ideally get rid of `Connector` repetition if we can make it clear from context.

samtranslator/model/sam_resources.py
```diff
@@ -2111,23 +2112,32 @@ def _add_connector_metadata(
...
+                "EmbeddedConnector": "aws:sam:embedded:connector" in original_metadata,
```
hoffa, line 2126: I still don't get what `aws:sam:embedded:connector` is. It doesn't show up in the final template, but looks misleadingly like something that does.

GavinZZ: It's used as an intermediate processing step. I intentionally removed it from the final template.

hoffa: Why intermediate?

GavinZZ: Because the final metadata hasn't been built yet at the time I added `aws:sam:embedded:connector`

tests/translator/output/aws-cn/connector_api_to_multiple_function.json
```diff
@@ -7,6 +7,8 @@
             "Destination": {
               "Type": "AWS::Lambda::Function"
             },
+            "Embedded": false,
+            "MultiDestination": true,
```
hoffa, line 11: This was originally a single connector, right? In the output now we have to `Metadata`s with `MultiDestination` being true. Is that the intention? How do we know how many times multi-destination was used?

GavinZZ: That's actually a great point. I was thinking that for multi-destination connector, each generated policy should count as one usage. But I should double check with Simon on this behaviour.

hoffa:
For posterity, `Metadata` was initially added so it's easy to see from output template which connector (which doesn't exist in output template) created the permissions resource(s).

PR #2877, opened 2023-02-09
ci: Add GH Action to detect compatibility breaking chages

aahung:

I created another PR draft to test it: https://github.com/aws/serverless-application-model/pull/2878#issuecomment-1423485136

image

The test GH action run: https://github.com/aws/serverless-application-model/actions/runs/4130070494

### Checklist

    - [ ] Using correct values
    - [ ] Using wrong values

### Examples?

### DIFF
--- a/.github/workflows/check_interface_compatibility.yml
+++ b/.github/workflows/check_interface_compatibility.yml
@@ -16,29 +16,31 @@ jobs:
         with:
           python-version: "3.10"
 
-      - name: Extract new public interfaces
+      - name: Extract original public interfaces
         run: |
           make init
-          bin/public_interface.py extract >> ${{ runner.temp }}/interfaces.new.json
+          bin/public_interface.py extract >> ${{ runner.temp }}/interfaces.original.json
 
-      - name: Backup bin/public_interface.py
+      - name: Backup original bin/public_interface.py
         run: |
           # Keep a copy of bin/public_interface.py to use on the base branch
           # So we are using the same bin/public_interface.py to process old/new codebase.
+          # Because the step has credential of GitHub token to post comments,
+          # always use the original version that is in our repository.
           cp bin/public_interface.py ${{ runner.temp }}/public_interface.py
 
-      - name: Checkout the base branch
+      - name: Checkout the PR
         uses: actions/checkout@v3
         with:
-          ref: ${{ github.base_ref }}
+          ref: "refs/pull/${{ github.event.number }}/merge"
 
-      - name: Extract original public interfaces
+      - name: Extract PR public interfaces
         run: |
-          echo "Original HEAD: $(git rev-parse --short HEAD)"
+          echo "HEAD: $(git rev-parse --short HEAD)"
           make init
           # Recover bin/public_interface.py
           cp ${{ runner.temp }}/public_interface.py bin/public_interface.py
-          bin/public_interface.py extract >> ${{ runner.temp }}/interfaces.original.json
+          bin/public_interface.py extract >> ${{ runner.temp }}/interfaces.new.json
 
       - name: Archive interface files for debugging
         uses: actions/upload-artifact@v3
--- a/.github/workflows/check_interface_compatibility.yml
+++ b/.github/workflows/check_interface_compatibility.yml
@@ -46,7 +46,7 @@ jobs:
         id: detect
         run: |
           exit_code=0
-          output=$(bin/public_interface.py check ${{ runner.temp }}/interfaces.original.json ${{ runner.temp }}/interfaces.new.json) | exit_code=$?
+          output=$(bin/public_interface.py check ${{ runner.temp }}/interfaces.original.json ${{ runner.temp }}/interfaces.new.json) || exit_code=$?
 
           echo "output=$output"
           echo "exit_code=$exit_code"
--- a/.github/workflows/check_interface_compatibility.yml
+++ b/.github/workflows/check_interface_compatibility.yml
@@ -36,7 +36,6 @@ jobs:
 
       - name: Extract PR public interfaces
         run: |
-          echo "HEAD: $(git rev-parse --short HEAD)"
           make init
           # Recover bin/public_interface.py
           cp ${{ runner.temp }}/public_interface.py bin/public_interface.py
@@ -48,20 +47,15 @@ jobs:
           exit_code=0
           output=$(bin/public_interface.py check ${{ runner.temp }}/interfaces.original.json ${{ runner.temp }}/interfaces.new.json) || exit_code=$?
 
-          echo "output=$output"
-          echo "exit_code=$exit_code"
+          # Show the output in job summary
           echo "$output" >> $GITHUB_STEP_SUMMARY
+          
+          # Save exit code to determine whether to post pull request comment or not
           echo "exit_code=$exit_code" >> $GITHUB_OUTPUT
-          echo "$output" >> ${{ runner.temp }}/output.txt
-
-      - name: Archive interface files for debugging
-        uses: actions/upload-artifact@v3
-        with:
-          name: interface-json-files
-          path: |
-            ${{ runner.temp }}/interfaces.*.json
-            output.txt
-          retention-days: 7  # Only storing it for a week in case of debugging needs.
+          
+          # Save the output for next step to process it,
+          # and saving it into a file can avoid the need of escaping in github-script.
+          echo "$output" >> ${{ runner.temp }}/output.md
 
       - name: Post GitHub PR comment
         if: ${{ steps.detect.outputs.exit_code != '0' }}
@@ -70,16 +64,17 @@ jobs:
           github-token: ${{secrets.GITHUB_TOKEN}}
           script: |
             const fs = require('fs');
-            const data = fs.readFileSync('${{ runner.temp }}/output.txt', 'utf8')
+            const output = fs.readFileSync('${{ runner.temp }}/output.md', 'utf8')
             github.rest.issues.createComment({
               issue_number: context.issue.number,
               owner: context.repo.owner,
               repo: context.repo.repo,
-              body: "hi" + data
+              body: output
             })
 
       - name: Determine the success/failure of the workflow
         run: |
           if [ ${{ steps.detect.outputs.exit_code }} -ne 0 ]; then
+            cat ${{ runner.temp }}/output.md
             exit ${{ steps.detect.outputs.exit_code }}
           fi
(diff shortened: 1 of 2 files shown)


.github/workflows/check_interface_compatibility.yml
```diff
@@ -0,0 +1,74 @@
...
+      - name: Backup bin/public_interface.py
+        run: |
+          # Keep a copy of bin/public_interface.py to use on the base branch
+          # So we are using the same bin/public_interface.py to process old/new codebase.
+          cp bin/public_interface.py ${{ runner.temp }}/public_interface.py
+
+      - name: Checkout the base branch
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ github.base_ref }}
```
hoffa, line 33: How about checking out to different directories instead? Currently it's kind of confusing with the `cp`s and such.

.github/workflows/check_interface_compatibility.yml
```diff
@@ -0,0 +1,76 @@
...
+
+      - name: Extract original public interfaces
+        run: |
+          echo "Original HEAD: $(git rev-parse --short HEAD)"
+          make init
+          # Recover bin/public_interface.py
+          cp ${{ runner.temp }}/public_interface.py bin/public_interface.py
+          bin/public_interface.py extract >> ${{ runner.temp }}/interfaces.original.json
```
hoffa, line 41: Why append?

.github/workflows/check_interface_compatibility.yml
```diff
@@ -0,0 +1,76 @@
...
+        id: detect
+        run: |
+          output=$(bash -c '(bin/public_interface.py check ${{ runner.temp }}/interfaces.original.json ${{ runner.temp }}/interfaces.new.json); exit $?')
+          exit_code=$?
+          echo "output=$output"
+          echo "exit_code=$exit_code"
+          echo "output=$output" >> $GITHUB_OUTPUT
+          echo "exit_code=$exit_code" >> $GITHUB_OUTPUT
+
+      - name: Post GitHub PR comment
```
hoffa, line 60: Why can't we just let it fail? We use auto-merge so much this won't work unless it's enforced.

.github/workflows/check_interface_compatibility.yml
```diff
@@ -0,0 +1,76 @@
...
+        uses: actions/upload-artifact@v3
+        with:
+          name: interface-json-files
+          path: ${{ runner.temp }}/interfaces.*.json
+          retention-days: 7  # Only storing it for a week in case of debugging needs.
+    
+      - name: Detect compatibility breaking changes
+        id: detect
+        run: |
+          output=$(bash -c '(bin/public_interface.py check ${{ runner.temp }}/interfaces.original.json ${{ runner.temp }}/interfaces.new.json); exit $?')
```
hoffa, line 53: Why `bash`? The script should already return the relevant exit code.

.github/workflows/check_interface_compatibility.yml
```diff
@@ -0,0 +1,76 @@
...
+
+      - name: Extract original public interfaces
+        run: |
+          echo "Original HEAD: $(git rev-parse --short HEAD)"
+          make init
+          # Recover bin/public_interface.py
+          cp ${{ runner.temp }}/public_interface.py bin/public_interface.py
+          bin/public_interface.py extract >> ${{ runner.temp }}/interfaces.original.json
+
+      - name: Archive interface files for debugging
```
hoffa, line 43: Why do we need this? Isn't the script output good enough?

aahung: This is still draft and you are reviewing the stuff used for my debugging.

.github/workflows/check_interface_compatibility.yml
aahung: not in final version

.github/workflows/check_interface_compatibility.yml
aahung: discussed offline, since it is not enforced a comment catches reviewers attention than a small cross, and it is more readable

.github/workflows/check_interface_compatibility.yml
aahung: This approach saves time in `make init`

aahung: the latest version has enough comment to explain it

.github/workflows/check_interface_compatibility.yml
aahung: not in the final version

.github/workflows/check_interface_compatibility.yml
hoffa: Why not enforce?

aahung: As it could be intentional

aahung: And this homemade script could generate false positives

GavinZZ (approved):

.github/workflows/check_interface_compatibility.yml
```diff
@@ -0,0 +1,80 @@
...
+
+      - name: Post GitHub PR comment
+        if: ${{ steps.detect.outputs.exit_code != '0' }}
+        uses: actions/github-script@v6
+        with:
+          github-token: ${{secrets.GITHUB_TOKEN}}
+          script: |
+            const fs = require('fs');
+            const output = fs.readFileSync('${{ runner.temp }}/output.md', 'utf8')
+            github.rest.issues.createComment({
```
acristin, line 68: Will this comment block merges until resolved, as a review comment would?

acristin (approved):

.github/workflows/check_interface_compatibility.yml
aahung: good idea, let me check if there is an API for that

.github/workflows/check_interface_compatibility.yml
```diff
@@ -0,0 +1,80 @@
...
+          make init
+          # Recover bin/public_interface.py
+          cp ${{ runner.temp }}/public_interface.py bin/public_interface.py
+          bin/public_interface.py extract > ${{ runner.temp }}/interfaces.new.json
+    
+      - name: Detect compatibility breaking changes
+        id: detect
+        run: |
+          exit_code=0
+          output=$(bin/public_interface.py check ${{ runner.temp }}/interfaces.original.json ${{ runner.temp }}/interfaces.new.json) || exit_code=$?
```
hoffa, line 48: AFAIK you don't need the initialization or `||` thing. Just `exit_code=$?` on separate line should be enough.

.github/workflows/check_interface_compatibility.yml
```diff
@@ -0,0 +1,80 @@
...
+
+permissions:
+  pull-requests: write
+
+jobs:
+  check-interface-compatibility:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
```
hoffa, line 14: For symmetry with the other checkout, a `name` here describing what it checks out would be nice.

.github/workflows/check_interface_compatibility.yml
```diff
@@ -0,0 +1,80 @@
...
+          cp bin/public_interface.py ${{ runner.temp }}/public_interface.py
+
+      - name: Checkout the PR
+        uses: actions/checkout@v3
+        with:
+          ref: "refs/pull/${{ github.event.number }}/merge"
+
+      - name: Extract PR public interfaces
+        run: |
+          make init
```
hoffa, line 39: Why is this needed again? Aren't all the dependencies from the base branch code, since that's the only script that's used?

.github/workflows/check_interface_compatibility.yml
aahung: to create a PR comment that requires resolution, `path` and `line` are needed. Maybe it is too much work

https://octokit.github.io/rest.js/v19#pulls-create-review-comment

.github/workflows/check_interface_compatibility.yml
```diff
@@ -0,0 +1,80 @@
...
+            github.rest.issues.createComment({
+              issue_number: context.issue.number,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              body: output
+            })
+
+      - name: Determine the success/failure of the workflow
+        run: |
+          if [ ${{ steps.detect.outputs.exit_code }} -ne 0 ]; then
```
hoffa, line 77: Can this be like L61? So the UI shows nicer that the step is skipped.

hoffa: nit: and consider using a regular 'boolean' like in https://github.com/aws/serverless-application-model/blob/53865b59fbd9a3caae57159236378420a4420276/.github/workflows/schema.yml#L32 so it's arguably easier to read, as we really only care about whether there's a breaking change, not about exit codes.

.github/workflows/check_interface_compatibility.yml
aahung: in case there is dependency change and module cannot be imported if samtranslator depends on a new pypi package.

.github/workflows/check_interface_compatibility.yml
aahung: updated

.github/workflows/check_interface_compatibility.yml
aahung: same issue as we saw in the release script, if `bin/public_interface.py` exit 1, the whole job stops without `||`

.github/workflows/check_interface_compatibility.yml
aahung: good point, updated.

aahung: ok it worked https://github.com/aws/serverless-application-model/actions/runs/4130330209/jobs/7136902172

.github/workflows/check_interface_compatibility.yml
acristin: Yeah, that's what I concluded as well.

Wow, that documentation seems much better than the official docs; happy to learn of it

.github/workflows/check_interface_compatibility.yml
```diff
@@ -0,0 +1,83 @@
+name: Check interface compatibility
```
hoffa, line 1: Actually I think we need a way to run this easily locally, e.g. `make compat`. Currently it'll probably get annoying to not be able to reproduce and have big comments in PRs.

We could store the JSON in the repository and compare against that, no?

hoffa (approved):

Issue #1656, opened 2020-07-15
Suppress lambda permissions


jvasiljevich:
**Description:**
I am attempting to define an Authorizer within an `AWS::Serverless::Api resource`. The Authorizer is linked to a Lambda that is not defined in my stack and is shared across multiple stacks within our account.

It appears that SAM's default behavior is to automatically add resource-based permissions to grant my API Gateway permission to invoke the authorizer. I can see `AWS::Lambda::Permission` resources being created in my Cloudformation output, though none are actually defined in my template.yaml file. Normally, this behavior would be fine, however, when this occurs across multiple stacks, it causes the policy document on the shared lambda to grow to a level that eventually exceeds AWS size limits.

To work around this, I am attempting to use an IAM role to grant access to the shared lambda instead of resource-based permission. However, even when I specify a role to be used to invoke my shared lambda, SAM is still creating `AWS::Lambda::Permission` resources in Cloudformation.

Is there any way to prevent SAM from auto-generating these resource-based permissions?

My API definition is below. You'll note that both the `FunctionInvokeRole` and `InvokeRole` properties are set. I wouldn't expect any permission resources to be created if these properties are set with an IAM role.

```
PublicGateway:
    Type: AWS::Serverless::Api
    Properties:
      StageName: latest
      EndpointConfiguration: REGIONAL
      Cors:
        AllowHeaders: '''Content-Type,Authorization,X-Amz-Date,X-Api-Key,X-Amz-Security-Token'''
        AllowMethods: '''DELETE,GET,HEAD,OPTIONS,PATCH,POST,PUT'''
        AllowOrigin: '''*'''
      Auth:
        InvokeRole:
          Fn::GetAtt:
          - MyCustomRole
          - Arn
        Authorizers:
          MyAuthorizer:
            FunctionArn:
              Fn::Sub: arn:aws:lambda:${AWS::Region}:${AWS::AccountId}:function:my-shared-lambda-authorizer
            FunctionInvokeRole:
              Fn::GetAtt:
              - MyCustomRole
              - Arn
            FunctionPayloadType: REQUEST
            Identity:
              Headers:
              - Authorization
              ReauthorizeEvery: 600
        DefaultAuthorizer: MyAuthorizer
        AddDefaultAuthorizerToCorsPreflight: false
      DefinitionBody:
        Fn::Transform:
          Name: AWS::Include
          Parameters:
            Location: s3://some-s3-bucket-full-of-swagger/swagger.json
      Variables:
[...]

PR #2885, opened 2023-02-10
feat: New DisableFunctionDefaultPermissions property to block the creation of permissions resource from SAM API Auth

aaythapa:
### Issue #, if available

#1656 

### Description of changes

A new opt-in boolean property `DisableFunctionDefaultPermissions` added to the lambda authorizers ([LambdaRequestAuthorizer](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-property-api-lambdarequestauthorizer.html) and [LambdaTokenAuthorizer](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-property-api-lambdatokenauthorizer.html)) in SAM API that will block the creation of the `AWS::Lambda::Permissions` resource that allows the API to invoke the authorizer function (its not created for the Cognito authorizer).

The type changes may make the PR a bit confusing to review so the main changes are in `api_generator.py` and `apigateway.py`
### Checklist

    - [x] Using correct values
    - [x] Using wrong values

### Examples?


ssenchenko (approved):

samtranslator/model/apigateway.py
```diff
@@ -167,7 +173,7 @@ def generate_swagger(self) -> Py27Dict:
 
         return swagger
 
-    def _add_prefixes(self, response_parameters):  # type: ignore[no-untyped-def]
+    def _add_prefixes(self, response_parameters: Dict[Any, Any]) -> Dict[str, str]:
```
hoffa, line 176: Is `response_parameters` originally JSON? If yes it's `Dict[str, Any]`.

schema_source/aws_serverless_api.py
```diff
@@ -77,6 +77,7 @@ class LambdaTokenAuthorizer(BaseModel):
     FunctionInvokeRole: Optional[str] = lambdatokenauthorizer("FunctionInvokeRole")
     FunctionPayloadType: Optional[Literal["TOKEN"]] = lambdatokenauthorizer("FunctionPayloadType")
     Identity: Optional[LambdaTokenAuthorizerIdentity] = lambdatokenauthorizer("Identity")
+    DisableDefaultFunctionPermissions: Optional[bool]  # TODO Add docs
```
hoffa, line 80: What's the rationale for having this as an authorizer property? Is it specific to an authorizer config (more than to a function)?

aaythapa: Yea the property specific to the LambdaTokenAuthorizer and the LambdaRequestAuthorizer only since they're the ones that create the Lambda permissions resource

tests/translator/input/api_rest_auth_disable_default_function_permissions.yaml
```diff
@@ -0,0 +1,47 @@
...
+  MyApiWithLambdaTokenAuth:
+    Type: AWS::Serverless::Api
+    Properties:
+      StageName: Prod
+      Auth:
+        DefaultAuthorizer: MyLambdaTokenAuth
+        Authorizers:
+          MyLambdaTokenAuth:
+            FunctionArn: !GetAtt MyAuthFn.Arn
+            DisableDefaultFunctionPermissions: true
```
hoffa, line 12: Can probably rename it to `DisableDefaultPermissions`, since in both cases (if I'm reading correctly), the context is about Lambda functions? And you could parse the property as either "disable default (function permissions)" or "disable (default function) permissions", which have a different meaning.

aaythapa: Added `Function` to the name since it repeats it in the other [props](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-property-api-lambdarequestauthorizer.html) (like `FuncitonArn`, `FunctionInvokeRole` and `FunctionPayloadType`). Wouldn't it be easier for customers to connect the new property with the existing ones in the Lambda authorizer if it has `Function` in the name?

hoffa: Yeah, possible. It's pretty clear either way, so happy to keep it as-is.

samtranslator/schema/schema.json
```diff
@@ -188427,6 +188427,10 @@
           "title": "AuthorizationScopes",
           "type": "array"
         },
+        "DisableDefaultFunctionPermissions": {
```
hoffa, line 188430: On second thought, how about `DisableFunctionDefaultPermissions`? To prevent confusion with the non-existent "default function".

hoffa (approved):

samtranslator/validator/value_validator.py
```diff
@@ -120,5 +120,12 @@ def to_be_an_integer(self, message: Optional[str] = "") -> int:
         """
         return cast(int, self.to_be_a(ExpectedType.INTEGER, message))
 
+    def to_be_a_bool(self, message: Optional[str] = "") -> int:
```
hoffa, line 123:
```
-> int:
```

?

samtranslator/schema/schema.json
aaythapa: Had a chat with Evan about the name. He agrees that it should be `FunctionDefault` rather than `DefaultFunction` but would also prefer if we didn't use `Disable` in the name (we're trying to move away from that word). 

Some suggestions he had was `BlockDefaultAuthPermissions: True` or `NoDefaultAuthPermissions: True`. Although I think keeping `Disable` will keep the property consistent with some other SAM properties and be more familiar to customers. Thoughts on this?

aaythapa: Talked offline, we decided to go with `DisableFunctionDefaultPermissions` to keep consistency and familiarity

samtranslator/model/api/api_generator.py
```diff
@@ -1168,7 +1169,7 @@ def _construct_authorizer_lambda_permission(self) -> List[LambdaPermission]:
 
         for authorizer_name, authorizer in authorizers.items():
             # Construct permissions for Lambda Authorizers only
-            if not authorizer.function_arn:
+            if not authorizer.function_arn or authorizer.disable_function_default_permissions:
```
aahung, line 1172: whenever I see a boolean check, I wonder are we raise exception whenever it is an intrinsic 🤔

aahung (commented):
overall lgtm, one comment

samtranslator/model/api/api_generator.py
aaythapa: It should raise an exception if its not boolean, I added an error check and error test to make sure the value is boolean. If they use an intrinsic then adding Language Extensions should fix the exception if it pops up

aahung: Oh just saw `tests/translator/input/error_api_invalid_auth.yaml`.

aahung (approved):

hoffa (approved):

PR #2889, opened 2023-02-10
chore: Add decorator @deprecated and add it to is_str()

aahung:

Create a file "bin/test_deprecation.py" with the content:
```py
from samtranslator.model.types import is_str

if __name__ == "__main__":
    is_str()("hi")

```

Then:

```sh
(samtranslator37) ~/P/serverless-application-model ❯❯❯ python bin/test_deprecation.py
bin/test_deprecation.py:4: DeprecationWarning: is_str is deprecated, please consider to use IS_STR
  is_str()("hi")
```

### Checklist

    - [ ] Using correct values
    - [ ] Using wrong values

### Examples?

### DIFF
--- a/samtranslator/internal/deprecation_control.py
+++ b/samtranslator/internal/deprecation_control.py
@@ -17,7 +17,7 @@ RT = TypeVar("RT")  # return type
 
 
 def _make_message(message: str, replacement: Optional[str]) -> str:
-    return f"{message}, please use {replacement}" if replacement else f"{message} and there is no replacement."
+    return f"{message}, please use {replacement}" if replacement else message
 
 
 def deprecated(replacement: Optional[str]) -> Callable[[Callable[..., RT]], Callable[..., RT]]:
--- a/samtranslator/internal/deprecation_control.py
+++ b/samtranslator/internal/deprecation_control.py
@@ -33,7 +33,9 @@ def deprecated(replacement: Optional[str]) -> Callable[[Callable[..., RT]], Call
     def decorator(func: Callable[..., RT]) -> Callable[..., RT]:
         @wraps(func)
         def wrapper(*args, **kwargs) -> RT:  # type: ignore
-            warning_message = _make_message(f"{func.__name__} is deprecated", replacement)
+            warning_message = _make_message(
+                f"{func.__name__} is deprecated and will be removed in a future release", replacement
+            )
             # Setting stacklevel=2 to let Python print the line that calls
             # this wrapper, not the line below.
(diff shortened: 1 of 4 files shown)


ssenchenko (approved):

acristin (approved):

samtranslator/internal/deprecation_control.py
```diff
@@ -0,0 +1,67 @@
...
+            # this wrapper, not the line below.
+            warnings.warn(warning_message, DeprecationWarning, stacklevel=2)
+            return func(*args, **kwargs)
+
+        return wrapper
+
+    return decorator
+
+
+def pending_deprecation(replacement: Optional[str]) -> Callable[[Callable[..., RT]], Callable[..., RT]]:
```
hoffa, line 49: Why do we need this? It's not intuitive to me what this even means or why I would care.

samtranslator/internal/deprecation_control.py
```diff
@@ -0,0 +1,67 @@
...
+from functools import wraps
+from typing import Callable, Optional, TypeVar
+
+RT = TypeVar("RT")  # return type
+
+
+def _make_message(message: str, replacement: Optional[str]) -> str:
+    if replacement:
+        return f"{message}, please consider to use {replacement}"
+    return f"{message} and there is no replacement."
```
hoffa, line 22: More stern, to the point:

```suggestion
        return f"{message}, please use {replacement}"
    return message
```

samtranslator/internal/deprecation_control.py
aahung: removeed

samtranslator/internal/deprecation_control.py
aahung: updated

samtranslator/internal/deprecation_control.py
```diff
@@ -0,0 +1,44 @@
...
+"""
+import warnings
+from functools import wraps
+from typing import Callable, Optional, TypeVar
+
+RT = TypeVar("RT")  # return type
+
+
+def _make_message(message: str, replacement: Optional[str]) -> str:
+    return f"{message}, please use {replacement}" if replacement else f"{message} and there is no replacement."
```
hoffa, line 20: I would get rid of the "and there is no replacement". It's normal to deprecate things and not expect it to be replaced. Also prevents confusion if someone just forgot to add the `replacement`.

aahung: 👍

hoffa (approved):

PR #2906, opened 2023-02-14
feat: make AppSync::DataSource mappings (DynamoDB)

pradhapanr:
### Description of changes
Created new resource `Serverless::GraphQLDataSource`. 

The properties map to the `AppSync::DataSource` resource in CFN.

DynamoDB is the only supported datasource with more to be added later.

I will be adding the connector changes in a follow up PR once we add `AppSync::DataSource` to connector profiles.

Created schema changes and ran against new transform tests.

### Checklist

    - [x] Using correct values
    - [ ] Using wrong values

### Examples?


samtranslator/model/appsync.py
```diff
@@ -40,3 +73,22 @@ class GraphQLSchema(Resource):
     ApiId: Intrinsicable[str]
     Definition: Optional[str]
     DefinitionS3Location: Optional[str]
+
+
+class DataSource(Resource):
+    resource_type = "AWS::AppSync::DataSource"
+    property_types = {
+        "ApiId": PropertyType(True, IS_STR),
```
aaythapa, line 81: Think we use `Property` and `PassThroughProperty` now instead of `PropertyType`. https://github.com/aws/serverless-application-model/blob/develop/DEVELOPMENT_GUIDE.md#development-guidelines

samtranslator/model/sam_resources.py
```diff
@@ -2194,3 +2202,108 @@ def _construct_appsync_schema(self, api_id: Intrinsicable[str]) -> GraphQLSchema
...
+        "ApiId": Property(True, one_of(IS_STR, IS_DICT)),
+        "Name": Property(False, IS_STR),
+        "Type": Property(True, IS_STR),
+        "Description": PassThroughProperty(False),
+        "ServiceRoleArn": Property(False, IS_STR),
+        "Permissions": Property(False, list_of(IS_STR)),
+        "DataSourceConfig": Property(True, IS_DICT),
+    }
+
+    ApiId: Intrinsicable[str]
```
aaythapa, line 2223: Think we don't resolve intrinsics anymore since we recommend adding [AWS::LanguageExtensions](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-languageextension-transform.html) to the template so this shouldn't be `Intrinsicable`

samtranslator/model/sam_resources.py
```diff
@@ -2194,3 +2202,108 @@ def _construct_appsync_schema(self, api_id: Intrinsicable[str]) -> GraphQLSchema
...
+        )
+
+        if self.Type not in SUPPORTED_DATASOURCES:
+            raise InvalidResourceException(self.logical_id, f"'{self.Type}' is not a supported data source type.")
+
+        datasource.Type = self.Type
+        datasource.Name = self.Name if self.Name else self.logical_id
+        datasource.Description = self.Description
+        datasource.ApiId = self.ApiId
+        # I will remove this ignore after, it will be fixed with pseudo-embedded connector implementation.
```
aaythapa, line 2304: Don't understand this comment. Whats the pseudo-embedded connector implementation?

tests/translator/output/error_graphqldatasource.json
```diff
@@ -0,0 +1,29 @@
+{
+  "errorMessage": "Invalid Serverless Application Specification document. Number of errors found: 8. Resource with id [NoApiId] is invalid. Missing required property 'ApiId'. Resource with id [NoBaseTableTTL] is invalid. Property 'DataSourceConfig.DeltaSync.BaseTableTTL' should be a string. Resource with id [NoDataSourceConfig] is invalid. Missing required property 'DataSourceConfig'. Resource with id [NoDeltaSyncTableName] is invalid. Property 'DataSourceConfig.DeltaSync.DeltaSyncTableName' should be a string. Resource with id [NoDeltaSyncTableTTL] is invalid. Property 'DataSourceConfig.DeltaSync.DeltaSyncTableTTL' should be a string. Resource with id [NoDynamoDBTableName] is invalid. Property 'DataSourceConfig.DynamoDB.TableName' should be a string. Resource with id [NoType] is invalid. Missing required property 'Type'. Resource with id [UnknownType] is invalid. 'WEIRD_TYPE' is not a supported data source type.",
+  "errors": [
```
aaythapa, line 3: don't need this `errors` list, only need `errorMessage`

schema_source/aws_serverless_graphqldatasource.py
```diff
@@ -0,0 +1,40 @@
...
+    DeltaSyncTableName: str
+    DeltaSyncTableTTL: str
+
+
+class DynamoDB(BaseModel):
+    TableName: str
+    Region: Optional[str]
+    UseCallerCredentials: Optional[bool]
+    Versioned: Optional[bool]
+    DeltaSync: Optional[DeltaSync]
```
hoffa, line 22: Aren't these all pass-throughs?

samtranslator/model/appsync.py
pradhapanr: Whoops, forgot to reflect these changes in this module like for `sam_resources.py`. good catch

samtranslator/model/sam_resources.py
pradhapanr: If the user does not define the `ServiceRoleArn` property, then we will create a connector from `AppSync::DataSource` to whatever resource is defined in their config. That's not implemented yet, but i typed `ServiceRoleArn` as optional so its flagging it in mypy. But this will change once I add the connector stuff and the `type: ignore` will be removed.

samtranslator/model/appsync.py
hoffa: For some context, for SAM properties, we want to avoid `PropertyType` since (1) it avoids [sneaky intrinsics-related bugs](https://github.com/aws/serverless-application-model/pull/2495#discussion_r976715242) since we no longer need to resolve them (thanks `AWS::LanguageExtensions`) and (2) ensures we don't wrongly fail valid templates by supporting any values supported by the underlying property (which we don't have control over).

For CloudFormation resources (like in this case), it's arguably a little different, since we're the producer and sometimes care and know exactly what types we expected. I don't have anything to propose currently, but thinking whether these should all be `PassThroughProperty`, whether we even care about strictly typing these, or whether `PropertyType` here makes sense... 🤔

schema_source/aws_serverless_graphqldatasource.py
pradhapanr: There's some validation logic that happens inside the transform to ensure the appropriate properties in `DataSourceConfig` exist. That would make them not passthrough right?

samtranslator/model/appsync.py
```diff
@@ -7,11 +7,44 @@
...
 
+class DynamoDBDataSource(TypedDict, total=False):
+    Region: str
+    TableName: str
+    UseCallerCredentials: bool
+    Versioned: bool
+    DeltaSync: DeltaSync
+
+
+class DataSourceConfig(TypedDict, total=False):
```
GavinZZ, line 35: Curious why is this class `total=False`? My understanding is that `total=False` doesn't require all keys present when instantiating a class.

samtranslator/model/sam_resources.py
```diff
@@ -2194,3 +2202,108 @@ def _construct_appsync_schema(self, api_id: Intrinsicable[str]) -> GraphQLSchema
...
+        if "Versioned" in ddb_properties:
+            ddb_config["Versioned"] = ddb_properties["Versioned"]
+
+        return cast(DynamoDBConfig, ddb_config)
+
+    def _validate_config_properties(self, datasource: DataSource) -> None:
+        # DataSourceConfig is quite large property so we require a lot of additional validation.
+        # The datasource object is modified by reference.
+
+        if len(self.DataSourceConfig) > 1:
```
GavinZZ, line 2281: Just want to confirm what happens if there is no `DataSourceConfig`

samtranslator/model/appsync.py
hoffa: So the problem with this whole `Resource` thing is that it's not statically typed. It's runtime type checks, which can cause runtime failures:

https://github.com/aws/serverless-application-model/blob/83306d38ff957152f7f274eef3b9b9abd873101f/samtranslator/model/types.py#L34-L40

That's why I'm a little more comfortable preferring lenient "types" such as `PropertyType` for producers which "supports" intrinsics by defaulting `self.supports_intrinsics` to true:

https://github.com/aws/serverless-application-model/blob/c39c2807bbf327255de8abed8b8150b18c60f053/samtranslator/model/__init__.py#L306-L308

It avoids risk of us prematurely failing something valid. And if there really is a problem, the CloudFormation resource knows better and will fail anyway.

So ideally the producer should be all regular, static Python types. Not sure how big of a task that is. But until then being more lenient (e.g. `PropertyType` or `PassThroughProperty`) is probably better.

samtranslator/model/sam_resources.py
```diff
@@ -2194,3 +2202,108 @@ def _construct_appsync_schema(self, api_id: Intrinsicable[str]) -> GraphQLSchema
...
+            raise InvalidResourceException(
+                self.logical_id, "Only one resource can be defined at a time in 'DataSourceConfig'."
+            )
+
+        if datasource.Type == "AMAZON_DYNAMODB":
```
GavinZZ, line 2286: Can we define a parameter for `AMAZON_DYNAMODB` and use it here and when we define `SUPPORTED_DATASOURCES` instead of hardcoding.

samtranslator/model/sam_resources.py
```diff
@@ -2194,3 +2202,108 @@ def _construct_appsync_schema(self, api_id: Intrinsicable[str]) -> GraphQLSchema
...
+    def _validate_dynamodb_datasource(self, ddb_properties: Dict[str, Any]) -> DynamoDBConfig:
+        ddb_config = {}
+
+        ddb_config["TableName"] = sam_expect(
+            ddb_properties.get("TableName"),
+            self.logical_id,
+            "DataSourceConfig.DynamoDB.TableName",
+        ).to_be_a_string()
+
+        ddb_config["AwsRegion"] = ddb_properties["Region"] if "Region" in ddb_properties else ref("AWS::Region")
```
GavinZZ, line 2246: Maybe I'm wrong, but I don't think `ref("AWS::Region")` will work.

samtranslator/model/sam_resources.py
```diff
@@ -2194,3 +2202,108 @@ def _construct_appsync_schema(self, api_id: Intrinsicable[str]) -> GraphQLSchema
...
+
+            ddb_config["DeltaSyncConfig"] = deltasync_properties
+
+        if "UseCallerCredentials" in ddb_properties:
+            ddb_config["UseCallerCredentials"] = ddb_properties["UseCallerCredentials"]
+
+        if "Versioned" in ddb_properties:
+            ddb_config["Versioned"] = ddb_properties["Versioned"]
+
+        return cast(DynamoDBConfig, ddb_config)
```
GavinZZ, line 2275: The name of this function maybe misleading, it says `validate...` but it's returning a value. Generally if a function is validating, I wouldn't expect it to return any value.

schema_source/aws_serverless_graphqldatasource.py
hoffa: Possibly. We should type as much as the transform logic cares about. So if it needs to parse or manipulate a value (not simply pass it to another property), it expects a certain structure, hence should be typed.

samtranslator/model/sam_resources.py
pradhapanr: I tested a transform and deployed it, worked fine and pointed to the correct table in the stack region.

samtranslator/model/appsync.py
pradhapanr: Currently that is true but we will be adding support for more data sources. I'm going to leave it for now since it will change in the near future and it's not impacting anything.

schema_source/aws_serverless_graphqldatasource.py
pradhapanr: I can make `UseCallerCredentials` and `Versioned` passthrough on second thought, but the others do have some parsing so I'll leave those alone.

samtranslator/model/sam_resources.py
pradhapanr: That's a good point. I think I'll change to `parse` prefix instead

schema_source/aws_serverless_graphqldatasource.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+# TODO: add docs
+class DeltaSync(BaseModel):
+    BaseTableTTL: str
+    DeltaSyncTableName: str
+    DeltaSyncTableTTL: str
+
+
+class DynamoDB(BaseModel):
```
ssenchenko, line 17: Let's have consistent names, ie this one is `DynamoDBConfig` in the model. It's confusing to see `DynamoDB` class

samtranslator/model/sam_resources.py
```diff
@@ -2194,3 +2202,108 @@ def _construct_appsync_schema(self, api_id: Intrinsicable[str]) -> GraphQLSchema
...
+        "Type": Property(True, IS_STR),
+        "Description": PassThroughProperty(False),
+        "ServiceRoleArn": Property(False, IS_STR),
+        "Permissions": Property(False, list_of(IS_STR)),
+        "DataSourceConfig": Property(True, IS_DICT),
+    }
+
+    ApiId: Intrinsicable[str]
+    Type: str
+    DataSourceConfig: DataSourceConfig
```
ssenchenko, line 2225: I think you want to wrap Configs in additional `dict` to make validation easier, but from UX perspective one more level doesn't make much sense 
```yaml
DataSourceConfig:
  DynamoDb:
    UseCallerCredentials: True
```
From `Type` field, we already know, which fields should be in the config. So 
```yaml
Config:
  UseCallerCredentials: True
```
It seems cleaner. And I'd call it simply `Config` because it's in `DataSource` already. If `Config` sounds too general, I guess `ConnectionConfig` would be more precise

samtranslator/model/sam_resources.py
```diff
@@ -2194,3 +2202,108 @@ def _construct_appsync_schema(self, api_id: Intrinsicable[str]) -> GraphQLSchema
...
+
+        ddb_config["TableName"] = sam_expect(
+            ddb_properties.get("TableName"),
+            self.logical_id,
+            "DataSourceConfig.DynamoDB.TableName",
+        ).to_be_a_string()
+
+        ddb_config["AwsRegion"] = ddb_properties["Region"] if "Region" in ddb_properties else ref("AWS::Region")
+
+        if "DeltaSync" in ddb_properties:
```
ssenchenko, line 2248: nit: Let's make DeltaSync validation a separate method. It'll look simpler

samtranslator/model/sam_resources.py
pradhapanr: That's an interesting idea. What would documentation look like for the `Config` property? I'm concerned that it would make that property bloated having it support all the different datasource properties.

schema_source/aws_serverless_graphqldatasource.py
pradhapanr: I was concerned with naming it `DynamoDB` in model since it was being imported into the `sam_resources` module, and it's strictly related to DataSource. I agree though, this naming is strange, let me play around with it  more and figure out something better.

samtranslator/model/appsync.py
pradhapanr: We're moving ahead with a different approach for type checking in CFN resources, since what we currently use can be  misleading. View this PR for more info:

https://github.com/aws/serverless-application-model/pull/2907

samtranslator/model/sam_resources.py
pradhapanr: This solution feels like a lot of constants to me for values that aren't being used frequently. For the most part, these types are only ever being used once. Why shouldn't we hardcode in that case?

samtranslator/model/sam_resources.py
```diff
@@ -2194,3 +2202,108 @@ def _construct_appsync_schema(self, api_id: Intrinsicable[str]) -> GraphQLSchema
...
+            deltasync_properties = ddb_properties["DeltaSync"]
+
+            sam_expect(
```
aaythapa, line 2251: aren't these all pass through properties? Do we need to run validation on them?

samtranslator/model/sam_resources.py
pradhapanr: I agree that adding another layer is not necessary. But I'm going to keep the configs separate. I just think the property gets too bloated and it could be more of a detriment to customers than a benefit.

samtranslator/model/sam_resources.py
pradhapanr: They're all required properties if `DeltaSync` is defined by the customer, so I added validation to confirm the fields exist.

samtranslator/model/sam_resources.py
GavinZZ: I agree that if it's only used once, there is no need to parameterize it. But I've already seen this value used twice in this PR and I suspect we're going to use `AMAZON_DYNAMODB` data source type in the future PRs as well..

pradhapanr: Fair point. I think I'll leave it as is for now and update in future PR if there's frequent reuse happening.

samtranslator/model/sam_resources.py
```diff
@@ -2194,3 +2202,108 @@ def _construct_appsync_schema(self, api_id: Intrinsicable[str]) -> GraphQLSchema
...
+    ServiceRoleArn: Optional[str]
+
+    @cw_timer
+    def to_cloudformation(self, **kwargs: Any) -> List[Resource]:
+        appsync_datasource = self._construct_appsync_datasource()
+        resources: List[Resource] = [appsync_datasource]
+
+        return resources
+
+    def _validate_dynamodb_datasource(self, ddb_properties: Dict[str, Any]) -> DynamoDBConfig:
```
hoffa, line 2237: Something to think about: how do we make this more generic, so that we don't always need to implement this logic? We already have a generic way of validating at the surface level (using `Property`, etc.).

pradhapanr: Me and @ssenchenko  discussed making a more generic way to validate nested properties like `DynamoDBConfig` using `TypedDict`, but it looked a bit too time consuming to tackle on myself at this early stage of the project. I did mark it down as a stretch goal though.

samtranslator/model/sam_resources.py
aaythapa: Discussed offline, these properties will have default values so we need validation incase the user tries to override them

samtranslator/model/appsync.py
```diff
@@ -7,9 +7,23 @@
...
 from samtranslator.utils.types import Intrinsicable
 
+# Data source constants can be found here under "Type" property:
+# https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-appsync-datasource.html
+SUPPORTED_DATASOURCES = {"AMAZON_DYNAMODB"}
 
-class Auth(TypedDict, total=False):
-    Type: str
+
+class DeltaSyncConfigType(TypedDict):
```
hoffa, line 15: This applies to any new classes/functions/variables. We should avoid putting stuff under anywhere else than `internal` if possible. People depend on the library in strangest of ways, so that's one obvious place that conveys we don't provide any guarantees.

pradhapanr: Offline discussion: Moving the `appsync` module to internal is trivial, so I have updated as such. Moving the `sam_resources` macros are not, the translator searches this module for all macros when resolving which makes separation difficult. I will look later to see if we can add multiple modules.

samtranslator/model/sam_resources.py
```diff
@@ -2194,3 +2201,106 @@ def _construct_appsync_schema(self, api_id: Intrinsicable[str]) -> GraphQLSchema
...
+        sam_expect(
+            deltasync_properties.get("DeltaSyncTableTTL"),
+            self.logical_id,
+            "DynamoDBConfig.DeltaSyncConfig.DeltaSyncTableTTL",
+        ).to_be_a_string()
+
+        return cast(DeltaSyncConfigType, deltasync_properties)
+
+    def _parse_dynamodb_datasource(self, ddb_properties: Dict[str, Any]) -> DynamoDBConfigType:
+        ddb_config: Dict[str, Any] = {}
```
hoffa, line 2255: IMO rather than doing this and `cast`ing, we should build the `TypedDict` directly.

pradhapanr: Like just change the typing on `ddb_config` to `DynamoDBConfigType` and return without `cast`?

hoffa: Maybe, but not sure if `mypy` is smart enough since it's also kind of an assertion. I was thinking something like:

```
return DynamoDBConfig(
    Versioned=...,
    ...
)
```

pradhapanr: I thought of doing something like that, but optional properties like `Versioned` aren't always defined, so creating the dict in just 1 call isn't possible I think.

If I was to set the value to `None` as a default, it shows up as `null` in the transformed template as well which is obviously not appropriate.

samtranslator/model/sam_resources.py
```diff
@@ -2194,3 +2201,106 @@ def _construct_appsync_schema(self, api_id: Intrinsicable[str]) -> GraphQLSchema
...
+    ServiceRoleArn: Optional[str]
+
+    @cw_timer
+    def to_cloudformation(self, **kwargs: Any) -> List[Resource]:
+        appsync_datasource = self._construct_appsync_datasource()
+        resources: List[Resource] = [appsync_datasource]
+
+        return resources
+
+    def _parse_deltasync_properties(self, deltasync_properties: Dict[str, Any]) -> DeltaSyncConfigType:
```
aaythapa, line 2235: TODO to add default values?

samtranslator/model/sam_resources.py
```diff
@@ -2194,3 +2201,106 @@ def _construct_appsync_schema(self, api_id: Intrinsicable[str]) -> GraphQLSchema
...
+        datasource = DataSource(
+            logical_id=self.logical_id, depends_on=self.depends_on, attributes=self.resource_attributes
+        )
+
+        if self.Type not in SUPPORTED_DATASOURCES:
+            raise InvalidResourceException(self.logical_id, f"'{self.Type}' is not a supported data source type.")
+
+        datasource.Type = self.Type
+        datasource.Name = self.Name if self.Name else self.logical_id
+        print("blahblahblahablahlahl")
```
ssenchenko, line 2298: Probably to be removed

samtranslator/model/sam_resources.py
hoffa: You can still assign stuff to typed variables before creating the final `TypedDict`.

> If I was to set the value to None as a default, it shows up as null in the transformed template as well which is obviously not appropriate.

How is that related to `TypedDict`?

samtranslator/model/sam_resources.py
pradhapanr: debugging gone wrong

samtranslator/internal/model/appsync.py
```diff
@@ -0,0 +1,74 @@
+from typing import Any, Dict, List, Optional
```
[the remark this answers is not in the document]
pradhapanr: Addressed here: https://github.com/aws/serverless-application-model/pull/2906#discussion_r1107410780

samtranslator/model/sam_resources.py
pradhapanr: Offline discussion: The `TypedDict` is partial, so we cant make one call that suits all combinations. I will change type of `ddb_config` to `DynamoDBConfigType` and remove the cast.

samtranslator/model/sam_resources.py
```diff
@@ -2149,7 +2156,7 @@ class SamGraphQLApi(SamResourceMacro):
         "SchemaUri": Property(False, IS_STR),
     }
 
-    Auth: Auth
+    Auth: Dict[str, Any]
```
hoffa, line 2159: Why? Is the value `Any`?

samtranslator/model/sam_resources.py
```diff
@@ -2194,3 +2201,106 @@ def _construct_appsync_schema(self, api_id: Intrinsicable[str]) -> GraphQLSchema
...
+
+        if self.Type not in SUPPORTED_DATASOURCES:
+            raise InvalidResourceException(self.logical_id, f"'{self.Type}' is not a supported data source type.")
+
+        datasource.Type = self.Type
+        datasource.Name = self.Name if self.Name else self.logical_id
+        datasource.Description = self.Description
+        datasource.ApiId = self.ApiId
+        # I will remove this ignore after, it will be fixed with pseudo-embedded connector implementation.
+        datasource.ServiceRoleArn = self.ServiceRoleArn  # type: ignore
```
hoffa, line 2302: > type: ignore

Why?

pradhapanr: `ServiceRoleArn` is an optional property for the SAM resource, but is required on CFN resource.

This is because we are going to generate the role and policy if the customer does not define it. But that has not been implemented yet, so we're technically assigning an optional value to a required one which is a typing error. Once I create the connector change and push that, this value will always be defined so the `# type: ignore` will be removed.

I could change all the types to account for this now and then change them back with the connector PR but I don't think it's necessary since there's no customer impact at this time.

samtranslator/model/sam_resources.py
pradhapanr: @aahung brought up a good point when I was talking with him. Typing nested SAM properties (dictionaries) is not great because we haven't verified at this step what is actually in the dictionary, so it can do more harm than good to type it as such. So I am just using a very broad dictionary typing instead.

The CloudFormation resource typings are fine to use because we have a bunch of validation before hand to confirm they are of appropriate shape, so those will remain.

ssenchenko (approved):

hoffa (approved):

PR #2912, opened 2023-02-15
Adding 'Region' and 'SetIdentifier' to Route53Configuration to support latency-based routing

davidcolclazier:
### Description of changes

This PR adds the necessary logic to support latency-based routing for Route53 record sets created by the following AWS entities:
    - AWS::Serverless::Api
    - AWS::Serverless::HttpApi

This PR adds two new properties - `SetIdentifier` and `Region` to the Route53Configuration within the CustomDomainConfiguration of the above entities. If both are provided, the record sets will have a latency-based routing policy. If only the Region is provided, an error is raised. If only SetIdentifier is provided, it is ignored. This allows for future policy implementation without modifying existing logic.

Validation included the following:
  - added unit tests for both entities - passing
  - added integration tests , deployed to AWS using personal domain/cert - passing
  - ensure code coverage remains over 95% - passing 
### Checklist

    - [x] Using correct values
    - [x] Using wrong values

### Examples?

  Happy to provide an example if necessary.


davidcolclazier:
Looks like the black-check process is failing - I ran `make black` successfully [(see here)](https://github.com/aws/serverless-application-model/pull/2912/commits/67fc00292c03203e11a776018ca8274cab494822), so not sure how to resolve this... Looking into it on my end! Any ideas?

hoffa:
> Looks like the black-check process is failing - I ran `make black` successfully [(see here)](https://github.com/aws/serverless-application-model/pull/2912/commits/67fc00292c03203e11a776018ca8274cab494822), so not sure how to resolve this... Looking into it on my end! Any ideas?

Did you activate the venv? Otherwise it'll use whatever `black` that shows up in your PATH, which could be a different version.

davidcolclazier:
> > Looks like the black-check process is failing - I ran `make black` successfully [(see here)](https://github.com/aws/serverless-application-model/pull/2912/commits/67fc00292c03203e11a776018ca8274cab494822), so not sure how to resolve this... Looking into it on my end! Any ideas?
> 
> Did you activate the venv? (e.g. `python3 -m venv .venv && source .venv/bin/activate && make init && make black`) Otherwise it'll use whatever `black` that shows up in your PATH, which could be a different version.

I believe so, but reverifying in a new environment now - I'll let you know!

davidcolclazier:
> > > Looks like the black-check process is failing - I ran `make black` successfully [(see here)](https://github.com/aws/serverless-application-model/pull/2912/commits/67fc00292c03203e11a776018ca8274cab494822), so not sure how to resolve this... Looking into it on my end! Any ideas?
> > 
> > 
> > Did you activate the venv? (e.g. `python3 -m venv .venv && source .venv/bin/activate && make init && make black`) Otherwise it'll use whatever `black` that shows up in your PATH, which could be a different version.
> 
> I believe so, but reverifying in a new environment now - I'll let you know!

Not sure how it was missed, but should be resolved now. Thanks!

davidcolclazier:
> > Looks like the black-check process is failing - I ran `make black` successfully [(see here)](https://github.com/aws/serverless-application-model/pull/2912/commits/67fc00292c03203e11a776018ca8274cab494822), so not sure how to resolve this... Looking into it on my end! Any ideas?
> 
> Did you activate the venv? (e.g. `python3 -m venv .venv && source .venv/bin/activate && make init && make black`) Otherwise it'll use whatever `black` that shows up in your PATH, which could be a different version.

What tool do you use for JSON formatting? the format check is failing for these files:
integration/config/file_to_s3_map_modified.json
integration/resources/expected/combination/api_with_custom_domains_regional_latency_routing.json
integration/resources/expected/combination/api_with_custom_domains_regional_latency_routing_ipv6.json

Yaml checks are passing, so I think this is the last piece - apologies for the confusion!

hoffa:
> > > Looks like the black-check process is failing - I ran `make black` successfully [(see here)](https://github.com/aws/serverless-application-model/pull/2912/commits/67fc00292c03203e11a776018ca8274cab494822), so not sure how to resolve this... Looking into it on my end! Any ideas?
> > 
> > 
> > Did you activate the venv? (e.g. `python3 -m venv .venv && source .venv/bin/activate && make init && make black`) Otherwise it'll use whatever `black` that shows up in your PATH, which could be a different version.
> 
> What tool do you use for JSON formatting? the format check is failing for these files: integration/config/file_to_s3_map_modified.json integration/resources/expected/combination/api_with_custom_domains_regional_latency_routing.json integration/resources/expected/combination/api_with_custom_domains_regional_latency_routing_ipv6.json


It should "just work" by running `make black`, as it also formats JSON/YAML files:

https://github.com/aws/serverless-application-model/blob/5c68e889cbb60c809998f68b260615af392d955c/Makefile#L20-L24

> Yaml checks are passing, so I think this is the last piece - apologies for the confusion!

No problem at all, the contribution is appreciated! And if the process doesn't work intuitively perhaps it can be a good sign that we can improve it further.

davidcolclazier:
> > > > Looks like the black-check process is failing - I ran `make black` successfully [(see here)](https://github.com/aws/serverless-application-model/pull/2912/commits/67fc00292c03203e11a776018ca8274cab494822), so not sure how to resolve this... Looking into it on my end! Any ideas?
> > > 
> > > 
> > > Did you activate the venv? (e.g. `python3 -m venv .venv && source .venv/bin/activate && make init && make black`) Otherwise it'll use whatever `black` that shows up in your PATH, which could be a different version.
> > 
> > 
> > What tool do you use for JSON formatting? the format check is failing for these files: integration/config/file_to_s3_map_modified.json integration/resources/expected/combination/api_with_custom_domains_regional_latency_routing.json integration/resources/expected/combination/api_with_custom_domains_regional_latency_routing_ipv6.json
> 
> It should "just work" by running `make black`, as it also formats JSON/YAML files:
> 
> https://github.com/aws/serverless-application-model/blob/5c68e889cbb60c809998f68b260615af392d955c/Makefile#L20-L24
> 
> > Yaml checks are passing, so I think this is the last piece - apologies for the confusion!
> 
> No problem at all, the contribution is appreciated! And if the process doesn't work intuitively perhaps it can be a good sign that we can improve it further.

Yep, figured it out - these files had Windows line endings (had to set them to unix in order to execute them):
- bin\json-format.py
- bin\yaml-format.py
- bin\run_cfn_lint.sh

integration/resources/templates/combination/api_with_custom_domains_regional_latency_routing.yaml
```diff
@@ -0,0 +1,56 @@
+Parameters:
```
hoffa, line 1: Please add both templates as transform tests as well, so that it's clear to what it transforms:

```
python bin/add_transform_test.py --template-file integration/resources/templates/combination/api_with_custom_domains_regional_latency_routing.yaml
python bin/add_transform_test.py --template-file integration/resources/templates/combination/api_with_custom_domains_regional_latency_routing_ipv6.yaml
```

samtranslator/model/api/api_generator.py
```diff
@@ -566,22 +568,32 @@ def _construct_record_sets_for_domain(
...
 
         return recordset_list
 
+    def _construct_route53_routing_policy_properties(
+        self, route53_config: Dict[str, Any], recordSet: Dict[str, Any]
+    ) -> Dict[str, Any]:
+        if route53_config.get("RoutingPolicy") is not None and route53_config.get("RoutingPolicy") == "Latency":
+            region = boto3.session.Session().region_name
+            recordSet["SetIdentifier"] = region
+            recordSet["Region"] = region
```
hoffa, line 593: Thoughts on simply adding `SetIdentifier` and `Region` instead? It's not clear to me why setting those enables latency-based routing, and it might turn out to be too much maintenance in the future, if for whatever reason a different combination is desired.

schema_source/aws_serverless_api.py
```diff
@@ -131,6 +131,7 @@ class Route53(BaseModel):
     HostedZoneId: Optional[PassThroughProp] = route53("HostedZoneId")
     HostedZoneName: Optional[PassThroughProp] = route53("HostedZoneName")
     IpV6: Optional[bool] = route53("IpV6")
+    RoutingPolicy: Optional[str] = route53("RoutingPolicy")
```
hoffa, line 134: You can remove the assignment, we'll add it once https://github.com/awsdocs/aws-sam-developer-guide is updated (otherwise it'll show an empty description in the schema):

```suggestion
    RoutingPolicy: Optional[str]  # TODO: add docs
```

schema_source/aws_serverless_httpapi.py
```diff
@@ -86,6 +86,7 @@ class Route53(BaseModel):
     HostedZoneId: Optional[PassThroughProp] = route53("HostedZoneId")
     HostedZoneName: Optional[PassThroughProp] = route53("HostedZoneName")
     IpV6: Optional[bool] = route53("IpV6")
+    RoutingPolicy: Optional[str] = route53("RoutingPolicy")
```
hoffa, line 89:
```suggestion
    RoutingPolicy: Optional[str]  # TODO: add docs
```

samtranslator/model/api/api_generator.py
davidcolclazier: This was my first pass, but I decided to change it for a couple of reasons:

 - The other properties (i.e. HostedZoneName) in Route53 config are technically for the Route53::RecordSetGroup, where these properties are for the Route53::RecordSet.. is it confusing to combine these properties into the same entity? If not, this may be the way to go - see below.
 
- According to [this documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-route53-recordset-1.html#:~:text=%3A%20No%20interruption-,Region,-Latency%2Dbased%20resource), `Region` is only valid for Latency-based routing policies. And [here](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-route53-recordset-1.html#:~:text=%3A%20No%20interruption-,SetIdentifier,-Resource%20record%20sets) you'll see that the `SetIdentifier` is provided for any policy type other than `Simple`. This allows us to add additional routing policies without requiring the user to look up what in particular is required for a given policy. 

However, it requires us to lock down the region, which some users may not like. In addition, it becomes more complicated with the other policies (think GeoIp). 

If you're ok with combining the RecordSetGroup and RecordSet properties, it covers everything - happy to refactor it as such.

davidcolclazier: That said, going the other route introduces an interesting predicament as well - setting `Region` and `GeoLocation` together is invalid (and is true for every policy-specific property) - should we be validating at this level as well?

schema_source/aws_serverless_httpapi.py
```diff
@@ -86,6 +86,8 @@ class Route53(BaseModel):
     HostedZoneId: Optional[PassThroughProp] = route53("HostedZoneId")
     HostedZoneName: Optional[PassThroughProp] = route53("HostedZoneName")
     IpV6: Optional[bool] = route53("IpV6")
+    SetIdentifier: Optional[str]  # TODO: add docs
+    Region: Optional[str]  # TODO: add docs
```
hoffa, line 90: These are both pass-throughs (i.e. the value, which could be any arbitrary intrinsic function, is passed as-is to the underlying CloudFormation which we don't control):

```suggestion
    SetIdentifier: Optional[PassThroughProp]  # TODO: add docs
    Region: Optional[PassThroughProp]  # TODO: add docs
```

tests/model/api/test_http_api_generator.py
```diff
@@ -325,6 +326,75 @@ def test_basic_route53(self):
         self.assertIsNotNone(route, None)
         self.assertEqual(domain.DomainNameConfigurations[0].get("EndpointType"), "REGIONAL")
 
+    @patch("boto3.session.Session.region_name", "eu-west-2")
```
hoffa, line 329: nit: we tend to prefer regular transform tests (i.e. check a SAM template transforms into an expected CloudFormation template), as the tests are more thorough (as technically others parts of the transform logic could transform the template as well), they're less coupled to implementation details, and they go through other tests as well (specifically, schema tests)

samtranslator/model/api/http_api_generator.py
```diff
@@ -435,6 +435,20 @@ def _construct_record_sets_for_domain(
...
 
+    def _construct_route53_routing_policy_properties(
+        self, route53_config: Dict[str, Any], recordset: Dict[str, Any]
+    ) -> Dict[str, Any]:
+        if route53_config.get("Region") is not None:
+            if route53_config.get("SetIdentifier") is None:
+                raise InvalidResourceException(
+                    self.logical_id,
+                    "SetIdentifier is required in order to enable latency based routing on Route53 recordsets.",
+                )
```
hoffa, line 446: Is this test necessary?

I have no strong preference either way (it's often fine to let the underlying resource figure out if it's wrong), but if we do add this, there are 2 minor caveats:

1. We should add a transform test for this error case (see https://github.com/aws/serverless-application-model/pull/2885 for an example, basically create a transform test template prefixed with `error_`, and in the output create a JSON with a top-level `errorMessage` with the expected error message)
2. If in the future `Region` becomes usable without `SetIdentifier`, it'll be a poor experience until we change the logic (releases can take some time)

davidcolclazier: As CloudFormation performs equivalent validation after the template is transformed upon second thought I don't think this test is necessary; especially given your second point. Thanks!

tests/model/api/test_http_api_generator.py
davidcolclazier: No worries - I'll remove these. Thanks!

davidcolclazier:
@hoffa I'm having an issue with the transform tests failing - the logical ids for various resources don't match. I'm trying to figure out how to resolve this and am coming up blank. Any ideas?

E                                                         'EndpointConfiguration': {'Types': ['REGIONAL']},
E                                                         'Parameters': {'endpointConfigurationTypes': 'REGIONAL'}},
E                                          'Type': 'AWS::ApiGateway::RestApi'},
E   -                'ServerlessRestApiDeployment0173db1f08': {'Properties': {'Description': 'RestApi '
E   ?                                            ^^ ^^ ^^^^
E
E   +                'ServerlessRestApiDeploymentd7cef5b922': {'Properties': {'Description': 'RestApi '
E   ?                                            ^ ^^^^ ^^^

GavinZZ:
@davidcolclazier Thanks for reporting the issue. There is a problem with the existing `add_transform_tests` script where it doesn't generate the correct deployment id for `cn, us-gov` partitions. I'm working on a fix, see https://github.com/aws/serverless-application-model/pull/2984.

PR #2003, opened 2021-04-27
feat(policy-templates) Add QLDBSendCommandPolicy

folosada:
*Issue:* #2004 

*Description of changes:* Add a new policy template "QLDBSendCommandPolicy"

*Description of how you validated changes:*

*Checklist:*

- [X] Write/update tests
- [ ] `make pr` passes
- [ ] Update documentation
- [ ] Verify transformed template deploys and application functions as expected

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #2918, opened 2023-02-17
Add Connector from Lambda to QLDB

awood45:
Addresses and replaces #2003


integration/resources/templates/combination/connector_function_to_qldb_read.yaml
```diff
@@ -0,0 +1,42 @@
+Resources:
```
hoffa, line 1: Can you add these as transform tests as well?

```bash
python bin/add_transform_test.py --template-file integration/resources/templates/combination/connector_function_to_qldb_read.yaml
python bin/add_transform_test.py --template-file integration/resources/templates/combination/connector_function_to_qldb_write.yaml
```

acristin (changes requested):
I think you need to add a QLDB item [here](https://github.com/aws/serverless-application-model/blob/13604cd2d9671cd6e774e5bfd610a03d82a08d76/integration/helpers/resource.py#L212)

acristin (approved):

hoffa (approved):

ssenchenko (approved):

GavinZZ (approved):

Issue #2844, opened 2023-02-01
SESBulkTemplatedCrudPolicy lacks template permission


sukrit007:
### Description

The SES Policy templates SESBulkTemplatedCrudPolicy lacks resource for template ARN when performing ses:SendTemplatedEmail .  

https://github.com/aws/serverless-application-model/blob/develop/samtranslator/policy_templates_data/policy_templates.json#L1890

### Steps to reproduce
Use the SESBulkTemplatedCrudPolicy and specify verified Identity as parameter.

### Observed result
Lambda fails to send email,  as it lacks ses:SendTemplatedEmail on template. 

After manually modifying the policy and also adding ARN for template,  it works

### Expected result

We need to add TemplateName as parameter to  SESBulkTemplatedCrudPolicy

### Additional environment details

1. OS:  Mac 12.6.2
2. If using the [SAM CLI](https://github.com/aws/aws-sam-cli), `sam --version`: 1.67.0
3. AWS region: us-east-2

awood45:
We'll probably need to make a new policy template for this, because unfortunately while we would need to add this to get the resource constraint applied, it would break any templates that have this and have applied a workaround. Looking in to adding a policy template for this.

berger501:
Hello,
  i discovered this issue because we are using sam with SESBulkTemplatedCrudPolicy and we received a SES Security notification because of that i'm assuming that the policy created by sam will not be sufficient to send emails after 4th of april because no resource for the template is created. If i understand it correctly, this issue should fix this?

acristin:
Hi, I'm not familiar with the specific campaign from SES, but we expect that the new policy template will suffice. This will be one of the acceptance criteria for the change.

PR #2929, opened 2023-02-22
feat: new SESBulkTemplatedCrudPolicy_v2 policy

aaythapa:
### Issue #, if available

#2844 

### Description of changes

Added a new version of the `SESBulkTemplatedCrudPolicy` policy. `SESBulkTemplatedCrudPolicy_v2` is same but with template arn added to the resource section.

Made an example template and manually changed the policy to include the template arn in the resource section. No errors showed.

### Checklist

    - [x] Using correct values
    - [ ] Using wrong values

### Examples?


hoffa (approved):

samtranslator/policy_templates_data/policy_templates.json
```diff
@@ -1920,6 +1920,54 @@
         }
       }
     },
+    "SESBulkTemplatedCrudPolicy_v2": {
```
ssenchenko, line 1923: I believe we decided that `_v2` should be a new file, not a new policy

ssenchenko (changes requested):

samtranslator/policy_templates_data/policy_templates.json
aaythapa: going to double check this

SimonCMoore: hmm actually I thought it was just a new policy with _v2 in the same file rather than a new file. (I was thinking each was in its own file rather than all in one big file, not sure we really want to copy paste the whole thing to add one more policy).  Others should weigh in.

hoffa: This is a regular additive change, I see no benefit to a separate file.

ssenchenko: There is a difference if we set up an automated check for policy templates as we have for profiles. Having all versions in one file will makes running those checks much harder in case an older version is considered problematic, and a new one has a fix and both are in the same file.

awood45: I think that's solvable - same file is going to be a lot simpler, although as we scale some sort of breaking apart (perhaps by logical groupings like service) will be easier to comprehend and to maybe add the possibility of lazy loading if/when performance becomes an issue.

samtranslator/policy_templates_data/policy_templates.json
```diff
@@ -1920,6 +1920,54 @@
...
+              "ses:SendRawEmail",
+              "ses:SendTemplatedEmail",
+              "ses:SendBulkTemplatedEmail",
+              "ses:VerifyEmailIdentity"
+            ],
+            "Effect": "Allow",
+            "Resource": [
+              {
+                "Fn::Sub": [
+                  "arn:${AWS::Partition}:ses:${AWS::Region}:${AWS::AccountId}:identity/${identityName}",
```
awood45, line 1939: Have we tested this? Many of these actions do not support a resource constraint, which might cause them to fail if it is present. You may need to divide the actions amongst resource-constrained and `*`-constrained depending on what resource types are supported in this table: https://docs.aws.amazon.com/service-authorization/latest/reference/list_amazonses.html

awood45 (changes requested):

samtranslator/policy_templates_data/policy_templates.json
aaythapa: Great catch, you were right. The actions `ses:GetIdentityVerificationAttributes` and `ses:VerifyEmailIdentity` don't support resource constraints so those action will fail. They're also part of the `SESBulkTemplatedCrudPolicy` and the `SESCrudPolicy` which both have resource constrains so they'll fail with those policies as well. 

Going to take those actions out of this policy and create a separate, *-constrained, policy for them.

ssenchenko (approved):
After discussions with @SimonCMoore and @hoffa I think it's okay to have it all in a single plain json object

aaythapa:
Waiting on security approval

awood45 (approved):

Issue #2927, opened 2023-02-22
AWS::Serverless::Function: Adding multiple event source of the same type results to an error


jsunico:
### Description
When creating a serverless function with multiple event sources of the same type, an error is thrown in Cloudformation.

Error: Multiple IAM policies have the same policy name: [SamAutoGeneratedAMQPolicy]. Specify a unique name for each policy and run the stack operation again.

![image](https://user-images.githubusercontent.com/65151494/220558205-54cb1c54-f3f6-48b1-88b0-25218734623c.png)

### Steps to reproduce
Create a resource with the following definition:

```
  MyFunction:
    Type: AWS::Serverless::Function
    Properties:
      FunctionName: Test
      CodeUri: path-to-code/
      Handler: app.handler
      Runtime: nodejs18.x
      MemorySize: 256
      Architectures:
        - x86_64
      VpcConfig:
        SecurityGroupIds:
          - !Ref SecurityGroup
        SubnetIds: !Ref SubnetIds
      Events:
        BrokerOne:
            Type: MQ
            Properties:
              Broker: !Ref Broker1
              BatchSize: 1
              Queues:
                - MyQueue
              SourceAccessConfigurations:
                - Type: BASIC_AUTH
                  URI: !Ref BrokerSecret
        BrokerTwo:
            Type: MQ
            Properties:
              Broker: !Ref Broker2
              BatchSize: 1
              Queues:
                - MyQueue
              SourceAccessConfigurations:
                - Type: BASIC_AUTH
                  URI: !Ref BrokerSecret
        BrokerThree:
            Type: MQ
            Properties:
              Broker: !Ref Broker3
              BatchSize: 1
              Queues:
                - MyQueue
              SourceAccessConfigurations:
                - Type: BASIC_AUTH
                  URI: !Ref BrokerSecret
```

### Observed result
Error in Cloudformation is thrown:

Multiple IAM policies have the same policy name: [SamAutoGeneratedAMQPolicy]. Specify a unique name for each policy and run the stack operation again.

### Expected result
Lambda functions are created with multiple source mapping of the same type.

### Additional environment details

1. OS: WSL Ubuntu 20.04.5 LTS
2. If using the [SAM CLI](https://github.com/aws/aws-sam-cli), `sam --version`: SAM CLI, version 1.73.0
3. AWS region: eu-west-1

PR #2942, opened 2023-02-24
fix: policy name for MQ event source

ssenchenko:
### Issue #, if available
#2927

### Description of changes
Made MQ source policy name include MQ resource logical ID, so policy names of multiple MQ event sources won't clash

Transformation tests

### Checklist

    - [x] Using correct values
    - [ ] Using wrong values

### Examples?


GavinZZ (commented):
Is this change backward-compatible?

tests/translator/output/aws-us-gov/function_with_amq.json
```diff
@@ -35,7 +37,9 @@
         "SourceAccessConfigurations": [
           {
             "Type": "BASIC_AUTH",
-            "URI": "arn:aws:secretsmanager:us-west-2:123456789012:secret:my-path/my-secret-name-1a2b3c"
+            "URI": {
+              "Fn::Sub": "arn:${AWS::Partition}:secretsmanager:us-west-2:123456789012:secret:my-path/my-secret-name-1a2b3c"
+            }
           }
```
hoffa, line 43: Output templates should not change so we can ensure backward compatibility. Please add new transform tests instead.

ssenchenko: I ran templates with Function + MQ with `bin/add_transform_test.py` and had the error like “hardcoded partition is detected, you shouldn’t do that” I suppose `add_transform_test` changed over time. So instead of `arn:aws:` in all urls I changed it to `!Sub 'arn:${AWS::Partition}:`. It's the only change to the existing transformation tests. 

Plus, I've added one new test for multiple MQ event sources.

hoffa: We should ensure existing templates that deploy successfully transform exactly the same, and either:

1. Add a property to opt-in to the new behavior (e.g. `DynamicPolicyName: true`... naming is hard).
2. Figure out a way to do it in a backward-compatible manner without new properties (might be tricky, need to be mindful of reading order, etc.).

ssenchenko:
> Is this change backward-compatible?

Issue #2533, opened 2022-10-11
Unsupported intrinsic functions


hoffa:
Creating this to help track and manage intrinsics-related issues.

### The problem

Some properties have limited support for [intrinsic functions](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference.html) (e.g. `Fn::If`, `Fn::FindInMap`, etc.). Using unsupported intrinsic functions in such properties causes the deployment to fail.

### Why it's a problem

AWS SAM is a [AWS CloudFormation macro](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/template-macros.html); it receives a [SAM template](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-specification-template-anatomy.html) as input (as-is, along with the intrinsic functions), and returns a [CloudFormation template](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/template-anatomy.html) which is then deployed by CloudFormation.

This means that unless the intrinsic function is statically resolvable (e.g. it only refers to [CloudFormation parameters](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/parameters-section-structure.html)), SAM has no way of resolving the intrinsic function. It's not an issue for properties that are passed as-is to properties of the underlying CloudFormation resources, but it becomes an issue when SAM must know the value for its transform logic (see e.g. the [`Enabled` property](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-property-function-schedule.html#sam-function-schedule-enabled) of `Schedule` events).

tom-mcclintock:
Chiming in to say I came here via #2377 and am very interested to see this get supported.

In my use case, I use `!Ref` to set VPC endpoint IDs:
```yaml
AWSTemplateFormatVersion: '2010-09-09'
Transform: AWS::Serverless-2016-10-31

Parameters:
  VPCE:
    Type: String

Resources:
  ApiGateway:
    Type: AWS::Serverless::Api
    Properties:
      StageName: somestagename
      EndpointConfiguration:
        Type: PRIVATE
        VPCEndpointIds:
          - !Ref VPCE
```

Running `sam validate` yields an error, but also ends with the green message `... is a valid SAM Template`:
```bash
[...]

PR #2943, opened 2023-02-24
feat: Support SAM API MergeDefinitions property

GavinZZ:
### Description of changes
Allow merging SAM generated definitions into customer defined defnitions.

all tests passed

### Checklist

    - [ ] Using correct values
    - [ ] Using wrong values

### Examples?

### DIFF
--- a/samtranslator/internal/schema_source/aws_serverless_api.py
+++ b/samtranslator/internal/schema_source/aws_serverless_api.py
@@ -158,6 +158,7 @@ class EndpointConfiguration(BaseModel):
 
 Name = Optional[PassThroughProp]
 DefinitionUriType = Optional[Union[str, DefinitionUri]]
+MergeDefinitions = Optional[bool]
 CacheClusterEnabled = Optional[PassThroughProp]
 CacheClusterSize = Optional[PassThroughProp]
 Variables = Optional[PassThroughProp]
@@ -184,6 +185,7 @@ class Properties(BaseModel):
     Cors: Optional[CorsType] = properties("Cors")
     DefinitionBody: Optional[DictStrAny] = properties("DefinitionBody")
     DefinitionUri: Optional[DefinitionUriType] = properties("DefinitionUri")
+    MergeDefinitions: Optional[MergeDefinitions]  # TODO: update docs when live
     Description: Optional[PassThroughProp] = properties("Description")
     DisableExecuteApiEndpoint: Optional[PassThroughProp] = properties("DisableExecuteApiEndpoint")
     Domain: Optional[Domain] = properties("Domain")
@@ -208,6 +210,7 @@ class Globals(BaseModel):
     DefinitionUri: Optional[PassThroughProp] = properties("DefinitionUri")
     CacheClusterEnabled: Optional[CacheClusterEnabled] = properties("CacheClusterEnabled")
     CacheClusterSize: Optional[CacheClusterSize] = properties("CacheClusterSize")
+    MergeDefinitions: Optional[MergeDefinitions]  # TODO: update docs when live
     Variables: Optional[Variables] = properties("Variables")
     EndpointConfiguration: Optional[PassThroughProp] = properties("EndpointConfiguration")
     MethodSettings: Optional[MethodSettings] = properties("MethodSettings")
--- a/samtranslator/model/api/api_generator.py
+++ b/samtranslator/model/api/api_generator.py
@@ -172,6 +172,7 @@ class ApiGenerator:
         stage_name: Optional[Intrinsicable[str]],
         shared_api_usage_plan: Any,
         template_conditions: Any,
+        merge_definitions: Optional[bool] = None,
         tags: Optional[Dict[str, Any]] = None,
         endpoint_configuration: Optional[Dict[str, Any]] = None,
         method_settings: Optional[List[Any]] = None,
@@ -221,6 +222,7 @@ class ApiGenerator:
         self.depends_on = depends_on
         self.definition_body = definition_body
         self.definition_uri = definition_uri
+        self.merge_definitions = merge_definitions
         self.name = name
         self.stage_name = stage_name
         self.tags = tags
@@ -254,6 +256,7 @@ class ApiGenerator:
         :returns: the RestApi to which this SAM Api corresponds
         :rtype: model.apigateway.ApiGatewayRestApi
         """
+        self._validate_properties()
         rest_api = ApiGatewayRestApi(self.logical_id, depends_on=self.depends_on, attributes=self.resource_attributes)
         # NOTE: For backwards compatibility we need to retain BinaryMediaTypes on the CloudFormation Property
         # Removing this and only setting x-amazon-apigateway-binary-media-types results in other issues.
@@ -267,16 +270,6 @@ class ApiGenerator:
             # Since this region does not support EDGE configuration, we explicitly set the endpoint type
             # to Regional which is the only supported config.
             self._set_endpoint_configuration(rest_api, "REGIONAL")
-
-        if self.definition_uri and self.definition_body:
-            raise InvalidResourceException(
-                self.logical_id, "Specify either 'DefinitionUri' or 'DefinitionBody' property and not both."
-            )
-
-        if self.open_api_version and not SwaggerEditor.safe_compare_regex_with_string(
-            SwaggerEditor.get_openapi_versions_supported_regex(), self.open_api_version
-        ):
-            raise InvalidResourceException(self.logical_id, "The OpenApiVersion value must be of the format '3.0.0'.")
 
         self._add_cors()
         self._add_auth()
@@ -310,6 +303,22 @@ class ApiGenerator:
             rest_api.ApiKeySourceType = self.api_key_source_type
 
         return rest_api
+
+    def _validate_properties(self) -> None:
+        if self.definition_uri and self.definition_body:
+            raise InvalidResourceException(
+                self.logical_id, "Specify either 'DefinitionUri' or 'DefinitionBody' property and not both."
+            )
+
+        if self.definition_uri and self.merge_definitions:
+            raise InvalidResourceException(
+                self.logical_id, "Cannot set 'MergeDefinitions' to True when using `DefinitionUri`."
+            )
+
+        if self.open_api_version and not SwaggerEditor.safe_compare_regex_with_string(
+            SwaggerEditor.get_openapi_versions_supported_regex(), self.open_api_version
+        ):
+            raise InvalidResourceException(self.logical_id, "The OpenApiVersion value must be of the format '3.0.0'.")
 
     def _add_endpoint_extension(self) -> None:
         """Add disableExecuteApiEndpoint if it is set in SAM
--- a/samtranslator/model/eventsources/push.py
+++ b/samtranslator/model/eventsources/push.py
@@ -25,6 +25,7 @@ from samtranslator.swagger.swagger import SwaggerEditor
 from samtranslator.translator import logical_id_generator
 from samtranslator.translator.arn_generator import ArnGenerator
 from samtranslator.utils.py27hash_fix import Py27Dict, Py27UniStr
+from samtranslator.utils.utils import InvalidValueType, dict_deep_get
 from samtranslator.validator.value_validator import sam_expect
 
 CONDITION = "Condition"
@@ -705,7 +706,7 @@ class Api(PushEventSource):
 
         explicit_api = kwargs["explicit_api"]
         api_id = kwargs["api_id"]
-        if explicit_api.get("__MANAGE_SWAGGER"):
+        if explicit_api.get("__MANAGE_SWAGGER") or explicit_api.get("MergeDefinitions"):
             self._add_swagger_integration(explicit_api, api_id, function, intrinsics_resolver)  # type: ignore[no-untyped-call]
 
         return resources
@@ -757,8 +758,13 @@ class Api(PushEventSource):
         :param model.apigateway.ApiGatewayRestApi rest_api: the RestApi to which the path and method should be added.
         """
         swagger_body = api.get("DefinitionBody")
+        merge_definitions = api.get("MergeDefinitions")
         if swagger_body is None:
             return
+        if merge_definitions:
+            # Use a skeleton swagger body for API event source to make sure the generated definition body
+            # is unaffected by the inline/customer defined DefinitionBody
+            swagger_body = SwaggerEditor.gen_skeleton()
 
         partition = ArnGenerator.get_partition_name()
         uri = _build_apigw_integration_uri(function, partition)  # type: ignore[no-untyped-call]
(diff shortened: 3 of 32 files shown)


samtranslator/model/sam_resources.py
```diff
@@ -1146,6 +1146,7 @@ class SamApi(SamResourceMacro):
         "Tags": PropertyType(False, IS_DICT),
         "DefinitionBody": PropertyType(False, IS_DICT),
         "DefinitionUri": PropertyType(False, one_of(IS_STR, IS_DICT)),
+        "MergeDefinitions": PropertyType(False, is_type(bool)),
```
xazhao, line 1149: `PropertyType` is deprecated IMO. Can we use `Property` here? https://github.com/aws/serverless-application-model/blob/c5830b63857f52e540fec13b29f029458edc539a/samtranslator/model/__init__.py#L36-L45

samtranslator/internal/schema_source/aws_serverless_api.py
```diff
@@ -158,6 +158,7 @@ class EndpointConfiguration(BaseModel):
 
 Name = Optional[PassThroughProp]
 DefinitionUriType = Optional[Union[str, DefinitionUri]]
+MergeDefinitions = Optional[SamIntrinsicable[bool]]
```
hoffa, line 161: Why `SamIntrinsicable`? If it's a property only relevant to SAM, we shouldn't deal with intrinsics as `AWS::LanguageExtensions` takes care of them before SAM. Applies to all types of `MergeDefinitions` in this PR.

hoffa (approved):

samtranslator/model/sam_resources.py
hoffa: Yes: https://github.com/aws/serverless-application-model/blob/develop/DEVELOPMENT_GUIDE.md#development-guidelines

In fact this PR in its current state has a bug: if you pass an intrinsic to `MergeDefinitions` it'll happily accept it since it's using `PropertyType`, and the code checks for truthyness, so any intrinsic function that should resolve to `False` now is treated as `true`.

samtranslator/internal/schema_source/aws_serverless_api.py
GavinZZ: How do you guarantee that the template also uses `AWS::LanguageExtensions`?

hoffa: We ask that customers that encounter an issue add it: https://github.com/aws/serverless-application-model/issues/2533

`AWS::LanguageExtensions` resolves intrinsics generically and more thoroughly than the unsustainable ad-hoc approach SAM had so far.

samtranslator/model/eventsources/push.py
```diff
@@ -921,7 +924,29 @@ def _add_swagger_integration(  # type: ignore[no-untyped-def] # noqa: too-many-s
                 path=self.Path, method_name=self.Method, request_parameters=parameters
             )
 
-        api["DefinitionBody"] = editor.swagger
+        if api.get("MergeDefinitions", False):
+            self._merge_definitions(api, editor.swagger)
+        else:
+            api["DefinitionBody"] = editor.swagger
+
+    def _merge_definitions(self, api: Dict[str, Any], generated_body: Dict[str, Any]) -> None:
```
xazhao, line 932: nit: I slightly prefer have it returns a new merged definition instead of mutating the input.

GavinZZ: Updated

xazhao (approved):

samtranslator/model/api/api_generator.py
```diff
@@ -172,6 +172,7 @@ def __init__(  # noqa: too-many-arguments
         stage_name: Optional[Intrinsicable[str]],
         shared_api_usage_plan: Any,
         template_conditions: Any,
+        merge_definitions: Optional[Intrinsicable[bool]] = None,
```
aahung, line 175: why it needs to be an intrinsic? Our merge logics depending on this can be resolved right?

samtranslator/model/eventsources/push.py
```diff
@@ -759,6 +759,9 @@ def _add_swagger_integration(  # type: ignore[no-untyped-def] # noqa: too-many-s
         swagger_body = api.get("DefinitionBody")
         if swagger_body is None:
             return
+        if api.get("MergeDefinitions", False):
+            # Use a skeleton swagger body for API event source
```
hoffa, line 763: Should explain _why_.

samtranslator/model/eventsources/push.py
aahung, line 932: nit: this method's signature is a bit hard to understand at a glance.
- it is not clear why we are merging an "api" with a "body." They are of different types. 
- it is not clear what is merged into what. 

Suggest: `def _merge_definitions(self, source_definition_body, dest_definition_body)`

samtranslator/model/api/api_generator.py
GavinZZ: i forgot to update this part... It shouldn't be intrinsic.

samtranslator/model/eventsources/push.py
```diff
@@ -921,7 +924,29 @@ def _add_swagger_integration(  # type: ignore[no-untyped-def] # noqa: too-many-s
...
+    def _merge_definitions(self, api: Dict[str, Any], generated_body: Dict[str, Any]) -> None:
+        """
+        Merge SAM generated swagger definition into inline DefinitionBody using the follow rule:
+        - for a conflicting key, use SAM generated value
+        - otherwise include key-value pairs from both definitions
+        """
+        paths_property = api["DefinitionBody"]["paths"]
+
+        if self.Path not in paths_property:
+            paths_property.setdefault(self.Path, {})
```
hoffa, line 941: I have a hard time getting an intuitive sense of what it's doing, because it mutates the inputs, and uses instance attributes (so I need to know what `self.Path`, etc. are). Can it be made pure (no mutations) instead?

samtranslator/model/api/api_generator.py
```diff
@@ -268,11 +271,6 @@ def _construct_rest_api(self) -> ApiGatewayRestApi:
             # to Regional which is the only supported config.
             self._set_endpoint_configuration(rest_api, "REGIONAL")
 
-        if self.definition_uri and self.definition_body:
-            raise InvalidResourceException(
-                self.logical_id, "Specify either 'DefinitionUri' or 'DefinitionBody' property and not both."
-            )
-
         if self.open_api_version and not SwaggerEditor.safe_compare_regex_with_string(
```
aahung, line 274: why this is not in `_validate_properties`

samtranslator/model/eventsources/push.py
```diff
@@ -921,7 +924,29 @@ def _add_swagger_integration(  # type: ignore[no-untyped-def] # noqa: too-many-s
...
+        if self.Method not in paths_property[self.Path]:
+            paths_property[self.Path].setdefault(self.Method, {})
+
+        path_method_body = paths_property[self.Path][self.Method]
+        generated_path_method_body = generated_body["paths"][self.Path][self.Method]
```
hoffa, line 946: Only the second "path method body" has a specifier (`generated`), so it's hard to understand what specifically is the first one. It's a lot easier to read when they're symmetrical.

samtranslator/model/eventsources/push.py
```diff
@@ -921,7 +924,29 @@ def _add_swagger_integration(  # type: ignore[no-untyped-def] # noqa: too-many-s
...
+        else:
+            api["DefinitionBody"] = editor.swagger
+
+    def _merge_definitions(self, api: Dict[str, Any], generated_body: Dict[str, Any]) -> None:
+        """
+        Merge SAM generated swagger definition into inline DefinitionBody using the follow rule:
+        - for a conflicting key, use SAM generated value
+        - otherwise include key-value pairs from both definitions
+        """
+        paths_property = api["DefinitionBody"]["paths"]
```
aahung, line 938: Is it possible this line of code cause `KeyError`?

samtranslator/model/eventsources/push.py
```diff
@@ -921,7 +924,29 @@ def _add_swagger_integration(  # type: ignore[no-untyped-def] # noqa: too-many-s
...
+
+        if self.Path not in paths_property:
```
aahung, line 940: do we have validation on the type of `paths_property` (to be a dict) anywhere since it is a customer input?

samtranslator/model/eventsources/push.py
```diff
@@ -921,7 +924,29 @@ def _add_swagger_integration(  # type: ignore[no-untyped-def] # noqa: too-many-s
...
+            paths_property.setdefault(self.Path, {})
+        if self.Method not in paths_property[self.Path]:
```
aahung, line 942: do we have a type validation (to be a dict) since it is a customer input?

samtranslator/model/eventsources/push.py
```diff
@@ -921,7 +924,29 @@ def _add_swagger_integration(  # type: ignore[no-untyped-def] # noqa: too-many-s
...
+            paths_property[self.Path].setdefault(self.Method, {})
```
aahung, line 943: `setdefault` doesn't do anything if `self.Method` is not a key. The `if` check can be removed

samtranslator/model/eventsources/push.py
```diff
@@ -921,7 +924,29 @@ def _add_swagger_integration(  # type: ignore[no-untyped-def] # noqa: too-many-s
...
+
+        path_method_body = paths_property[self.Path][self.Method]
+        generated_path_method_body = generated_body["paths"][self.Path][self.Method]
+
+        # this guarantees to use SAM generated value for a conflicting key
+        paths_property[self.Path][self.Method] = {**path_method_body, **generated_path_method_body}
```
aahung, line 949: do we have a type check on `path_method_body` since it is a customer input?

samtranslator/model/eventsources/push.py
```diff
@@ -921,7 +926,39 @@ def _add_swagger_integration(  # type: ignore[no-untyped-def] # noqa: too-many-s
...
+        - for a conflicting key, use SAM generated value
+        - otherwise include key-value pairs from both definitions
+        """
+        merged_definition_body = source_definition_body.copy()
+        source_body_paths = merged_definition_body.get("paths", {})
+        if (
+            not isinstance(source_body_paths, dict)
+            or not isinstance(source_body_paths.get(self.Path, {}), dict)
+            or not isinstance(source_body_paths.get(self.Path, {}).get(self.Method, {}), dict)
+        ):
```
hoffa, line 948: Don't we have some method to easily deepget from a dict?

So this all could be replaced with e.g.

```python
if not isinstance(deepget(source_body_path, [self.Path, self.Method], {}), dict):
```

samtranslator/model/eventsources/push.py
```diff
@@ -921,7 +926,39 @@ def _add_swagger_integration(  # type: ignore[no-untyped-def] # noqa: too-many-s
...
+            # Do nothing if swagger body is invalid. We are already raising exception later by validating swagger body
```
aahung, line 949: we can still raise exceptions here. Relying on knowledge like this is not reliable. Imagine in the future that validation mentioned changes behavior this comment will be outdated.

samtranslator/model/eventsources/push.py
```diff
@@ -921,7 +926,39 @@ def _add_swagger_integration(  # type: ignore[no-untyped-def] # noqa: too-many-s
...
+
+        path_method_body = source_body_paths.get(self.Path, {}).get(self.Method, {})
+        generated_path_method_body = dest_definition_body["paths"][self.Path][self.Method]
+        # this guarantees that the merged definition use SAM generated value for a conflicting key
+        merged_path_method_body = {**path_method_body, **generated_path_method_body}
+
+        if self.Path in source_body_paths:
+            source_body_paths[self.Path][self.Method] = merged_path_method_body
+        else:
+            source_body_paths[self.Path] = {self.Method: merged_path_method_body}
```
hoffa, line 960: IMO clearer:

```python
if self.Path not in source_body_paths:
    source_body_paths[self.Path] = {}
source_body_paths[self.Path][self.Method] = merged_path_method_body
```

samtranslator/model/eventsources/push.py
```diff
@@ -921,7 +927,36 @@ def _add_swagger_integration(  # type: ignore[no-untyped-def] # noqa: too-many-s
...
+        - for a conflicting key, use SAM generated value
+        - otherwise include key-value pairs from both definitions
+        """
+        merged_definition_body = source_definition_body.copy()
+        source_body_paths = merged_definition_body.get("paths", {})
+
+        try:
+            path_method_body = dict_deep_get(source_body_paths, [self.Path, self.Method]) or {}
+        except InvalidValueType as e:
+            raise InvalidResourceException(self.logical_id, f"Property 'DefinitionBody' is invalid: {str(e)}") from e
```
aahung, line 949: I think this should not be `self.logical_id` as `self.logical_id` refers to the function I believe
```suggestion
            raise InvalidResourceException(api_id, f"Property 'DefinitionBody' is invalid: {str(e)}") from e
```

samtranslator/model/eventsources/push.py
```diff
@@ -921,7 +927,36 @@ def _add_swagger_integration(  # type: ignore[no-untyped-def] # noqa: too-many-s
...
+
+        generated_path_method_body = dest_definition_body["paths"][self.Path][self.Method]
+        # this guarantees that the merged definition use SAM generated value for a conflicting key
+        merged_path_method_body = {**path_method_body, **generated_path_method_body}
```
aahung, line 953: if `path_method_body` is not a dict, will it be a dip?

samtranslator/model/eventsources/push.py
GavinZZ: good catch

samtranslator/model/eventsources/push.py
GavinZZ: no, exception is caught.

GavinZZ: Had to add another check to make sure no dip.

samtranslator/model/eventsources/push.py
```diff
@@ -946,7 +946,10 @@ def _get_merged_definitions(
         try:
             path_method_body = dict_deep_get(source_body_paths, [self.Path, self.Method]) or {}
         except InvalidValueType as e:
-            raise InvalidResourceException(self.logical_id, f"Property 'DefinitionBody' is invalid: {str(e)}") from e
+            raise InvalidResourceException(api_id, f"Property 'DefinitionBody' is invalid: {str(e)}") from e
+
+        if not isinstance(path_method_body, dict):
+            raise InvalidResourceException(api_id, "Property 'DefinitionBody' is invalid")
```
aahung, line 952: This error message is not actionable:
```
sam_expect(path_method_body, api_id, f"DefinitionBody.paths.{self.Path}.{self.Method}").to_be_a_map()
```

GavinZZ: Nice, looks cleaner.

aahung (approved):
Thanks for addressing the comments!

PR #2953, opened 2023-02-27
fix: multiple mq source event policy name (add `DynamicPolicyName`)

ssenchenko:
### Issue #, if available
#2927

### Description of changes
Historically, policy name was hardcoded as `SamAutoGeneratedAMQPolicy` but it led to a policy name clash and failure to deploy, if a Function had at least 2 MQ event sources. Since policy is attached to the Lambda execution role, policy name should be based on MQ logical ID not to clash with policy names of other MQ event sources. However, to support backwards compatibility, we need to keep policy `SamAutoGeneratedAMQPolicy` by default, because customers might have code which relies on that policy name consistance.

To support both old policy name and ability to have more than one MQ event source, we introduce new `DynamicPolicyName` which when set to true will use MQ logical ID to genertae policy name.

**Q**: Why to introduce a new field and not to make policy name dynamic by default if there are multiple MQ event sources?
 **A**: Since a customer could have a single MQ source and rely on it's policy name in their code. If that customer decides to add a new MQ source, they don't want to change the policy name for the first MQ all over their code base. But they can opt in using a dynamic policy name for all other MQ sources they add.

**Q**: Why not use dynamic policy names automatically for all MQ event sources but first?
 **A**: SAM-T doesn't have state and doesn't know what was the CFN resource attribute in a previous transformation. Hence, trying to "use dynamic policy names automatically for all MQ event sources but first" can rely only on event source order. If a customer added a new MQ source __before__ an old one, an old one would receive a dynamic name and would break (potentially) customer's code.

### Checklist

    - [x] Using correct values
    - [ ] Using wrong values

### Examples?


samtranslator/model/eventsources/pull.py
```diff
@@ -355,9 +355,46 @@ class MQ(PullEventSource):
...
         **PullEventSource.property_types,
         "Broker": PassThroughProperty(True),
+        "DynamicPolicyName": Property(False, is_type(bool)),
     }
 
     Broker: PassThrough
+    DynamicPolicyName: Optional[bool]
+
+    @property
+    def policy_name(self) -> str:
```
hoffa, line 365: nit: rename to `_policy_name` so less likely be mistaken as a public function

samtranslator/internal/schema_source/aws_serverless_function.py
```diff
@@ -385,6 +385,7 @@ class MSKEvent(BaseModel):
 class MQEventProperties(BaseModel):
     BatchSize: Optional[PassThroughProp] = mqeventproperties("BatchSize")
     Broker: PassThroughProp = mqeventproperties("Broker")
+    DynamicPolicyName: Optional[bool]
```
hoffa, line 388:
```suggestion
    DynamicPolicyName: Optional[bool]  # TODO: add docs
```

samtranslator/model/eventsources/pull.py
ssenchenko: That private access in python is just an illusion anyway and I'm not fond of us using them, but why should it be private? it's a read only property, what's wrong with it being public?

hoffa: It's to minimize public (by convention, yes there's no "real" private) API footprint so that we have the freedom to change things in the future. We've had instances of e.g. SAM CLI depending on functions in our codebase that we changed, causing errors. And since we never publicized a library-level backward compatibility policy, we err toward being cautious and assuming everything by convention public is depended on.

ssenchenko: Sure,  I'll change ig

hoffa (approved):

samtranslator/internal/schema_source/aws_serverless_function.py
```diff
@@ -405,6 +405,7 @@ class MSKEvent(BaseModel):
 class MQEventProperties(BaseModel):
     BatchSize: Optional[PassThroughProp] = mqeventproperties("BatchSize")
     Broker: PassThroughProp = mqeventproperties("Broker")
+    DynamicPolicyName: Optional[bool]  # TODO: add docs
```
awood45, line 408: Resolve the TODO?

samtranslator/model/eventsources/pull.py
```diff
@@ -424,9 +424,46 @@ class MQ(PullEventSource):
...
+
+        Historically, policy name was hardcoded as `SamAutoGeneratedAMQPolicy` but it led to a policy name clash
+        and failure to deploy, if a Function had at least 2 MQ event sources.
+        Since policy is attached to the Lambda execution role,
+        policy name should be based on MQ logical ID not to clash with policy names of other MQ event sources.
+        However, to support backwards compatibility, we need to keep policy `SamAutoGeneratedAMQPolicy` by default,
+        because customers might have code which relys on that policy name consistancy.
+
+        To support both old policy name and ability to have more than one MQ event source, we introduce new field
+        `DynamicPolicyName` which when set to true will use MQ logical ID to genertae policy name.
```
awood45, line 448: Nitpick: Typo on generate/genertae - a few typos in the doc string here.

aahung (approved):

awood45 (changes requested):

samtranslator/internal/schema_source/aws_serverless_function.py
ssenchenko: Discussed it with @hoffa We add docs when it's in released and deployed everywhere. After that I'll resolve it

aahung: cannot until the doc is out

hoffa: We only add it once https://github.com/awsdocs/aws-sam-developer-guide is updated (i.e. the change has deployed).

awood45 (approved):

Issue #2941, opened 2023-02-24
Deployment Error for generated EB Rules due to long name


hawflau:
### Description
we got an error `QueryForAvailabilityWithUserExceptionQueryForAvailabilityWithUserExceptionEventLambdaTarget' at 'targets.1.member.id' failed to satisfy constraint: Member must have length less than or equal to 64 when deploying a changeset`. It was resulted from have a long Function Logical ID and a long Event key:

As discussed, the fix would be to: when generating the rule name, if it's longer than its max length, generate a shorter name. This should avoid any breaking change as it doesn't change any generated rule name if its length is within 64 characters.

### Steps to reproduce

```yml
QueryForAvailabilityWithUserException:
    Type: AWS::Serverless::Function
    Properties:
      Handler: app.query_for_command_availability_with_user_exception
      CodeUri: telemetry
      Role:
        Fn::GetAtt:
        - DefaultRole
        - Arn
      Events:
        QueryForAvailabilityWithUserExceptionEvent:
          Type: Schedule
          Properties:
            Schedule: cron(05 12 * * ? *)
```

### Observed result

Error seen: `QueryForAvailabilityWithUserExceptionQueryForAvailabilityWithUserExceptionEventLambdaTarget' at 'targets.1.member.id' failed to satisfy constraint: Member must have length less than or equal to 64 when deploying a changeset`

### Expected result

Deployment succeeds

### Additional environment details

1. OS:
2. If using the [SAM CLI](https://github.com/aws/aws-sam-cli), `sam --version`:
3. AWS region:

GavinZZ:
Hi there, thanks for reporting the issue. Will look into the problem and work on a fix.

hoffa:
If this limit has been there since the beginning (to be verified), we should be able to generically trim all logical IDs without concern for backward compatibility.

GavinZZ:
I have drafted a PR for the fix. The changes are currently in pending-release state and you should expect to see the change live in about 2 weeks. Closing this issue.

PR #2967, opened 2023-02-27
fix: Truncate Event Bridge Rule if Id is over 64 characters

GavinZZ:
### Issue #, if available
https://github.com/aws/serverless-application-model/issues/2941

### Description of changes
According to the `Id` property in this documentation https://docs.aws.amazon.com/eventbridge/latest/APIReference/API_Target.html, the max length is 64. Truncate it if it's too long.

### Checklist

    - [ ] Using correct values
    - [ ] Using wrong values

### Examples?

### DIFF
--- a/samtranslator/model/events.py
+++ b/samtranslator/model/events.py
@@ -5,8 +5,8 @@ from samtranslator.model.intrinsics import fnGetAtt, ref
 
 # Event Rule Targets Id and Logical Id has maximum 64 characters limit
 # https://docs.aws.amazon.com/eventbridge/latest/APIReference/API_Target.html
-EVENT_RULE_ID_MAX_LENGTH = 64
-EVENT_RULE_ID_EVENT_SUFFIX = "Event"
+EVENT_RULE_LOGICAL_ID_MAX_LENGTH = 64
+EVENT_RULE_LOGICAL_ID_EVENT_SUFFIX = "Event"
 
 
 class EventsRule(Resource):
@@ -33,19 +33,19 @@ class EventsRule(Resource):
     ) -> None:
         super().__init__(logical_id, relative_id, depends_on, attributes)
 
-        if len(self.logical_id) > EVENT_RULE_ID_MAX_LENGTH:
+        if len(self.logical_id) > EVENT_RULE_LOGICAL_ID_MAX_LENGTH:
             # Truncate logical id to satisfy the EVENT_RULE_ID_MAX_LENGTH limit
             self.logical_id = _truncate_with_suffix(
-                self.logical_id, EVENT_RULE_ID_MAX_LENGTH, EVENT_RULE_ID_EVENT_SUFFIX
+                self.logical_id, EVENT_RULE_LOGICAL_ID_MAX_LENGTH, EVENT_RULE_LOGICAL_ID_EVENT_SUFFIX
             )
 
 
 def generate_valid_target_id(logical_id: str, suffix: str) -> str:
     """Truncate Target Id if it is exceeding EVENT_RULE_ID_MAX_LENGTH limi."""
-    if len(logical_id) + len(suffix) <= EVENT_RULE_ID_MAX_LENGTH:
+    if len(logical_id) + len(suffix) <= EVENT_RULE_LOGICAL_ID_MAX_LENGTH:
         return logical_id + suffix
 
-    return _truncate_with_suffix(logical_id, EVENT_RULE_ID_MAX_LENGTH, suffix)
+    return _truncate_with_suffix(logical_id, EVENT_RULE_LOGICAL_ID_MAX_LENGTH, suffix)
 
 
 def _truncate_with_suffix(s: str, length: int, suffix: str) -> str:
--- a/samtranslator/model/events.py
+++ b/samtranslator/model/events.py
@@ -1,5 +1,12 @@
+from typing import Any, Dict, List, Optional
+
 from samtranslator.model import GeneratedProperty, Resource
 from samtranslator.model.intrinsics import fnGetAtt, ref
+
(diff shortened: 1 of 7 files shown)


samtranslator/model/events.py
```diff
@@ -16,3 +20,18 @@ class EventsRule(Resource):
...
+        depends_on: Optional[List[str]] = None,
+        attributes: Optional[Dict[str, Any]] = None,
+    ) -> None:
+        super().__init__(logical_id, relative_id, depends_on, attributes)
+
+        if len(self.logical_id) > EVENT_RULE_ID_MAX_LENGTH:
+            # According to documentation, event rule id has max length of 64 characters limit
+            # https://docs.aws.amazon.com/eventbridge/latest/APIReference/API_Target.html
+            # Truncate logical id to 59 characters and append 'Event' as a workaround
+            self.logical_id = self.logical_id[:59] + "Event"
```
aaythapa, line 37:
```suggestion
            suffix = "Event"
            self.logical_id = self.logical_id[:(64-len(suffix))] + suffix
```
Might be a bit cleaner

samtranslator/model/events.py
```diff
@@ -1,6 +1,10 @@
+from typing import Any, Dict, List, Optional
+
 from samtranslator.model import GeneratedProperty, Resource
 from samtranslator.model.intrinsics import fnGetAtt, ref
 
+EVENT_RULE_ID_MAX_LENGTH = 64
```
aahung, line 6: nit: need reference here

samtranslator/model/events.py
aahung: `suffix` can be a module constant or class `Final` attribute

hoffa: Or make it a function. It seems like a fairly generic operation for something like SAM.

```python
def truncate_suffix(length: int, suffix: Optional[str]):
```

aahung: ^ `truncate_with_max_length_and_add_a_suffix_if_the_input_exceeds_the_max_length()`

samtranslator/model/events.py
```diff
@@ -16,3 +23,27 @@ class EventsRule(Resource):
...
+        attributes: Optional[Dict[str, Any]] = None,
+    ) -> None:
+        super().__init__(logical_id, relative_id, depends_on, attributes)
+
+        if len(self.logical_id) > EVENT_RULE_ID_MAX_LENGTH:
+            # Truncate logical id to satisfy the EVENT_RULE_ID_MAX_LENGTH limit
+            self.logical_id = (
+                self.logical_id[: (EVENT_RULE_ID_MAX_LENGTH - len(EVENT_RULE_ID_EVENT_SUFFIX))]
+                + EVENT_RULE_ID_EVENT_SUFFIX
+            )
```
hoffa, line 41: Shouldn't this call the function underneath?

GavinZZ: No, because the original implementation uses `logical_id`, not `logical_id + "Events"`. If we use the fucntion below, it could be a breaking change.

hoffa: I would really enjoy a

```python
def _truncate_with_suffix(s: str, length: int, suffix: str):
    return s[: length - len(suffix)] + suffix
```

🥺

The logic is simple but it's deceivingly similar yet formatting is different it's hard to pattern match at a glance to figure out the difference.

samtranslator/utils/utils.py
```diff
@@ -31,6 +31,13 @@ def __init__(self, relative_path: str) -> None:
             super().__init__("It should be a map")
 
 
+def truncate_with_suffix(s: str, length: int, suffix: str) -> str:
```
hoffa, line 34: If this needs to be a generic util, can we move to `internal` instead?

GavinZZ: Do you mean the whole utils file?

hoffa: Just this function (FWIW originally I just meant to add a private function in `events.py`, but generic function works as well).

hoffa (approved):

samtranslator/utils/utils.py
GavinZZ: Moving it back to `events.py` as a private function.

samtranslator/model/events.py
```diff
@@ -1,6 +1,13 @@
+from typing import Any, Dict, List, Optional
+
 from samtranslator.model import GeneratedProperty, Resource
 from samtranslator.model.intrinsics import fnGetAtt, ref
 
+# Event Rule Targets Id and Logical Id has maximum 64 characters limit
+# https://docs.aws.amazon.com/eventbridge/latest/APIReference/API_Target.html
+EVENT_RULE_ID_MAX_LENGTH = 64
```
aahung, line 8: nit

```suggestion
EVENT_RULE_LOGICAL_ID_MAX_LENGTH = 64
```

samtranslator/model/events.py
```diff
@@ -1,6 +1,13 @@
...
+EVENT_RULE_ID_EVENT_SUFFIX = "Event"
```
aahung, line 9: nit

```suggestion
EVENT_RULE_LOGICAL_ID_EVENT_SUFFIX = "Event"
```

aahung (approved):

samtranslator/model/events.py
hoffa: It's not logical ID though 👀

GavinZZ: It is the logical id for `AWS::Events::Rule` resource.

hoffa: Where does it say that in https://docs.aws.amazon.com/eventbridge/latest/APIReference/API_Target.html?

PR #2917, opened 2023-02-16
Add resource validator using schema models

GavinZZ:
### Description of changes
Use models from SAM schema for property validation.

### Description of how you validated changes
Added unit tests and all tests passed

### Checklist

- [ ] Adheres to the [development guidelines](https://github.com/aws/serverless-application-model/blob/develop/DEVELOPMENT_GUIDE.md#development-guidelines)
- [ ] Add/update [transform tests](https://github.com/aws/serverless-application-model/blob/develop/DEVELOPMENT_GUIDE.md#unit-testing-with-multiple-python-versions)
    - [ ] Using correct values
    - [ ] Using wrong values
- [ ] Add/update [integration tests](https://github.com/aws/serverless-application-model/blob/develop/INTEGRATION_TESTS.md)

### Examples?

Please reach out in the comments if you want to add an example. Examples will be 
added to `sam init` through [aws/aws-sam-cli-app-templates](https://github.com/aws/aws-sam-cli-app-templates).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

Issue #1590, opened 2020-05-11
Resources created through the Serverless Transform contain hardcoded partitions in arns


dontirun:
**Description:**

I am writing [custom linting scripts](https://github.com/aws-cloudformation/cfn-python-lint/issues/1522) to detect whether my cloudformation template has hardcoded Partitions and/or Regions. Templates processed by the SAM transform create ARnss with hardcoded Partitions.

For example this AWS::Lambda::Permission

```
    "rCfnNagNotficationLambdaDevelopCfnNagPermission": {
      "Type": "AWS::Lambda::Permission",
      "Properties": {
        "Action": "lambda:InvokeFunction",
        "Principal": "logs.amazonaws.com",
        "FunctionName": {
          "Ref": "rCfnNagNotficationLambda"
        },
        "SourceArn": {
          "Fn::Sub": [
            "arn:aws:logs:${AWS::Region}:${AWS::AccountId}:log-group:${__LogGroupName__}:*",
            {
              "__LogGroupName__": {
                "Fn::GetAtt": [
                  "rDevelopPipeline",
                  "Outputs.oCfnNagLogGroupName"
                ]
              }
            }
          ]
        }
      }
```
Or this ManagedPolicyArn in this AWS::IAM::Role

```
    "rUploadCustomCfnNagRole": {
      "Type": "AWS::IAM::Role",
      "Properties": {
        "AssumeRolePolicyDocument": {
          "Version": "2012-10-17",
          "Statement": [
            {
              "Action": [
                "sts:AssumeRole"
              ],
              "Effect": "Allow",
              "Principal": {
                "Service": [
                  "lambda.amazonaws.com"
                ]
              }
            }
          ]
        },
        "ManagedPolicyArns": [
          "arn:aws:iam::aws:policy/service-role/AWSLambdaBasicExecutionRole"
        ],
        "Policies": [
          {
            "PolicyName": "rUploadCustomCfnNagRolePolicy0",
            "PolicyDocument": {
              "Version": "2012-10-17",
              "Statement": [
                {
                  "Action": [
                    "s3:PutObject",
                    "s3:DeleteObject"
                  ],
                  "Resource": [
                    {
                      "Fn::Sub": "${rCustomCFNNagRulesBucket.Arn}/*"
                    }
                  ],
                  "Effect": "Allow"
                },
                {
                  "Action": [
                    "logs:CreateLogGroup",
                    "logs:CreateLogStream",
                    "logs:PutLogEvents"
[...]

PR #2971, opened 2023-02-27
feat: make Logging property for GraphQLApi

pradhapanr:
### Description of changes
Created new property `Logging` in `Serverless::GraphQLApi` resource, which allows customers to specify their logging implementation in the SAM template.

Maps to the `LogConfig` property of the `AppSync::GraphQLApi` resource. Includes defaults for 2 properties:
- `CloudWatchLogsRoleArn`: If not defined, creates a role with an AWS managed AppSync CloudWatch policy and attaches to the `GraphQLApi` resource.
- `FieldLogLevel`: Defaults to `ALL`.

This change also includes a small refactor of the error transform tests for `GraphQLApi`, bundling them all in `error_graphqlapi.yaml`.

Modified and created transform tests to work with `Logging` property.
### Checklist

    - [x] Using correct values
    - [x] Using wrong values

### Examples?


GavinZZ (commented):
Overall looks good. Just some nit comments.

samtranslator/internal/model/appsync.py
```diff
@@ -25,19 +25,27 @@ class DynamoDBConfigType(TypedDict, total=False):
     DeltaSyncConfig: DeltaSyncConfigType
 
 
+class LogConfigType(TypedDict, total=False):
+    CloudWatchLogsRoleArn: Intrinsicable[str]
```
aaythapa, line 29: We don't do intrinsicable anymore

samtranslator/model/sam_resources.py
```diff
@@ -2144,26 +2147,76 @@ class SamGraphQLApi(SamResourceMacro):
...
 
         api.AuthenticationType = sam_expect(self.Auth.get("Type"), self.logical_id, "Auth.Type").to_be_a_string()
         api.Name = self.Name if self.Name else self.logical_id
         api.XrayEnabled = self.XrayEnabled
 
+        cloudwatch_role = None  # conditionally generated, if left as None then its not added to resources
+
+        # Logging has 3 possible types: dict, bool, and None.
+        # GraphQLApi will not include logging if and only if the user explicity sets Logging as false boolean.
+        # It will for every other value (including true boolean which is essentially same as None).
```
aaythapa, line 2175: Bit confused here, why would setting Logging to None be the same as setting it to True? Shouldn't None mean theres no logging

samtranslator/model/sam_resources.py
```diff
@@ -2144,26 +2147,76 @@ class SamGraphQLApi(SamResourceMacro):
...
-        return api
+        return api, cloudwatch_role
+
+    def _parse_logging_properties(self) -> LogConfigType:
+        log_config: LogConfigType = {}
+
+        if type(self.Logging) == dict and "CloudWatchLogsRoleArn" in self.Logging:
+            log_config["CloudWatchLogsRoleArn"] = self.Logging["CloudWatchLogsRoleArn"]
+        else:
+            log_config["CloudWatchLogsRoleArn"] = fnGetAtt(f"{self.logical_id}CloudWatchRole", "Arn")
```
aaythapa, line 2195: Would it make more sense to call `self._construct_cloudwatch_role()` here instead and then set arn?

samtranslator/model/sam_resources.py
```diff
@@ -2144,26 +2147,76 @@ class SamGraphQLApi(SamResourceMacro):
...
         if self.Tags:
             api.Tags = get_tag_list(self.Tags)
 
-        return api
+        return api, cloudwatch_role
+
+    def _parse_logging_properties(self) -> LogConfigType:
+        log_config: LogConfigType = {}
+
+        if type(self.Logging) == dict and "CloudWatchLogsRoleArn" in self.Logging:
```
aaythapa, line 2192:
```suggestion
        if isinstance(self.Logging, dict):
           if "CloudWatchLogsRoleArn" in self.Logging:
           ....
           if "ExcludeVerboseContent" in self.Logging:
           ...
           if "FieldLogLevel" in self.Logging
           ....
```

samtranslator/model/sam_resources.py
GavinZZ: I think `Logging` to None is basically when customers don't define the property. We would want logging by default.

samtranslator/internal/model/appsync.py
hoffa: If it's to static-type the output fields and intrinsics are used it's fine/needed.

samtranslator/model/sam_resources.py
```diff
@@ -2144,26 +2147,76 @@ class SamGraphQLApi(SamResourceMacro):
...
+        role = IAMRole(
+            logical_id=f"{self.logical_id}CloudWatchRole",
+            depends_on=self.depends_on,
+            attributes=self.resource_attributes,
+        )
+        role.AssumeRolePolicyDocument = IAMRolePolicies.construct_assume_role_policy_for_service_principal(
+            "appsync.amazonaws.com"
+        )
+        role.ManagedPolicyArns = [
+            f"arn:{ArnGenerator.get_partition_name()}:iam::aws:policy/service-role/AWSAppSyncPushToCloudWatchLogs"
```
hoffa, line 2217: We shouldn't have to fetch the partition ourselves as it causes `cfn-lint` (see https://github.com/aws/serverless-application-model/pull/2550 and https://github.com/aws/serverless-application-model/issues/1590) to complain about hardcoded partitions. You should use `${AWS::Partition}` instead.

samtranslator/internal/model/appsync.py
pradhapanr: I thought we discussed that for the CFN types they were fine to use?

samtranslator/model/sam_resources.py
pradhapanr: What Gavin said pretty much. Code wise, None is equivalent to True boolean in the case of Logging.

samtranslator/model/sam_resources.py
pradhapanr: Good to know, I'll change that

samtranslator/model/sam_resources.py
```diff
@@ -2144,26 +2147,76 @@ class SamGraphQLApi(SamResourceMacro):
...
+        cloudwatch_role = None  # conditionally generated, if left as None then its not added to resources
+
+        # Logging has 3 possible types: dict, bool, and None.
+        # GraphQLApi will not include logging if and only if the user explicity sets Logging as false boolean.
+        # It will for every other value (including true boolean which is essentially same as None).
+        if type(self.Logging) != bool or self.Logging is True:
+            api.LogConfig = self._parse_logging_properties()
+
+            # We create a CloudWatch role for the user if the Logging property is not a dictionary
+            # that contains the "CloudWatchLogsRoleArn" property.
```
hoffa, line 2180: Why check for `CloudWatchLogsRoleArn`?

pradhapanr: If the user already has a Role, then there's no need for us to create it

samtranslator/model/sam_resources.py
pradhapanr: Yep, I'll add this change as well

samtranslator/model/sam_resources.py
```diff
@@ -2144,26 +2147,76 @@ class SamGraphQLApi(SamResourceMacro):
...
+
+    def _parse_logging_properties(self) -> LogConfigType:
+        log_config: LogConfigType = {}
+
+        if type(self.Logging) == dict and "CloudWatchLogsRoleArn" in self.Logging:
+            log_config["CloudWatchLogsRoleArn"] = self.Logging["CloudWatchLogsRoleArn"]
+        else:
+            log_config["CloudWatchLogsRoleArn"] = fnGetAtt(f"{self.logical_id}CloudWatchRole", "Arn")
+
+        if type(self.Logging) == dict and "ExcludeVerboseContent" in self.Logging:
```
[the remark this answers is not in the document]
pradhapanr: `ExcludeVerboseContent` is an optional value. I have tested deploying without it, CFN defaults it as False. `CloudWatchLogsRoleArn` and `FieldLogLevel` are not optional values (although CFN documentation suggests they are).

samtranslator/model/sam_resources.py
aaythapa: Ah I see thanks I thought the customers were setting the value of Logging to None. Thanks for the clarification

samtranslator/model/sam_resources.py
pradhapanr: Yea I really wanted to group this particular logic together if possible but one function is responsible for creating the CFN property and the other is responsible for making the role. It felt weird combining all of that functionality together and then returning the Role resource sometimes. Let me try taking another look at this and see if I can write it more elegantly

samtranslator/internal/model/appsync.py
```diff
@@ -25,19 +25,27 @@ class DynamoDBConfigType(TypedDict, total=False):
...
+    ExcludeVerboseContent: bool
```
GavinZZ, line 30: This should be `Optional[bool]`?

aahung: You can import `NotRequired` from https://pypi.org/project/typing-extensions/

samtranslator/model/sam_resources.py
pradhapanr: While trying to refactor this, I realized we can’t simply abstract away this condition into its own check.
There is two cases where the key should be defaulted in Logging:
- Logging is a Bool / None type
- Logging is a dictionary, but does not contain a key for the defaulted property.

We could have 2 separate if statements, one to handle when Logging is Bool/None, and one where it’s a dictionary, but there would be even more duplicated code and I’m not convinced that’s better clarity wise. So I’m leaning towards keeping this implementation, but replacing `type` with `isinstance`.

samtranslator/model/sam_resources.py
```diff
@@ -2144,26 +2147,76 @@ class SamGraphQLApi(SamResourceMacro):
...
         api.AuthenticationType = sam_expect(self.Auth.get("Type"), self.logical_id, "Auth.Type").to_be_a_string()
         api.Name = self.Name if self.Name else self.logical_id
         api.XrayEnabled = self.XrayEnabled
 
+        cloudwatch_role = None  # conditionally generated, if left as None then its not added to resources
+
+        # Logging has 3 possible types: dict, bool, and None.
+        # GraphQLApi will not include logging if and only if the user explicity sets Logging as false boolean.
+        # It will for every other value (including true boolean which is essentially same as None).
+        if type(self.Logging) != bool or self.Logging is True:
```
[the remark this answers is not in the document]
pradhapanr: It’s hard to pull off a simple negation for this since there’s two cases where we don’t need to create a role, Logging disabled, or if customer already has a Role. But in the case that customer already has a role, we still need to create the LogConfig property, so we’ll have some additional if statements.

I’m working on a change that would extract the role creation logic to the `to_cloudformation()` function, so then each function will just return a single resource and I can get rid of these conditional tuple returns. It should be a bit more clear as to when and why a CloudWatch role is created for the customer.

samtranslator/model/sam_resources.py
```diff
@@ -2144,26 +2147,76 @@ class SamGraphQLApi(SamResourceMacro):
...
 
-    def _construct_appsync_api(self) -> GraphQLApi:
+    def _construct_appsync_api_resources(self) -> Tuple[GraphQLApi, Optional[IAMRole]]:
         api = GraphQLApi(logical_id=self.logical_id, depends_on=self.depends_on, attributes=self.resource_attributes)
 
         api.AuthenticationType = sam_expect(self.Auth.get("Type"), self.logical_id, "Auth.Type").to_be_a_string()
         api.Name = self.Name if self.Name else self.logical_id
         api.XrayEnabled = self.XrayEnabled
 
+        cloudwatch_role = None  # conditionally generated, if left as None then its not added to resources
```
ssenchenko, line 2171: nit: this way of declaring has a couple of disadvantages
- type is missing
- it sucks to add type manually and then to get it return from the function anyway
- more difficult to read because actual assignment happens later

Suggestion: instead of declaring it here, do it when you return it from the function like this
```python
should_construct_new_role = isinstance(self.Logging, dict) and  "CloudWatchLogsRoleArn" not in self.Logging
cloudwatch_role = self._construct_cloudwatch_role() if should_construct_new_role else None
```

samtranslator/model/sam_resources.py
ssenchenko: +1 to add a separate function which will always return role arn and incapsulate the logic how we get this arn

tests/translator/input/error_graphqlapi.yaml
```diff
@@ -0,0 +1,32 @@
+Transform: AWS::Serverless-2016-10-31
+Resources:
+  NoAuthAPI:
+    Type: AWS::Serverless::GraphQLApi
+    Properties:
+      XrayEnabled: true
+      Tags:
+        Key1: Value1
+        Key2: Value2
```
ssenchenko, line 9: please let's add an empty line between resources, it's too overwhelming right now

samtranslator/internal/model/appsync.py
pradhapanr: The only problem I have with `NotRequired` is that I can't instantiate the TypedDict as empty and then add values to it as I create it. My original idea was to create a `Dict[str, Any]` dictionary and then unpack it in another dictionary with the appropriate TypedDict to get static type checking, but that is not a supported behaviour (see https://github.com/python/mypy/issues/4122).

So if we want to use `NotRequired` now, I think we'd need to use a condition to check if the optional key exists, and then we can create a TypedDict with that optional key or a TypedDict without it. That's a lot of extra code IMO that will suck to add for all of the resources I'm creating, especially those with multiple optional values.

Is there any other way we could go about constructing the dictionary to have static type checking that doesn't introduce so much bloat?

pradhapanr:
The most recent commits should address the majority of comments that were focused on type checking and reducing branching.

I separated the logic for when `Logging` property is a boolean type or a dictionary type, so they now have two separate functions: `create_logging_default()` and `parse_logging_properties()`.

This reduced the nested if branches and repeated type checks, and also separates the logic in what should be a more readable manner.

samtranslator/model/sam_resources.py
```diff
@@ -2163,7 +2170,63 @@ def _construct_appsync_api(self) -> GraphQLApi:
...
+        # GraphQLApi will not include logging if and only if the user explicity sets Logging as false boolean.
+        # It will for every other value (including true boolean which is essentially same as None).
+        if isinstance(self.Logging, bool) and self.Logging is False:
+            return api, None
+
+        api.LogConfig, cloudwatch_role = (
+            self._parse_logging_properties() if isinstance(self.Logging, dict) else self._create_logging_default()
+        )
+
+        return api, cloudwatch_role
```
GavinZZ, line 2183: Nice, looks a lot cleaner.

GavinZZ (approved):

samtranslator/model/sam_resources.py
```diff
@@ -2163,7 +2170,63 @@ def _construct_appsync_api(self) -> GraphQLApi:
...
+        return log_config, cloudwatch_role
+
+    def _parse_logging_properties(self) -> Tuple[LogConfigType, Optional[IAMRole]]:
+        """
+        Parse logging properties from SAM template, and use defaults if required keys dont exist.
+
+        This function should only be called when "Logging" property is a dictionary type.
+        """
+        logging = cast(Dict[str, Any], self.Logging)
+        log_config: LogConfigType = {}
```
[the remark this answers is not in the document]
pradhapanr: With nested properties, customers can actually add a bunch of random values and we have no way of validating the exact shape (this is the case for all SAM resources). For example, this input is technically valid and would transform:
```
Resources:
  MyGraphQLApi:
    Properties:
      ...
      Logging:
        CloudWatchLogsRoleArn: some-role
        ThisPropertyDoesntExist: blah
```
There's no validation for these "extra" properties, so we can't copy over the entire input. This was the main reason why I got rid of the typings for SAM resources and left them only for CFN resources where we can be 100% confident.

samtranslator/model/sam_resources.py
```diff
@@ -2163,7 +2170,63 @@ def _construct_appsync_api(self) -> GraphQLApi:
...
+
+        log_config["FieldLogLevel"] = logging["FieldLogLevel"] if "FieldLogLevel" in logging else "ALL"
+        if "ExcludeVerboseContent" in logging:
+            log_config["ExcludeVerboseContent"] = logging["ExcludeVerboseContent"]
+
+        cloudwatch_role = self._construct_cloudwatch_role() if "CloudWatchLogsRoleArn" not in logging else None
+        log_config["CloudWatchLogsRoleArn"] = (
+            cloudwatch_role.get_runtime_attr("arn") if cloudwatch_role else logging["CloudWatchLogsRoleArn"]
+        )
+        return log_config, cloudwatch_role
```
hoffa, line 2215: Why return `cloudwatch_role` separately if it's already part of `LogConfigType`?

ssenchenko (approved):

hoffa (approved):

samtranslator/model/sam_resources.py
pradhapanr: cloudwatch_role is an IAMRole resource so we can create it, the field in `LogConfigType` is just the arn

samtranslator/model/sam_resources.py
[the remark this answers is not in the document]
aaythapa: hmm but if the customer were to deploy with the extra properties then it'd fail no? I do like that with the way its implemented now we know exactly whats going to be in the transformed template so I don't mind keeping it the way it is

aaythapa (dismissed):

hoffa (dismissed):

samtranslator/model/sam_resources.py
```diff
@@ -2163,7 +2170,63 @@ def _construct_appsync_api(self) -> GraphQLApi:
...
+        log_config: LogConfigType = {}
+
+        log_config["FieldLogLevel"] = logging["FieldLogLevel"] if "FieldLogLevel" in logging else "ALL"
+        if "ExcludeVerboseContent" in logging:
+            log_config["ExcludeVerboseContent"] = logging["ExcludeVerboseContent"]
+
+        cloudwatch_role = self._construct_cloudwatch_role() if "CloudWatchLogsRoleArn" not in logging else None
+        log_config["CloudWatchLogsRoleArn"] = (
+            cloudwatch_role.get_runtime_attr("arn") if cloudwatch_role else logging["CloudWatchLogsRoleArn"]
+        )
```
hoffa, line 2214: We really need to use https://github.com/aws/serverless-application-model/pull/2917. Can do later. Currently it doesn't fail on garbage input.

hoffa (approved):

samtranslator/model/sam_resources.py
pradhapanr: Just so everyone knows, I'll be incorporating @GavinZZ's PR with my next batch of changes, so we should have much better validation of customer input once that is in effect.

aaythapa (approved):

GavinZZ (approved):

ssenchenko (approved):

Issue #2975, opened 2023-02-28
cyclic import on ServerlessAppPlugin with 1.60.0


alexrashed:
### Description
With the release of [1.60.0](https://github.com/aws/serverless-application-model/releases/tag/v1.60.0), it is not possible anymore to import from `samtranslator.plugins.application.serverless_app_plugin` due to a cyclic import introduced with be58556829aa8642c3505c54fbc0d053298e7456.

### Steps to reproduce
I used the following script to verify my tests (and to identify the commit which introduced the issue using it as a run script for `git bisect`):
```
#!/bin/bash
set -eo pipefail
rm -rf .venv
python -m venv .venv
source .venv/bin/activate
pip install -e .
python -c "from samtranslator.plugins.application.serverless_app_plugin import ServerlessAppPlugin"
```

### Observed result
This script mentioned above fails on the `v1.60.0` tag, as well as on the latest commit on `develop`, with the following error:
```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/localstack/Repos/serverless-application-model/samtranslator/plugins/application/serverless_app_plugin.py", line 11, in <module>
    from samtranslator.intrinsics.actions import FindInMapAction
  File "/home/localstack/Repos/serverless-application-model/samtranslator/intrinsics/actions.py", line 5, in <module>
    from samtranslator.model.exceptions import InvalidDocumentException, InvalidTemplateException
  File "/home/localstack/Repos/serverless-application-model/samtranslator/model/__init__.py", line 7, in <module>
    from samtranslator.intrinsics.resolver import IntrinsicsResolver
  File "/home/localstack/Repos/serverless-application-model/samtranslator/intrinsics/resolver.py", line 4, in <module>
    from samtranslator.intrinsics.actions import Action, GetAttAction, RefAction, SubAction
ImportError: cannot import name 'Action' from partially initialized module 'samtranslator.intrinsics.actions' (most likely due to a circular import) (/home/localstack/Repos/serverless-application-model/samtranslator/intrinsics/actions.py)
```

### Expected result
With b3b6c79704d0ae2a39645662e16ce2d9110dd016, the import does not fail (which I would expect).

### Additional environment details

1. OS: Ubuntu 22.04, Python 3.10
2. If using the [SAM CLI](https://github.com/aws/aws-sam-cli), `sam --version`: -
3. AWS region: -

aahung:
[...]

PR #2977, opened 2023-02-28
fix: Decouple samtranslator.models and *.intrinsics and add import tests

aahung:
### Description of changes

#2975

`samtranslator.models.__init__` should not import `samtranslator.intrinsics.*` which in turn import `samtranslator.models.__init__`

Added a new import test, without the fix, it will fail:

image

### Checklist

    - [ ] Using correct values
    - [ ] Using wrong values

### Examples?

### DIFF
--- a/samtranslator/internal/intrinsics.py
+++ b/samtranslator/internal/intrinsics.py
@@ -1,0 +1,23 @@
+from typing import Any, Dict, Optional, Union
+
+from samtranslator.intrinsics.resolver import IntrinsicsResolver
+from samtranslator.model.exceptions import InvalidResourceException
+
+
+def resolve_string_parameter_in_resource(
+    logical_id: str,
+    intrinsics_resolver: IntrinsicsResolver,
+    parameter_value: Optional[Union[str, Dict[str, Any]]],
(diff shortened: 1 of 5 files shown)


samtranslator/schema/schema.json
```diff
@@ -193550,10 +193550,10 @@
         "InvokeRole": {
           "anyOf": [
             {
-              "type": "string"
+              "type": "object"
```
acristin, line 193553: I don't understand why the schema is changing.

GavinZZ: curious too

aahung: no idea. in gh action it decided to generate with different order: https://github.com/aws/serverless-application-model/actions/runs/4296629078/jobs/7488579979

aahung: possibly pydantic orders the keys according to the hash of our module `samtranslator`

GavinZZ (approved):

acristin (approved):

hoffa (approved):

Issue #1786, opened 2020-11-09
Defining CORS when ApiKeyRequired is true results in an OPTIONS method that requires an API key


stopmachine:
**Description:**

When setting up an API Gateway with {proxy+} integration through SAM, having a required API key prevents the CORS definition from properly responding to OPTIONS requests, since they require an API key as well.

**Steps to reproduce the issue:**
1. Define an Api that requires an API key:
```
Resources:
  BasicAWSApiGateway:
    Type: AWS::Serverless::Api
    Properties:
      StageName: !Ref Environment
      Auth:
        ApiKeyRequired: true
      Cors: "'*'"
      DefinitionBody:
        swagger: "2.0.0"
        info:
          version: 1.0.0
        paths:
          /{proxy+}:
            x-amazon-apigateway-any-method:
              produces:
                - application/json
              parameters:
                - in: path
                  name: proxy
                  required: true
                  type: string
              x-amazon-apigateway-integration:
                uri: !Sub "arn:aws:apigateway:${AWS::Region}:lambda:path/2015-03-31/functions/arn:aws:lambda:${AWS::Region}:${AWS::AccountId}:function:${LambdaName}-${Environment}/invocations"
                httpMethod: POST
                type: aws_proxy
                passthroughBehavior: when_no_match
                credentials: !GetAtt ApiGatewayExecutionRole.Arn
```

**Observed result:**
The `OPTIONS` method in API Gateway requires an API Key, blocking CORS since browsers don't add a `x-api-key` header.

**Expected result:**
The `OPTIONS` method should not require an API Key.

aahung:
In https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-property-api-apiauth.html#sam-api-apiauth-adddefaultauthorizertocorspreflight

We have a `AddDefaultAuthorizerToCorsPreflight` to control whether `DefaultAuthorizer` is applied to `OPTIONS` endpoint or not, but we don't have one for `ApiKeyRequired`. 

According to [Fetch Standard](https://fetch.spec.whatwg.org/#http-cors-protocol)

> For a CORS-preflight request, request’s credentials mode is always "same-origin", i.e., it excludes credentials

OPTIONS endpoint should work without the need for credential

pisaacs:
While this issue is being resolved, are there a suggested work-arounds?

eh-admin:
> While this issue is being resolved, are there suggested work-arounds?

potential "workarounds":
[...]

PR #2981, opened 2023-03-01
fix: Defining CORS when ApiKeyRequired is true results in an OPTIONS method that requires an API key

ConnorRobertson:
### Issue #, if available
#1786 
### Description of changes
Created a property of AWS::Serverless::Api::Auth called AddApiKeyRequiredToCorsPreflight. When set to false does not apply ApiKeyRequired to Cors Preflight requests. By Default set to true.

Transformed a template that had the issue where ApiKeyRequired was being applied to CorsPreflight requests. Set AddApiKeyRequired to Cors Preflight to false. Checked if the transformed template required an ApiKey in the Cors section.
### Checklist

    - [x] Using correct values
    - [ ] Using wrong values

### Examples?


integration/combination/test_api_with_cors.py
```diff
@@ -12,10 +12,7 @@
 @skipIf(current_region_does_not_support([REST_API]), "Rest API is not supported in this testing region")
 class TestApiWithCors(BaseTest):
     @parameterized.expand(
-        [
-            "combination/api_with_cors",
-            "combination/api_with_cors_openapi",
-        ]
+        ["combination/api_with_cors", "combination/api_with_cors_openapi", "combination/api_with_cors_and_apikey"]
```
hoffa, line 15: nit: can you add back the trailing comma so `black` can format on multiple lines? easier to review what changed

samtranslator/swagger/swagger.py
```diff
@@ -612,14 +612,16 @@ def set_path_default_authorizer(  # noqa: too-many-branches
                 if "AWS_IAM" in method_definition["security"][0]:
                     self.add_awsiam_security_definition()
 
-    def set_path_default_apikey_required(self, path: str) -> None:
+    def set_path_default_apikey_required(self, path: str, AddApiKeyRequiredToCorsPreflight: bool = True) -> None:
```
hoffa, line 615: Prefer consistent naming for regular variables, can be confusing to see at a glance otherwise.

samtranslator/swagger/swagger.py
```diff
@@ -695,6 +700,10 @@ def add_auth_to_method(self, path: str, method_name: str, auth: Dict[str, Any],
             self._set_method_authorizer(path, method_name, method_authorizer, authorizers, method_scopes)  # type: ignore[no-untyped-call]
 
         method_apikey_required = auth and auth.get("ApiKeyRequired")
+
+        if auth.get("AddApiKeyRequiredToCorsPreflight") and method_name == "options":
```
hoffa, line 704: I don't see any type validation from what I can see, so I'm assuming even `"false"` will evaluate as `true` here?

samtranslator/swagger/swagger.py
```diff
@@ -673,6 +675,9 @@ def set_path_default_apikey_required(self, path: str) -> None:
 
             security = existing_non_apikey_security + apikey_security
 
+            if method_name == "options" and not AddApiKeyRequiredToCorsPreflight:
+                security = existing_non_apikey_security
```
hoffa, line 679: What's the difference with this and the change in `add_auth_to_method`?

samtranslator/swagger/swagger.py
ConnorRobertson: I think this actually doesn't have to be here at all I will remove it and test.

ConnorRobertson: Okay, so this is necessary if they define AddApiKeyRequiredToCorsPreflight: false at the individual Api level. I did miss the logic flip so I need to change it to if not auth.get("AddApiKeyRequiredToCorsPreflight") and method_name == "options":

samtranslator/swagger/swagger.py
ConnorRobertson: This matters if they define AddApiKeyRequiredToCorsPreflight in the globals instead of at the individual Api level.

integration/combination/test_api_with_cors.py
ConnorRobertson: Will do!

samtranslator/swagger/swagger.py
ConnorRobertson: What's a better name I could use?

samtranslator/swagger/swagger.py
hoffa: Okay, can you add (transform?) tests as well that check this logic?

samtranslator/swagger/swagger.py
hoffa: Why?

samtranslator/swagger/swagger.py
hoffa: Use `snake_case`. Here the name is quite complex (I assume only because of precedent with similar name). Maybe something like `require_options_api_key`?

samtranslator/internal/schema_source/aws_serverless_api.py
```diff
@@ -100,6 +100,7 @@ class UsagePlan(BaseModel):
 
 class Auth(BaseModel):
     AddDefaultAuthorizerToCorsPreflight: Optional[bool] = auth("AddDefaultAuthorizerToCorsPreflight")
+    AddApiKeyRequiredToCorsPreflight: Optional[bool]  # TODO Add Docs
```
GavinZZ, line 103: Should also add this property to `Globals` schema.

samtranslator/model/api/api_generator.py
```diff
@@ -1224,9 +1225,11 @@ def _set_default_authorizer(
                 add_default_auth_to_preflight=add_default_auth_to_preflight,
             )
 
-    def _set_default_apikey_required(self, swagger_editor: SwaggerEditor) -> None:
+    def _set_default_apikey_required(
+        self, swagger_editor: SwaggerEditor, AddApiKeyRequiredToCorsPreflight: bool
```
GavinZZ, line 1229: Please rename `AddApiKeyRequiredToCorsPreflight` to use snake_case naming stype.

samtranslator/swagger/swagger.py
ConnorRobertson: Should I do this in the api_generator.py file too?

samtranslator/internal/schema_source/aws_serverless_api.py
ConnorRobertson: Will do!

samtranslator/swagger/swagger.py
```diff
@@ -695,6 +700,10 @@ def add_auth_to_method(self, path: str, method_name: str, auth: Dict[str, Any],
             self._set_method_authorizer(path, method_name, method_authorizer, authorizers, method_scopes)  # type: ignore[no-untyped-call]
 
         method_apikey_required = auth and auth.get("ApiKeyRequired")
+
+        if not auth.get("AddApiKeyRequiredToCorsPreflight") and method_name == "options":
```
hoffa, line 704: So it's symmetrical (and easier to read) with the other very similar check:

```suggestion
        if method_name == "options" and not auth.get("AddApiKeyRequiredToCorsPreflight"):
```

hoffa, line 704: Is the type of `AddApiKeyRequiredToCorsPreflight` checked anywhere?

Currently if it's `"False"` (a string), `not "False"` will evaluate `False`.

ConnorRobertson: I don't think it is checked anywhere. Wouldn't "False" (a string) be an incorrect input or should that still be an acceptable input?

hoffa: I don't know. If you're unsure, add tests for that case.

samtranslator/swagger/swagger.py
```diff
@@ -612,14 +612,16 @@ def set_path_default_authorizer(  # noqa: too-many-branches
                 if "AWS_IAM" in method_definition["security"][0]:
                     self.add_awsiam_security_definition()
 
-    def set_path_default_apikey_required(self, path: str) -> None:
+    def set_path_default_apikey_required(self, path: str, required_options_api_key: bool = True) -> None:
```
aaythapa, line 615: is `required_options_api_key` optional with True as the default value?

aaythapa (dismissed):

samtranslator/swagger/swagger.py
ConnorRobertson: Yeah. Default value should be True and it is optional.

aaythapa: Should be Optional[bool] = True then i think

ssenchenko: Let's take a step back to Python without types. `required_options_api_key=True` means if `required_options_api_key` not passed to the function, set it to `True`. Compare it to `required_options_api_key=None`. It'd mean, if not passed to the function, set it to `None`. Now, `Optional` is equivalent to `Union[some_type, None]`. So `Optional[bool]` means `Union[bool, None]` and it actually means 3 possible values `True`, `False`, `None`. While `required_options_api_key: bool = True` still means only 2 possible values `True` and `False` with `True` if the variable is not passed to the function (it actually allows not to pass a variable). Connor is right.
P.S I hate typing in Python (and Typescript too)

ssenchenko (approved):

samtranslator/swagger/swagger.py
ConnorRobertson: That was a really good explanation. I'll change it back.

aaythapa: yep, thanks for the explanation slava

samtranslator/swagger/swagger.py
```diff
@@ -691,10 +696,15 @@ def add_auth_to_method(self, path: str, method_name: str, auth: Dict[str, Any],
         method_scopes = auth and auth.get("AuthorizationScopes")
         api_auth = api and api.get("Auth")
         authorizers = api_auth and api_auth.get("Authorizers")
+        required_options_api_key = method_name == "options" and (auth.get("AddApiKeyRequiredToCorsPreflight") is False)
```
hoffa, line 699: To improve legibility, code should be closest to where it's used. So unless it's needed elsewhere having this right above the `if` is preferred.

hoffa, line 699: We should fail if it's not a boolean. Otherwise values such as `"false"` won't enable the setting, but will quietly transform without error.

You can use this:

https://github.com/aws/serverless-application-model/blob/1034e333cf1020a15a35329dd110446e9bc70cfb/samtranslator/validator/value_validator.py#L123-L128

hoffa (approved):

PR #2993, opened 2023-03-03
feat: Add a new property `SeparateRecordSetGroup` to disable merging into record set group

xazhao:
### Issue #, if available
Following up change of https://github.com/aws/serverless-application-model/pull/2408
When there are multiple APIs share the same `HostedZoneId`, the router53 record set will be merged into a record set group. A `Condition` in one of the APIs could be overrode in merging process.
With the new property `SeparateRecordSets` under Route53, users can assign to different `Condition`s to each API so it's more flexible.

### Checklist

    - [ ] Using correct values
    - [ ] Using wrong values

### Examples?


samtranslator/model/api/api_generator.py
```diff
@@ -564,6 +566,15 @@ def _construct_api_domain(  # noqa: too-many-branches
             logical_id = "RecordSetGroup" + logical_id_suffix
 
             record_set_group = route53_record_set_groups.get(logical_id)
+
+            if self.individual_recordset:
+                return (
```
ssenchenko, line 571: nit: because we return now 4 of them, making it a named Tuple will be more convinient for future maintenance

samtranslator/model/api/api_generator.py
```diff
@@ -634,7 +679,17 @@ def to_cloudformation(self, redeploy_restapi_parameters, route53_record_set_grou
         permissions = self._construct_authorizer_lambda_permission()
         usage_plan = self._construct_usage_plan(rest_api_stage=stage)
 
-        return rest_api, deployment, stage, permissions, domain, basepath_mapping, route53, usage_plan
+        return (
```
ssenchenko, line 682: nit: those long tuples are easier to maintain when they are named, no need to double-check that the order is not broken

samtranslator/model/api/api_generator.py
xazhao: I like the idea. Will update it.

samtranslator/model/api/api_generator.py
```diff
@@ -442,12 +444,12 @@ def _construct_stage(
 
     def _construct_api_domain(  # noqa: too-many-branches
         self, rest_api: ApiGatewayRestApi, route53_record_set_groups: Any
-    ) -> Tuple[Optional[ApiGatewayDomainName], Optional[List[ApiGatewayBasePathMapping]], Any]:
+    ) -> Tuple[Optional[ApiGatewayDomainName], Optional[List[ApiGatewayBasePathMapping]], Any, Any]:
```
hoffa, line 447: Is `Any` correct?

samtranslator/internal/schema_source/aws_serverless_api.py
```diff
@@ -173,6 +173,7 @@ class EndpointConfiguration(BaseModel):
 TracingEnabled = Optional[PassThroughProp]
 OpenApiVersion = Optional[Union[float, str]]  # TODO: float doesn't exist in documentation
 AlwaysDeploy = Optional[bool]
+IndividualRecordSet = Optional[bool]
```
hoffa, line 176: Should be plural?

hoffa, line 176: Would under `Domain` or `Route53` make more sense? Or does it make sense even when they aren't set?

xazhao: I think under `Route53` makes more sense.  Make it unset by default is for backward compatibility

xazhao: This means the API will create a standalone Route53 RecordSet. So I feel it's not plural.

xazhao: Oh nvm it could be two. You're right.

hoffa: I wonder if `SeparateRecordSets` would be clearer. 🤔 I can't articulate well why "individual" feels slightly off... maybe because it's often used to refer to people? ("an individual")

samtranslator/model/api/api_generator.py
```diff
@@ -626,14 +676,31 @@ def _construct_alias_target(self, domain: Dict[str, Any], api_domain_name: str,
...
+        ApiGatewayRestApi,
+        ApiGatewayDeployment,
+        ApiGatewayStage,
+        List[LambdaPermission],
+        Optional[ApiGatewayDomainName],
+        Optional[List[ApiGatewayBasePathMapping]],
+        Optional[Route53RecordSetGroup],
+        Optional[Any],
+        Optional[List[Route53RecordSet]],
+    ]:
```
aahung, line 691: omg this is... can we improve this function to make it return `List[SamResource]` or `List[Any]`

samtranslator/model/api/api_generator.py
```diff
@@ -565,6 +574,15 @@ def _construct_api_domain(  # noqa: too-many-branches
             logical_id = "RecordSetGroup" + logical_id_suffix
 
             record_set_group = route53_record_set_groups.get(logical_id)
+
+            if route53.get("SeparateRecordSets"):
```
aahung, line 578: `SeparateRecordSets` cannot be intrinsic right? Otherwise this might be an intrinsic resolving to `False`

samtranslator/model/api/api_generator.py
xazhao: Sure that's ugly.

xazhao: Maybe a dataclass? With a `List` we have to remember what resource that index matches to.

aahung: I don't think there is flexibility. `to_cloudformation` has to return an iterable of CFN resources

hoffa: Do the specific types matter? Shouldn't they all be by design `Resource`s?

xazhao: Oh right! Yes a List of `Resource` should work

samtranslator/model/api/api_generator.py
hoffa: Correct, this might wrongly enable it. Need to validate type as it's a nested value. :/

samtranslator/model/api/api_generator.py
hoffa: Looking at the code it does depend on the specific value though, so IMO `Tuple` makes the most sense without large refactor. 🤷‍♂️ Your call.

samtranslator/model/api/api_generator.py
xazhao: Added type validation and intrinsic check.

samtranslator/model/api/api_generator.py
xazhao: Did a little refactoring so it can return a list now.

samtranslator/model/api/api_generator.py
```diff
@@ -646,7 +706,21 @@ def to_cloudformation(self, redeploy_restapi_parameters, route53_record_set_grou
         permissions = self._construct_authorizer_lambda_permission()
         usage_plan = self._construct_usage_plan(rest_api_stage=stage)
 
-        return rest_api, deployment, stage, permissions, domain, basepath_mapping, route53, usage_plan
+        generated_resources: List[Any] = []
```
hoffa, line 709: Why `Any`? I'd rather have a verbose, well-typed tuple than no typing.

samtranslator/model/sam_resources.py
```diff
@@ -1276,28 +1275,18 @@ def to_cloudformation(self, **kwargs):  # type: ignore[no-untyped-def]
...
-        if route53:
-            resources.extend([route53])
-        # contains usage plan, api key and usageplan key resources
-        if usage_plan_resources:
-            resources.extend(usage_plan_resources)
+        resources: List[Any] = []
+
+        for resource in generated_api_resources:
+            if resource:
+                if isinstance(resource, (list, tuple)):
```
hoffa, line 1286: Why does `to_cloudformation` return both collections and individual `Resource`s?

samtranslator/model/api/api_generator.py
xazhao: The goal is to make it as a `List[Resource]` but that requires more refactoring. E.g. `permissions` is a list and `usage_plan` is a tuple so it's mixed type. So I use `Any` for now and will make it `List[Resource]` when the refactoring is done.

samtranslator/model/sam_resources.py
xazhao: See my above comment. The tuple it was using before includes both collections and individual resources. Will refactor that.

samtranslator/model/api/api_generator.py
hoffa: Then is it `List[Union[Resource, List[Resource], Tuple[Resource]]]`?

xazhao: Good point! Added types to the list

samtranslator/model/api/api_generator.py
```diff
@@ -565,6 +575,18 @@ def _construct_api_domain(  # noqa: too-many-branches
             logical_id = "RecordSetGroup" + logical_id_suffix
 
             record_set_group = route53_record_set_groups.get(logical_id)
+
+            if route53.get("SeparateRecordSets") and not is_intrinsic(route53.get("SeparateRecordSets")):
+                sam_expect(
+                    route53.get("SeparateRecordSets"), self.logical_id, "Domain.Route53.SeparateRecordSets"
+                ).to_be_a_string
```
hoffa, line 582: Dead code. Also untested.

xazhao: Good catch. Fixed and added tests

samtranslator/model/api/api_generator.py
```diff
@@ -565,6 +575,18 @@ def _construct_api_domain(  # noqa: too-many-branches
             logical_id = "RecordSetGroup" + logical_id_suffix
 
             record_set_group = route53_record_set_groups.get(logical_id)
+
+            if route53.get("SeparateRecordSets") and not is_intrinsic(route53.get("SeparateRecordSets")):
```
hoffa, line 579: Isn't the

```
and not is_intrinsic(route53.get("SeparateRecordSets"))
```

unnecessary? Since it checks `to_be_a_bool`.

samtranslator/model/api/api_generator.py
```diff
@@ -576,7 +598,39 @@ def _construct_api_domain(  # noqa: too-many-branches
...
+        logical_id_ipv6 = "RecordSetIpv6" + logical_id_suffix
+
+        recordset_list = []
+
+        recordset = Route53RecordSet(logical_id, attributes=self.passthrough_resource_attributes)
+        recordset.Name = domain.get("DomainName")
+        recordset.Type = "A"
+        recordset.AliasTarget = self._construct_alias_target(domain, api_domain_name, route53)
+
+        if route53.get("HostedZoneId"):
```
hoffa, line 619: nit: could avoid the repetition with `HostedZoneId`, `HostedZoneName` and `DomainName` by assigning to variable at start

samtranslator/model/api/api_generator.py
xazhao: good point. Removed.

samtranslator/model/api/api_generator.py
xazhao: Updated

samtranslator/model/api/api_generator.py
```diff
@@ -576,7 +598,39 @@ def _construct_api_domain(  # noqa: too-many-branches
...
+            recordset.HostedZoneId = route53.get("HostedZoneId")
+        else:
+            recordset.HostedZoneName = route53.get("HostedZoneName")
+
+        recordset_list.extend([recordset])
+
+        if route53.get("IpV6"):
+            recordset_ipv6 = Route53RecordSet(logical_id_ipv6, attributes=self.passthrough_resource_attributes)
```
hoffa, line 628: Thoughts on having this as part of the `RecordSetGroup` for the domain (i.e. both IPv4 and IPv6 under same group)? Seems neat, and might help with preventing hitting the 500-resource limit.

samtranslator/model/api/api_generator.py
```diff
@@ -626,14 +681,21 @@ def _construct_alias_target(self, domain: Dict[str, Any], api_domain_name: str,
...
 
         :returns: a tuple containing the RestApi, Deployment, and Stage for an empty Api.
         :rtype: tuple
         """
         rest_api = self._construct_rest_api()
-        domain, basepath_mapping, route53 = self._construct_api_domain(rest_api, route53_record_set_groups)
+        apiDomainResponse = self._construct_api_domain(rest_api, route53_record_set_groups)
+        domain = apiDomainResponse.Domain
+        basepath_mapping = apiDomainResponse.ApiGWBasePathMappingList
+        route53_recordsetGroup = apiDomainResponse.RecordSetsGroup
```
hoffa, line 696: Mixing both snake case and camel case. 😩 Let's use `snake_case` everywhere.

samtranslator/model/api/api_generator.py
```diff
@@ -576,7 +598,40 @@ def _construct_api_domain(  # noqa: too-many-branches
...
+        logical_id_suffix = LogicalIdGenerator("", [hostedZoneId or hostedZoneName, domainName]).gen()
+        logical_id = "RecordSet" + logical_id_suffix
+        logical_id_ipv6 = "RecordSetIpv6" + logical_id_suffix
+
+        recordset_list = []
+
+        recordset = Route53RecordSet(logical_id, attributes=self.passthrough_resource_attributes)
+        recordset.Name = domain.get("DomainName")
+        recordset.Type = "A"
+        recordset.AliasTarget = self._construct_alias_target(domain, api_domain_name, route53)
```
hoffa, line 618: Can reuse the target for IPv6 record as well.

xazhao: Updated

samtranslator/model/api/api_generator.py
xazhao: lol updated

samtranslator/model/api/api_generator.py
xazhao: Synced offline. Decided to use `RecordSetGroup` to combine IPv4/IPv6

samtranslator/model/route53.py
```diff
@@ -15,3 +15,19 @@ class Route53RecordSetGroup(Resource):
     HostedZoneId: Optional[Intrinsicable[str]]
     HostedZoneName: Optional[Intrinsicable[str]]
     RecordSets: Optional[List[Any]]
+
+
+class Route53RecordSet(Resource):
```
hoffa, line 20: Unused?

samtranslator/model/sam_resources.py
```diff
@@ -1276,28 +1275,18 @@ def to_cloudformation(self, **kwargs):  # type: ignore[no-untyped-def]
...
-        if domain:
-            resources.extend([domain])
-        if basepath_mapping:
-            resources.extend(basepath_mapping)
-        if route53:
-            resources.extend([route53])
-        # contains usage plan, api key and usageplan key resources
-        if usage_plan_resources:
-            resources.extend(usage_plan_resources)
+        resources: List[Any] = []
```
hoffa, line 1282: If this type is necessary, I'd rather use whatever type is correct, than ignore it.

samtranslator/model/api/api_generator.py
```diff
@@ -576,27 +596,45 @@ def _construct_api_domain(  # noqa: too-many-branches
...
+        return ApiDomainResponse(domain, basepath_resource_list, record_set_group)
+
+    def _construct_single_record_set_group(
+        self, domain: Dict[str, Any], api_domain_name: str, route53: Any
+    ) -> Route53RecordSetGroup:
+        hostedZoneId = route53.get("HostedZoneId")
+        hostedZoneName = route53.get("HostedZoneName")
+        domainName = domain.get("DomainName")
+        logical_id_suffix = LogicalIdGenerator("", [hostedZoneId or hostedZoneName, domainName]).gen()
+        logical_id = "RecordSetGroup" + logical_id_suffix
```
hoffa, line 608: Would this be identical?


```suggestion
        logical_id = LogicalIdGenerator("RecordSetGroup", [hostedZoneId or hostedZoneName, domainName]).gen()
```

samtranslator/internal/schema_source/aws_serverless_api.py
```diff
@@ -134,6 +134,7 @@ class Route53(BaseModel):
     IpV6: Optional[bool] = route53("IpV6")
     SetIdentifier: Optional[PassThroughProp]  # TODO: add docs
     Region: Optional[PassThroughProp]  # TODO: add docs
+    SeparateRecordSetGroup: Optional[bool]  # TODO: add docs
```
hoffa, line 137: If it's plural "separate record set groups", it just means record set groups are separated (whatever it means). If it's singular "separate record set group", my first question is "separate what record set group?".

So IMO either we should keep it singular but specify what record set group it separates, or keep it plural so it's more generically just saying record set groups are separated.

hoffa: Oh I see, so it's a single separate record set group for this API, but if it's diabled then it'll do some "optimizations" and join records from multiple APIs?

xazhao: exactly!

samtranslator/model/route53.py
xazhao: Deleted

samtranslator/model/api/api_generator.py
```diff
@@ -626,14 +665,20 @@ def _construct_alias_target(self, domain: Dict[str, Any], api_domain_name: str,
         return alias_target
 
     @cw_timer(prefix="Generator", name="Api")
-    def to_cloudformation(self, redeploy_restapi_parameters, route53_record_set_groups):  # type: ignore[no-untyped-def]
+    def to_cloudformation(
+        self, redeploy_restapi_parameters: Optional[Any], route53_record_set_groups: Dict[str, Route53RecordSetGroup]
+    ) -> List[Any]:
```
hoffa, line 670: Any better types?

hoffa (approved):

samtranslator/model/api/api_generator.py
xazhao: That reminds me we can move un-nest step into the `to_cloudformation` so we can use `List<Resource>` here. Updated.

aahung (approved):

PR #3000, opened 2023-03-06
fix: Refactor Serverless::GraphQLDataSource

pradhapanr:
### Description of changes
`Serverless::GraphQLDataSource` functionalities will now be grouped underneath the `Serverless::GraphQLApi` resource.

These changes remove the `Serverless::GraphQLDataSource` and add their respective properties to `GraphQLApi`.

Currently, only DynamoDB data sources are supported with these changes. The property mappings have changed slightly as well, now each datasource will have its own property (i.e `DynamoDBDataSources`) where they can then create all the data sources they want.
Wrote transform tests. Passes `make pr`.

### Checklist

    - [ ] Using correct values
    - [ ] Using wrong values

### Examples?


samtranslator/model/sam_resources.py
```diff
@@ -2279,89 +2284,43 @@ def _construct_appsync_schema(
...
+            )
 
-    ApiId: PassThrough
-    Type: str
-    DynamoDBConfig: Optional[Dict[str, Any]]
-    Name: Optional[str]
-    Description: Optional[PassThrough]
-    ServiceRoleArn: Optional[str]
+            cfn_datasource.ServiceRoleArn = ddb_datasource.ServiceRoleArn
+            cfn_datasource.Name = ddb_datasource.Name if ddb_datasource.Name else relative_id
```
hoffa, line 2298: nit: a common pattern that avoids the repetition is `a or b`

samtranslator/model/sam_resources.py
```diff
@@ -2279,89 +2284,43 @@ def _construct_appsync_schema(
...
+            cfn_datasource.Type = "AMAZON_DYNAMODB"
+            cfn_datasource.ApiId = api_id
 
-    @cw_timer
-    def to_cloudformation(self, **kwargs: Any) -> List[Resource]:
-        model = self.validate_properties_and_return_model(aws_serverless_graphqldatasource.Properties)
+            if ddb_datasource.Description:
+                cfn_datasource.Description = cast(str, ddb_datasource.Description)
```
hoffa, line 2303: The model says it's a pass-through, so we don't know that it's a string. It could be a complex intrinsic.

samtranslator/model/sam_resources.py
```diff
@@ -2279,89 +2284,43 @@ def _construct_appsync_schema(
...
 
-    def _parse_dynamodb_datasource(
-        self, ddb_properties: aws_serverless_graphqldatasource.DynamoDBConfig
-    ) -> DynamoDBConfigType:
+    def _parse_ddb_config(self, ddb_datasource: aws_serverless_graphqlapi.DynamoDBDataSource) -> DynamoDBConfigType:
         ddb_config: DynamoDBConfigType = {}
 
-        ddb_config["TableName"] = ddb_properties.TableName
-        ddb_config["AwsRegion"] = ddb_properties.Region or ref("AWS::Region")
+        ddb_config["AwsRegion"] = cast(str, ddb_datasource.Region) or ref("AWS::Region")
```
hoffa, line 2313: Same as above, per the model we don't know that it's a string.

Applies to all `cast`s in this PR.

pradhapanr: What should our resolution be here then? Typing passthrough properties as also `Instrinsicable` in the CFN resource type and then casting as `Intrinsicable[Type]` comes to mind right away.

@GavinZZ I remember you put down some TODOs for the type casting for `Passthrough` stuff. Any ideas?

pradhapanr: Resolved with @hoffa offline: So it looks like `mypy` doesn't recognize that `PassThroughProp` is basically just `Any` type. So for now, we are going to cast to `Passthrough` in our code and see if we can find a resolution for this otherwise.

hoffa (approved):

ssenchenko (approved):

PR #3006, opened 2023-03-08
feat: Add "Implied" Connector in DataSource

pradhapanr:
### Description of changes
Before, the only way a customer could assign permissions to a DDB DataSource was by declaring the `ServiceRoleArn` property.

With these changes, we add two new properties: `TableArn` and `Properties` which allow customers to define permissions within a template, which we will then use to create a DataSource -> Table connector.

Because the DataSource property naturally defines a lot of the properties required for the creation of a connector, we only need these two new properties to build it for them.

Added transform tests. Passes `make pr`.
### Checklist

    - [x] Using correct values
    - [x] Using wrong values

### Examples?


samtranslator/model/sam_resources.py
```diff
@@ -2300,9 +2301,43 @@ def _construct_ddb_datasources(
                 cfn_datasource.Description = cast(PassThrough, ddb_datasource.Description)
 
             cfn_datasource.DynamoDBConfig = self._parse_ddb_config(ddb_datasource)
-            datasources.append(cfn_datasource)
 
-        return datasources
+            if ddb_datasource.ServiceRoleArn:
+                cfn_datasource.ServiceRoleArn = cast(PassThrough, ddb_datasource.ServiceRoleArn)
+            else:
+                # If the user doesn't have their own role, then we will create for them if TableArn and Permissions is defined.
```
hoffa, line 2308: This is a good candidate for its own separate function that early-returns, avoiding the deeply nested (hence typically hard-to-read) blocks of code.

samtranslator/model/sam_resources.py
```diff
@@ -2321,3 +2356,32 @@ def _parse_ddb_config(self, ddb_datasource: aws_serverless_graphqlapi.DynamoDBDa
             ddb_config["DeltaSyncConfig"] = cast(PassThrough, deltasync_properties)
 
         return ddb_config
+
+    def _construct_ddb_datasource_connector_resources(  # noqa: PLR0913
+        self,
```
hoffa, line 2361: Unused.

samtranslator/internal/schema_source/aws_serverless_graphqlapi.py
```diff
@@ -25,9 +25,10 @@ class DeltaSync(BaseModel):
 
 
 class DynamoDBDataSource(BaseModel):
-    # TableArn, Permissions properties for connector
-    ServiceRoleArn: PassThroughProp  # TODO: make optional when we ship connector changes
     TableName: str
+    ServiceRoleArn: Optional[PassThroughProp]
+    TableArn: Optional[str]
+    Permissions: Optional[List[str]]
```
hoffa, line 31: Should this be one of:

https://github.com/aws/serverless-application-model/blob/c29f64a069ead2658a2e5412df53002b85bde3ee/samtranslator/internal/schema_source/aws_serverless_connector.py#L10

samtranslator/internal/model/appsync.py
```diff
@@ -1,11 +1,13 @@
-from typing import Any, Dict, List, Optional, Union
+from typing import Any, Dict, List, Optional, Tuple, Union
 
 from typing_extensions import TypedDict
 
 from samtranslator.model import GeneratedProperty, Resource
 from samtranslator.model.intrinsics import fnGetAtt
 from samtranslator.utils.types import Intrinsicable
 
+ConnectorIdAndProperties = Tuple[str, Dict[str, Any]]
```
hoffa, line 9: Unused.

samtranslator/model/sam_resources.py
pradhapanr: What's unused here?

samtranslator/internal/model/appsync.py
pradhapanr: nice catch

samtranslator/model/sam_resources.py
hoffa: `self`

samtranslator/internal/schema_source/aws_serverless_graphqlapi.py
pradhapanr: Yep I think thats more correct, let me use that typing instead

samtranslator/model/sam_resources.py
```diff
@@ -2300,9 +2302,55 @@ def _construct_ddb_datasources(
...
+                resources.append(role)
+
+            if connector_resources:
+                resources.extend(connector_resources)
+
+            resources.append(cfn_datasource)
+
+        return resources
+
+    def parse_datasource_role(
```
hoffa, line 2320: Prefix with `_` so it's private by convention, otherwise consumers might be too eager to depend on it.

samtranslator/model/sam_resources.py
```diff
@@ -2300,9 +2302,55 @@ def _construct_ddb_datasources(
...
-        return datasources
+            cfn_datasource.ServiceRoleArn, role, connector_resources = self.parse_datasource_role(
+                ddb_datasource, cfn_datasource.get_runtime_attr("arn"), relative_id, kwargs
+            )
+
+            if role:
+                resources.append(role)
+
+            if connector_resources:
+                resources.extend(connector_resources)
```
hoffa, line 2314: I assume no need to return `role` and `connector_resources` separately if they're all added to an opaque `resources` anyway?

samtranslator/model/sam_resources.py
pradhapanr: Nice catch, that is definitely a typo thanks

samtranslator/model/sam_resources.py
pradhapanr: Yea exactly. So we'll move our `ServiceRoleArn` check back into the loop, and then only execute this function when we know we need to generate `role` and `connector_resources`. Should make this much more intuitive

ssenchenko (dismissed):

samtranslator/model/sam_resources.py
```diff
@@ -2282,16 +2283,18 @@ def _construct_appsync_schema(
         return schema
 
     def _construct_ddb_datasources(
-        self, ddb_datasources: Dict[str, aws_serverless_graphqlapi.DynamoDBDataSource], api_id: Intrinsicable[str]
-    ) -> List[DataSource]:
-        datasources: List[DataSource] = []
+        self,
+        ddb_datasources: Dict[str, aws_serverless_graphqlapi.DynamoDBDataSource],
+        api_id: Intrinsicable[str],
+        kwargs: Dict[str, Any],
```
hoffa, line 2289: Caller says `**kwargs` is `Any`.

pradhapanr: I think when `kwargs` is unpacked with the `**` operator, it turns into `Dict[str, Any]`

samtranslator/model/sam_resources.py
```diff
@@ -2300,9 +2303,49 @@ def _construct_ddb_datasources(
...
+            attributes=self.resource_attributes,
+        )
+        role.AssumeRolePolicyDocument = IAMRolePolicies.construct_assume_role_policy_for_service_principal(
+            "appsync.amazonaws.com"
+        )
+        role.RoleName = role_id
+        role_arn = role.get_runtime_attr("arn")
+
+        connector_resources = self._construct_ddb_datasource_connector_resources(
+            relative_id, datasource_arn, table_arn, permissions, role_id, kwargs
```
hoffa, line 2345: The role logical ID is also its name?

samtranslator/model/sam_resources.py
```diff
@@ -2300,9 +2303,49 @@ def _construct_ddb_datasources(
...
+        role_id = f"{relative_id}Role"
+        role = IAMRole(
+            logical_id=role_id,
+            depends_on=self.depends_on,
+            attributes=self.resource_attributes,
+        )
+        role.AssumeRolePolicyDocument = IAMRolePolicies.construct_assume_role_policy_for_service_principal(
+            "appsync.amazonaws.com"
+        )
+        role.RoleName = role_id
```
hoffa, line 2341: We shouldn't set names (especially global ones such as for IAM) unless necessary. This means two stacks deploying the same template will fail due to naming conflict. Instead we should pass an intrinsic that resolves to whatever CloudFormation decides.

samtranslator/model/sam_resources.py
pradhapanr: https://stackoverflow.com/questions/57897649/are-the-keys-of-a-kwargs-argument-to-python-function-guaranteed-to-be-type-strin

samtranslator/model/sam_resources.py
```diff
@@ -2321,3 +2363,32 @@ def _parse_ddb_config(self, ddb_datasource: aws_serverless_graphqlapi.DynamoDBDa
...
+    ) -> List[Resource]:
+        logical_id = f"{datasource_id}ToTableConnector"
+        connector_dict = {
+            "Type": "AWS::Serverless::Connector",
+            "Properties": {
+                "Source": {"Type": "AWS::AppSync::DataSource", "Arn": source_arn, "RoleName": role_name},
+                "Destination": {
+                    "Type": "AWS::DynamoDB::Table",
+                    "Arn": destination_arn,
+                },
```
hoffa, line 2384: Why not just pass the `Id`?

hoffa (approved):

ssenchenko (approved):

PR #3026, opened 2023-03-14
chore: update properties in schema with new docs

aaythapa:
### Issue #, if available

N/A

### Description of changes

Updated schema by adding docs for new properties

### Description of how you validated changes

`make schema`

### Checklist

- [x] Adheres to the [development guidelines](https://github.com/aws/serverless-application-model/blob/develop/DEVELOPMENT_GUIDE.md#development-guidelines)
- [ ] Add/update [transform tests](https://github.com/aws/serverless-application-model/blob/develop/DEVELOPMENT_GUIDE.md#unit-testing-with-multiple-python-versions)
    - [ ] Using correct values
    - [ ] Using wrong values
- [ ] Add/update [integration tests](https://github.com/aws/serverless-application-model/blob/develop/INTEGRATION_TESTS.md)

### Examples?

Please reach out in the comments if you want to add an example. Examples will be 
added to `sam init` through [aws/aws-sam-cli-app-templates](https://github.com/aws/aws-sam-cli-app-templates).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #3027, opened 2023-03-14
ci: Exclude samtranslator.internal from interface scan

aahung:
### Description of changes

See #3026, an example of unnecessary workflow failure

### Checklist

    - [ ] Using correct values
    - [ ] Using wrong values

### Examples?

### DIFF
--- a/bin/public_interface.py
+++ b/bin/public_interface.py
@@ -15,17 +15,20 @@ import os.path
 import pkgutil
 import sys
 from pathlib import Path
-from typing import Any, Dict, List, NamedTuple, Set, Union
+from typing import Any, Dict, List, NamedTuple, Optional, Set, Union
 
 _ARGUMENT_SELF = {"kind": "POSITIONAL_OR_KEYWORD", "name": "self"}
 
 
 class InterfaceScanner:
-    def __init__(self) -> None:
+    def __init__(self, skipped_modules: Optional[List[str]] = None) -> None:
         self.signatures: Dict[str, Union[inspect.Signature]] = {}
         self.variables: Set[str] = set()
+        self.skipped_modules: Set[str] = set(skipped_modules) or set()
 
     def scan_interfaces_recursively(self, module_name: str) -> None:
+        if module_name in self.skipped_modules:
+            return
         self._scan_interfaces_in_module(module_name)
         for submodule in pkgutil.iter_modules([module_name.replace(".", os.path.sep)]):
             submodule_name = module_name + "." + submodule.name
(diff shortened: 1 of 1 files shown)


bin/public_interface.py
```diff
@@ -212,13 +215,19 @@ def main() -> None:
     subparsers = parser.add_subparsers(dest="command")
     extract = subparsers.add_parser("extract", help="Extract public interfaces")
     extract.add_argument("--module", help="The module to extract public interfaces", type=str, default="samtranslator")
+    extract.add_argument(
+        "--skipped-modules",
+        help="The modules separated by ',' that should be skipped",
```
acristin, line 220: Minor nit: `"The modules, separated by ',', that should be skipped"`

bin/public_interface.py
```diff
@@ -212,13 +215,19 @@ def main() -> None:
...
+        default="samtranslator.internal",
+    )
     check = subparsers.add_parser("check", help="Check public interface changes")
     check.add_argument("original_json", help="The original public interface JSON file", type=Path)
     check.add_argument("new_json", help="The new public interface JSON file", type=Path)
     args = parser.parse_args()
 
     if args.command == "extract":
-        scanner = InterfaceScanner()
+        scanner = InterfaceScanner(skipped_modules=args.skipped_modules.split(","))
```
acristin, line 230: Should we trim the resulting strings in case the customer adds spaces between the comma and the next item?

acristin (approved):
Overall looks good; minor suggestion around UX that you're free to ignore or put off until later.

xazhao (approved):

bin/public_interface.py
```diff
@@ -212,13 +215,19 @@ def main() -> None:
     subparsers = parser.add_subparsers(dest="command")
     extract = subparsers.add_parser("extract", help="Extract public interfaces")
     extract.add_argument("--module", help="The module to extract public interfaces", type=str, default="samtranslator")
+    extract.add_argument(
+        "--skipped-modules",
+        help="The modules separated by ',' that should be skipped",
+        type=str,
+        default="samtranslator.internal",
```
hoffa, line 222: You can use `nargs='+'` so you don't need the `split`.

aaythapa (approved):

hoffa (approved):

bin/public_interface.py
aahung: nargs="*"

bin/public_interface.py
aahung: no longer apply due to nargs

Issue #2581, opened 2022-11-08
Transform fails when `DeletionProtection` specified in `AWS::Cognito::UserPool` with Cognito event


jjaimez:
### Describe your idea/feature/enhancement
 I wish the SAM Translator would support [DeletionProtection](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-cognito-userpool.html#cfn-cognito-userpool-deletionprotection) in Cognito User Pool definition.

### Proposal

```
MyUserPool:
    Type: AWS::Cognito::UserPool
    Properties:
      DeletionProtection: ACTIVE
```

Things to consider:

### Additional Details
I tried to do it and i got the following error.  
`property DeletionProtection not defined for resource of type AWS::cognito::UserPool`

xazhao:
Thanks for the feature request. I will discuss this with the team and post updates here.

ssenchenko:
`DeletionProtection` seems to be quite a fresh feature. [The blog](https://aws.amazon.com/about-aws/whats-new/2022/10/amazon-cognito-console-user-pool-deletion-protection/) which mentions it, is dated of Oct 2022.
It explains why `DeletionProtection` is not currently supported by CloudFormation https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-cognito-userpool.html#cfn-cognito-userpool-deletionprotection
We can add it after CloudFormation starts supporting it.

ssenchenko:
Feel free to reopen this issue if you notice that CloudFormation has started to support this property

JP-tech-sh:
[...]

PR #3041, opened 2023-03-17
fix: add `DeletionProtection` to `AWS::Cognito::UserPool`

hoffa:
### Issue #, if available

https://github.com/aws/serverless-application-model/issues/2581

### Description of changes

This is a quick fix to the issue, but we should fix the root cause which is that it validates the resource specified by a customer (see https://github.com/aws/serverless-application-model/issues/2581#issuecomment-1474422567). So this will keep happening for every new property. I'm toying with a longer-term fix in https://github.com/aws/serverless-application-model/pull/3040.

### Checklist

    - [x] Using correct values
    - [ ] Using wrong values

### Examples?


aahung (approved):
oh that sucks

acristin (approved):

PR #3088, opened 2023-04-05
fix: Connector "Type" cannot be Serverless resource types

GavinZZ:
### Description of changes
There is a connector bug when using `RoleName` + `Type`.
```
MyReadConnector:
    Type: AWS::Serverless::Connector
    Properties:
      Source:
        RoleName: MyRole
      Destination:
        Id: MyTable
      Permissions:
      - Read
 ```
Because we didn't input `Id`, it will directly use `Type` and try to find a profile using `AWS::Serverless::Function` to `AWS::DynamoDB::Table`. It will fail because not such profile is found. We should be using `AWS::lambda::Function` instead.

Added tests

### Checklist

    - [ ] Using correct values
    - [ ] Using wrong values

### Examples?

### DIFF
--- a/samtranslator/model/connector/connector.py
+++ b/samtranslator/model/connector/connector.py
@@ -28,7 +28,7 @@ ConnectorResourceReference = namedtuple(
     ],
 )
 
-SAM_TO_CFN_RESOURCE_TYPE = {
+_SAM_TO_CFN_RESOURCE_TYPE = {
     SamResourceType.Function.value: LambdaFunction.resource_type,
     SamResourceType.StateMachine.value: StepFunctionsStateMachine.resource_type,
     SamResourceType.Api.value: ApiGatewayRestApi.resource_type,
@@ -105,11 +105,6 @@ def _is_valid_resource_reference(obj: Dict[str, Any]) -> bool:
     return id_provided != non_id_provided
 
 
-def get_underlying_cfn_resource_type(resource_type: str) -> str:
-    """profiles.json only support CFN resource type. We need to convert SAM resource types to corresponding CFN resource type"""
-    return SAM_TO_CFN_RESOURCE_TYPE.get(resource_type, resource_type)
-
-
 def get_resource_reference(
     obj: Dict[str, Any], resource_resolver: ResourceResolver, connecting_obj: Dict[str, Any]
 ) -> ConnectorResourceReference:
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Review by ssenchenko — APPROVED** [review]

File: samtranslator/model/connector/connector.py [hunk 1]
```diff
@@ -89,6 +89,17 @@ def _is_valid_resource_reference(obj: Dict[str, Any]) -> bool:
...
 
 
+def get_underlying_cfn_resource_type(resource_type: str) -> str:
+    sam_to_cfn_types = {
+        "AWS::Serverless::Function": "AWS::Lambda::Function",
+        "AWS::Serverless::StateMachine": "AWS::StepFunctions::StateMachine",
+        "AWS::Serverless::Api": "AWS::ApiGateway::RestApi",
+        "AWS::Serverless::SimpleTable": "AWS::DynamoDB::Table",
+        "AWS::Serverless::HttpApi": "AWS::ApiGatewayV2::Api",
+    }
```
aahung, line 99: this is a constant, it should be at the module level

aahung: Can we avoid re-typing them to avoid typos? Like `LambdaFunction.resource_type` and `SamResourceType.Function`.

GavinZZ: Good suggestion!

xazhao (approved):

samtranslator/model/connector/connector.py
```diff
@@ -20,6 +28,14 @@
     ],
 )
 
+SAM_TO_CFN_RESOURCE_TYPE = {
```
aahung, line 31: nit: 
```suggestion
_SAM_TO_CFN_RESOURCE_TYPE = {
```

samtranslator/model/connector/connector.py
```diff
@@ -89,6 +105,11 @@ def _is_valid_resource_reference(obj: Dict[str, Any]) -> bool:
     return id_provided != non_id_provided
 
 
+def get_underlying_cfn_resource_type(resource_type: str) -> str:
```
aahung, line 108: nit: 

```suggestion
def _get_underlying_cfn_resource_type(resource_type: str) -> str:
```

aahung (approved):

samtranslator/model/connector/connector.py
aahung: this function doesn't feel necessary anymore as it only has 1 LOC

Issue #3109, opened 2023-04-17
AWS::Scheduler::Schedule fails on creation when using same template on multiple stacks


cignuz-jp:
### Description

When creating a new stack from a template that has `AWS::Serverless::Function` using the `ScheduleV2` Event source type, change set execution fails with error: `<function-event> already exists in stack arn:aws:cloudformation:us-east-2:************:stack/firstScheduler/d4db7f60-dccb-11ed-aaac-02795b774f1e`

The [documentation page](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-property-function-schedulev2.html#sam-function-schedulev2-name) for `Name` property of the `ScheduleV2` event type mentions: 
>The name of the schedule. If you don't specify a name, AWS SAM generates a name in the format Function-Logical-IDEvent-Source-Name and uses that ID for the schedule name. 

From above statement it can be concluded that if the template is used on more than one stack the Name for the Event will not be unique which will lead to the error. 

 

### Steps to reproduce

1. Create a SAM application that has an AWS::Serverless::Function resource with an Event of type `ScheduleV2`. The following is a change I've made to the a SAM application based on the `quick-start-cloudwatch-events` template.

```yaml
  ScheduledEventLogger:
    Type: AWS::Serverless::Function
    Properties:
      Description: A Lambda function that logs the payload of messages sent to an associated SQS queue.
      Runtime: nodejs18.x
      Architectures:
        - x86_64
      Handler: src/handlers/scheduled-event-logger.scheduledEventLoggerHandler
      Events:
        EventBridgeScheduler:
          Type: ScheduleV2
          Properties:
            Description: scheduler to consume notifications
            ScheduleExpression: 'cron(0 0 ? * 2 *)'
            ScheduleExpressionTimezone: America/Los_Angeles
            FlexibleTimeWindow:
              Mode: 'OFF'
      MemorySize: 128
      Timeout: 100
```

2. Build and deploy the sam application to a new stack (e.g: firstScheduler)
3. Deploy the sam application to a different stack (e.g: secondScheduler).
### Observed result

The SAM application fails to deploy on 3rd step showing the error referencing the resource exists in the stack deployed on step 2.

### Expected result

The SAM application to be successfully deployed.

### Additional environment details

1. OS: MacOS Monterrey (12.6.1) on M1 chip
2. If using the [SAM CLI](https://github.com/aws/aws-sam-cli), `sam --version`: SAM CLI, version 1.79.0
[...]

PR #3120, opened 2023-04-19
fix: Add ScheduleV2 OmitName to omit generated Name in Scheduler::Schedule

aahung:
### Issue #, if available

https://github.com/aws/serverless-application-model/issues/3109

### Description of changes

This fixes a bug causing SAM always generates `AWS::Scheduler::Schedule`'s Name. However, we cannot break backward compatibility and make existing templates generate different results after this change. This PR introduces a new property `OmitName` so it doesn't generate `Name` in `AWS::Scheduler::Schedule` when it is set.

Added tests

### Checklist

    - [ ] Using correct values
    - [ ] Using wrong values

### Examples?


GavinZZ (approved):

ConnorRobertson (approved):

samtranslator/model/eventsources/scheduler.py
```diff
@@ -138,8 +138,17 @@ def _construct_scheduler_schedule_without_target(
...
-        # Scheduler schedule's Name is a required property
-        scheduler_schedule.Name = self.Name or self.logical_id
+        if self.Name:
+            scheduler_schedule.Name = self.Name
+        elif self.Name is False:
+            # When `Name` is `False` explicitly,
+            # we set `AWS::Scheduler::Schedule`'s Name to None
+            # and `Name` is not specified, to keep backward compatibility,
+            # we use name generated by SAM instead of passing `None`:
+            # https://github.com/aws/serverless-application-model/issues/3109
```
acristin, line 148: This comment is a bit confusing; the second part of the sentence (starting with "and \`Name\` is not specified")  refers to behavior in a different branch of the conditional. 
It might be clearer if you move the whole comment to right before the `if`, or move the second part of the sentence to the `else` block.

acristin (approved):
Comment is only a nit about a comment, so consider it non-blocking.

samtranslator/model/eventsources/scheduler.py
aahung: good idea

aahung: updated

GavinZZ (approved):

samtranslator/model/eventsources/scheduler.py
```diff
@@ -57,19 +58,20 @@ class SchedulerEventSource(ResourceMacro):
...
     FlexibleTimeWindow: Optional[Dict[str, Any]]
-    Name: Optional[str]
-    State: Optional[str]
-    Description: Optional[str]
-    StartDate: Optional[str]
-    EndDate: Optional[str]
-    ScheduleExpressionTimezone: Optional[str]
-    GroupName: Optional[str]
-    KmsKeyArn: Optional[str]
+    Name: Optional[Union[PassThrough, bool]]
```
xazhao, line 61: Do we still need `bool` here?

aahung: no we don't, nice catch

aahung: fixed

xazhao (approved):

samtranslator/model/eventsources/scheduler.py
aahung: no longer applicable

samtranslator/internal/schema_source/aws_serverless_statemachine.py
```diff
@@ -72,6 +72,7 @@ class ScheduleV2EventProperties(BaseModel):
     ScheduleExpressionTimezone: Optional[PassThroughProp] = scheduleeventv2properties("ScheduleExpressionTimezone")
     StartDate: Optional[PassThroughProp] = scheduleeventv2properties("StartDate")
     State: Optional[PassThroughProp] = scheduleeventv2properties("State")
+    OmitName: Optional[bool]  # TODO: add doc
```
hoffa, line 75: What would the docs look like?

samtranslator/model/eventsources/scheduler.py
```diff
@@ -49,6 +49,7 @@ class SchedulerEventSource(ResourceMacro):
         "RoleArn": PropertyType(False, IS_STR),
         "DeadLetterConfig": PropertyType(False, IS_DICT),
         "RetryPolicy": PropertyType(False, IS_DICT),
+        "OmitName": PropertyType(False, IS_BOOL, supports_intrinsics=False),
```
hoffa, line 52:
```suggestion
        "OmitName": Property(False, IS_BOOL),
```

acristin (approved):

samtranslator/internal/schema_source/aws_serverless_statemachine.py
aahung: > By default, the `Name` is always set for `AWS::Scheduler::Schedule` for backward compatibility. Set this property to true to omit the `Name`.

samtranslator/model/eventsources/scheduler.py
aahung: updated

samtranslator/internal/schema_source/aws_serverless_statemachine.py
hoffa: It doesn't _really_ omit the name from a customer perspective, as it'll still have a name, just whatever CloudFormation decides (instead of SAM). We can probably figure out some better wording for it, so I'm fine with this although it's a little confusing.

hoffa (approved):

samtranslator/internal/schema_source/aws_serverless_statemachine.py
aahung: true, "omit the name in SAM transform"

PR #3132, opened 2023-04-26
feat: actionable error on non-ARN policy

hoffa:
### Description of changes

Value is passed to [`ManagedPolicyArns`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-iam-role.html#cfn-iam-role-managepolicyarns) which must be ARNs:

https://github.com/aws/serverless-application-model/blob/b154e380d33cb0c28c9d015c11f5a5e929d6d99f/samtranslator/model/role_utils/role_constructor.py#L153-L154

https://github.com/aws/serverless-application-model/blob/b154e380d33cb0c28c9d015c11f5a5e929d6d99f/samtranslator/model/role_utils/role_constructor.py#L164

If it's a string but not an ARN, the deployment will fail either way, so it's backward compatible for us to fail instead with a more helpful error.

### Checklist

    - [x] Using correct values
    - [x] Using wrong values

### Examples?


samtranslator/model/role_utils/role_constructor.py
```diff
@@ -57,7 +57,7 @@ def _get_managed_policy_arn(
             if arn:
                 return arn
 
-    return name
+    return None
```
aahung, line 60: what if it is an intrinsic? We should return as it is

hoffa: `name` is guaranteed a string:

https://github.com/aws/serverless-application-model/blob/b154e380d33cb0c28c9d015c11f5a5e929d6d99f/samtranslator/model/role_utils/role_constructor.py#L142-L148

aahung: nvm we have `if isinstance(policy_arn, str):` in the caller.

aahung (approved):

ssenchenko (approved):

xazhao (approved):

PR #3141, opened 2023-05-02
chore: mark `test_request_parameters_open_api` as `@nonblocking`

hoffa:
### Description of changes

We had some transient errors with this test, so marking as `xfail`. The test itself was added nearly 2 years ago, so it's been tested on all regions.

As follow-up, we should most likely add a Pytest hook that checks if any test `xfail`ed, and publish an EMF metric if it did. This way we can soft-alarm on service issues or regressions.

Ran the test with expired credentials:

```
% pytest integration -k test_request_parameters_open_api
...
====== 188 deselected, 1 xfailed, 2 warnings in 4.69s =======
```

### Checklist

    - [ ] Using correct values
    - [ ] Using wrong values

### Examples?


ssenchenko (approved):

ssenchenko:
Why do you need to add a new transform test though?

hoffa:
> Why do you need to add a new transform test though?

It's identical to the one used by the integration test.

The integration test verifies the transformed output makes sense, and the transform test ensures we don't accidentally break that working template.

integration/helpers/base_test.py
```diff
@@ -54,6 +54,8 @@
     "api_with_custom_domains_regional_ownership_verification",
 ]
 
+nonblocking = pytest.mark.xfail
```
hoffa, line 57:
```suggestion
# In general, we should only mark integration tests as @nonblocking if
#   1. The test succeeded every region (this ensures the transformed output makes sense)
#   2. The test resources are defined in a single template without Parameters or other CloudFormation macros (this ensures we can represent transform using a transform test)
#   3. An identical transform test exists for the integration test template (this ensures we keep transforming the working template the same way)
nonblocking = pytest.mark.xfail
```

xazhao (approved):

aaythapa (approved):

PR #3163, opened 2023-05-15
Generate lambda authoriser permissions

ssenchenko:
### Description of changes
Adding Connector and generating its resources for a GraphQLApi authoriser

### Checklist

    - [ ] Using correct values
    - [ ] Using wrong values

### Examples?


aaythapa (approved):

schema_source/sam.schema.json
```diff
@@ -186,6 +156,36 @@
       "title": "AppSyncResolver",
       "type": "object"
     },
+    "Authoriser": {
+      "additionalProperties": false,
+      "properties": {
+        "LambdaAuthorizer": {
+          "$ref": "#/definitions/LambdaAuthoriserConfig"
```
awood45, line 163: This doesn't match the GraphQL resource - why are we using "Authoriser" instead of "Authorizer"?

awood45 (changes requested):

samtranslator/model/sam_resources.py
```diff
@@ -2266,53 +2273,64 @@ def _construct_appsync_api_resources(
...
 
         # This would be much easier and type-safe if accessing the properties in Resource
         # was a dictionary. Currently, top level properties are class attributes, but nested
         # properties are dictionaries...
         # TODO: Access properties in Resource class as dict?
-        if name:
+        if name and auth_dict:
             setattr(api, name, auth_dict)
+            if name == "LambdaAuthorizerConfig":
+                lambda_auth_arns.append(cast(LambdaAuthorizerConfigType, auth_dict)["AuthorizerUri"])
```
hoffa, line 2307: Will there be more of these `name` types? I notice this is duplicated below.

hoffa (approved):

samtranslator/model/sam_resources.py
```diff
@@ -2204,12 +2204,16 @@ def __init__(
     @cw_timer
     def to_cloudformation(self, **kwargs: Any) -> List[Resource]:
```
hoffa, line 2205: I don't see any transform failure tests; is it expected that there are no new transform failure cases in the code?

samtranslator/model/sam_resources.py
ssenchenko: Nope. I didn't want to go in additional refactoring here. But 2 this is it. The only reason to use it because Python3.7 which we still support doesn't support `isinstance(obj, SubTypeOfTypedDict)` :(

samtranslator/model/sam_resources.py
ssenchenko: There have been transform error tests for authorisers. I haven't change them

samtranslator/model/sam_resources.py
hoffa: Would it simplify things if `_validate_and_extract_authorizer_config` returned a `Tuple` of `Optional`s for each type, instead of a single tuple and then branching based off of `name`?

ssenchenko: Sorry, I'm not following what exactly are you suggesting

hoffa: Currently it returns (name, object), and in caller we do specific logic based on the name. What if instead we return (object, object, object, ...), one for each possible object. And instead of checking the name, we check whether the optional is not `None`.

ssenchenko: I see. Sure. I'll do it in a separate pr, not to mess up with what I'm doing now

awood45 (approved):

PR #3176, opened 2023-05-19
chore: Get CFN resource properties from profiles.json for SAM Connectors

GavinZZ:
### Description of changes
We want to read from profiles.json and get CFN resource properties for SAM Connectors

Added new tests, all tests passed.

### Checklist

    - [ ] Using correct values
    - [ ] Using wrong values

### Examples?

### DIFF
--- a/samtranslator/model/connector/connector.py
+++ b/samtranslator/model/connector/connector.py
@@ -6,8 +6,9 @@ from typing_extensions import TypeGuard
 from samtranslator.model import ResourceResolver
 from samtranslator.model.apigateway import ApiGatewayRestApi
 from samtranslator.model.apigatewayv2 import ApiGatewayV2HttpApi
+from samtranslator.model.connector_profiles.profile import replace_cfn_resource_properties
 from samtranslator.model.dynamodb import DynamoDBTable
-from samtranslator.model.intrinsics import fnGetAtt, get_logical_id_from_intrinsic, ref
+from samtranslator.model.intrinsics import get_logical_id_from_intrinsic, ref
 from samtranslator.model.lambda_ import (
     LambdaFunction,
 )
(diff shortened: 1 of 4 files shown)


samtranslator/model/connector/connector.py
```diff
@@ -174,30 +179,44 @@ def get_resource_reference(
...
-        for target in properties.get("Targets", []):
-            target_arn = target.get("Arn")
-            target_logical_id = get_logical_id_from_intrinsic(target_arn)
-            if (target_logical_id and target_logical_id == connecting_obj_id) or (
-                connecting_obj_arn and target_arn == connecting_obj_arn
-            ):
-                return target.get("RoleArn")
+    role_property = cfn_resource_properties.get("Inputs", {}).get("Role")
+    if not role_property:
+        return None
```
ssenchenko, line 203: Seems to me like the same as `return None` in the last row

ssenchenko:
Oh I missed that it's a draft

ssenchenko (approved):

samtranslator/model/connector/connector.py
```diff
@@ -174,30 +179,43 @@ def get_resource_reference(
...
-            if (target_logical_id and target_logical_id == connecting_obj_id) or (
-                connecting_obj_arn and target_arn == connecting_obj_arn
-            ):
-                return target.get("RoleArn")
+    role_property = cfn_resource_properties.get("Inputs", {}).get("Role")
+
+    if isinstance(role_property, str):
+        return properties.get(role_property)
+
+    if isinstance(role_property, dict) and role_property.get("Function") == "GetEventsRuleRole":
```
hoffa, line 206: So this only supports `Function` on `Role`? Is there a more generic approach we can take (that doesn't make the code too complex) so we don't need to change multiple places in code later if we support more `Function`s?

GavinZZ: I don't see how other properties would every need a custom function.

hoffa: Yeah I see only input we have is for role. And in any case we can adjust the code later; any issues should be caught by transform tests when we add new connectors.

hoffa (approved):

samtranslator/model/connector/connector.py
```diff
@@ -208,39 +226,26 @@ def _get_resource_role_name(
     return ref(logical_id)
 
 
-def _get_resource_queue_url(logical_id: str, resource_type: str) -> Optional[Dict[str, Any]]:
-    if resource_type == "AWS::SQS::Queue":
-        return ref(logical_id)
-    return None
+def _get_resource_queue_url(properties: Dict[str, Any]) -> Optional[Any]:
+    return properties.get("Outputs", {}).get("Url")
```
hoffa, line 230: Since we're using a ambiguous "properties" anyway, perhaps make it `outputs`, so we don't need to keep repeating the fallback logic?

xazhao (approved):

samtranslator/model/connector/connector.py
GavinZZ: updated.

PR #3180, opened 2023-05-24
Defaults support

ssenchenko:
### Description of changes
Support `Defaults` property as per design. The difference with `Globals` is that we override lists and dicts instead of merging.

The set of properties to reference in `Defaults` will be restricted to fit most common use cases:
- `Runtime` for the common section (Resolvers and Functions) only
- `InlineCode` , `CodeUri`, and `Pipeline` for Resolvers, in case customers need to override default code we generate
- `DataSource` and `DataSourceName` for Functions

We can always add more if customers ask. For now, I'm not sure that properties like `Sync`, `MaxBatchSize`, `Caching` have a lot of use. And some properties like `Id`, `Name`, `Description` don't make any sense in `Defaults` at all.

An example with `Defaults` like this
```yaml
SamGraphQL:
    Type: "AWS::Serverless::GraphQLApi"
    Propeties:
      ...
      Defaults:
        Runtime: # applied to both Resolvers and Functions
          Name: APPSYNC_JS
          RuntimeVersion: 1.0.0
        # applied to Functions only
        Functions:
          DataSource: NotesTableDataSource
      
      Resolovers:
        Query:
          getNote:
            Pipeline:
            - getNoteFunc
        Mutation:
          deleteNote:
            Pipeline:
            - deleteNoteFunc
            - notifyFunc
          saveNote:
            Pipeline:
            - saveNoteFunc
            - notifyFunc
      
      Functions:
        getNote:
          CodeUri: "function/getNote.js"
        deleteNote:
          CodeUri: "functions/deleteNote.js"
        saveNote:
          CodeUri: "functions/saveNote.js"
        notifyFunc:
          DataSource: None #override Defaults
          CodeUri: "functions/notify.js"
```

Expands to that
```yaml
SamGraphQL:
    Type: "AWS::Serverless::GraphQLApi"
    Propeties:
      ...
      Resolovers:
        Query:
          getNote:
            Runtime:
              Name: APPSYNC_JS
              RuntimeVersion: "1.0.0"
            Pipeline:
            - getNoteFunc
        Mutation:
          deleteNote:
            Runtime:
              Name: APPSYNC_JS
              RuntimeVersion: "1.0.0"
            Pipeline:
            - deleteNoteFunc
            - notifyFunc
          saveNote:
            Runtime:
              Name: APPSYNC_JS
              RuntimeVersion: "1.0.0"
            Pipeline:
            - saveNoteFunc
            - notifyFunc
      
      Functions:
        getNoteFunc:
          Runtime:
            Name: APPSYNC_JS
            RuntimeVersion: "1.0.0"
          DataSource: NotesTableDataSource
          CodeUri: "function/getNote.js"
        deleteNoteFunc:
          Runtime:
            Name: APPSYNC_JS
            RuntimeVersion: "1.0.0"
          DataSource: NotesTableDataSource
          CodeUri: "functions/deleteNote.js"
        saveNoteFunc:
          Runtime:
            Name: APPSYNC_JS
            RuntimeVersion: "1.0.0"
          DataSource: NotesTableDataSource
          CodeUri: "functions/saveNote.js"
        notifyFunc:
          Runtime:
            Name: APPSYNC_JS
            RuntimeVersion: "1.0.0"
          DataSource: None
          CodeUri: "functions/notify.js"
```

Also I had to do a couple of bug fixes like `MaxBatchSize` property forgotten for Resolver and still dangling `DataSource` for Resolvers.

*Why a transformation test `*generate_code*` deleted?*  There is no need in it, Resolver generated code is present in every test with Resolvers

*Why no new `errors_` tests?*  No specific `Defaults` related errors. The only error possible is using a field that's not supported by `Defaults` but it's covered by schema.

### Checklist

    - [x] Using correct values
    - [ ] Using wrong values

### Examples?


samtranslator/internal/schema_source/aws_serverless_graphqlapi.py
```diff
@@ -152,7 +152,19 @@ class Cache(BaseModel):
...
 
+class Defaults(BaseModel):
+    # properties common for Resolvers and Functions
+    Runtime: Optional[Runtime]
+    MaxBatchSize: Optional[PassThroughProp]
+    Sync: Optional[Sync]
+    # only Resolver properties
+    Resolvers: Optional[Resolver]
+    # only Function properties
+    Functions: Optional[Function]
```
hoffa, line 163: What does it mean to have a default function?

ssenchenko: I means, we can reference here all Function properties

samtranslator/internal/schema_source/aws_serverless_graphqlapi.py
```diff
@@ -152,7 +152,19 @@ class Cache(BaseModel):
     TransitEncryptionEnabled: Optional[PassThroughProp]
 
 
+class Defaults(BaseModel):
+    # properties common for Resolvers and Functions
+    Runtime: Optional[Runtime]
+    MaxBatchSize: Optional[PassThroughProp]
+    Sync: Optional[Sync]
```
aaythapa, line 159: Could there be a case where customers would want the defaults to only apply to Resolvers or only to Functions? What would they do in that case?

samtranslator/model/graphql/graphql.py
```diff
@@ -0,0 +1,60 @@
...
+    ):
+        self.resolver = resolver
+        self.function = function
+
+    @classmethod
+    def from_sam_graphql(cls, model: Properties) -> "Defaults":
+        """Create Defaults from SAM GraphQLApi model.
+
+        SAM GraphQLApi object has optional Defaults property which allows to set defaults for all
+        Functions, Resolvers, or some properties like Runtime for both Functioins and Resolvers.
```
aaythapa, line 29: nit: `Functioins` typo in comments

samtranslator/model/graphql/graphql.py
aaythapa, line 1: Can the `Defaults` be handled in the `sam_resources.py` file or is there a good reason it needs its own file?

samtranslator/internal/schema_source/aws_serverless_graphqlapi.py
ssenchenko: Why not? Even properties like `CodeUri` or `InlineCode` have a use case to be declared in `Defaults`. If a customer uses Lambda Resolvers mostly, they might want to override the code we provide by default, but they still might want to have the same code for all Resolvers and Functions.

samtranslator/internal/schema_source/aws_serverless_graphqlapi.py
ssenchenko: Yes, they will put those properties under `Resolvers` or `Functions` property:
```yaml
Defaults:
  MaxBatchSize: 10 # that's for both Resolvers and Functions
  Resolvers:
    MaxBatchSize: 20 # that's Resolvers only, it overrides '10' from above
  Functions:
    Sync:
      # this Sync setting is only for functions
```

samtranslator/model/graphql/graphql.py
ssenchenko: IMO a couple of reasons:
1. `sam_resources.py` is already pretty long and inconvenient because of its size
2. Having `Defaults` as class seems cleaner than function + NamedTuple (or simple Tuple or other data structure)
3. Having a class which is not a SAM resource in `sam_resources.py` doesn't sound right, even setting file size aside
4. Having `Defaults` as a nested class in `sam_resources.py` leads to extra verbosity and extra ugliness partially because we still support Python 3.7
5. Having a separate package with helper functions and data structures is a practice for other resources
6. I'd move more stuff from GraphQLApi resource to that separate graphql package, and hopefully, one day, I'll get it done :)

tests/translator/input/graphqlapi_resolver_ref_function_by_logical_and_function_id.yaml
```diff
@@ -28,7 +28,6 @@ Resources:
           saveNote:
             InlineCode: blahblahblah
             FieldName: my_field
-            DataSource: NONE
```
hoffa, line 31: Why was this removed?

samtranslator/internal/schema_source/aws_serverless_graphqlapi.py
```diff
@@ -152,7 +152,19 @@ class Cache(BaseModel):
...
+    # only Resolver properties
+    Resolvers: Optional[Resolver]
```
hoffa, line 161: nit: technically these aren't correct, as this means if `Resolvers` is provided every property required by `Resolver` must be specified. It's more of a "`Resolver` but all its properties are optional", similar to TypeScript's [`Partial`](https://www.typescriptlang.org/docs/handbook/utility-types.html#partialtype), but not sure if there's anything similar

samtranslator/internal/schema_source/aws_serverless_graphqlapi.py
hoffa, line 159: Is it guaranteed `Sync` will always be exactly the same for resolvers and functions?

aaythapa: Ah that makes sense, so whatever they define in the `Resolvers` will take precedence

samtranslator/model/graphql/graphql.py
aaythapa: Thanks for the clarification it makes sense to me. I'd say yea to moving more of the GraphQLApi resource to the graphql package because my only issue is that `Default` is the only thing thats separated from `sam_resources.py`

aaythapa (approved):

samtranslator/internal/schema_source/aws_serverless_graphqlapi.py
ssenchenko: There will be no Sync in new version

samtranslator/internal/schema_source/aws_serverless_graphqlapi.py
ssenchenko: That's a good catch. Though all props are optional now in Resolver, I thought at some point to have a separate type just in case. Well, it's a separate type in a new version anyway

tests/translator/input/graphqlapi_resolver_ref_function_by_logical_and_function_id.yaml
ssenchenko: ah, no DataSources in Resolver. They where there to be applied to all Functions in the resolver, but now we have `Defaults`

samtranslator/internal/schema_source/aws_serverless_graphqlapi.py
ssenchenko: Restricting properties as discussed

hoffa (approved):

PR #3191, opened 2023-05-30
chore: make custom domains api test nonblocking

aaythapa:
### Description of changes

Encountering certificate issues with these tests; added equivalent transform tests and marking as non-blocking since we've already tested it.

The integration tests mostly check that the right resources are made and that some of the properties are transformed as expected. This can be verified through the transform tests. 

https://github.com/aws/serverless-application-model/blob/2c06b57402ab6000a3f0e3c2085babac54aa2648/integration/combination/test_custom_http_api_domains_test.py#L32-L37

https://github.com/aws/serverless-application-model/blob/2c06b57402ab6000a3f0e3c2085babac54aa2648/integration/combination/test_custom_rest_api_domains.py#L28-L31

There are cases where the integration tests check inside files of the deployed s3 bucket
https://github.com/aws/serverless-application-model/blob/2c06b57402ab6000a3f0e3c2085babac54aa2648/integration/combination/test_custom_rest_api_domains.py#L56-L57
But i think we can ignore this case as the property (`TruststoreUri`) is a pass through and shouldn't be affected by SAM. 

Ensured new transform tests are same as the integration ones and that the output is expected. 

### Checklist

    - [ ] Using correct values
    - [ ] Using wrong values

### Examples?

### DIFF
--- a/integration/combination/test_custom_http_api_domains_test.py
+++ b/integration/combination/test_custom_http_api_domains_test.py
@@ -3,12 +3,14 @@ from unittest.case import skipIf
 from integration.config.service_names import CUSTOM_DOMAIN
 from integration.helpers.base_internal_test import BaseInternalTest
 from integration.helpers.resource import current_region_not_included
+from integration.helpers.base_test import nonblocking
 
 
 @skipIf(
     current_region_not_included([CUSTOM_DOMAIN]),
     "CustomDomain is not supported in this testing region",
 )
+@nonblocking
 class TestCustomHttpApiDomains(BaseInternalTest):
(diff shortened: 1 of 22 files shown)


hoffa (approved):

GavinZZ (commented):
Changes look good. Just wondering why are there 5 transform tests added when only 2 integ tests removed?

GavinZZ (approved):

aaythapa:
> Why are there 5 transform tests added when only 2 integ tests removed?

The integ tests test on multiple files so I added all them all as transform tests

PR #3193, opened 2023-05-31
Allow SAM Api and HttpApi to Propagate Tags to Generated Resources

GavinZZ:
### Description of changes
There is a request to allow tags to propagate to sub-resources. Support it by introducing `PropagateTags` property.

Added multiple transform tests and all passed.
Added integ tests to verify tags exist.

### Checklist

    - [ ] Using correct values
    - [ ] Using wrong values

### Examples?


samtranslator/model/__init__.py
```diff
@@ -536,6 +536,25 @@ def _construct_tag_list(
...
+        :param resources: List of resources
+        :param tags: dictionary of tags to propagate to the resources.
+
+        :return: None
+        """
+        if not propagate_tags:
+            return
+
+        tags_list = get_tag_list(tags)
+        for resource in resources:
```
ssenchenko, line 554: What if some of them cannot have Tags?

samtranslator/model/__init__.py
```diff
@@ -536,6 +536,25 @@ def _construct_tag_list(
         # customer's knowledge.
         return get_tag_list(sam_tag) + get_tag_list(additional_tags) + get_tag_list(tags)
 
+    def propagate_tags(
+        self, propagate_tags: Optional[bool], resources: List[Resource], tags: Optional[Dict[str, Any]]
```
hoffa, line 540: Can be static or a separate resource.

samtranslator/model/apigateway.py
ssenchenko, line 1: What about SAM Function, it doesn't generate any resources?

samtranslator/plugins/globals/globals.py
aaythapa, line 1: I thought we were going to add a Globals section to enable tag propagation for all resources? I only see globals for httpapi and restapi specifically.

samtranslator/model/__init__.py
```diff
@@ -536,6 +536,25 @@ def _construct_tag_list(
...
+
+        :return: None
+        """
+        if not propagate_tags:
+            return
+
+        tags_list = get_tag_list(tags)
+        for resource in resources:
+            if "Tags" in resource.property_types:
+                resource.Tags = tags_list
```
hoffa, line 556: This assumes (1) the property is always called `Tags`, and (2) the property format is always the same.

Neither are true, for example:

1. In `AWS::Backup::Framework` it's called `FrameworkTags`: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-backup-framework.html#cfn-backup-framework-frameworktags
2. `AWS::AmplifyUIBuilder::Component` uses map for `Tags`: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-amplifyuibuilder-component.html#cfn-amplifyuibuilder-component-tags

For this reason I'd rather at the very least be abundantly clear on the assumptions, or somehow make it explicit.

samtranslator/plugins/globals/globals.py
GavinZZ: `Globals` only support resource type level property.

samtranslator/model/__init__.py
GavinZZ: That is checked on line 555. It checks if resource has `Tags` property.

ssenchenko: Actually, withdrawing this comment, Chris has much more specific feedback

samtranslator/model/__init__.py
GavinZZ: Will add comments to clearly state the assumptions I made here.

samtranslator/model/__init__.py
ssenchenko: I think `propogate_tags` instead of a parameter can be `self.PropogateTags` because a function with an `Optional[bool]` without default value looks strange

samtranslator/plugins/globals/globals.py
aaythapa: yea we can probably add it later too if there's demand for it.

samtranslator/model/__init__.py
hoffa: Are you sure comment is enough?

I think the current PR will fail deployment; `AWS::ApiGatewayV2::Api` takes a map for `Tags`: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-apigatewayv2-api.html#cfn-apigatewayv2-api-tags

ssenchenko: I suggest we add a separate method for that assignment like
```python
@staticmethod
def assign_tags(reciever_resource, tags_list):
    if "Tags" in resource.property_types:
        resource.Tags = tags_list
```
So if we need to overload it to handle specific format for some sub-resources, we overload only it and don't copy-paste other code which is general

tests/translator/output/httpapi_with_propagate_tags.json
```diff
@@ -0,0 +1,157 @@
...
+              "*"
+            ]
+          }
+        },
+        "FailOnWarnings": true,
+        "Tags": [
+          {
+            "Key": "TagKey1",
+            "Value": "Value1"
+          },
```
hoffa, line 65: Surprised `cfn-lint` didn't catch this since this shouldn't even deploy. 🤔

ssenchenko: ha-ha `cfn-lint` probably doesn't check because it's unstandardized mess

ssenchenko: Seems like we need non-blocking integration tests for each SAM resource with Tags otherwise, never can be sure

hoffa: But it should be correctly in the schema already that `cfn-lint` uses to check format. 😢

samtranslator/model/__init__.py
GavinZZ: I like it

samtranslator/model/__init__.py
```diff
@@ -433,6 +434,10 @@ def get_passthrough_resource_attributes(self) -> Dict[str, Any]:
                 attributes[resource_attribute] = self.resource_attributes.get(resource_attribute)
         return attributes
 
+    def assign_tags(self, tags: Dict[str, Any]) -> None:
```
hoffa, line 437: We should state assumptions and why this exists.

samtranslator/model/__init__.py
```diff
@@ -536,6 +541,30 @@ def _construct_tag_list(
...
+    @staticmethod
+    def propagate_tags(
+        resources: List[Resource], tags: Optional[Dict[str, Any]], propagate_tags: Optional[bool] = False
+    ) -> None:
+        """
+        Propagates tags to the resources. This function assumes that generated resources always have
+        the tags property called `Tags`
+
+        TODO: There are CFN resources that do not follow the assumption such as 'AWS::Backup::Framework'
+        We may need to modify the code to account for thia scenario in the future.
```
hoffa, line 553: What is required to make it robust? Doesn't the current approach of overloading work with any situation?

GavinZZ: Yes, the comment is outdated. Will update it.

hoffa (approved):

samtranslator/model/__init__.py
```diff
@@ -433,6 +434,21 @@ def get_passthrough_resource_attributes(self) -> Dict[str, Any]:
                 attributes[resource_attribute] = self.resource_attributes.get(resource_attribute)
         return attributes
 
+    def assign_tags(self, tags: Dict[str, Any]) -> None:
+        """
+        Assigns tags to the resource. This function assumes that generated resources always have
+        the tags property called `Tags`
+
+        There are CFN resources that do not follow the assumption such as 'AWS::Backup::Framework'
+        We may need to modify the code to account for thia scenario in the future. We could do this
```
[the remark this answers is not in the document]
GavinZZ: If the tag property is called something else.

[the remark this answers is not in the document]
GavinZZ: but the function overrides will still work, just need to use the exact tag name.

samtranslator/model/__init__.py
```diff
@@ -433,6 +434,21 @@ def get_passthrough_resource_attributes(self) -> Dict[str, Any]:
...
+        by function overriding this function in the class definition of that resource. An example
+        is provided in 'AWS::ApiGatewayV2::Api'
+
+        :param tags: Dictionary of tags to be assigned to the resource
```
hoffa, line 447: Maybe so it's a little clearer it's by design and not a limitation that needs to be fixed:

```suggestion
        Assigns tags to the resource. This function assumes that generated resources always have
        the tags property called `Tags` that takes a list of key-value objects.

        Override this function if the above assumptions do not apply to the resource (e.g. different
        property name or type (see e.g. 'AWS::ApiGatewayV2::Api').

        :param tags: Dictionary of tags to be assigned to the resource
```

tests/translator/input/api_with_propagate_tags.yaml
```diff
@@ -0,0 +1,17 @@
+Parameters:
+  TagValueParam:
+    Type: String
+    Default: value
+
+Resources:
+  MyApiWithPropagateTags:
```
aaythapa, line 7: I don't think this template results in all the generated resources from [here](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-specification-generated-resources-api.html) being created right? We should have a test that has all generated resources and make sure the tags propagate correctly.

GavinZZ: I have a complete set of tests in the upcoming PR which includes function/sfn. Will ignore this comment for now.

aaythapa (approved):

samtranslator/model/__init__.py
```diff
@@ -536,6 +552,25 @@ def _construct_tag_list(
         # customer's knowledge.
         return get_tag_list(sam_tag) + get_tag_list(additional_tags) + get_tag_list(tags)
 
+    @staticmethod
+    def propagate_tags(
+        resources: List[Resource], tags: Optional[Dict[str, Any]], propagate_tags: Optional[bool] = False
```
ssenchenko, line 555: We have `tags` in the Resource class, so there is no need to pass them as a parameter and keep this method static.

GavinZZ: Good point. Going to change this in next PR that includes function/sfn changes.

GavinZZ: Actually, this method is already a static method, it cannot access any class attributes.

Issue #1541, opened 2020-03-28
API Usage Plan Generates (Unnecessary?) 'DependsOn'


thbishop-intuit:
**Description:**
Specifying an API usage plan will populate 'DependsOn' with resources that are already included as a `Ref` in `Properties`. There are a couple of places where this currently occurs. To be clear, this doesn't break functionality but it does report as a warning in [cfn-lint](aws-cloudformation/cfn-python-lint) (which is how I came across this).

Is this extra `DependsOn` perhaps needed to handle a very specific use case? I'm more than happy to submit a fix if these extra declarations are not needed.

**Steps to reproduce the issue:**
1. This is demonstrated within the translator tests. For example, this [input](https://github.com/awslabs/serverless-application-model/blob/1b2d6f25514780d918b1bebf262ea05b89c5e646/tests/translator/input/api_with_usageplans.yaml) (which is head of `develop` at the time of this bug).

**Observed result:**
These are from the translator test outputs:

A couple of examples for usage plan resource:
https://github.com/awslabs/serverless-application-model/blob/1b2d6f25514780d918b1bebf262ea05b89c5e646/tests/translator/output/api_with_usageplans.json#L392-L430

https://github.com/awslabs/serverless-application-model/blob/1b2d6f25514780d918b1bebf262ea05b89c5e646/tests/translator/output/api_with_usageplans.json#L365-L391

And a couple of examples for usage plan key resource:
https://github.com/awslabs/serverless-application-model/blob/1b2d6f25514780d918b1bebf262ea05b89c5e646/tests/translator/output/api_with_usageplans.json#L83-L97

https://github.com/awslabs/serverless-application-model/blob/1b2d6f25514780d918b1bebf262ea05b89c5e646/tests/translator/output/api_with_usageplans.json#L530-L544

**Expected result:**
I would expect the usage plan resource to look like:

```
    "MyApiTwoUsagePlan": {
      "Type": "AWS::ApiGateway::UsagePlan",
      "Properties": {
        "ApiStages": [
          {
            "ApiId": {
              "Ref": "MyApiTwo"
            },
            "Stage": {
              "Ref": "MyApiTwoProdStage"
            }
          }
        ],
        "Description": "Description for usage plan",
        "Tags": [
          {
            "Value": "value1",
            "Key": "key1"
          },
          {
            "Value": "value2",
            "Key": "key2"
          }
        ],
        "Quota": {
          "Limit": 10,
          "Period": "MONTH",
[...]

PR #3200, opened 2023-06-02
Allow SAM Function and SAM StateMachines to Propagate Tags to Generated Resources

GavinZZ:
### Description of changes
In the previous [PR](https://github.com/aws/serverless-application-model/pull/3193), we have supported `PropagateTags` in SAM API and SAM HttpApi. This PR targets to support SAM Function and SAM StateMachine.

Note that SAM Function and SAM StateMachine support API as an event source. 
* When ApiId is provided in the API event source, customers have to manually define a API resource. We would not consider this API resource as a generated resource.
* When ApiId is not provided, then SAM-T generates an implicit API for customers and this implicit API should be considered as a generated resource and hence propagate tags.
(I have covered this edge case in the state_machine_with_api_and_propagate_tags test where we have two state machine with api event source. One defining `ApiId` and a custom API resource; the other don't. The expected result is tags should be propagate to the implicit API `ServerlessRestApi` while the custom defined one doesn't have tags)

We have complete set of unit tests covering changes. There is transform tests covering each `PropagateTags` supported SAM resource and make sure the generated resource's tags are expected. The generated resources matches our [documentation](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-specification-generated-resources.html)
1. `api_with_propagate_tags` covers all generated api resources, including `AWS::ApiGateway::RestApi`, `AWS::ApiGateway::Stage`, `AWS::ApiGateway::Deployment`, `AWS::ApiGateway::DomainName`, `AWS::ApiGateway::UsagePlan`, `AWS::ApiGateway::UsagePlanKey`, `AWS::ApiGateway::ApiKey`. 
2. `httpapi_with_propagate_tags` (in previous [PR](https://github.com/aws/serverless-application-model/pull/3193)) covers all generated http api resources, including `AWS::ApiGatewayV2::Api`, `AWS::ApiGatewayV2::Stage`, `AWS::ApiGatewayV2::DomainName`.
3. `function_with_events_and_propagate_tags` covers all generated resources that can have tags, including `AWS::Lambda::Function`, `AWS::Lambda::Alias`, `AWS::Lambda::Version`, `AWS::IAM::Role`, `AWS::CodeDeploy::Application`, `AWS::CodeDeploy::DeploymentGroup`, `AWS::ApiGateway::RestApi`, `AWS::ApiGatewayV2::Api`, `AWS::Lambda::EventSourceMapping`, `AWS::Events::Rule`, `AWS::IoT::TopicRule`, `AWS::Lambda::EventInvokeConfig`, `AWS::SNS::Topic`, `AWS::SQS::Queue`.
4. `state_machine_with_api_and_propagate_tags` covers all resources that can have tags `AWS::StepFunctions::StateMachine`, `AWS::IAM::Role`, `AWS::ApiGateway::RestApi`.

Verified that SAM will propagate the tags defined to the ones with `Tags` property.

### Checklist

    - [ ] Using correct values
    - [ ] Using wrong values

### Examples?


samtranslator/plugins/api/implicit_api_plugin.py
```diff
@@ -93,6 +93,19 @@ def _generate_implicit_api_resource(self) -> Dict[str, Any]:
         Helper function implemented by child classes that create a new implicit API resource
         """
 
+    def _add_tags_to_implicit_api_if_necessary(self, resource: SamResource, template: SamTemplate) -> None:
```
hoffa, line 96: Is there no cleaner way to do it? This feels very duct tape-y.

.cfnlintrc.yaml
```diff
@@ -123,6 +123,7 @@ ignore_templates:
   - tests/translator/output/**/managed_policies_minimal.json  # Intentionally has non-existent managed policy name
   - tests/translator/output/**/function_with_mq.json  # Property "EventSourceArn" can Fn::GetAtt to a resource of types [AWS::DynamoDB::GlobalTable, AWS::DynamoDB::Table, AWS::Kinesis::Stream, AWS::Kinesis::StreamConsumer, AWS::SQS::Queue]
   - tests/translator/output/**/function_with_mq_using_autogen_role.json  # Property "EventSourceArn" can Fn::GetAtt to a resource of types [AWS::DynamoDB::GlobalTable, AWS::DynamoDB::Table, AWS::Kinesis::Stream, AWS::Kinesis::StreamConsumer, AWS::SQS::Queue]
+  - tests/translator/output/**/api_with_propagate_tags.json # Obsolete DependsOn on resource
```
hoffa, line 126: Why? Or is it https://github.com/aws/serverless-application-model/issues/1541?

samtranslator/plugins/api/implicit_api_plugin.py
GavinZZ: It would be much cleaner if we can move `GlobalsPlugin` to be in front of the implicit api plugin. I tried it, but it caused backward incompatible issue.

.cfnlintrc.yaml
GavinZZ: Yes, it's this issue. CFN lint gives warning on unnecessary depends on.

GavinZZ: I can look into this issue as a separate PR. Seems pretty easy to fix.

samtranslator/internal/schema_source/aws_serverless_function.py
```diff
@@ -618,6 +618,7 @@ class Properties(BaseModel):
     SnapStart: Optional[SnapStart] = prop("SnapStart")
     RuntimeManagementConfig: Optional[RuntimeManagementConfig] = prop("RuntimeManagementConfig")
     Tags: Optional[Tags] = prop("Tags")
+    PropagateTags: Optional[bool]  # TODO: add docs
```
xazhao, line 621: When `PropagateTags` is set without setting `Tags`, do we fail the transform or just ignore the property?

GavinZZ: I believe the behaviour should be ignoring the property. I will add a test covering this scenario.

xazhao (approved):

samtranslator/plugins/api/implicit_api_plugin.py
```diff
@@ -93,6 +93,19 @@ def _generate_implicit_api_resource(self) -> Dict[str, Any]:
...
+        """
+        Decides whether to add tags to the implicit api resource.
+        :param dict template_dict: SAM template dictionary
+        """
+        implicit_api_resource = template.get(self.IMPLICIT_API_LOGICAL_ID)
+        globals_var = template.get_globals().get(SamResourceType(resource.type).name, {})
+        should_propagate_tags = resource.properties.get("PropagateTags") or globals_var.get("PropagateTags")
+
+        if implicit_api_resource and resource.properties.get("Tags") and should_propagate_tags:
+            implicit_api_resource.properties.setdefault("Tags", {}).update(resource.properties["Tags"])
```
hoffa, line 106: I don't really like how there's so many assumptions and mixing of concerns. It takes a generic resource as input, yet assumes the tags is `Tags`, the propagate tags is `PropagateTags`, but there's nothing guaranteeing it's correct from this function's perspective.

samtranslator/plugins/api/implicit_rest_api_plugin.py
```diff
@@ -61,6 +61,12 @@ def _process_api_events(  # noqa: too-many-arguments
 
             sam_expect(event_properties, event_id, "", is_sam_event=True).to_be_a_map("Properties should be a map.")
 
+            # if the API ID is not provided in the event source properties, this implies that SAM-T will
+            # construct an implicit API resource for this API event source, and we need to add tags to the
+            # implicit API resource if customers specify `PropagateTags` property
+            if self.API_ID_EVENT_PROPERTY not in event_properties:
+                self._add_tags_to_implicit_api_if_necessary(function, template)
+
             self._add_implicit_api_id_if_necessary(event_properties)  # type: ignore[no-untyped-call]
```
hoffa, line 70: Shouldn't these be together? i.e. check once and do all implicit API related stuff once.

samtranslator/plugins/api/implicit_api_plugin.py
GavinZZ: This `resource` is a SAM resource not a generic CFN resource. We can safely assume that all SAM resources have use `Tags` property isntead of an arbitrary name.

hoffa (approved):

PR #3220, opened 2023-06-20
fix: Support resolve simple ref intrinsics for Tracing Properties

GavinZZ:
### Description of changes
Allow resolving simple refs intrinsics in `Tracing` properties. Also fix a bug where we originally enabled tracing as long as tracing is not empty. However, according to documentation, this should be enabled if the value is set to `Active`. 

In `gloabls_for_function` transform test, the property `Tracing` is set to `PassThrough`, but in the output json, it still contains xray policy. 

All tests passed

### Checklist

    - [ ] Using correct values
    - [ ] Using wrong values

### Examples?


samtranslator/model/sam_resources.py
```diff
@@ -624,7 +627,9 @@ def _construct_role(
         )
 
         managed_policy_arns = [ArnGenerator.generate_aws_managed_policy_arn("service-role/AWSLambdaBasicExecutionRole")]
-        if self.Tracing:
+
+        tracing = intrinsics_resolver.resolve_parameter_refs(self.Tracing)
+        if tracing == TRACING_CONFIG_ACTIVE:
```
hoffa, line 632: This will likely cause redeployment right? Since anyone so far that has used `Tracing` to something other than `Active` will now have the policy removed.

GavinZZ: hmmm actually good point. It would.

PR #3224, opened 2023-06-21
fix: read from default boto3 session if present.

sriram-mv:
### Issue #, if available

https://github.com/aws/aws-sam-cli/issues/5390

### Checklist

    - [ ] Using correct values
    - [ ] Using wrong values

### Examples?


xazhao (approved):

tests/unit/test_region_configuration.py
```diff
@@ -51,6 +52,16 @@ def test_when_apigw_edge_configuration_is_not_supported(self, partition):
     def test_is_service_supported_positive(self, service, region):
         self.assertTrue(RegionConfiguration.is_service_supported(service, region))
 
+    @parameterized.expand(
+        [
+            ["ec2", None],
+        ]
+    )
+    def test_is_service_supported_positive_boto3_default_session(self, service, region):
```
GavinZZ, line 60: It seems that `region`  isn't used at all.

GavinZZ (approved):
LGTM except for a quick question.

tests/unit/test_region_configuration.py
sriram-mv: Good catch! let me remove that.

hoffa (approved):

PR #3240, opened 2023-07-05
Set runtime for resolvers which use generated code

ssenchenko:
### Description of changes

Resolvers provide by default generated code to pass request to functions and receive response. There is no need to ask a customer for Runtime when default code is used because we know ourselves which Runtime should be used for the generated code.

### Checklist

    - [x] Using correct values
    - [x] Using wrong values

### Examples?

### DIFF
--- a/samtranslator/model/sam_resources.py
+++ b/samtranslator/model/sam_resources.py
@@ -14,6 +14,7 @@ from samtranslator.feature_toggle.feature_toggle import FeatureToggle
 from samtranslator.internal.intrinsics import resolve_string_parameter_in_resource
 from samtranslator.internal.model.appsync import (
     APPSYNC_PIPELINE_RESOLVER_JS_CODE,
+    APPSYNC_PIPELINE_RESOLVER_JS_RUNTIME,
     AdditionalAuthenticationProviderType,
     ApiCache,
     ApiKey,
@@ -2187,7 +2188,7 @@ class SamConnector(SamResourceMacro):
 
 
 class SamGraphQLApi(SamResourceMacro):
-    """SAM GraphQL API Macro (WIP)."""
+    """SAM GraphQL API Macro."""
 
     resource_type = "AWS::Serverless::GraphQLApi"
     property_types = {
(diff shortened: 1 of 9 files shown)


tests/translator/input/error_graphqlapi.yaml
aaythapa, line 1: the output of this file doesn't need to change?

samtranslator/model/sam_resources.py
```diff
@@ -2973,11 +2974,13 @@ def _construct_appsync_resolver_resources(
                 # to a default snippet which has basic definition of request/response functions.
                 if not cfn_resolver.Code and not cfn_resolver.CodeS3Location:
                     cfn_resolver.Code = APPSYNC_PIPELINE_RESOLVER_JS_CODE
+                    cfn_resolver.Runtime = APPSYNC_PIPELINE_RESOLVER_JS_RUNTIME
+                else:
+                    cfn_resolver.Runtime = self._parse_runtime(resolver, relative_id)
```
aaythapa, line 2979: Im guessing if code or s3location is set then runtime needs to be set? Do we have an error thats thrown if runtime isn't set?

tests/translator/input/graphqlapi_resolver_function_with_lambda_datasource.yaml
```diff
@@ -26,9 +26,20 @@ Resources:
...
+              - key1
+              - key2
+            Pipeline:
+            - MyFunction
+          myResolverWithCustomCode:
+            FieldName: my_field
+            InlineCode: custom code
             Runtime:
-              Name: some-runtime
+              Name: APPSYNC_JS
```
hoffa, line 40: nit: I'd prefer adding entirely new transform tests instead of modifying old ones, otherwise it's harder to verify at a glance that it contains no backward incompatible changes

hoffa (approved):

tests/translator/input/error_graphqlapi.yaml
ssenchenko: Actually not, because I modified the existing use case and still raises and error when there is a custom Resolver code

samtranslator/model/sam_resources.py
ssenchenko: Yes in `_parse_runtime()`

tests/translator/input/graphqlapi_resolver_function_with_lambda_datasource.yaml
ssenchenko: Thanks. I agree. I'll keep it in mind for future changes. As for this change, there is no incompatible changes. The change above is because using "some-runtime" is actually incorrect value, a random text

PR #3249, opened 2023-07-09
Requirements: Bump Pydantic from 1.8 to 2.0.2.

ohadbenita:
### Checklist

- [ ] Adheres to the [development guidelines](https://github.com/aws/serverless-application-model/blob/develop/DEVELOPMENT_GUIDE.md#development-guidelines)
- [ ] Add/update [transform tests](https://github.com/aws/serverless-application-model/blob/develop/DEVELOPMENT_GUIDE.md#unit-testing-with-multiple-python-versions)
    - [ ] Using correct values
    - [ ] Using wrong values
- [ ] Add/update [integration tests](https://github.com/aws/serverless-application-model/blob/develop/INTEGRATION_TESTS.md)

### Examples?

Please reach out in the comments if you want to add an example. Examples will be 
added to `sam init` through [aws/aws-sam-cli-app-templates](https://github.com/aws/aws-sam-cli-app-templates).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #3282, opened 2023-08-01
chore: add support for pydantic 2.x

yan12125:
### Issue #, if available

This PR supersedes https://github.com/aws/serverless-application-model/pull/3249

### Description of changes

This change brings pydantic 2.x support without breaking 1.x support by utilizing the compatibility package pydantic.v1

Run the test suite with both pydantic 1.10.9 and 2.0.2, and both are green

### Checklist

- [ ] Add/update [transform tests](https://github.com/aws/serverless-application-model/blob/develop/DEVELOPMENT_GUIDE.md#unit-testing-with-multiple-python-versions) - irrelevant
    - [ ] Using correct values
    - [ ] Using wrong values
- [ ] Add/update [integration tests](https://github.com/aws/serverless-application-model/blob/develop/INTEGRATION_TESTS.md) - irrelevant

### Examples?


yan12125:
The second commit introduces a `# type: ignore` commit to work-around [a mypy issue](https://github.com/python/mypy/issues/1153). It violates [development guidelines](https://github.com/aws/serverless-application-model/blob/develop/DEVELOPMENT_GUIDE.md#development-guidelines) but cannot figure out a better way.

requirements/base.txt
```diff
@@ -3,4 +3,4 @@ jsonschema<5,>=3.2  # TODO: evaluate risk of removing jsonschema 3.x support
 typing_extensions>=4.4,<5 # 3.7 doesn't have Literal
 
 # resource validation & schema generation
-pydantic~=1.8
+pydantic>=1.8
```
ssenchenko, line 6: Is there any specific reason to use `>=1.8` instead of `2.1`?

samtranslator/compat.py
```diff
@@ -0,0 +1,8 @@
+try:
+    from pydantic import v1 as pydantic
```
ssenchenko, line 2: If there is no need in keeping v1.8 as a constraint, then there is no need in this `try... except` and `compat.py`

requirements/base.txt
yan12125: There is a call for keeping pydantic 1.x support: https://github.com/aws/serverless-application-model/pull/3249#issuecomment-1634028897

hoffa, line 6: nit: better convey intent:

```suggestion
pydantic>=1.8,<3
```

ssenchenko: I'm still not sure we should keep supporting versions <2.0

yan12125: Thank you, I applied this first so that it will not be forgotten when a consensus about keeping pydantic 1.x is reached. If there is a different consensus later, I will change it to `pydantic>=2,<3`.

ConnorRobertson (approved):

ssenchenko (approved):

yan12125:
Thank you for reviewing. I noticed this commit is not included in 1.73.0. Is it forgotten?

aaythapa:
Hi @yan12125 this PR was merged yesterday, it needs to go through our release process which can take some time. Should be released within the coming week or 2.

yan12125:
Got it, thanks!

mahyess:
Not released in v1.74.0 either. Anyone knows why?

aaythapa:
this was part of [v1.75.0](https://github.com/aws/serverless-application-model/releases/tag/v1.75.0), sorry about the delay!

PR #3300, opened 2023-08-11
chore: Upgrade ruff/black/mypy and fix violations

aahung:
### Description of changes

- `noqa` only supports rule code, not name, so previously `noqa` suppresses every rule if used 🤕 
- Upgrade ruff/black/mypy to latest version 
- Fix violations if possible
- Disable a rule with too many violations

### Checklist

    - [ ] Using correct values
    - [ ] Using wrong values

### Examples?


integration/combination/test_state_machine_with_cwe_dlq_generated.py
```diff
@@ -110,6 +110,6 @@ def test_state_machine_with_cwe(self):
 
 
 def get_first_key_value_pair_in_dict(dictionary):
-    key = list(dictionary.keys())[0]
+    key = next(iter(dictionary.keys()))
```
aahung, line 113: One risk of this is we have special handling of `IndexError` and it will break if it becomes `StopIteration` error. I searched the codebase and didn't find any, so it should be safe

integration/helpers/deployer/deployer.py
```diff
@@ -388,7 +388,7 @@ def _display_stack_outputs(self, stack_outputs, **kwargs):
                 ("Value", output.get("OutputValue")),
             ]:
                 pprint_columns(
-                    columns=["{k:<{0}}{v:<{0}}".format(MIN_OFFSET, k=k, v=v)],
+                    columns=[f"{k:<{MIN_OFFSET}}{v:<{MIN_OFFSET}}"],
```
aahung, line 391:
```
>>> k = 'K'
>>> v = 'V'
>>> MIN_OFFSET=5
>>> f"{k:<{MIN_OFFSET}}{v:<{MIN_OFFSET}}"
'K    V    '
>>> "{k:<{0}}{v:<{0}}".format(MIN_OFFSET, k=k, v=v)
'K    V    '
```

samtranslator/intrinsics/resolver.py
```diff
@@ -71,7 +71,10 @@ def resolve_sam_resource_refs(
             references supported in this SAM template, along with the value they should resolve to.
         :return list errors: List of dictionary containing information about invalid reference. Empty list otherwise
         """
-        return self._traverse(_input, supported_resource_refs, self._try_resolve_sam_resource_refs)
+        # The _traverse() return type is the same as the input. Here the input is Dict[str, Any]
+        return cast(
+            Dict[str, Any], self._traverse(_input, supported_resource_refs, self._try_resolve_sam_resource_refs)
```
ssenchenko, line 76: `_traverse` can work on `str` and `list` as well but it's always the same type as the type of `input_value` in `_traverse` call.
Since it's already huge PR with different improvements, I suggest you make `_traverse` a template generic function with a generic type for `input_value` then we don't need this ugly `cast`. 
Otherwise, please leave `TODO` here and in `_traverse` to remove cast and to use generic type

ssenchenko (commented):
Though I'm aware of performance tuning for `%s` in logger, I believe that f-strings are more readable. I'm not sure that in our case we should prioritize performance over readability. I'd keep f-strings and disable the rule that says otherwise. I wonder what others think

aahung:
> Thank you for the update! It's great to see your PR!
> 
> Though I'm aware of performance tuning for `%s` in logger, I believe that f-strings are more readable. I'm not sure that in our case we should prioritize performance over readability. I'd keep f-strings and disable the rule that says otherwise. I wonder what others think

No strong opinion on that, reverted those

samtranslator/intrinsics/resolver.py
aahung: Did a quick experiment, no easy way to make it generic without introducing more casting. Added a TODO.

aahung:
Thanks for the quick review man @ssenchenko 😸

aaythapa (approved):

PR #3310, opened 2023-08-17
feat: Add TimeoutInMillis Property to API Event Source

gracelu0:
### Issue #, if available

Feature request discussion: [https://github.com/aws/serverless-application-model/discussions/3287](https://github.com/aws/serverless-application-model/discussions/3287)
- Both Global level and Function level support are requested, but only Function level support is implemented in this pull request.

### Description of changes

Add TimeoutInMillis Property to API Event Source (currently only supported for HTTP API Event Source). 

Example:
```
MyFunction:
    Properties:
      Handler: src/handlers/code.handler
      Events:
        MyApiEvent:
          Type: Api  # API Event Source
          Properties:
            Path: /hello
            Method: GET
            TimeoutInMillis: 5000 # Now supported

```

Validated with transform tests and integration tests. Transformed template with added TimeoutInMillis property, then deployed it successfully. Added integration test also verifies that the TimeoutInMillis property is set to the expected value from the template.

### Checklist

    - [x] Using correct values
    - [ ] Using wrong values

### Examples?

### DIFF
--- a/integration/combination/test_function_with_implicit_api_with_timeout.py
+++ b/integration/combination/test_function_with_implicit_api_with_timeout.py
@@ -9,3 +9,13 @@ from integration.helpers.resource import current_region_does_not_support
 class TestFunctionWithImplicitApiWithTimeout(BaseTest):
     def test_function_with_implicit_api_with_timeout(self):
         self.create_and_verify_stack("combination/function_with_implicit_api_with_timeout")
+
+        # verify that TimeoutInMillis is set to expected value in the integration
+        expected_timeout = 5000
+        apigw_client = self.client_provider.api_client
+        rest_api_id = self.get_physical_id_by_type("AWS::ApiGateway::RestApi")
+        resources = apigw_client.get_resources(restApiId=rest_api_id)["items"]
+
+        method = apigw_client.get_method(restApiId=rest_api_id, resourceId=resources[0]["id"], httpMethod="GET")
+        method_integration = method["methodIntegration"]
+        self.assertEqual(method_integration["timeoutInMillis"], expected_timeout)
--- a/integration/combination/test_function_with_implicit_api_with_timeout.py
+++ b/integration/combination/test_function_with_implicit_api_with_timeout.py
@@ -1,0 +1,11 @@
+from unittest.case import skipIf
+
+from integration.config.service_names import REST_API
+from integration.helpers.base_test import BaseTest
+from integration.helpers.resource import current_region_does_not_support
+
+
+@skipIf(current_region_does_not_support([REST_API]), "REST API is not supported in this testing region")
+class TestFunctionWithImplicitApiWithTimeout(BaseTest):
+    def test_function_with_implicit_api_with_timeout(self):
+        self.create_and_verify_stack("combination/function_with_implicit_api_with_timeout")
--- a/samtranslator/internal/schema_source/aws_serverless_function.py
+++ b/samtranslator/internal/schema_source/aws_serverless_function.py
@@ -278,7 +278,7 @@ class ApiEventProperties(BaseModel):
     RequestModel: Optional[RequestModel] = apieventproperties("RequestModel")
     RequestParameters: Optional[RequestModelProperty] = apieventproperties("RequestParameters")
     RestApiId: Optional[Union[str, Ref]] = apieventproperties("RestApiId")
-    TimeoutInMillis: Optional[SamIntrinsicable[int]]  # TODO: add doc
+    TimeoutInMillis: Optional[PassThroughProp]  # TODO: add doc
 
 
 class ApiEvent(BaseModel):
--- a/samtranslator/internal/schema_source/aws_serverless_function.py
+++ b/samtranslator/internal/schema_source/aws_serverless_function.py
@@ -278,6 +278,7 @@ class ApiEventProperties(BaseModel):
     RequestModel: Optional[RequestModel] = apieventproperties("RequestModel")
     RequestParameters: Optional[RequestModelProperty] = apieventproperties("RequestParameters")
     RestApiId: Optional[Union[str, Ref]] = apieventproperties("RestApiId")
+    TimeoutInMillis: Optional[SamIntrinsicable[int]]  # TODO: add doc
 
 
 class ApiEvent(BaseModel):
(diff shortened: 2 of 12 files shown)


integration/combination/test_function_with_implicit_api_with_timeout.py
```diff
@@ -0,0 +1,11 @@
...
+
+from integration.config.service_names import REST_API
+from integration.helpers.base_test import BaseTest
+from integration.helpers.resource import current_region_does_not_support
+
+
+@skipIf(current_region_does_not_support([REST_API]), "REST API is not supported in this testing region")
+class TestFunctionWithImplicitApiWithTimeout(BaseTest):
+    def test_function_with_implicit_api_with_timeout(self):
+        self.create_and_verify_stack("combination/function_with_implicit_api_with_timeout")
```
aaythapa, line 11: We're only checking if the stack deploys here. Could we add some code to check if the property is working as expected? Maybe something [like this](https://github.com/aws/serverless-application-model/blob/develop/integration/combination/test_http_api_with_cors.py#L40-L48)

samtranslator/internal/schema_source/aws_serverless_function.py
```diff
@@ -278,6 +278,7 @@ class ApiEventProperties(BaseModel):
     RequestModel: Optional[RequestModel] = apieventproperties("RequestModel")
     RequestParameters: Optional[RequestModelProperty] = apieventproperties("RequestParameters")
     RestApiId: Optional[Union[str, Ref]] = apieventproperties("RestApiId")
+    TimeoutInMillis: Optional[SamIntrinsicable[int]]  # TODO: add doc
```
aaythapa, line 281: Why `SamIntrinsicable` ? I think we use `PassThroughProp` now

tests/translator/input/implicit_api_with_timeout.yaml
```diff
@@ -0,0 +1,41 @@
...
+      CodeUri: s3://sam-demo-bucket/todo_list.zip
+      Handler: index.gethtml
+      Runtime: nodejs12.x
+      Policies: AmazonDynamoDBReadOnlyAccess
+      Events:
+        GetHtml:
+          Type: Api
+          Properties:
+            Path: /{proxy+}
+            Method: any
```
aaythapa, line 41: Don't think you need this function for this test case

samtranslator/model/eventsources/push.py
```diff
@@ -1072,6 +1076,20 @@ def add_auth_to_swagger(  # noqa: PLR0912, PLR0913
...
+        Adds a timeout to this path/method.
+
+        :param path: string of path name
+        :param method_name: string of method name
+        :param timeout: int of timeout duration in milliseconds
+        :param editor: SwaggerEditor object
+
+        """
+        for method_definition in editor.iter_on_method_definitions_for_path_at_method(path, method_name):
+            method_definition[editor._X_APIGW_INTEGRATION]["timeoutInMillis"] = timeout
```
aaythapa, line 1091: Any reason this method is here and not in the `swagger.py` file? Should probably be moved to keep consistency with how http api defines `add_timeout_to_method`

samtranslator/model/eventsources/push.py
```diff
@@ -829,6 +831,8 @@ def _add_swagger_integration(  # type: ignore[no-untyped-def] # noqa: PLR0912, P
             self.add_auth_to_swagger(
                 self.Auth, api, api_id, self.relative_id, self.Method, self.Path, stage, editor, intrinsics_resolver
             )
+        if self.TimeoutInMillis:
+            self.add_timeout_to_swagger(self.Path, self.Method, self.TimeoutInMillis, editor)
```
aaythapa, line 835: I see in [this doc](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-swagger-extensions-integration.html) the `timeoutInMillis` can't be less than 50 and more than 29000. Maybe we can throw an error here if someone puts a num  <50 or >29000

integration/combination/test_function_with_implicit_api_with_timeout.py
gracelu0: yep good idea

samtranslator/model/eventsources/push.py
xazhao: If we consider it as a`Passthrough` property, we don't really need any validations here.

samtranslator/internal/schema_source/aws_serverless_function.py
xazhao: Yep we can use `PassThroughProp` here

samtranslator/model/eventsources/push.py
xazhao: I see this is already implemented in `OpenApiEditor`. Is it possible to reuse it by moving it to `BaseEditor`?

gracelu0:
> I think globals can be implemented in this PR too, just to have everything roll out at once. Unless we have a good reason not to?

`TimeoutInMillis` is not a supported global property for HTTP APIs currently (https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-specification-template-anatomy-globals.html), so for consistency I first implemented the function level support.

samtranslator/model/eventsources/push.py
gracelu0: Right, the two functions are identical and only use BaseEditor constants so I could refactor and move it to `BaseEditor`. Is the use case too specific to be moved there though?

xazhao: We have 2 editor extends `BaseEditor` and both of them require this function. So it makes sense to move there IMO.

gracelu0: Updated so `add_timeout_to_method` is now inherited from `BaseEditor`

xazhao (approved):

xazhao:
LGTM!

tests/translator/input/implicit_api_with_timeout.yaml
```diff
@@ -0,0 +1,27 @@
+Resources:
+  RestApiFunction:
+    Type: AWS::Serverless::Function
+    Properties:
+      CodeUri: s3://sam-demo-bucket/todo_list.zip
+      Handler: index.restapi
+      Runtime: nodejs12.x
```
aaythapa, line 7: Super nitpick: nodejs12 is deprecated

aaythapa (approved):

tests/translator/input/implicit_api_with_timeout.yaml
gracelu0: ah I'll note that when writing future tests

PR #3311, opened 2023-08-17
feat: appsync datasource to eventbus connector

paulhcsun:
### Issue #, if available

https://issues.amazon.com/issues/SAMT-637

### Description of changes

Added Connector from AppSync DataSource to EventBus

- transformation tests

### Checklist

    - [x] Using correct values
    - [ ] Using wrong values

### Examples?


samtranslator/model/connector_profiles/profiles.json
```diff
@@ -790,6 +790,23 @@
             }
           }
         }
+      },
+      "AWS::Events::EventBus": {
+        "Type": "AWS_IAM_ROLE_MANAGED_POLICY",
+        "Properties": {
+          "SourcePolicy": true,
+          "AccessCategories": {
+            "Write": {
```
aaythapa, line 799: Do we not have read permissions?

paulhcsun: I discussed briefly about this with Slava at the start and we were thinking that the general workflow for eventBuses doesn't really make sense to allow you to read from it because we just send events through the eventBus to wherever we want it to go. The other existing connector for Function -> EventBus also only allows for writing and not reading. Do you think there are any scenarios where we'd want to be able to read from it?

aaythapa: Yea I agree with that, since its just routing events there really is no need to read. Thanks for the explanation.

aaythapa (approved):

integration/resources/templates/combination/connector_appsync_to_eventbus.yaml
xazhao, line 1: This template seems not included in our integration test. Do we have a plan to include it later?

paulhcsun: Oh you're right, i will push that file as well.

integration/resources/templates/combination/connector_appsync_to_eventbus.yaml
```diff
@@ -0,0 +1,173 @@
...
+                Resources:[
+                  sayHelloResponse['sayHello']
+                ],
+                DetailType:"myDetailType"
+              }
+            ]
+          }
+
+          var eventbridge = new AWS.EventBridge();
+          await eventbridge.putEvents(put_params).promise();
```
ssenchenko, line 167: The test expects asn explicit [response](https://github.com/ssenchenko/serverless-application-model/blob/00245b65ca324a418ebe713359350fab829b964a/integration/combination/test_connectors.py#L73)

We need to process the result, check that there are no errors and return 200. Or return errors too (see the test)

integration/resources/templates/combination/connector_appsync_to_eventbus.yaml
```diff
@@ -0,0 +1,173 @@
...
+            const data = body.data?.[queryName];
+            
+            if (body.errors !== undefined) {
+              throw JSON.stringify(body.errors);
+            }
+            
+            return body.data;
+          };
+
+          const sayHelloResponse = await makeRequest("sayHello");
```
ssenchenko, line 150: Is it okay to call it before PutEvent()?

integration/resources/templates/combination/connector_appsync_to_eventbus.yaml
```diff
@@ -0,0 +1,173 @@
...
+    Properties:
+      ApiId: !GetAtt AppSyncApi.ApiId
+      TypeName: Query
+      FieldName: sayHello
+      DataSourceName: !GetAtt AppSyncEventBusDataSource.Name
+      RequestMappingTemplate: |
+        {
+          "version" : "2018-05-29", 
+          "operation" : "PutEvents",
+          "events" : [{}]
```
ssenchenko, line 61: Is it okay that `events` are empty? Can we test with that?

integration/resources/templates/combination/connector_appsync_to_eventbus.yaml
paulhcsun: I'll take a look at ways we can verify that the event is actually in the eventBus after calling `putEvents()` and add that to this test case.

integration/resources/templates/combination/connector_appsync_to_eventbus.yaml
paulhcsun: Related to https://github.com/aws/serverless-application-model/pull/3311/files#r1300549275, will be resolved with verification of event being in the eventBus.

integration/resources/templates/combination/connector_appsync_to_eventbus.yaml
paulhcsun: Related to https://github.com/aws/serverless-application-model/pull/3311/files#r1300549275, will be resolved with verification of event being in the eventBus.

integration/resources/templates/combination/connector_appsync_to_eventbus.yaml
paulhcsun: Updated the test case to verify that events are succesfully sent to the eventBus via the putEvents call

ssenchenko (approved):

paulhcsun:
Previously blocked by China region marketing launch. Ready to merge now.

PR #3316, opened 2023-08-22
fix: Fix availability dip in PropagateTags property

GavinZZ:
### Description of changes
Dip fix on `PropagateTags` property.

All tests passed. 

### Checklist

    - [ ] Using correct values
    - [ ] Using wrong values

### Examples?

### DIFF
--- a/samtranslator/plugins/api/implicit_api_plugin.py
+++ b/samtranslator/plugins/api/implicit_api_plugin.py
@@ -107,7 +107,7 @@ class ImplicitApiPlugin(BasePlugin, Generic[T], metaclass=ABCMeta):
             return
 
         implicit_api_resource = template.get(self.IMPLICIT_API_LOGICAL_ID)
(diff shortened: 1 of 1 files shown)


awood45 (approved):

GavinZZ:
The reason I did not add a test is because We do validation on template which caused 
```
    def on_before_transform_template(self, template_dict):  # type: ignore[no-untyped-def]
        """
        Hook method that runs before a template gets transformed. In this method, we parse and process Globals section
        from the template (if present).
    
        :param dict template_dict: SAM template as a dictionary
        """
        try:
            global_section = Globals(template_dict)
        except InvalidGlobalsSectionException as ex:
>           raise InvalidDocumentException([ex]) from ex
E           samtranslator.model.exceptions.InvalidDocumentException: [InvalidGlobalsSectionException('Globals', 'Value of ${section} must be a dictionary')]```

I don't think it's worth the trouble to add this test.

aaythapa (approved):

PR #3328, opened 2023-08-31
[fix] Auth Override not working with DefinitionBody fix

paulhcsun:
### Description of changes
Adding new optional opt-in property `OverrideApiAuth` with a `boolean` value to the `AWS::Serverless::Function` `Events` level. 

Transformation test

### Checklist

    - [x] Using correct values
    - [ ] Using wrong values

### Examples?


samtranslator/internal/schema_source/aws_serverless_function.py
```diff
@@ -280,6 +281,7 @@ class ApiEventProperties(BaseModel):
     RequestParameters: Optional[RequestModelProperty] = apieventproperties("RequestParameters")
     RestApiId: Optional[Union[str, Ref]] = apieventproperties("RestApiId")
     TimeoutInMillis: Optional[PassThroughProp]  # TODO: add doc
+    TestAuthOverride: Optional[PassThroughProp]  # TODO: add doc
```
aaythapa, line 284: Why's it here aswell? Isn't it just under auth?

samtranslator/model/eventsources/push.py
```diff
@@ -668,6 +669,7 @@ class Api(PushEventSource):
     RequestModel: Optional[Dict[str, Any]]
     RequestParameters: Optional[List[Any]]
     TimeoutInMillis: Optional[PassThrough]
+    TestAuthOverride: Optional[bool]
```
aaythapa, line 672: Same comment as above, Shouldn't this go under the list of [auth properties](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-property-function-apifunctionauth.html)?

samtranslator/model/sam_resources.py
```diff
@@ -179,6 +179,7 @@ class SamFunction(SamResourceMacro):
         "SnapStart": PropertyType(False, IS_DICT),
         "FunctionUrlConfig": PropertyType(False, IS_DICT),
         "RuntimeManagementConfig": PassThroughProperty(False),
+        "TestAuthOverride": PropertyType(False, IS_BOOL),
```
aaythapa, line 182: Don't need to add this here, this is only for list of SAM function properites

tests/model/eventsources/test_api_event_source.py
```diff
@@ -21,10 +22,16 @@ def setUp(self):
 
         self.stage = "Prod"
         self.suffix = "123"
+        self.kwargs = {
+            "function": self.func,
+            "explicit_api": {},
+            "api_id": "RestApi",
+            "intrinsics_resolver": IntrinsicsResolver({}),
```
aaythapa, line 29: what's the reason for messing with kwargs here (can't deduce from the code) rather than testing it how it was before?

tests/translator/input/api_with_swagger_authorizer_none_test_auth_override.yaml
```diff
@@ -0,0 +1,142 @@
...
+            Path: /lambda-token
+        LambdaTokenAnyMethod:
+          Type: Api
+          Properties:
+            RestApiId: !Ref MyApiWithLambdaTokenAuth
+            Method: any
+            Auth:
+              Authorizer: NONE
+            Path: /any/lambda-token
+        LambdaRequest:
```
aaythapa, line 112: having the events name state which ones overriding vs which ones not would help to easily compare with the output

tests/translator/input/api_with_swagger_authorizer_none_test_auth_override.yaml
aaythapa, line 1: this test looks a little bloated, lot of the template isn't needed or relevant to what you're testing

tests/model/eventsources/test_api_event_source.py
paulhcsun: This is because we changed the way we get the `intrinsics_resolver` in the `to_cloudformation()` method from `intrinsics_resolver = kwargs.get("intrinsics_resolver")` to `intrinsics_resolver: IntrinsicsResolver = kwargs["intrinsics_resolver"]` and caused those tests to fail because we now expect there to always be an `IntrinsicsResolver`. We changed this because the `self.add_auth_to_swagger()` call compalins about `intrinsics_resolver` being returned as `Optional` by the `.get()` method. It is returned similarly in the `sam_resources.py` file.

samtranslator/model/sam_resources.py
paulhcsun: Good catch, i'll remove this.

samtranslator/internal/schema_source/aws_serverless_function.py
paulhcsun: Good catch, will remove this.

samtranslator/model/eventsources/push.py
```diff
@@ -743,6 +745,24 @@ def to_cloudformation(self, **kwargs):  # type: ignore[no-untyped-def]
         if explicit_api.get("__MANAGE_SWAGGER") or explicit_api.get("MergeDefinitions"):
             self._add_swagger_integration(explicit_api, api_id, function, intrinsics_resolver)  # type: ignore[no-untyped-call]
 
+        swagger_body = explicit_api.get("DefinitionBody")
+
+        if swagger_body and self.Auth and self.Auth.get("TestAuthOverride") is True:
```
GavinZZ, line 750: I thought the name `TestAuthOverride` is still pending UX decision based on the doc I saw yesterday. Has that decision been made?

tests/translator/input/api_with_swagger_authorizer_none_test_auth_override.yaml
paulhcsun: trimmed these down so this should be fine now as there is just one event per api.

samtranslator/model/eventsources/push.py
paulhcsun: Not yet, I've signed up for tomorrows UX office hour and will push a change to update the property name once that decision is made. The current PR is open just for general review and will not be merged until the property name is updated.

samtranslator/model/eventsources/push.py
```diff
@@ -743,6 +743,24 @@ def to_cloudformation(self, **kwargs):  # type: ignore[no-untyped-def]
         if explicit_api.get("__MANAGE_SWAGGER") or explicit_api.get("MergeDefinitions"):
             self._add_swagger_integration(explicit_api, api_id, function, intrinsics_resolver)  # type: ignore[no-untyped-call]
 
+        swagger_body = explicit_api.get("DefinitionBody")
+
+        if swagger_body and self.Auth and self.Auth.get("OverrideApiAuth") is True:
```
bentvelj, line 748: [nit] Is the `is True` necessary?

paulhcsun: Good catch, I don't think it is. I'll push a change to remove this.

samtranslator/model/eventsources/push.py
```diff
@@ -727,7 +727,7 @@ def to_cloudformation(self, **kwargs):  # type: ignore[no-untyped-def]
         resources = []
 
         function = kwargs.get("function")
-        intrinsics_resolver = kwargs.get("intrinsics_resolver")
+        intrinsics_resolver: IntrinsicsResolver = kwargs["intrinsics_resolver"]
```
ssenchenko, line 730: Why haven't we been sure that it's not `None` but now we are?

tests/model/eventsources/test_api_event_source.py
ssenchenko: So we're doing that only because of the static type check?
Then we should initialize `intrinsic_resolver` to an empty object in `push.py` and don't touch the tests as it can be a backward compat issue

GavinZZ (approved):
Changes LGTM, but my understanding is that we won't merge this until UX review. Just want to make sure we don't accidentally merge it.

samtranslator/model/eventsources/push.py
ssenchenko: As I see below, we probably have to create an empty `IntrinsicResolver` here if we need it but... still we need to be very cautious about that. Is it `self.add_auth_to_swagger` which requires the resolver not to be `None`? Can we handle `None` value inside it?

samtranslator/model/eventsources/push.py
```diff
@@ -743,6 +743,24 @@ def to_cloudformation(self, **kwargs):  # type: ignore[no-untyped-def]
...
+                explicit_api,
+                api_id,
+                self.relative_id,
+                self.Method,
+                self.Path,
+                stage,
+                editor,
+                intrinsics_resolver,
+            )
+            explicit_api["DefinitionBody"] = editor.swagger
```
aaythapa, line 763: Can you add some comments here? Bit confused about what its actually doing

tests/model/eventsources/test_api_event_source.py
aaythapa: > We changed this because the self.add_auth_to_swagger() call compalins about intrinsics_resolver being returned as Optional by the .get() method.

what was complaining exactly? (if you have the err msg that'd be helpful)

tests/translator/input/api_with_swagger_authorizer_none_override_api_auth.yaml
aaythapa, line 1: tests look good, can you add one more event without `OverrideApiAuth` just to show what output looks like without it being overridden

paulhcsun: Good idea, i'll add that in. Also in response to the above comment, here is the error for `intrinsics_resolver`:

`samtranslator/model/eventsources/push.py:762: error: Argument 9 to "add_auth_to_swagger" of "Api" has incompatible type "Optional[Any]"; expected "IntrinsicsResolver"  [arg-type]`

samtranslator/model/eventsources/push.py
paulhcsun: For sure, i'll add some comments in the next commit. Essentially it's just creating the resources needed to call `self.add_auth_to_swagger()` and checking if the `OverrideApiAuth` property is set. 

`add_auth_to_swagger()` is needed because it will call `add_auth_to_method()` in `swagger.py` which checks event source `Auth` and updates the `DefinitionBody` with any security overrides.

then the line `explicit_api["DefinitionBody"] = editor.swagger` saves that change to `DefinitionBody` so it can be detected when transforming the `Api` resource.

paulhcsun:
> Changes LGTM, but my understanding is that we won't merge this until UX review. Just want to make sure we don't accidentally merge it.

UX review is finished, the property name has been approved and is okay to merge once this PR is approved. I will remove the [DO NOT MERGE].

samtranslator/model/eventsources/push.py
paulhcsun: `self.add_auth_to_swagger` expects `IntrinsicsResolver` but the `.get()` is returning an `Optional[Any]`. 

I don't think we can handle this inside of `add_auth_to_swagger()` but it seems like if i add a check on `intrinsics_resolver` before calling `add_auth_to_swagger()` it is happy.

i.e. `if swagger_body and self.Auth and intrinsics_resolver and self.Auth.get("OverrideApiAuth"):`

ssenchenko: Sounds good. But do we really need an intrinsic resolver to be non null to make override happen?

paulhcsun:
Due to [limited support for intrinsic functions in SAM templates](https://github.com/aws/serverless-application-model/issues/2533), we will not be supporting intrinsics for the new property `OverrideApiAuth`.

aaythapa (approved):

ssenchenko (approved):

PR #3337, opened 2023-09-05
fix: Append CodeSha256 Value to Description

GavinZZ:
### Description of changes
Append the value of CodeSha256 to end of Lambda Function description. This would cause an update to the existing users who use codesha256 but since it will always create a new lambda version, it wouldn't cause any downtime for users. 

```
When you deploy a new version of an AWS Lambda function, it does not cause downtime for the existing function. AWS Lambda is designed to allow seamless deployments without affecting the current running version
```

Additionally add `DeletionPolicy` and `UpdateReplacePolicy` to an integ test.

All tests passed.

### Checklist

    - [ ] Using correct values
    - [ ] Using wrong values

### Examples?

### DIFF
--- a/samtranslator/model/sam_resources.py
+++ b/samtranslator/model/sam_resources.py
@@ -297,6 +297,8 @@ class SamFunction(SamResourceMacro):
                 description = intrinsics_resolver.resolve_parameter_refs(self.Description)
                 if not description or isinstance(description, str):
                     lambda_function.Description = f"{description} {code_sha256}" if description else code_sha256
+                else:
+                    lambda_function.Description = {"Fn::Join": [" ", [description, code_sha256]]}
             lambda_version = self._construct_version(
                 lambda_function, intrinsics_resolver=intrinsics_resolver, code_sha256=code_sha256
             )
--- a/samtranslator/model/sam_resources.py
+++ b/samtranslator/model/sam_resources.py
@@ -291,9 +291,9 @@ class SamFunction(SamResourceMacro):
                         "AutoPublishCodeSha256 must be a string",
                     )
                 # Lambda doesn't create a new version if the code in the unpublished version is the same as the
-                # previous published version. To handle scenario users only update content in 'CodeUri', CloudFormation
-                # won't be able to detect any change in lambda function in the template, which would cause deployment
-                # failure. Append to description to handle this issue
+                # previous published version. To address situations where users modify only the 'CodeUri' content, 
+                # CloudFormation might not detect any changes in the Lambda function within the template, leading 
+                # to deployment issues. To resolve this, we'll append codesha256 value to the description.
                 description = intrinsics_resolver.resolve_parameter_refs(self.Description)
                 if not description or isinstance(description, str):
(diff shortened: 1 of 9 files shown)


samtranslator/model/sam_resources.py
```diff
@@ -290,6 +290,13 @@ def to_cloudformation(self, **kwargs):  # type: ignore[no-untyped-def]
                         self.logical_id,
                         "AutoPublishCodeSha256 must be a string",
                     )
+                # Lambda doesn't create a new version if the code in the unpublished version is the same as the
+                # previous published version. To handle scenario users only update content in 'CodeUri', CloudFormation
```
paulhcsun, line 294: nit: "To handle this scenario, users..."

samtranslator/model/sam_resources.py
```diff
@@ -290,6 +290,13 @@ def to_cloudformation(self, **kwargs):  # type: ignore[no-untyped-def]
...
+                # won't be able to detect any change in lambda function in the template, which would cause deployment
+                # failure. Append to description to handle this issue
+                description = intrinsics_resolver.resolve_parameter_refs(self.Description)
```
ssenchenko, line 297: What if `Description` uses other intrinsics which we fail to resolve?
Would it be safer to use `!Join` intrinsic to add `code_sha_256`?

samtranslator/model/sam_resources.py
GavinZZ: I don't think that's what I'm trying to convey. I want to say to handle the scenario that users only updated content in `CodeUri`, ..... Let me rephrase it.

ssenchenko (commented):
Do I understand it correctly that you're saying that updating `Description`will cause redeployment?

samtranslator/model/sam_resources.py
GavinZZ: Sorry my description is not accurate. I meant it will cause an update to users who use CodeSha256 in their template even when there's no change made.

GavinZZ: If `Description` uses other intrinsics, we would not resolve it here. `!Join` is definitely a viable solution. The reason I didn't use `!Join` is that I feel like it's an overkill since not many users would use `Description` with other intrinsics and use `AutoPublishCodeSha256` at the same time. What do you think?

ssenchenko: I agree that it sounds like an overkill and in 99.90% there will be no need in `Join` but I was surprised when I got a ticket from users using intrinsics on Name in GraphQL to override LogicalId. It'd say it's better to be safe.

SimonCMoore:
How about adding a test for when there is already a description and one for a description with an intrinsic in it?

GavinZZ:
Added

ssenchenko (approved):

paulhcsun (approved):

ramosbugs:
Btw, this was a breaking change since this caused some of our Lambda function descriptions to now be longer than the max length of 256, and deployment failed. Now we have to shorten our descriptions to compensate... something to be aware of for the future.

PR #3353, opened 2023-09-26
fix: Catch network interface deletion failure

GavinZZ:
### Description of changes
Catch network interface deletion failure. We should not throw an error when we failed to delete an unused network interface.

N/A

### Checklist

    - [ ] Using correct values
    - [ ] Using wrong values

### Examples?

### DIFF
--- a/integration/conftest.py
+++ b/integration/conftest.py
@@ -75,7 +75,10 @@ def _delete_unused_network_interface_by_subnet(ec2_client, subnet_id):
         network_interface_ids += [ni["NetworkInterfaceId"] for ni in page["NetworkInterfaces"]]
 
     for ni_id in network_interface_ids:
-        ec2_client.delete_network_interface(NetworkInterfaceId=ni_id)
+        try:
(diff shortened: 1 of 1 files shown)


ssenchenko (approved):

integration/conftest.py
```diff
@@ -75,7 +75,10 @@ def _delete_unused_network_interface_by_subnet(ec2_client, subnet_id):
         network_interface_ids += [ni["NetworkInterfaceId"] for ni in page["NetworkInterfaces"]]
 
     for ni_id in network_interface_ids:
-        ec2_client.delete_network_interface(NetworkInterfaceId=ni_id)
+        try:
+            ec2_client.delete_network_interface(NetworkInterfaceId=ni_id)
```
xazhao, line 79: Shall we do the detach first before deleting? `You must detach the network interface before you can delete it.`

GavinZZ: There is a check above that checks whether the status is `available`. And `if the network interface is not attached to an instance, the status is available`

xazhao (approved):

PR #3363, opened 2023-10-03
feat: allow user to provide role for schedule

sidhujus:
### Issue #, if available
feature request discussion [#3290](https://github.com/aws/serverless-application-model/discussions/3290
)
### Description of changes
Add new Role property to Schedule for user to pass in their own role to be used with the generated event rule
 Example 
```Resources:

  MyStateMachine:
    Type: AWS::Serverless::StateMachine
    Properties:
      Definition:
        Comment: A Hello World example of the Amazon States Language using Pass states
        StartAt: Hello
        States:
          Hello:
            Type: Pass
            Result: Hello
            Next: World
          World:
            Type: Pass
            Result: World
            End: true
      Policies:
      - Version: '2012-10-17'
        Statement:
        - Effect: Deny
          Action: '*'
          Resource: '*'

      Events:
        CWSchedule:
          Type: Schedule
          Properties:
            Schedule: rate(1 minute)
            Description: test schedule
            Enabled: false
            Role: with-role-MyStateMachineCWScheduleRole-RI4SGRI1ORL9
```
Validated with tests and tansformed a template with the new Role property and succesfully deployed and had it execute the state machine 
### Checklist

    - [x] Using correct values
    - [ ] Using wrong values

### Examples?

### DIFF
--- a/samtranslator/model/stepfunctions/events.py
+++ b/samtranslator/model/stepfunctions/events.py
@@ -1,6 +1,6 @@
 import json
 from abc import ABCMeta
-from typing import Any, Dict, List, Optional, cast
+from typing import Any, Dict, List, Optional, Union, cast
 
 from samtranslator.metrics.method_decorator import cw_timer
 from samtranslator.model import Property, PropertyType, Resource, ResourceMacro
@@ -144,7 +144,7 @@ class Schedule(EventSource):
         events_rule.Name = self.Name
         events_rule.Description = self.Description
 
-        role = None
+        role = self.RoleArn
         if self.RoleArn is None:
             role = self._construct_role(resource, permissions_boundary, prefix=None)
             resources.append(role)
@@ -157,27 +157,30 @@ class Schedule(EventSource):
                 self, source_arn, passthrough_resource_attributes
             )
             resources.extend(dlq_resources)
-        events_rule.Targets = [self._construct_target(resource, role, self.RoleArn, dlq_queue_arn)]
+        events_rule.Targets = [self._construct_target(resource, role, dlq_queue_arn)]
 
         return resources
 
     def _construct_target(
         self,
         resource: StepFunctionsStateMachine,
-        role: Optional[IAMRole],
-        provided_role: Optional[str],
+        role: Union[IAMRole, Optional[PassThrough]],
         dead_letter_queue_arn: Optional[str],
     ) -> Dict[str, Any]:
-        """Constructs the Target property for the EventBridge Rule.
+        """_summary_
 
-        parameters:
-        resource: The State Machine resource associated with the event
-        role: An IAMRole resource
-        provided_role: User provided role ARN to be used instead of a newly generated IAMRole resource
-        dead_letter_queue_arn: ARN of the dead letter queue
+        Parameters
+        ----------
+        resource
+            StepFunctionsState machine resource to be generated
+        role
+            The role to be used by the Schedule event resource either generated or user provides arn
+        dead_letter_queue_arn
+            Dead letter queue associated with the resource
 
-        :returns: the Target property
-        :rtype: dict
+        Returns
+        -------
+            _description_
         """
         target_id = (
             self.Target["Id"]
@@ -185,18 +188,14 @@ class Schedule(EventSource):
             else generate_valid_target_id(self.logical_id, EVENT_RULE_SFN_TARGET_SUFFIX)
         )
 
-        if provided_role:
-            target = {
-                "Arn": resource.get_runtime_attr("arn"),
-                "Id": target_id,
-                "RoleArn": provided_role,
-            }
-        elif role:
-            target = {
-                "Arn": resource.get_runtime_attr("arn"),
-                "Id": target_id,
-                "RoleArn": role.get_runtime_attr("arn"),
-            }
+        target = {
+            "Arn": resource.get_runtime_attr("arn"),
+            "Id": target_id,
+        }
+
+        if role is not None:
+            target["RoleArn"] = role if isinstance(role, str) else role.get_runtime_attr("arn")
+
         if self.Input is not None:
             target["Input"] = self.Input
 
--- a/samtranslator/model/stepfunctions/events.py
+++ b/samtranslator/model/stepfunctions/events.py
@@ -93,6 +93,7 @@ class Schedule(EventSource):
         "DeadLetterConfig": PropertyType(False, IS_DICT),
         "RetryPolicy": PropertyType(False, IS_DICT),
         "Target": Property(False, IS_DICT),
+        "Role": Property(False, IS_STR),
     }
 
     Schedule: PassThrough
@@ -104,6 +105,7 @@ class Schedule(EventSource):
     DeadLetterConfig: Optional[Dict[str, Any]]
     RetryPolicy: Optional[PassThrough]
     Target: Optional[PassThrough]
+    Role: Optional[PassThrough]
 
     @cw_timer(prefix=SFN_EVETSOURCE_METRIC_PREFIX)
     def to_cloudformation(self, resource, **kwargs):  # type: ignore[no-untyped-def]
@@ -135,8 +137,10 @@ class Schedule(EventSource):
         events_rule.Name = self.Name
         events_rule.Description = self.Description
 
-        role = self._construct_role(resource, permissions_boundary)  # type: ignore[no-untyped-call]
-        resources.append(role)
+        role = None
+        if self.Role is None:
+            role = self._construct_role(resource, permissions_boundary)  # type: ignore[no-untyped-call]
+            resources.append(role)
 
         source_arn = events_rule.get_runtime_attr("arn")
         dlq_queue_arn = None
@@ -146,11 +150,11 @@ class Schedule(EventSource):
                 self, source_arn, passthrough_resource_attributes
             )
             resources.extend(dlq_resources)
-        events_rule.Targets = [self._construct_target(resource, role, dlq_queue_arn)]  # type: ignore[no-untyped-call]
+        events_rule.Targets = [self._construct_target(resource, role, self.Role, dlq_queue_arn)]  # type: ignore[no-untyped-call]
 
         return resources
 
-    def _construct_target(self, resource, role, dead_letter_queue_arn=None):  # type: ignore[no-untyped-def]
+    def _construct_target(self, resource, role, provided_role=None, dead_letter_queue_arn=None):  # type: ignore[no-untyped-def]
         """Constructs the Target property for the EventBridge Rule.
 
         :returns: the Target property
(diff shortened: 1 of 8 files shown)


GavinZZ:
nit: we usually name the PR 
`feat: xxxxx` for feature request
`fix: xxxxx` for bug fix
`chore: xxxxx` for chore task

GavinZZ (commented):
Noticed that this currently only supports hardcoded `Role` arn. SAM generally supports both hardcoded RoleArn or intrinsics `!Ref` to a role resource.

samtranslator/model/stepfunctions/events.py
```diff
@@ -135,8 +137,10 @@ def to_cloudformation(self, resource, **kwargs):  # type: ignore[no-untyped-def]
         events_rule.Name = self.Name
         events_rule.Description = self.Description
 
-        role = self._construct_role(resource, permissions_boundary)  # type: ignore[no-untyped-call]
-        resources.append(role)
+        role = None
+        if self.Role is None:
+            role = self._construct_role(resource, permissions_boundary)  # type: ignore[no-untyped-call]
```
ssenchenko, line 142: Generally we try to avoid adding `# type: ignore` in new code. Have you tried to type `_construct_role()`? `# type: ignore` is the last resort if typing the invoked API would take too long.

samtranslator/model/stepfunctions/events.py
```diff
@@ -146,11 +150,11 @@ def to_cloudformation(self, resource, **kwargs):  # type: ignore[no-untyped-def]
...
             )
             resources.extend(dlq_resources)
-        events_rule.Targets = [self._construct_target(resource, role, dlq_queue_arn)]  # type: ignore[no-untyped-call]
+        events_rule.Targets = [self._construct_target(resource, role, self.Role, dlq_queue_arn)]  # type: ignore[no-untyped-call]
 
         return resources
 
-    def _construct_target(self, resource, role, dead_letter_queue_arn=None):  # type: ignore[no-untyped-def]
+    def _construct_target(self, resource, role, provided_role=None, dead_letter_queue_arn=None):  # type: ignore[no-untyped-def]
         """Constructs the Target property for the EventBridge Rule.
```
ssenchenko, line 159: Please update docstring

samtranslator/model/stepfunctions/events.py
```diff
@@ -146,11 +150,11 @@ def to_cloudformation(self, resource, **kwargs):  # type: ignore[no-untyped-def]
                 self, source_arn, passthrough_resource_attributes
             )
             resources.extend(dlq_resources)
-        events_rule.Targets = [self._construct_target(resource, role, dlq_queue_arn)]  # type: ignore[no-untyped-call]
+        events_rule.Targets = [self._construct_target(resource, role, self.Role, dlq_queue_arn)]  # type: ignore[no-untyped-call]
 
         return resources
 
-    def _construct_target(self, resource, role, dead_letter_queue_arn=None):  # type: ignore[no-untyped-def]
+    def _construct_target(self, resource, role, provided_role=None, dead_letter_queue_arn=None):  # type: ignore[no-untyped-def]
```
ssenchenko, line 157: Let's type it. It's okay to use smth like `Dict[str, Any]` It's still better than no typings at all

samtranslator/model/stepfunctions/events.py
ssenchenko: And no need to add types to the docstring. for new code we use Numpy docstring format

samtranslator/model/stepfunctions/events.py
ssenchenko, line 157: It doesn't seem that we use `role` parameter at all

samtranslator/model/stepfunctions/events.py
```diff
@@ -161,11 +165,20 @@ def _construct_target(self, resource, role, dead_letter_queue_arn=None):  # type
...
             else generate_valid_target_id(self.logical_id, EVENT_RULE_SFN_TARGET_SUFFIX)
         )
-        target = {
-            "Arn": resource.get_runtime_attr("arn"),
-            "Id": target_id,
-            "RoleArn": role.get_runtime_attr("arn"),
-        }
+
+        if provided_role:
+            role_arn = fnSub("arn:aws:iam::${AWS::AccountId}:role/${Role}", {"Role": provided_role})
```
aaythapa, line 170: I see that [ScheduleV2 event](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-property-statemachine-statemachineschedulev2.html#sam-statemachine-statemachineschedulev2-rolearn) has a `RoleArn` property that is a straight pass through. Maybe we should do the same here to match that event + we don't have to worry about what the value is, we can just pass it through.

samtranslator/model/stepfunctions/events.py
```diff
@@ -167,13 +186,12 @@ def _construct_target(self, resource, role, provided_role=None, dead_letter_queu
...
         if provided_role:
-            role_arn = fnSub("arn:aws:iam::${AWS::AccountId}:role/${Role}", {"Role": provided_role})
             target = {
                 "Arn": resource.get_runtime_attr("arn"),
                 "Id": target_id,
-                "RoleArn": role_arn,
+                "RoleArn": provided_role,
             }
-        else:
+        elif role:
```
aaythapa, line 194:
```suggestion
  target = {
      "Arn": resource.get_runtime_attr("arn"),
      "Id": target_id,
      "RoleArn": role_arn,
      "RoleArn": provided_role if provided_role else role.get_runtime_attr("arn"),
  }
```
Think this is cleaner

samtranslator/model/stepfunctions/events.py
```diff
@@ -50,7 +51,13 @@ def _generate_logical_id(self, prefix, suffix, resource_type):  # type: ignore[n
         generator = logical_id_generator.LogicalIdGenerator(prefix + resource_type, suffix)
         return generator.gen()
 
-    def _construct_role(self, resource, permissions_boundary=None, prefix=None, suffix=""):  # type: ignore[no-untyped-def]
+    def _construct_role(
+        self,
+        resource: StepFunctionsStateMachine,
+        permissions_boundary: Optional[str],
+        prefix: Optional[str],
+        suffix: str = "",
```
aaythapa, line 59: why `str = ""` here rather than `Optional[str]`

samtranslator/model/stepfunctions/events.py
```diff
@@ -161,11 +184,19 @@ def _construct_target(self, resource, role, dead_letter_queue_arn=None):  # type
...
+                "Arn": resource.get_runtime_attr("arn"),
+                "Id": target_id,
+                "RoleArn": provided_role,
+            }
+        elif role:
+            target = {
+                "Arn": resource.get_runtime_attr("arn"),
+                "Id": target_id,
+                "RoleArn": role.get_runtime_attr("arn"),
+            }
```
aaythapa, line 199:
```suggestion
        target = {
            "Arn": resource.get_runtime_attr("arn"),
            "Id": target_id,
            "RoleArn": provided_role if provided_role else role.get_runtime_attr("arn"),
        }
```
Think something like this looks cleaner

tests/translator/input/state_machine_with_schedule_role.yaml
```diff
@@ -0,0 +1,15 @@
...
+    Properties:
+      DefinitionUri: s3://sam-demo-bucket/my_state_machine.asl.json
+      Events:
+        CWSchedule:
+          Type: Schedule
+          Properties:
+            Schedule: rate(1 minute)
+            Description: test schedule
+            Enabled: false
+            RoleArn: FakeArn
```
aaythapa, line 15: should look like an arn at least. `!Sub 'arn:${AWS::Partition}:iam::role/yoyo'` will do

tests/translator/input/state_machine_with_schedule_role.yaml
```diff
@@ -0,0 +1,15 @@
+Resources:
+
+  MyStateMachine:
+    Type: AWS::Serverless::StateMachine
+
+    Properties:
+      DefinitionUri: s3://sam-demo-bucket/my_state_machine.asl.json
+      Events:
```
aaythapa, line 8: should add an Event without the `RoleArn` property as well

samtranslator/model/stepfunctions/events.py
sidhujus: I tried doing that but the linting fails then because roles type is Optional and it basically requires us to explicitly check that role is defined. A way around that is that we can generate a dummy IAMRole and pass that in, but then it makes less sense to read from the outside.

samtranslator/model/stepfunctions/events.py
sidhujus: https://github.com/aws/serverless-application-model/pull/3363/files#diff-e4eab90ce32328ebbbb90bad6d4e5ea21e06f241708a08eb9354c88146301f9dR49 it checks whether the suffix is alphanumerical but doesnt do that for the prefix. So I tried to avoid changing the logic there and just left it to default to the empty string.

samtranslator/model/stepfunctions/events.py
sidhujus: changed it so that that theres just one role parameter now and it checks the type to decide how to fill in the template

samtranslator/model/stepfunctions/events.py
sidhujus: Talked offline and decided that changing _generate_logical_id() was not worth it

GavinZZ (approved):
Great work!

samtranslator/model/stepfunctions/events.py
```diff
@@ -135,8 +144,12 @@ def to_cloudformation(self, resource, **kwargs):  # type: ignore[no-untyped-def]
         events_rule.Name = self.Name
         events_rule.Description = self.Description
 
-        role = self._construct_role(resource, permissions_boundary)  # type: ignore[no-untyped-call]
-        resources.append(role)
+        role: Union[IAMRole, str]
```
aaythapa, line 147: I think you need to add `Dict` here as well. If the user puts intrinsics then the type will be `Dict[str, Any]`

samtranslator/model/stepfunctions/events.py
```diff
@@ -146,26 +159,45 @@ def to_cloudformation(self, resource, **kwargs):  # type: ignore[no-untyped-def]
...
 
-    def _construct_target(self, resource, role, dead_letter_queue_arn=None):  # type: ignore[no-untyped-def]
-        """Constructs the Target property for the EventBridge Rule.
-
-        :returns: the Target property
-        :rtype: dict
+    def _construct_target(
+        self,
+        resource: StepFunctionsStateMachine,
+        role: Union[IAMRole, str],
```
aaythapa, line 169: same comment as above

ssenchenko (approved):

Issue #1874, opened 2020-12-13
Allow Globals to be ignored by Functions and not applied.


brysontyrrell:
### Describe your idea/feature/enhancement

I'm experimenting with using the `PackageType: Image` option for Functions and came across this error when I attempted to deploy my test app:

```
Error: Failed to create changeset for the stack: *****, ex: Waiter ChangeSetCreateComplete failed: Waiter encountered a terminal failure state Status: FAILED. Reason: Transform AWS::Serverless-2016-10-31 failed with: Invalid Serverless Application Specification document. Number of errors found: 1. Resource with id [*****] is invalid. Runtime, Handler, Layers cannot be present when PackageType is of type `Image`
```

Example template:

```yaml
AWSTemplateFormatVersion: 2010-09-09
Transform: AWS::Serverless-2016-10-31

Globals:

  Function:
    Runtime: python3.8
    Handler: index.lambda_handler
    MemorySize: 128

Resources:

  ZipFunction:
    Type: AWS::Serverless::Function
    Properties:
      CodeUri: ./src/zip

  ImageFunction:
    Type: AWS::Serverless::Function
    Properties:
      PackageType: Image
    Metadata:
      Dockerfile: Dockerfile
      DockerContext: ./src/image

```

If I have a template of Zip and Image type Functions I can no longer use Globals because this error is thrown. I will have to define these properties on each Zip type individually.

### Proposal

Either an addition to Globals or AWS::Serverless::Function that prevents globals from being applied to the resource(s).

### Additional Details

SAM CLI, version 1.13.2

si3mshady:
SAM sucks

sapessi:
> SAM sucks

We'd love to try and change your mind. What can we change/fix to make SAM work for your use-case?

hindmanj:
It feels like the core issue is actually how Globals is specified and applied to resources.  Specifically, it's not possible to 'subtype' deep enough.  Conceptually, the better solution is something like one of these for Globals:

```
Globals:
  Function:
    ZipPackageType:
      Runtime:
      Layers:
    ImagePackageType:
      ImageUri:

Globals:
  Function:
    Runtime:
    Layers:
  ImageFunction:
    ImageUri:
```

[...]

PR #3386, opened 2023-10-18
feat: Support Ignore Globals resource attribute

GavinZZ:
### Issue #, if available
https://github.com/aws/serverless-application-model/issues/1874

### Description of changes
Introduce `IgnoreGlobals` property on SAM Function. It's default to false, but when set to true, it will ignore the Function globals section.

All existing and new tests pass

### Checklist

    - [ ] Using correct values
    - [ ] Using wrong values

### Examples?


ssenchenko (approved):

ConnorRobertson (approved):

samtranslator/schema/schema.json
```diff
@@ -250245,6 +250245,20 @@
...
+              "type": "string"
+            },
+            {
+              "items": {
+                "type": "string"
+              },
+              "type": "array"
+            }
+          ],
+          "title": "Ignoreglobals"
```
paulhcsun, line 250260: nit: can we capitalize the G in globals to keep consistent with the other titles

paulhcsun (approved):

samtranslator/schema/schema.json
GavinZZ: This is auto-generated so I can't control how the schema generates.

samtranslator/plugins/globals/globals.py
```diff
@@ -111,12 +112,57 @@ def __init__(self, template) -> None:  # type: ignore[no-untyped-def]
...
+        :return dict: processed template globals
+        """
+        if not ignore_globals:
+            return self.template_globals[resource_type]
+
+        if isinstance(ignore_globals, str) and ignore_globals == "*":
+            return GlobalProperties({})
+
+        if isinstance(ignore_globals, list):
+            global_props: GlobalProperties = copy.deepcopy(self.template_globals[resource_type])
```
xazhao, line 139: Why do we need a deepcopy here?

GavinZZ: We need a copy because this attribute is resource level. We don't want to actually remove any item from the globals for other resources.

xazhao (approved):

tirumerla:
Hello @GavinZZ thank you for this feature we are wondering when this change will be rolled out, looking at the commit history looks like this is not part of the latest release?

GavinZZ:
@tirumerla This change should be rolling out by end of this week with the next PyPI release.

GavinZZ:
https://github.com/aws/serverless-application-model/releases/tag/v1.81.0
This change is live now!

PR #3392, opened 2023-10-24
Fix: Raise Error When UpdateReplacePolicy is not correctly defined.

ConnorRobertson:
### Description of changes
Raises error when UpdateReplacePolicy is not Retain, Snapshot, Delete or None.

Created a template to test and ran make pr.
### Checklist

    - [ ] Using correct values
    - [ ] Using wrong values

### Examples?

### DIFF
--- a/samtranslator/plugins/api/implicit_api_plugin.py
+++ b/samtranslator/plugins/api/implicit_api_plugin.py
@@ -499,7 +499,5 @@ class ImplicitApiPlugin(BasePlugin, Generic[T], metaclass=ABCMeta):
         api_dict_deletion = self.api_deletion_policies.setdefault(api_id, set())
         api_dict_deletion.add(deletion_policy)
 
-        api_dict_update_replace = self.api_update_replace_policies.setdefault(api_id, set())
-        if update_replace_policy not in ["Retain", "Snapshot", "Delete", None]:
-            raise ValueError("UpdateReplacePolicy must be equal to Retain, Snapshot, Delete, or must be None.")
-        api_dict_update_replace.add(update_replace_policy)
+        api_dict_update_replace = self.api_update_replace_policies.setdefault(api_id, list())
+        api_dict_update_replace.append(update_replace_policy)
--- a/samtranslator/plugins/api/implicit_api_plugin.py
+++ b/samtranslator/plugins/api/implicit_api_plugin.py
@@ -501,5 +501,5 @@ class ImplicitApiPlugin(BasePlugin, Generic[T], metaclass=ABCMeta):
 
         api_dict_update_replace = self.api_update_replace_policies.setdefault(api_id, set())
         if update_replace_policy not in ["Retain", "Snapshot", "Delete", None]:
-            raise ValueError("UpdateReplacePolicy must be equal to Retain, Snapshot, or Delete, or must be None.")
+            raise ValueError("UpdateReplacePolicy must be equal to Retain, Snapshot, Delete, or must be None.")
         api_dict_update_replace.add(update_replace_policy)
--- a/samtranslator/plugins/api/implicit_api_plugin.py
+++ b/samtranslator/plugins/api/implicit_api_plugin.py
@@ -500,4 +500,6 @@ class ImplicitApiPlugin(BasePlugin, Generic[T], metaclass=ABCMeta):
         api_dict_deletion.add(deletion_policy)
 
         api_dict_update_replace = self.api_update_replace_policies.setdefault(api_id, set())
+        if update_replace_policy not in ["Retain", "Snapshot", "Delete", None]:
+            raise ValueError("UpdateReplacePolicy must be equal to Retain, Snapshot, or Delete, or must be None.")
         api_dict_update_replace.add(update_replace_policy)
--- a/samtranslator/plugins/api/implicit_api_plugin.py
+++ b/samtranslator/plugins/api/implicit_api_plugin.py
@@ -496,8 +496,8 @@ class ImplicitApiPlugin(BasePlugin, Generic[T], metaclass=ABCMeta):
         method_conditions = api_dict_condition.setdefault(path, {})
         method_conditions[method] = condition
 
-        api_dict_deletion = self.api_deletion_policies.setdefault(api_id, set())
-        api_dict_deletion.add(deletion_policy)
+        api_dict_deletion = self.api_deletion_policies.setdefault(api_id, [])
+        api_dict_deletion.append(deletion_policy)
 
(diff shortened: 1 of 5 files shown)


paulhcsun (approved):

samtranslator/plugins/api/implicit_api_plugin.py
```diff
@@ -500,4 +500,6 @@ def _update_resource_attributes_from_api_event(  # noqa: PLR0913
         api_dict_deletion.add(deletion_policy)
 
         api_dict_update_replace = self.api_update_replace_policies.setdefault(api_id, set())
+        if update_replace_policy not in ["Retain", "Snapshot", "Delete", None]:
+            raise ValueError("UpdateReplacePolicy must be equal to Retain, Snapshot, Delete, or must be None.")
```
aaythapa, line 504: small error test for this would be nice

samtranslator/plugins/api/implicit_api_plugin.py
```diff
@@ -500,4 +500,6 @@ def _update_resource_attributes_from_api_event(  # noqa: PLR0913
         api_dict_deletion.add(deletion_policy)
 
         api_dict_update_replace = self.api_update_replace_policies.setdefault(api_id, set())
+        if update_replace_policy not in ["Retain", "Snapshot", "Delete", None]:
```
aaythapa, line 503: whats the usecase where people would set it to `None`?

ConnorRobertson: I'll see if I can remove that. I was thinking it might be necessary if update_replace_policy didn't exist.

GavinZZ (commented):
Why is this check in implicit api plugin?

samtranslator/plugins/api/implicit_api_plugin.py
ssenchenko: Then 
```python
if update_replace_policy and update_replace_policy not in ["Retain", "Snapshot", "Delete"]:
```
it sounds more idiomatic

ssenchenko: And also do we use those values "Retain", "Snapshot" etc? If so, let's make them enum or literal constants

GavinZZ: Also I vaguely remember that it’s possible to have intrinsic in DeletionPolicy, not sure if it’s the same for UpdateDeplacePolicy?

ConnorRobertson: If it isn't defined then it is None. So None has to be there.

ConnorRobertson:
> Why is this check in implicit api plugin?

This is where the error was failing do you have an idea where it should be.

samtranslator/plugins/api/implicit_api_plugin.py
ConnorRobertson: > Also I vaguely remember that it’s possible to have intrinsic in DeletionPolicy, not sure if it’s the same for UpdateDeplacePolicy?

Yeah I just checked. Some intrinsics seems supported as well as some pseudo-parameters.
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/function-refs-in-policy-attributes.html

GavinZZ:
> > Why is this check in implicit api plugin?
> 
> This is where the error was failing do you have an idea where it should be.

Have you tried with the same `UpdateReplacePolicy` on a SAM Function resource? Would it fail as well?

aaythapa:
> > > Why is this check in implicit api plugin?
> > 
> > 
> > This is where the error was failing do you have an idea where it should be.
> 
> Have you tried with the same `UpdateReplacePolicy` on a SAM Function resource? Would it fail as well?

piggy backing on this, i see [DeletionPolicy](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-deletionpolicy.html#aws-attribute-deletionpolicy-options) also has options, we should add an error check for that as well

ConnorRobertson:
> > > > Why is this check in implicit api plugin?
> > > 
> > > 
> > > This is where the error was failing do you have an idea where it should be.
> > 
> > 
> > Have you tried with the same `UpdateReplacePolicy` on a SAM Function resource? Would it fail as well?
> 
> piggy backing on this, i see [DeletionPolicy](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-deletionpolicy.html#aws-attribute-deletionpolicy-options) also has options, we should add an error check for that as well

Can confirm that DeletionPolicy would also give this availability dip I will cover that too!

ConnorRobertson:
> > > Why is this check in implicit api plugin?
> > 
> > 
> > This is where the error was failing do you have an idea where it should be.
> 
> Have you tried with the same `UpdateReplacePolicy` on a SAM Function resource? Would it fail as well?

When using other resources they seem to pass Transform but fails to deploy due to a Validation Error.

xazhao:
These are pass through props. Should we avoid validating and simply pass the value? Same as other resources.

aaythapa:
> These are pass through props. Should we avoid validating and simply pass the value? Same as other resources.

I think not validating is what caused the dip in the first place

xazhao:
I mean not validating the value. As long as it's a string, it will be fine.

GavinZZ:
Was discussing with Xia. We could choose either one of the following.
1. Try catch `TypeError` the [entire statement](https://github.com/aws/serverless-application-model/blob/develop/samtranslator/plugins/api/implicit_api_plugin.py#L502) and re-raise an invalid resource attribute exception to make it a customer exception, or
2. We can update the [code here](https://github.com/aws/serverless-application-model/blob/develop/samtranslator/plugins/api/implicit_api_plugin.py#L502) to use `list()` instead of `set()`. There's not much reason to use set here. Consider this template
```
Transform:
  - AWS::LanguageExtensions
  - AWS::Serverless-2016-10-31

Resources:
  MyFunction:
    UpdateReplacePolicy: Delete
    Properties:
      ......
      Events:
        None:
          Type: Api
          Properties:
            Method: get
            Path: /method
  MyFunction2:
    UpdateReplacePolicy: Delete
    Properties:
      ......
      Events:
        Hi:
          Type: Api
          Properties:
            Method: post
            Path: /method
```
If the `UpdateReplacePolicy` is different, either using `set()` or `list()` would cause the final implicit API resource to use the last `UpdateReplacePolicy` in set or list.
If the `UpdateReplacePolicy` is the same, `set()` has tiny benefit of iterating less items, but the result is the same.

GavinZZ (approved):

PR #3396, opened 2023-10-31
fix: LayerVersion cannot be the target of DependsOn due to being hashed during transformation

ConnorRobertson:
### Description of changes
Changed it so that DependsOn will resolve to the hashed logical id of a target that has a logical id that changed during transformation.

Added a transform test

### Checklist

    - [ ] Using correct values
    - [ ] Using wrong values

### Examples?

### DIFF
--- a/samtranslator/translator/translator.py
+++ b/samtranslator/translator/translator.py
@@ -238,6 +238,7 @@ class Translator:
             del template["Transform"]
 
         if len(self.document_errors) == 0:
+            template = traverse_template_depends_on(template, changed_logical_ids)
             template = intrinsics_resolver.resolve_sam_resource_id_refs(template, changed_logical_ids)
             return intrinsics_resolver.resolve_sam_resource_refs(template, supported_resource_refs)
         raise InvalidDocumentException(self.document_errors)
@@ -396,6 +397,81 @@ class Translator:
             del properties["SourceReference"]
 
         return SamConnector.from_dict(full_connector_logical_id, connector)
+
+
+def traverse_template_depends_on(input_value: Dict[str, Any], resolution_data: Dict[str, str]) -> Dict[str, Any]:
+    """
+    Driver method that performs the actual traversal of input and calls _resolve_depends_on when
+    needed to perform the resolution.
+    :param input_value: Any primitive type  (dict, array, string etc) whose value might contain an intrinsic function
+    :param resolution_data: Data that will help with resolution i.e. old and new logical ids after transformation
+    :return: Modified `input` with DependsOn resolved
+    """
+    if len(resolution_data) == 0:
+        return input_value
+
+        #
+        # Traversal Algorithm:
+        #
+        # Imagine the input dictionary/list as a tree. We are doing a Pre-Order tree traversal here where we first
+        # process the root node before going to its children. Dict and Lists are the only two iterable nodes.
+        # Everything else is a leaf node. More description can be found on line 126 in _traverse in resolver.py
+        #
+
+    _resolve_depends_on(input_value, resolution_data)
+    if isinstance(input_value, dict):
+        traverse_dict(input_value, resolution_data)
+    return input_value
+
+
+def _resolve_depends_on(input_dict: Dict[str, Any], resolution_data: dict[str, str]) -> Dict[str, Any]:
+    """
+    Resolve DependsOn when logical ids get changed when transforming (ex: AWS::Serverless::LayerVersion)
+
+    :param input_dict: Chunk of the template that is attempting to be resolved
+    :param resolution_data: Dictionary of the original and changed logical ids
+    :return: Modified dictionary with values resolved
+    """
+    # Checks if input dict is resolvable
+    if input_dict is None or not canHandle(input_dict=input_dict):
+        return input_dict
+    # Checks if DependsOn is valid
+    if not (isinstance(input_dict["DependsOn"], (list, str))):
+        return input_dict
+    # Check if DependsOn matches the original value of a changed_logical_id key
+    for old_logical_id, changed_logical_id in resolution_data.items():
+        # Done like this as there is no other way to know if this is a DependsOn vs some value named the
+        # same as the old logical id. (ex LayerName is commonly the old_logical_id)
+        if isinstance(input_dict["DependsOn"], list):
+            for index, value in enumerate(input_dict["DependsOn"]):
+                if value == old_logical_id:
+                    input_dict["DependsOn"][index] = changed_logical_id
+        elif input_dict["DependsOn"] == old_logical_id:
+            input_dict["DependsOn"] = changed_logical_id
+    return input_dict
+
+
+def traverse_dict(input_dict: Dict[str, Any], resolution_data: Dict[str, str]) -> Dict[str, Any]:
+    """
+    Traverse a dictionary to resolve intrinsic functions on every value
+
+    :param input_dict: Input dictionary to traverse
+    :param resolution_data: Data needed to resolve DependsOn
+    :return: Modified dictionary with values resolved
+    """
+    for key, value in input_dict.items():
+        input_dict[key] = traverse_template_depends_on(value, resolution_data)
+    return input_dict
+
+
+def canHandle(input_dict: Dict[str, Any]) -> bool:
+    """
+    Checks if the input dictionary is of length one and contains "DependsOn"
+
+    :param input_dict: the Dictionary that is attempting to be resolved
+    :return boolean value of validation attempt
+    """
+    return isinstance(input_dict, dict) and "DependsOn" in input_dict
 
 
 def prepare_plugins(plugins: Optional[List[BasePlugin]], parameters: Optional[Dict[str, Any]] = None) -> SamPlugins:
--- a/tests/translator/input/layer_version_depends_on.yaml
+++ b/tests/translator/input/layer_version_depends_on.yaml
@@ -1,0 +1,26 @@
+Transform: AWS::Serverless-2016-10-31
+Resources:
+  Layer1:
+    Type: AWS::Serverless::LayerVersion
+    Properties:
+      ContentUri:
+        Bucket: test
+        Key: test.zip
+
+  Layer2:
+    Type: AWS::Serverless::LayerVersion
+    DependsOn: Layer1
+    Properties:
+      ContentUri:
+        Bucket: test
+        Key: test.zip
+
+  Layer3:
+    Type: AWS::Serverless::LayerVersion
+    DependsOn:
+    - Layer1
+    - Layer2
+    Properties:
+      ContentUri:
+        Bucket: test
+        Key: test.zip
--- a/tests/translator/output/aws-cn/layer_version_depends_on.json
+++ b/tests/translator/output/aws-cn/layer_version_depends_on.json
@@ -1,0 +1,42 @@
+{
+  "Resources": {
+    "Layer1d45b36fd2d": {
+      "DeletionPolicy": "Retain",
+      "Properties": {
+        "Content": {
+          "S3Bucket": "test",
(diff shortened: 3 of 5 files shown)


samtranslator/translator/translator.py
```diff
@@ -398,6 +399,81 @@ def _get_generated_connector(
...
+    :param input_dict: Input dictionary to traverse
+    :param resolution_data: Data needed to resolve DependsOn
+    :return: Modified dictionary with values resolved
+    """
+    for key, value in input_dict.items():
+        input_dict[key] = traverse_template_depends_on(value, resolution_data)
+    return input_dict
+
+
+def canHandle(input_dict: Dict[str, Any]) -> bool:
```
ssenchenko, line 467: Not python naming style

samtranslator/translator/translator.py
```diff
@@ -398,6 +399,81 @@ def _get_generated_connector(
         return SamConnector.from_dict(full_connector_logical_id, connector)
 
 
+def traverse_template_depends_on(input_value: Dict[str, Any], resolution_data: Dict[str, str]) -> Dict[str, Any]:
```
ssenchenko, line 402: we already have a traversal  of resources graph, can we use an existing one and just add their what we need to do instead of traversing it again?

GavinZZ (commented):
I think there's no need to recursively traverse & update logical id since we only cares about `DependsOn` on every single resource in `Resources`.

ConnorRobertson:
> I think there's no need to recursively traverse & update logical id since we only cares about `DependsOn` on every single resource in `Resources`.

Slava and I discussed this and I think the better solution might be to make a more generic transversal class in case we have a future situation we need this specific thing.

GavinZZ:
That would also work, but make sure you don't update any content that you shouldn't update.

samtranslator/translator/translator.py
xazhao: +1 If we can get all things done in one traversal, that would reduce the overall latency.

samtranslator/translator/translator.py
```diff
@@ -398,6 +399,81 @@ def _get_generated_connector(
...
+    """
+    Driver method that performs the actual traversal of input and calls _resolve_depends_on when
+    needed to perform the resolution.
+    :param input_value: Any primitive type  (dict, array, string etc) whose value might contain an intrinsic function
```
xazhao, line 406: Isn't `input_data` a dict already?

samtranslator/translator/translator.py
ConnorRobertson: We can't because the intrinsic resolve is it's own class.

samtranslator/utils/traverse.py
```diff
@@ -0,0 +1,142 @@
+from typing import Any, Callable, Dict, List
+
+
+class Traverse:
```
ssenchenko, line 4: Let's decouple traversal algorithm from action performed while traversing, in that way it would be easier to add more actions if we need them in the future

samtranslator/utils/traverse.py
```diff
@@ -0,0 +1,142 @@
...
+        :param resolver_method_name: Name of the method that will be called to actually resolve an the function. This method
+            is called with the parameters `(input, resolution_data)`.
+        :return: Modified `input` with values resolved
+        """
+        if resolver_method_name == "DependsOn":
+            return self._traverse(input_value, resolution_data, self._resolve_depends_on)
+
+        return input_value
+
+    # here is where I would put constant variables
```
xazhao, line 25: What do you mean here?

samtranslator/utils/traverse.py
```diff
@@ -0,0 +1,142 @@
...
+    def _traverse(
```
xazhao, line 26: These functions are same as IntrinsicResolver. Would be good to share these functions between 2 classes.

samtranslator/utils/traverse.py
GavinZZ, line 4: I think ideally we would have an abstract `Resolver` class that has the generic traversing logics. Then we could have `IntrinsicsResolver` that inherits `Resolver` class. We would also have a `DependsOnResolver` class that inherids `Resolver` class.

samtranslator/utils/traverse.py
```diff
@@ -0,0 +1,142 @@
...
+        Wraps the Driver Method which allows the resolver method to be changed as needed
+
+        :param input_value: Any primitive type  (dict, array, string etc) whose value might contain a changed value
+        :param resolution_data: Data that will help with resolution. For example, changed logical ids from hashing
+        LayerVersion
+        :param resolver_method_name: Name of the method that will be called to actually resolve an the function. This method
+            is called with the parameters `(input, resolution_data)`.
+        :return: Modified `input` with values resolved
+        """
+        if resolver_method_name == "DependsOn":
```
GavinZZ, line 20: Instead of if-else condition in `traverse-wrapper` I think using an abstract class + inheritance makes more sense here. We could do this as another refactor in the future.

samtranslator/utils/resolve_actions.py
```diff
@@ -0,0 +1,58 @@
+from abc import ABC, abstractmethod
+from typing import Any, Dict
+
+
+class ResolveAction(ABC):
```
ssenchenko, line 5: It's a very general class, let's keep `Resolve` out of name in case we won't need it in the future

ssenchenko: And same stays for the file name.

samtranslator/utils/resolve_actions.py
```diff
@@ -0,0 +1,58 @@
...
+
+        :param input_dict: Chunk of the template that is attempting to be resolved
+        :param resolution_data: Dictionary of the original and changed logical ids
+        :return: Modified dictionary with values resolved
+        """
+        # Checks if input dict is resolvable
+        if template is None or not self._can_handle_depends_on(input_dict=template):
+            return template
+        # Checks if DependsOn is valid
+        if not (isinstance(template["DependsOn"], (list, str))):
```
ssenchenko, line 37: Based on the number of times you have to use `"DependsOn"` it's better to make a class property

samtranslator/utils/traverse.py
```diff
@@ -0,0 +1,104 @@
...
+
+from samtranslator.utils.resolve_actions import ResolveAction
+
+
+def traverse(template: Dict[str, Any], actions: List[ResolveAction]) -> Any:
+    """
+    Calls the correct traversal method
+
+    :param template: The template that needs modifying
+    :param actions: That actions that need to be performed to the template
```
gracelu0, line 11: nit: `The` actions?

samtranslator/utils/traverse.py
```diff
@@ -0,0 +1,104 @@
...
+
+def _execute_actions(template: Dict[str, Any], actions: List[ResolveAction]) -> Any:
+    """
+    Calls the resolver methods when they are in the list of ResolveActions. Then calls the actions passed to traverse.
+
+    :param template: The template that needs modifying
+    :param actions: actions to be executed.
+    :return: Modified Template
+    """
+    if len(actions) == 0:
```
ssenchenko, line 27: no need in this, the loop below will not run if array is empty

samtranslator/utils/traverse.py
```diff
@@ -0,0 +1,104 @@
...
+    :param template: The template that needs modifying
+    :param actions: That actions that need to be performed to the template
+    :return: Modified Template
+
+    TODO: Add multiple strategies and choose between those with an Optional value
+    """
+    return _preorder_traverse(template, actions)
+
+
+def _execute_actions(template: Dict[str, Any], actions: List[ResolveAction]) -> Any:
```
ssenchenko, line 19: not sure if need this function, seems like a single loop

samtranslator/utils/traverse.py
```diff
@@ -0,0 +1,104 @@
...
+    """
+    Calls the correct traversal method
+
+    :param template: The template that needs modifying
+    :param actions: That actions that need to be performed to the template
+    :return: Modified Template
+
+    TODO: Add multiple strategies and choose between those with an Optional value
+    """
+    return _preorder_traverse(template, actions)
```
ssenchenko, line 16: There is no need to have separate function for pre-order.
Since we have > 2 leaves, in-order is not possible and all we possibly can have is pre-order and post-order. The difference is only in applying actions before traversing children or after. It can be implemented by have 2 different actions parameters `preorder_actions` and `postorder_actions` instead of `actions`. But I suppose that postorder would require some data from children, otherwise, why would we care about applying actions post-factum. Since we don't  have a use case now, instead of getting too abstract, lets implement preorder only, right in the `travers()` function

samtranslator/utils/traverse.py
```diff
@@ -0,0 +1,104 @@
...
+    Driver method that performs the actual traversal of input and calls the appropriate `resolver_method` when
+    to perform the resolution.
+
+    :param input_value: Any primitive type  (dict, array, string etc) whose value might contain a changed value
+    :param actions: Method that will be called to actually resolve the function.
+    :return: Modified `input` with values resolved
+    """
+
+    #
+    # Traversal Algorithm:
```
ssenchenko, line 50: This should go into doc string. And also we should speak about traversing in general instead of resolving

samtranslator/utils/resolve_actions.py
ConnorRobertson: What should the name be? Ideally I would use Actions but that's already used for intrinsic functions and I don't want overlap.

GavinZZ: shouldn't the traverse logic in this abstract class?

samtranslator/utils/traverse.py
ConnorRobertson: My idea with this here was it would make it easier to implement with separate strategies .

samtranslator/utils/traverse.py
```diff
@@ -0,0 +1,104 @@
...
+    if len(actions) == 0:
+        return template
+
+    for action in actions:
+        action.execute(template)
+
+    return template
+
+
+def _preorder_traverse(
```
xazhao, line 36: this and `_traverse_dict/_traverse_list` could be in abstract class so that they can be reused.

ConnorRobertson: Will leave to a later refactor due to differences in input values.

ssenchenko (approved):

paulhcsun (approved):
small update to the traverse function description that we discussed, otherwise lgtm

samtranslator/utils/actions.py
```diff
@@ -0,0 +1,60 @@
...
+        return template
+
+    def _can_handle_depends_on(self, input_dict: Dict[str, Any]) -> bool:
+        """
+        Checks if the input dictionary is of length one and contains "DependsOn"
+
+        :param input_dict: the Dictionary that is attempting to be resolved
+        :return boolean value of validation attempt
+        """
+        return isinstance(input_dict, dict) and self.DependsOn in input_dict
```
xazhao, line 60: why do we need to validate `input_dict` is a dict?

GavinZZ (approved):

samtranslator/utils/actions.py
```diff
@@ -0,0 +1,60 @@
...
+        :return: Modified dictionary with values resolved
+        """
+        # Checks if input dict is resolvable
+        if template is None or not self._can_handle_depends_on(input_dict=template):
+            return template
+        # Checks if DependsOn is valid
+        if not (isinstance(template[self.DependsOn], (list, str))):
+            return template
+        # Check if DependsOn matches the original value of a changed_logical_id key
+        for old_logical_id, changed_logical_id in self.resolution_data.items():
```
xazhao, line 42: Instead of looping through the dict, we can fetch the value directly from the dict.

Issue #3426, opened 2023-11-17
Bug: AWS SAM tag value on AWS::Serverless::Function empty when value is False


vandem9:
### Description:
When setting tags on AWS::Serverless::Function and the value of the tag is False instead of "False" it becomes empty on the deployed resource. This is tested with the tags defined on Globals.

### Steps to reproduce:
Create a SAM template with a serverless function, provide it with tags on the globals object (not tested with tags on the resource itself). When you provide a tag with the value False and not "False" it becomes empty.

![Screenshot 2023-11-17 at 15 29 48](https://github.com/aws/aws-sam-cli/assets/38071372/232b0047-b9b5-4cef-a80f-28ca95907a7a)
![Screenshot 2023-11-17 at 15 29 05](https://github.com/aws/aws-sam-cli/assets/38071372/8a52a5c9-7de1-4cdb-8bee-af4f5bb44625)

ConnorRobertson:
The PR has been merged so closing this issue, please let us know if this issue persists.

PR #3429, opened 2023-11-22
fix: Tags with value of False got evaluated to ""

ConnorRobertson:
### Issue #, if available
https://github.com/aws/serverless-application-model/issues/3426
### Description of changes
When the value of a tag was False it would be set to the default "" value as the if statement evaluated the value rather than checking if it was None.

Tested by transforming template and looking at tags

### Checklist

    - [ ] Using correct values
    - [ ] Using wrong values

### Examples?

### DIFF
--- a/tests/translator/input/simple_function_with_tags_false.yaml
+++ b/tests/translator/input/simple_function_with_tags_false.yaml
@@ -1,0 +1,24 @@
+Resources:
+  MyLambdaFunction:
+    Type: AWS::Serverless::Function
+    Properties:
+      Handler: index.handler
+      Runtime: nodejs18.x
+      InlineCode: |
+        exports.handler = async (event, context, callback) => {
+          return {
+            statusCode: 200,
+            body: 'Success'
+          }
+        }
+      MemorySize: 128
+      Policies:
+      - AWSLambdaRole
+      - AmazonS3ReadOnlyAccess
+      Tags:
+        TagKey1: TagValue1
+        TagKey2: ''
+        TagKey3: false
+        TagKey4: true
+Metadata:
+  SamTransformTest: true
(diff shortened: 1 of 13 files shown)


tests/translator/input/simple_function_with_tags_false.yaml
```diff
@@ -0,0 +1,24 @@
...
+        }
+      MemorySize: 128
+      Policies:
+      - AWSLambdaRole
+      - AmazonS3ReadOnlyAccess
+      Tags:
+        TagKey1: TagValue1
+        TagKey2: ''
+        TagKey3: false
+        TagKey4: true
```
aaythapa, line 22: `TagKey5: 0` missing here?

moelasmar (approved):

tests/translator/input/simple_function_with_tags_false.yaml
```diff
@@ -0,0 +1,24 @@
+Resources:
+  MyLambdaFunction:
```
aaythapa, line 2: Can you add some other resources besides SAM function in this tests as well? Just to be thorough

aaythapa (commented):
I remember we had the `PropogateTags` property for some resources, will this error pop up for that property?

ConnorRobertson:
> PropogateTags

This is a good question. I will investigate.

ConnorRobertson:
This same issue would have appeared in `PropagateTags` as well but this fix as is fixes that.

tests/translator/input/simple_function_with_tags_false.yaml
ConnorRobertson: Globals and in function seem to tag using the same function.

tests/translator/input/simple_function_with_tags_false.yaml
ConnorRobertson: Table is here to test another case but I think all the resources refer to the same code when tagging.

aaythapa (approved):

PR #3470, opened 2023-12-13
fix: clearer error when a property that modifies an OpenApiDefinition is set and a DefinitionBody is defined

ConnorRobertson:
### Issue #, if available
When a property that modifies the OpenApi Definition is set like `DisableExecuteApiEndpoint` and a `DefinitionBody` is defined outside of the template an error is created because we can't tell if it is a valid OpenApi definition.
Previously the error was really nebulous if you had an unchanged `DefinitionBody` and added a property like `DisableExecuteApiEndpoint` you would get an error "Invalid Swagger Document." which doesn't make sense from that perspective.
### Description of changes
Add information to help make an error easy to understand in the above described situation.
### Checklist

    - [ ] Using correct values
    - [ ] Using wrong values

### Examples?


gracelu0 (approved):

samtranslator/swagger/swagger.py
```diff
@@ -62,7 +62,13 @@ def __init__(self, doc: Optional[Dict[str, Any]]) -> None:
         """
 
         if not doc or not SwaggerEditor.is_valid(doc):
-            raise InvalidDocumentException([InvalidTemplateException("Invalid Swagger document")])
+            raise InvalidDocumentException(
+                [
+                    InvalidTemplateException(
+                        "Invalid Swagger document. May be caused by having the OpenApi definition defined outside of the template."
```
GavinZZ, line 68:
```suggestion
                        "Invalid Swagger document or the Swagger document is not explicitly defined in `DefinitionBody`."
```

aaythapa (approved):

ssenchenko (approved):

GavinZZ (approved):

PR #3522, opened 2024-02-06
chore: Fix KeyError when accessing PolicyDocument

jfuss:
If an intrinsic is within the Function.Policies Property, the `d["PolicyDocument"]` will fail because the valid policy is backed within the `FN:If`. Instead, we access `PolicyDocument` through a `.get()` call, which allows us to handle the intrinsic case.

### Issue #, if available
None

### Description of changes
See above

Wrote a test to validate the template is transformable.

### Checklist

    - [x] Using correct values
    - [x] Using wrong values

### Examples?

### DIFF
--- a/tests/translator/input/function_with_event_dest_basic.yaml
+++ b/tests/translator/input/function_with_event_dest_basic.yaml
@@ -39,14 +39,14 @@ Resources:
       Runtime: nodejs12.x
       MemorySize: 1024
       Policies:
-        - Fn::If:
-          - SomeCondition
-          - Version: '2012-10-17'
-            Statement:
-            - Effect: Deny
-              Action:
-              - logs:CreateLogGroup
-              - logs:CreateLogStream
-              - logs:PutLogEvents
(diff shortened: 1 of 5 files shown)


gracelu0 (approved):

tests/translator/input/function_with_event_dest_basic.yaml
ssenchenko, line 1: Unless a test is conceptually wrong we don't modify an existing test to ensure that current behaviour is still supported. 
It'd be clearer to add a new transform case instead.

jfuss: It's not conceptually wrong but also doesn't mutate the intent of the example either.

ssenchenko: I guess you're right. It's the test name which is confusing.
However adding a new test would save us time looking into intent of the existing test. Easier to review.

ssenchenko (approved):

PR #3551, opened 2024-02-27
fix: normalize the open api version

xazhao:
### Issue #, if available
In Serverless Api, if `OpenApiVersion` is different from `openapi` value in DefinitionBody, SAM will missconfigure the updated openApi definition body.

### Description of changes
When `openapi` is in the DefinitionBody, we use that value instead of `OpenApiVersion` property in the resource.

Transform tests covering both cases.

### Checklist

    - [x] Using correct values
    - [x] Using wrong values

### Examples?


tests/translator/input/api_open_api_version_match.yaml
```diff
@@ -0,0 +1,457 @@
+Transform:
```
GavinZZ, line 1: Why do we use such a complex template to test? Should we use a simple and short template so that we can easily know what we're trying to test here?

samtranslator/model/api/api_generator.py
```diff
@@ -1128,6 +1128,15 @@ def _openapi_postprocess(self, definition_body: Dict[str, Any]) -> Dict[str, Any
         if definition_body.get("openapi") is not None and self.open_api_version is None:
             self.open_api_version = definition_body.get("openapi")
 
+        if (
+            definition_body.get("openapi") is not None
+            and self.open_api_version is not None
+            and definition_body.get("openapi") != self.open_api_version
```
iali17, line 1134: Checking the major version might be better as major versions should allow for the template to be parsed properly. This would make it so `3.0.1` and `3.0.0` would not match and that might be too strict for this use case.

tests/translator/input/api_open_api_version_match.yaml
xazhao: It's a template when I reproduced the issue. I just deleted some unnecessary resources in the template.

samtranslator/model/api/api_generator.py
xazhao: Just decided to use a different approach to address the issue.

samtranslator/model/api/api_generator.py
```diff
@@ -400,7 +399,7 @@ def _construct_deployment(self, rest_api: ApiGatewayRestApi) -> ApiGatewayDeploy
             self.logical_id + "Deployment", attributes=self.passthrough_resource_attributes
         )
         deployment.RestApiId = rest_api.get_runtime_attr("rest_api_id")
-        if not self.remove_extra_stage:
```
GavinZZ, line 403: Mind want to add a comment here to explain why we add a stage when open api version is not specified. https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-resource-api.html#sam-api-openapiversion

GavinZZ (commented):
I don't see how we handle the scenario when open api version mismatches between definition body and api property.

xazhao:
> I don't see how we handle the scenario when open api version mismatches between definition body and api property.

Basically when `openapi` exists in definition body, we use that value and ignore `OpenApiVersion` property with the following code:
```
normalized_open_api_version = definition_body.get("openapi", self.open_api_version)
```

jfuss (approved):

samtranslator/model/api/api_generator.py
xazhao: I'm not sure why we made this change because it's 5 years ago. But I will look into that and add later.

PR #3574, opened 2024-04-03
fix: add new partition to arn generation

aaythapa:
### Description of changes

Added the `${AWS::Partition}` psuedo parameter before so that we didn't have to maintain a static list but there is a lot of code in SAM that relies on that static list and its causing tests in the new region to fail.

We can use `${AWS::Partition}` in the future but would require a bigger code change. To unblock us in the new region we can just add the new region's partition into the static list for now.

### Checklist

    - [ ] Using correct values
    - [ ] Using wrong values

### Examples?

### DIFF
--- a/samtranslator/translator/arn_generator.py
+++ b/samtranslator/translator/arn_generator.py
@@ -29,6 +29,10 @@ def _region_to_partition(region: str) -> str:
         if region_string.startswith(key):
             return value
 
+    # Using the ${AWS::Partition} placeholder so that we don't have to add new regions to the static list above
+    if "iso" in region_string:
+        return "${AWS::Partition}"
+
     return "aws"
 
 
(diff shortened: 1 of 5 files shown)


samtranslator/translator/arn_generator.py
```diff
@@ -23,6 +23,7 @@ def _region_to_partition(region: str) -> str:
         "us-isob": "aws-iso-b",
         "us-gov": "aws-us-gov",
         "eu-isoe": "aws-iso-e",
+        "us-isof": "aws-iso-f",
```
paulhcsun, line 26: not really sure how it works but do we need to remove the use of `${AWS::Partition}` in this file too?

aaythapa: Yea you're right, I'll remove it

GavinZZ (approved):

paulhcsun (approved):

PR #3613, opened 2024-06-12
fix: Remove ENI manual deletion

jfuss:
### Issue #, if available
None

### Description of changes
Removes the code that deletes ENI. This was added because MQ was not deleting the ENIs which is not happening.

`make pr`

### Checklist

    - [ ] Using correct values
    - [ ] Using wrong values

### Examples?

### DIFF
--- a/integration/conftest.py
+++ b/integration/conftest.py
@@ -1,5 +1,4 @@
 import logging
-import time
 from pathlib import Path
 
 import boto3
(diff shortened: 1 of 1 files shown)


aaythapa:
I see from the old commit this was originally added because
>Number of unused Network Interfaces can accumulate and fail future test runs when there’s no more available IP addresses in the pre-created subnets.

will we see the same errors after removing it?

jfuss:
@aaythapa 

> will we see the same errors after removing it?

We shouldn't. MQ reached out because they are attempting to delete ENIs that no longer exist, and the running assumption is it is caused by this.

godwingrs22 (approved):

aaythapa (approved):

PR #3614, opened 2024-06-17
fix: change endpoint configuration for state machine integ test

aaythapa:
### Issue #, if available

N/A

### Description of changes

We've seen this test fail in a couple of regions with the error `Request to ... failed with status: 403, expected status: 200`. This is also a known flaky test which we've seen fail before and succeed after a simple retry.

After investigation we concluded that the error is likely because the endpoint configuration is automatically set to EDGE which routes requests to a CloudFront Point of Presence. We suspect the test is calling the endpoint right after the stack is deployed but before the routing can take place. To mitigate this we set the endpoint configuration to REGIONAL for now.

Ran the test locally, it deployed and everything passed as before.

### Checklist

    - [ ] Using correct values
    - [ ] Using wrong values

### Examples?

### DIFF
--- a/integration/helpers/base_test.py
+++ b/integration/helpers/base_test.py
@@ -567,9 +567,9 @@ class BaseTest(TestCase):
             )
         return response
 
-    def verify_post_request(self, url: str, body_obj, expected_status_code: int):
+    def verify_post_request(self, url: str, body_obj, expected_status_code: int, headers=None):
         """Return response to POST request and verify matches expected status code."""
-        response = self.do_post_request(url, body_obj)
+        response = self.do_post_request_with_logging(url, body_obj, headers)
         if response.status_code != expected_status_code:
             raise StatusCodeError(
                 f"Request to {url} failed with status: {response.status_code}, expected status: {expected_status_code}"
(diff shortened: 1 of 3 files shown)


GavinZZ (approved):

paulhcsun (approved):

integration/helpers/base_test.py
```diff
@@ -567,9 +567,9 @@ def verify_options_request(self, url, expected_status_code, headers=None):
             )
         return response
 
-    def verify_post_request(self, url: str, body_obj, expected_status_code: int):
+    def verify_post_request(self, url: str, body_obj, expected_status_code: int, headers=None):
```
godwingrs22, line 570: I think we need to pass the headers [here](https://github.com/aws/serverless-application-model/blob/20643062084d3a1d3b3c7e626687bccaf9650a64/integration/single/test_basic_api.py#L128) to the calling method. If not it will be empty i think

godwingrs22 (commented):
Small comment related to logging the headers

integration/helpers/base_test.py
aaythapa: My understanding is [this part](https://github.com/aws/serverless-application-model/blob/f19a827ef4444e1a4e611b05d7f664b8f756daf7/integration/helpers/base_test.py#L656) of the code (whose class is defined [here](https://github.com/aws/serverless-application-model/blob/f19a827ef4444e1a4e611b05d7f664b8f756daf7/integration/helpers/request_utils.py#L13)) will match the headers outputted by the response to the headers defined in amazon headers method. For example if there is an error the `x-amzn-errortype` header will be set. Do you know of any other headers we should be passing in here?

godwingrs22: Thanks @aaythapa. Seems like the headers passed to the method are request headers which will be used while making the api call and not for logging. The headers we log are from the api response and these headers which we already defined [here](https://github.com/aws/serverless-application-model/blob/f19a827ef4444e1a4e611b05d7f664b8f756daf7/integration/helpers/request_utils.py#L4-L10) are good. 
May be it is good to rename `headers` variable to `requestHeaders` so it will be more clear. Its nonblocking and good to update in next PR. 

LGTM. Thanks @aaythapa for the fix.

godwingrs22 (approved):

PR #3615, opened 2024-06-17
fix: prevent redundant tags being added to AWS::ApiGatewayV2::Api

gracelu0:
### Issue #, if available

Currently if a user sets `PropagateTags` to True in the "AWS::Serverless::HttpApi" resource and specifies the DefinitionBody property with an OpenApi definition, they will encounter the following deploy-time error:
```
Redundant fields [tags={"TagKey":"TagValue","httpapi:createdBy":"SAM"}] provided when either Body or BodyS3Location is not empty
```

CloudFormation does not allow both the `Tags` property being set and [x-amazon-apigateway-tag-value](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-openapi-extensions-x-amazon-apigateway-tag-value.html) being set in the `Body` property.

### Description of changes
Update `assign_tags()` method for `ApiGatewayV2HttpApi` class to return (no-op). 

For context, `Tags` can only be defined when `DefinitionBody` is specified or else we raise an [error](https://github.com/aws/serverless-application-model/blob/develop/samtranslator/model/api/http_api_generator.py#L499). Also, `DefinitionBody` (OpenApi definition) will always be defined because we have a [plugin](https://github.com/aws/serverless-application-model/blob/develop/samtranslator/plugins/api/default_definition_body_plugin.py#L37) that generates a skeleton OpenApi definition before the transform if the user does not define one. Additionally `Tags` always gets [added](https://github.com/aws/serverless-application-model/blob/develop/samtranslator/model/api/http_api_generator.py#L522) to the `DefinitionBody` (see [x-amazon-apigateway-tag-value property](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-openapi-extensions-x-amazon-apigateway-tag-value.html)). 

For AWS::ApiGatewayV2::Api resource, CloudFormation does not allow `Tags` to be defined if the `Body` property is specified. SAM ensures `Body` will always be defined. Therefore no need to add `Tags` property (doing so will only cause the error specified above).

Added transform tests and updated existing transform tests which would fail on deployment currently. Ran `make test` and all tests pass.

Also reproduced error and deploying the updated template succeeds without errors

### Checklist

    - [ ] Using correct values
    - [ ] Using wrong values

### Examples?

### DIFF
--- a/samtranslator/model/apigatewayv2.py
+++ b/samtranslator/model/apigatewayv2.py
@@ -25,7 +25,6 @@ class ApiGatewayV2HttpApi(Resource):
     }
 
     runtime_attrs = {"http_api_id": lambda self: ref(self.logical_id)}
-    Tags: Optional[PassThrough]
 
     def assign_tags(self, tags: Dict[str, Any]) -> None:
         """Overriding default 'assign_tags' function in Resource class
@@ -34,8 +33,8 @@ class ApiGatewayV2HttpApi(Resource):
         :param tags: Tags to be assigned to the resource
 
         """
-        if tags is not None and "Tags" in self.property_types and "Body" not in self.property_types:
-            self.Tags = tags
+        # Tags are already defined in Body so they do not need to be assigned here
+        return
 
 
 class ApiGatewayV2Stage(Resource):
--- a/samtranslator/model/apigatewayv2.py
+++ b/samtranslator/model/apigatewayv2.py
@@ -34,7 +34,7 @@ class ApiGatewayV2HttpApi(Resource):
         :param tags: Tags to be assigned to the resource
 
         """
-        if tags is not None and "Tags" in self.property_types:
+        if tags is not None and "Tags" in self.property_types and "Body" not in self.property_types:
             self.Tags = tags
 
 
--- a/tests/translator/output/aws-cn/http_api_openapi_with_propagate_tags.json
+++ b/tests/translator/output/aws-cn/http_api_openapi_with_propagate_tags.json
@@ -1,0 +1,112 @@
+{
+  "AWSTemplateFormatVersion": "2010-09-09",
+  "Resources": {
+    "MyHttpApi": {
+      "Properties": {
+        "Body": {
+          "info": {
+            "title": "My API",
+            "version": "1.0.0"
+          },
+          "openapi": "3.0.1",
+          "paths": {
+            "/": {
+              "get": {
+                "responses": {
+                  "200": {
+                    "description": "OK"
+                  }
+                },
+                "x-amazon-apigateway-integration": {
+                  "httpMethod": "POST",
+                  "type": "aws_proxy",
+                  "uri": {
+                    "Fn::Sub": "arn:aws:apigateway:${AWS::Region}:lambda:path/2015-03-31/functions/${MyLambdaFunction.Arn}/invocations"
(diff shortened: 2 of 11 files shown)


tests/translator/output/aws-cn/http_api_openapi_with_propagate_tags.json
```diff
@@ -0,0 +1,112 @@
+{
+  "AWSTemplateFormatVersion": "2010-09-09",
+  "Resources": {
+    "MyHttpApi": {
```
aaythapa, line 4: For my own understanding, before the `Tags` property for this resource was being set which caused the deployment error because `x-amazon-apigateway-tag-value` is also set in the Body?

aaythapa (approved):

tests/translator/output/aws-cn/http_api_openapi_with_propagate_tags.json
gracelu0: Yep exactly. We already add the tags to the OpenApi definition (`Body` property) [here](https://github.com/gracelu0/serverless-application-model/blob/develop/samtranslator/model/api/http_api_generator.py#L522) when constructing the resource. We were additionally adding the `Tags` property to `AWS::ApiGatewayV2::Api` when `PropagateTags` is set to `True` which is what caused the error.

samtranslator/model/apigatewayv2.py
```diff
@@ -34,7 +34,7 @@ def assign_tags(self, tags: Dict[str, Any]) -> None:
         :param tags: Tags to be assigned to the resource
 
         """
-        if tags is not None and "Tags" in self.property_types:
+        if tags is not None and "Tags" in self.property_types and "Body" not in self.property_types:
```
GavinZZ, line 37: `Body` is always in the property_types, so this is essentially do nothing.

GavinZZ (changes requested):

samtranslator/model/apigatewayv2.py
GavinZZ: I also notice my code `"Tags" in self.property_types` is a useless check too.

gracelu0: Hmm that is a good point, I kind of misunderstood/misread while debugging. Just dug around the code a bit more and it looks like `Tags` can only be defined when `DefinitionBody` is specified or else we raise an [error](https://github.com/gracelu0/serverless-application-model/blob/develop/samtranslator/model/api/http_api_generator.py#L499). Also, `DefinitionBody` will always be defined because we have a [plugin](https://github.com/gracelu0/serverless-application-model/blob/develop/samtranslator/plugins/api/default_definition_body_plugin.py#L37) that generates a skeleton OpenApi definition before the transform. Additionally `Tags` always gets [added](https://github.com/gracelu0/serverless-application-model/blob/develop/samtranslator/model/api/http_api_generator.py#L522) to the `DefinitionBody`. 

So it seems like there is no case where we would need to add `Tags` at all which means `assign_tags()` in ApiGatewayV2 should just do nothing. Does that make sense?

GavinZZ: I think that makes sense.

GavinZZ (approved):
Noice

Issue #3617, opened 2024-06-20
Several tests fail with Pydantic 1.10.17 (latest 1.x)


musicinmybrain:
### Description

After I reported https://github.com/aws/serverless-application-model/issues/3600 about Pydantic 1.10.15, the release of Pydantic 1.10.16 fixed the problem – but then 1.10.17 broke it again.

### Steps to reproduce

```
$ gh repo clone aws/serverless-application-model
$ cd serverless-application-model
$ python3 -m venv _e
$ . _e/bin/activate
(_e) $ pip install -e .[dev]
(_e) $ pip install pydantic==1.10.17
(_e) $ make test
```

Interestingly, I can also reproduce this on the `develop` branch, which includes https://github.com/aws/serverless-application-model/pull/3611 – using either `pydantic==1.10.15` or `pydantic==1.10.17` – so it doesn’t look like that PR works as intended.

### Observed result

```
=========================================================================================== short test summary info ============================================================================================
FAILED tests/model/test_resource_validator.py::TestResourceValidatorFailure::test_connector_with_empty_properties - AttributeError: module 'pydantic.v1' has no attribute 'error_wrappers'
FAILED tests/model/test_resource_validator.py::TestResourceValidatorFailure::test_connector_with_invalid_permission - AttributeError: module 'pydantic.v1' has no attribute 'error_wrappers'
FAILED tests/model/test_resource_validator.py::TestResourceValidatorFailure::test_connector_with_invalid_permission_type - AttributeError: module 'pydantic.v1' has no attribute 'error_wrappers'
FAILED tests/model/test_resource_validator.py::TestResourceValidatorFailure::test_connector_without_source - AttributeError: module 'pydantic.v1' has no attribute 'error_wrappers'
FAILED tests/translator/test_translator.py::test_transform_invalid_document[error_graphqlapi] - AttributeError: module 'pydantic.v1' has no attribute 'error_wrappers'
============================================================================== 5 failed, 3967 passed, 5 rerun in 99.69s (0:01:39) ==============================================================================
```

### Expected result

```
(_e) $ pip install pydantic==1.10.16
(_e) $ make test
======================================================================================= 3972 passed in 103.12s (0:01:43) =======================================================================================
```

### Additional environment details

1. OS: Fedora 40 `x86_64`
2. If using the [SAM CLI](https://github.com/aws/aws-sam-cli), `sam --version`: N/A
3. AWS region: N/A

aaythapa:
[...]

PR #3618, opened 2024-06-24
fix: new pydantic version causing errors with unit tests

aaythapa:
### Issue #, if available

#3617 

### Description of changes

pydantic's new version has an issue where the error_wrappers attribute isn't included so the exception can't be caught, this causes a couple of tests to fail. Catching all exceptions for now until pydantic fixes their issue (opened a ticket with them here: https://github.com/pydantic/pydantic/issues/9742).

This PR should be reverted once pydantic releases a fix.

`make pr` passed after the change

### Checklist

    - [x] Using correct values
    - [ ] Using wrong values

### Examples?


samtranslator/model/__init__.py
```diff
@@ -345,15 +346,15 @@ def validate_properties_and_return_model(self, cls: Type[RT]) -> RT:
...
-            ) from e
+        # TODO there is an issue on pydantic's side where we get an "module 'pydantic.v1' has no attribute 'error_wrappers'"
+        # Once pydantic fixes their issue we should change the bottom line to catch pydantic.error_wrappers.ValidationError instead of all exceptions
+        # except pydantic.error_wrappers.ValidationError as e:
+        #     error_properties: str = ""
+        #     with suppress(KeyError):
+        #         error_properties = ".".join(str(x) for x in e.errors()[0]["loc"])
+        #     raise InvalidResourceException(self.logical_id, f"Property '{error_properties}' is invalid.") from e
+        except Exception as e:
+            raise InvalidResourceException(self.logical_id, "An unexpected error occurred during validation.") from e
```
paulhcsun, line 357: Is there a better error message we can be using here?

aaythapa, line 357: The diff looks confusing but just commented out everything related to `pydantic.error_wrappers.ValidationError` and left a todo. Plus, caught all `Exception` instead of catching `AttributeError` because `AttributeError` was not being thrown by whats in the try block but by the `except pydantic.error_wrappers.ValidationError as e:`

aaythapa: made it general since its catching all exceptions and its temporary, any suggestions for a better msg?

paulhcsun: Ohh I see, just read the other comment. In that case it makes sense for a catch all, but I think we could add an extra detail to make it a bit more specific like `An unexpected error occurred when validating resource properties.`

tests/translator/output/error_graphqlapi.json
```diff
@@ -3,11 +3,11 @@
...
     "Number of errors found: 23. ",
     "Resource with id [AdditionalAuthTypeLambdaAuthorizerWithNoConfig] is invalid. ",
-    "Property 'Auth.Additional' is invalid. ",
+    "An unexpected error occurred during validation. ",
     "Resource with id [AdditionalAuthTypeOpenIDWithNoConfig] is invalid. ",
-    "Property 'Auth.Additional' is invalid. ",
+    "An unexpected error occurred during validation. ",
     "Resource with id [AdditionalAuthUserPoolWithNoConfig] is invalid. ",
-    "Property 'Auth.Additional' is invalid. ",
+    "An unexpected error occurred during validation. ",
```
GavinZZ, line 10: This doesn't seem ideal. :pensive:

samtranslator/model/__init__.py
```diff
@@ -345,15 +346,15 @@ def validate_properties_and_return_model(self, cls: Type[RT]) -> RT:
...
             return cls.parse_obj(self._generate_resource_dict()["Properties"])
-        except pydantic.error_wrappers.ValidationError as e:
-            error_properties: str = ""
-            with suppress(KeyError):
-                error_properties = ".".join(str(x) for x in e.errors()[0]["loc"])
-            raise InvalidResourceException(self.logical_id, f"Property '{error_properties}' is invalid.") from e
-        except AttributeError as e:
-            raise InvalidResourceException(
-                self.logical_id, "Module 'pydantic.v1' has no attribute 'error_wrappers'"
-            ) from e
```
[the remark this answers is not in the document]
aaythapa: This doesn't work because the `AttributeError` isn't thrown by the try block but by `except pydantic.error_wrappers.ValidationError as e:`. So if we keep `except pydantic.error_wrappers.ValidationError as e:` there its going to throw an error that won't be caught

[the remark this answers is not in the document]
GavinZZ: Oops I think copy & paste and indentation messed up with the code I had above. I meant to have it as nested try except where we try and except `AttributeError` in the outer try-catch and the inner try-catch would be the original code with 
```
except pydantic.error_wrappers.ValidationError as e:
            error_properties: str = ""
            with suppress(KeyError):
                error_properties = ".".join(str(x) for x in e.errors()[0]["loc"])
            raise InvalidResourceException(self.logical_id, f"Property '{error_properties}' is invalid.") from e
```

aaythapa:
Talked offline with the team, we won't change anything on our end since pydantic released a breaking change. We'll wait on pydantic to fix on their side, for now customers should avoid using pydantic version 1.10.15 and 1.10.17 as the unit tests will fail.

PR #3620, opened 2024-06-24
fix: exclude breaking pydantic version

aaythapa:
### Issue #, if available
#3617

### Description of changes

Pydantic versions 1.10.15 and 1.10.17 introduced breaking changes. Exclude those versions from being installed.

Ran `make init` with `pydantic>=1.8,!=1.10.15,!=1.10.17,<2` (to insure pydantic V1 got installed). `pydantic 1.10.16` was installed

### Checklist

    - [ ] Using correct values
    - [ ] Using wrong values

### Examples?

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -3,4 +3,5 @@ jsonschema<5,>=3.2  # TODO: evaluate risk of removing jsonschema 3.x support
 typing_extensions>=4.4 # 3.8 doesn't have Required, TypeGuard and ParamSpec
 
 # resource validation & schema generation
-pydantic>=1.8,<3
+# 1.10.15 and 1.10.17 included breaking change from pydantic, more info: https://github.com/aws/serverless-application-model/issues/3617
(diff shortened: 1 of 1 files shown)


xazhao:
Should we pin the version as 1.10.16 or < 1.10.15? If they release a new version without fixing it, we will have to change it again.

aaythapa:
> Should we pin the version as 1.10.16 or < 1.10.15? If they release a new version without fixing it, we will have to change it again.

Even if we pin a version would we not have to introduce this change eventually to block ppl from using 1.10.15 and 1.10.17? Imo if they release another version without fixing this issue then we can pin a version 😅

xazhao:
Yeah if we pin a version now we will need to unpin eventually. There will be a little additional work but it's safer to make sure we don't use the bad version in the future (we can test the version before unpinning). Not a merge blocker though.

xazhao (approved):

GavinZZ (approved):

shikha372 (approved):

Issue #1796, opened 2020-11-13
KMSEncryptPolicy cannot be used to encrypt


john-aws:
Related to #1186, is the KMSEncryptPolicy policy template actually usable as written? It can't be used to publish KMS-encrypted data to an SNS topic, for example, because KMSEncryptPolicy doesn't include `kms:GenerateDataKey`.

sanwar-oc:
I second this. I'm facing the exact same issue.

astroanu:
what's the workaround for this? can someone help?

john-aws:
@wchengru Hi, what kind of example are you looking for?

I'm looking for a more usable out-of-the-box SAM policy template, for example:

```
kms:Encrypt
kms:ReEncrypt*
kms:GenerateDataKey*
kms:DescribeKey
```

elmaimbo:
I've submitted a pull request #3657 which adds kms:GenerateDataKey.

PR #3657, opened 2024-09-30
Added kms:GenerateDataKey action to KMSEncryptPolicy policy

elmaimbo:
### Issue #
#1796 KMSEncryptPolicy cannot be used to encrypt

### Description of changes
The KMSEncryptPolicy policy lacks the permission required to encrypt data using a data key. This trivial fix adds the missing permission.

Copied and pasted the change into the policy on AWS (that had been deployed using SAM), and confirmed that this allowed the encrypt function to work.


samtranslator/policy_templates_data/policy_templates.json
```diff
@@ -1253,7 +1253,10 @@
       "Definition": {
         "Statement": [
           {
-            "Action": "kms:Encrypt",
+            "Action": [
+              "kms:Encrypt",
+              "kms:GenerateDataKey"
+            ],
```
aaythapa, line 1259: This will break backwards compatibility for people using `KMSEncryptPolicy` unfortunately. We can create a new policy called `KMSEncryptPolicy_v2` with the new permissions. Example PR you can follow: https://github.com/aws/serverless-application-model/pull/2929

samtranslator/policy_templates_data/policy_templates.json
```diff
@@ -1253,7 +1253,10 @@
       "Definition": {
         "Statement": [
           {
-            "Action": "kms:Encrypt",
+            "Action": [
```
aaythapa, line 1256: Can we add some unit tests for this? You can follow this guide: https://github.com/aws/serverless-application-model/blob/develop/DEVELOPMENT_GUIDE.md#transform-tests

aaythapa (commented):
Thank you for contributing! Will do a review of this internally

aaythapa:
I see in the issue comments that someone requested these Actions
```
kms:Encrypt
kms:ReEncrypt*
kms:GenerateDataKey*
kms:DescribeKey
```
Any reason why we shouldn't include `kms:ReEncrypt*` and `kms:DescribeKey` here?

samtranslator/policy_templates_data/policy_templates.json
elmaimbo: My change is backward compatible because I'm only adding a new permission, not removing anything.

elmaimbo:
I noticed that almost all other policies listed actions explicitly, rather than using "*" (wildcard), so I've done the same in the latest commit.

samtranslator/policy_templates_data/policy_templates.json
aaythapa: Will double check this with another team member but I believe adding a new permissions will also be backwards incompatible. If a user has a SAM Function with `KMSEncryptPolicy` attached then the transformed template will have role with 
```
PolicyDocument": {
       "Statement": [
        {
         "Action": "kms:Encrypt",
         "Effect": "Allow",
         "Resource": {
          ...
         }
```
with your change the transformed template will be
```
PolicyDocument": {
       "Statement": [
        {
         "Action": [
          "kms:Encrypt",
          "kms:GenerateDataKey"
         ],
         "Effect": "Allow",
         "Resource": {
          ...
         }
```
The user didn't make any changes to their SAM template but the CFN template change which causes re-deployments which we consider backwards incompatible

aaythapa:
> I think the pertinent question is whether it is more appropriate to include kms:ReEncryptFrom, ReEncryptTo and kms:DescribeKey in the KMSEncryptPolicy definition, or create separate policy definitions for those? (I don't believe I'm qualified to answer that question myself, and TBH I have no need of these additional permissions.) 

I can bring this up during our internal review

aaythapa:
We've reviewed the PR internally and the actions that were added are all `GET` operations so nothing sensitive. 

We're good to merge after the change is made backwards compatible, I explained [here](https://github.com/aws/serverless-application-model/pull/3657#discussion_r1788417800) why this change would break compatibility.

@elmaimbo

elmaimbo:
> We've reviewed the PR internally and the actions that were added are all `GET` operations so nothing sensitive.
> 
> We're good to merge after the change is made backwards compatible, I explained [here](https://github.com/aws/serverless-application-model/pull/3657#discussion_r1788417800) why this change would break compatibility.
> 
> @elmaimbo

This is now complete in the latest commit.

Please let me know if anything is still outstanding?

Thanks,
Nick.

aaythapa (approved):

GavinZZ (approved):

PR #3719, opened 2025-02-13
feat: custom domain name support for private endpoints

Vandita2020:
### Description of changes
The AWS API Gateway now supports the ability to use custom domain names for private endpoints. Two new resources have been introduced to enable this feature - [AWS::ApiGateway::DomainNameV2](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-apigateway-domainnamev2.html) and [AWS::ApiGateway::BasePathMappingV2](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-apigateway-basepathmappingv2.html).
### Checklist

    - [ ] Using correct values
    - [ ] Using wrong values

### Examples?


samtranslator/model/api/api_generator.py
```diff
@@ -603,6 +612,129 @@ def _construct_api_domain(  # noqa: PLR0912, PLR0915
 
         return ApiDomainResponse(domain, basepath_resource_list, record_set_group)
 
+    def _construct_api_domain_v2(
+        self, rest_api: ApiGatewayRestApi, route53_record_set_groups: Any
+    ) -> ApiDomainResponseV2:
```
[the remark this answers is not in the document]
mbfreder: Updated.

samtranslator/model/api/api_generator.py
```diff
@@ -603,6 +612,129 @@ def _construct_api_domain(  # noqa: PLR0912, PLR0915
...
+                # contain letters, numbers, and one of "$-_.+!*'()"
+                path = "".join(e for e in basepath if e.isalnum())
+                logical_id = "{}{}{}".format(self.logical_id, path, "BasePathMapping")
+                basepath_mapping = ApiGatewayBasePathMappingV2(
+                    logical_id, attributes=self.passthrough_resource_attributes
+                )
+                basepath_mapping.DomainNameArn = ref(domain_name_arn)
+                basepath_mapping.RestApiId = ref(rest_api.logical_id)
+                basepath_mapping.Stage = ref(rest_api.logical_id + ".Stage")
+                basepath_mapping.BasePath = path if normalize_basepath else basepath
```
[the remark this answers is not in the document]
mbfreder: Updated.

samtranslator/model/api/api_generator.py
```diff
@@ -603,6 +612,129 @@ def _construct_api_domain(  # noqa: PLR0912, PLR0915
...
+                    self._construct_single_record_set_group(self.domain, domain_name, route53),
+                )
+
+            if not record_set_group:
+                record_set_group = Route53RecordSetGroup(logical_id, attributes=self.passthrough_resource_attributes)
+                if "HostedZoneId" in route53:
+                    record_set_group.HostedZoneId = route53.get("HostedZoneId")
+                if "HostedZoneName" in route53:
+                    record_set_group.HostedZoneName = route53.get("HostedZoneName")
+                record_set_group.RecordSets = []
```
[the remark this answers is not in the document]
mbfreder: Updated.

samtranslator/model/api/api_generator.py
```diff
@@ -603,6 +612,135 @@ def _construct_api_domain(  # noqa: PLR0912, PLR0915
...
+                " one of {}.".format(["EDGE", "REGIONAL", "PRIVATE"]),
+            )
+
+        domain.CertificateArn = certificate_arn
+
+        domain.EndpointConfiguration = {"Types": [endpoint]}
+
+        self._set_optional_domain_properties(domain)
+
+        basepaths: Optional[List[str]] = self._get_basepaths()
```
valerena, line 655: You should call this in the original `_construct_api_domain` (non v2) method too.

samtranslator/model/api/api_generator.py
```diff
@@ -667,6 +805,17 @@ def _construct_alias_target(self, domain: Dict[str, Any], api_domain_name: str,
             alias_target["DNSName"] = route53.get("DistributionDomainName")
         return alias_target
 
+    def _create_basepath_mapping(
```
valerena, line 808: call it `_v2`, because in theory you could be doing the same for v1, but this one explicitly uses v2.

(It would be good if you actually create another method that does this for v1 and use it in the other version)

samtranslator/model/api/api_generator.py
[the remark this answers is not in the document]
valerena: This comment meant that you should have created a function similar to what you did with `create_basepath_mapping`, but that you can actually use it here too (probably you need to pass a few more parameters, to make sure that the same function works for both cases).

samtranslator/model/api/api_generator.py
```diff
@@ -603,6 +612,135 @@ def _construct_api_domain(  # noqa: PLR0912, PLR0915
...
+                    route53.get("SeparateRecordSetGroup"), self.logical_id, "Domain.Route53.SeparateRecordSetGroup"
+                ).to_be_a_bool()
+                return ApiDomainResponseV2(
+                    domain,
+                    basepath_resource_list,
+                    self._construct_single_record_set_group(self.domain, domain_name, route53),
+                )
+
+            if not record_set_group:
+                record_set_group = self._get_record_set_group(logical_id, route53)
```
valerena, line 709: You can use this function in the original `_construct_api_domain` too.

samtranslator/model/api/api_generator.py
```diff
@@ -603,6 +612,135 @@ def _construct_api_domain(  # noqa: PLR0912, PLR0915
...
+                route53_record_set_groups[logical_id] = record_set_group
+
+            record_set_group.RecordSets += self._construct_record_sets_for_domain(self.domain, domain_name, route53)
+
+        return ApiDomainResponseV2(domain, basepath_resource_list, record_set_group)
```
valerena, line 714: It pains me to see that all this code it's the same as the original `_construct_api_domain` except by returning `ApiDomainResponseV2` instead of V1 (and most of the full function really)

I would really like to just take this to another function that we can call in both places. 
- either a function that can returns either a v1 or a v2 resource
- a function where you pass the "Class" of the object that needs to be created.
- a function that returns "the parameters", and then you create the object outside.

Now the problem here is that the domain and basepath mapping resource can also be either v1 or v2, so it gets harder to get the types right.

Maybe we can chat about it to see if there's a good solution.

mbfreder: Discussed offline that it's better to leave this as is for now. Will investigate further refactoring opportunities, but that will be part of another PR.

mbfreder:
Test failures are unrelated.

samtranslator/model/api/api_generator.py
```diff
@@ -676,10 +829,19 @@ def to_cloudformation(
...
-        basepath_mapping = api_domain_response.apigw_basepath_mapping_list
+        if isinstance(self.domain, dict) and self.domain.get("EndpointConfiguration") == "PRIVATE":
+            api_domain_response = self._construct_api_domain_v2(rest_api, route53_record_set_groups)
+            domain = api_domain_response.domain
+            basepath_mapping = api_domain_response.apigw_basepath_mapping_list
+        else:
+            api_domain_response = self._construct_api_domain(rest_api, route53_record_set_groups)
+            domain = api_domain_response.domain
+            basepath_mapping = api_domain_response.apigw_basepath_mapping_list
+
```
valerena, line 844: It feels cleaner if you only separate the `api_domain_response` definition, considering the other ones are the same. And it's conceptually much easier to understand (and the diff is smaller)

```
        if isinstance(self.domain, dict) and self.domain.get("EndpointConfiguration") == "PRIVATE":
            api_domain_response = self._construct_api_domain_v2(rest_api, route53_record_set_groups)
        else:
            api_domain_response = self._construct_api_domain(rest_api, route53_record_set_groups)
        domain = api_domain_response.domain
        basepath_mapping = api_domain_response.apigw_basepath_mapping_list
```

samtranslator/model/api/api_generator.py
```diff
@@ -702,7 +864,9 @@ def to_cloudformation(
                 List[Resource],
                 Tuple[Resource],
                 List[LambdaPermission],
+                List[Resource],
```
valerena, line 867: There's already another `List[Resource]`, you don't need this one again. (The `Union` here means that it could be of any of these types, so you don't need to repeat something that already exists)

samtranslator/model/sam_resources.py
```diff
@@ -1310,6 +1311,7 @@ class SamApi(SamResourceMacro):
         "Stage": ApiGatewayStage.resource_type,
         "Deployment": ApiGatewayDeployment.resource_type,
         "DomainName": ApiGatewayDomainName.resource_type,
+        "DomainNameV2": ApiGatewayDomainNameV2.resource_type,
```
valerena, line 1314: Do you know what this does? `referable_properties` sounds like things you can `!Ref`. and I don't know if this means we need to do something else for this to actually work, or if it just does.

valerena (commented):
I added a few comments. I'll finish the review later

tests/translator/input/api_with_custom_domains_private.yaml
```diff
@@ -0,0 +1,62 @@
+Parameters:
+  DomainName:
+    Type: String
+    Default: private.example.com
+    Description: Custom domain name for the API
+
+  CertificateArn:
+    Type: String
+    Default: arn:aws:acm:us-west-2:123456789:certificate/abcd-000-1234-0000-000000abcd
```
valerena, line 9: This is just a test, so it's not crazy important, but we probably shouldn't have a certificate in `us-west-2`, because this template is supposed to work for all regions, and there are outputs for the 3 partitions (`aws` plus `aws-cn` and `aws-us-gov`) below.

So instead of doing a fancy thing, we can just use a completely fake ARN instead, because we don't validate its value. Like this other test: https://github.com/aws/serverless-application-model/blob/81cdfc28b2ba98ab823ed9fdaae687c336ee6721/tests/translator/input/api_with_basic_custom_domain_intrinsics.yaml#L7-L9

Or just don't have a default at all, like here:
https://github.com/aws/serverless-application-model/blob/81cdfc28b2ba98ab823ed9fdaae687c336ee6721/tests/translator/input/http_api_with_custom_domains_regional.yaml#L4-L5

Any of these work for me, it's just weird to see `us-west-2` in the output of an `aws-cn` template.

samtranslator/model/api/api_generator.py
mbfreder: Updated

samtranslator/model/sam_resources.py
mbfreder: I don't think there's anything else that needs to be done for this to work. We can ref it using <api-logicalId>.DomainNameV2 property. https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-specification-generated-resources-api.html

valerena (approved):

vicheey (approved):

PR #3727, opened 2025-02-28
feat: skipping test if Layers is not available

Vandita2020:
### Description of changes
To skip the integration test for times when "Layer" functionality is not available in regions that are not GA yet. 

### Checklist

    - [ ] Using correct values
    - [ ] Using wrong values

### Examples?

### DIFF
--- a/integration/combination/test_function_with_alias.py
+++ b/integration/combination/test_function_with_alias.py
@@ -3,7 +3,7 @@ from unittest.case import skipIf
 
 from botocore.exceptions import ClientError
 
-from integration.config.service_names import REST_API
+from integration.config.service_names import LAYERS, REST_API
 from integration.helpers.base_test import LOG, BaseTest
 from integration.helpers.common_api import get_function_versions
 from integration.helpers.resource import current_region_does_not_support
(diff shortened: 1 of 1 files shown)


integration/combination/test_function_with_alias.py
```diff
@@ -160,6 +161,7 @@ def test_alias_with_event_sources_get_correct_permissions(self):
         function_policy = json.loads(function_policy_str)
         self.assertEqual(len(function_policy["Statement"]), len(permission_resources))
 
+    @skipIf(current_region_does_not_support([LAYERS]), "Layers is not supported in this testing region")
```
seshubaws, line 164: do we have to say layer versions specifically?

seshubaws (approved):

valerena (approved):

valerena (changes requested):
There's an issue with `make pr` because of the import (probably need to add both `LAYERS, REST_API` on the same line instead of separate)

integration/combination/test_function_with_alias.py
Vandita2020: "Layers" supersets the "Layer Version". Keeping it generic here.

valerena (approved):

PR #3737, opened 2025-03-24
fix(lambda): correcting changeset detection

Vandita2020:
### Description of changes
With the updates to the parameter values, the changes were not reflected when the changeSet got created, the issue was with the condition used to check if the update is being made. 

Previously, the code used `type == "Update"` to validate if the action was an update, but it should have been `changeset_type == "Update"`.
### Checklist

    - [ ] Using correct values
    - [ ] Using wrong values

### Examples?


vicheey:
Do you mind providing links to any related changes made prior to this for context?

valerena:
It's interesting that this *is* a bug, but it's not related to the change that we were talking before. So I don't know exactly what this will do (because this only affects SAM's integration tests, and nothing else.

seshubaws:
> It's interesting that this _is_ a bug, but it's not related to an issue we discussed offline. So I don't know exactly what this will do (because this only affects SAM's integration tests, and nothing else).

@valerena do you a link/sim to the issue we discussed offline?

PR #3740, opened 2025-03-25
fix: BasePath not included in AWS::ApiGateway::BasePathMapping when defined as ["/"]

mbfreder:
### Issue #, if available
The "AWS::Serverless::Api" SAM resource type creates an "AWS::ApiGateway::BasePathMapping" resource when processed.
When the 'BasePath' sub-property of this resource type is defined as ["/"] , the "AWS::ApiGateway::BasePathMapping" resource in the processed template does NOT include the 'BasePath' property.

### Description of changes
transform ["/"] into "" in the processed template.

Updated unit tests to cover the edge case

### Checklist

    - [ ] Using correct values
    - [ ] Using wrong values

### Examples?


vicheey:
For context, this is the previous PR that contain change that this current PR try to address : https://github.com/aws/serverless-application-model/pull/3719

valerena:
It's probably better to add **new** test files instead of modifying the existing ones. It's always good to keep old tests relatively steady, so we can confirm that we're not breaking any of the old cases that were already working fine.

samtranslator/model/api/api_generator.py
```diff
@@ -805,8 +805,7 @@ def _create_basepath_mapping(
         basepath_mapping.DomainName = ref(api_domain_name)
         basepath_mapping.RestApiId = ref(rest_api.logical_id)
         basepath_mapping.Stage = ref(rest_api.logical_id + ".Stage")
-        if basepath:
-            basepath_mapping.BasePath = basepath
+        basepath_mapping.BasePath = basepath
```
valerena, line 808: I think we still need this if, because we use this method passing `None` as `basepath` (and `basepath` is optional), so we should probably actually update this to `if basepath is not None:` instead of removing the `if`. 

I think that's a different edge case that we might not be covering in the test cases either, which is the first branch of this `if`: https://github.com/aws/serverless-application-model/blob/develop/samtranslator/model/api/api_generator.py#L546-L548

vicheey, line 808:
```suggestion
        # Skip checking the basepath; We need to set "BasePath": "" if no basepath is provided to satisfy CFN. 
        basepath_mapping.BasePath = basepath
```

valerena (approved):

vicheey (approved):

PR #3783, opened 2025-06-19
Merge main to develop

vicheey:
This PR & commit is automatically created by aws-sam-cli-bot for SamTranslator@1.99.0 release.


.cfnlintrc.yaml
```diff
@@ -139,6 +139,8 @@ ignore_templates:
   - tests/translator/output/**/managed_policies_everything.json # intentionally contains wrong arns
   - tests/translator/output/**/function_with_provisioned_poller_config.json
   - tests/translator/output/**/function_with_metrics_config.json
+  - tests/translator/output/**/function_with_self_managed_kafka_and_schema_registry.json # cfnlint is not updated to recognize the SchemaRegistryConfig property
+  - tests/translator/output/**/function_with_msk_with_schema_registry_config.json # cfnlint is not updated to recognize the SchemaRegistryConfig property
```
roger-zhangg, line 143: we should remove these 2 as a fast followup

requirements/dev.txt
```diff
@@ -26,6 +26,9 @@ ruamel.yaml==0.17.21  # It can parse yaml while perserving comments
 mypy~=1.3.0
 
 # types
-boto3-stubs[appconfig,serverlessrepo]>=1.19.5,==1.*
+boto3-stubs[appconfig,serverlessrepo]>=1.34.0,<2.0.0
 types-PyYAML~=6.0
 types-jsonschema~=3.2
+
+# CloudFormation CLI tools
+cloudformation-cli>=0.2.39,<0.3.0
```
roger-zhangg, line 34: as a followup let's check why this is required.

valerena: It was added with Schema Registry. It might not be required, but at least it's only dev dependencies.

.cfnlintrc.yaml
valerena: And we should probably also check if we can remove more of this list. I feel like most of them are added before the feature is launched, but then never removed.

requirements/dev.txt
vicheey: Yes.

.cfnlintrc.yaml
vicheey: Ack

PR #3805, opened 2025-08-22
Add resource property validation framework to SAM translator. 

vicheey:
### Issue #, if available
None 

### Description of changes
This PR introduces a comprehensive validation framework for SAM resource properties,  and enhancing the translator's ability to catch configuration errors early in the deployment process.

-  **Added property rule validation method** in samtranslator/model/__init__.py 
-  **Enhanced add propagate tags logic** in samtranslator/model/__init__.py 
-  **Add enhance property rule validation logic** samtranslator/validator/property_rule.py
- **Updated SAM resources** in samtranslator/model/sam_resources.py to integrate 
validation
- **Comprehensive test coverage** in tests/test_model.py 
- **Comprehensive test coverage** in tests/validator/test_property_rule.py
- 
- Added extensive test suite covering validation scenarios
- All existing tests continue to pass

### Checklist

    - [x] Using correct values
    - [x] Using wrong values


vicheey:
The Tests failed because of lower coverage percentage. (94.82% instead of 95%). Added validation logic push the overall total line increase and lower the coverage percentage.  

Low coverage files:
• samtranslator/validator/validator.py - 27% coverage 
• samtranslator/model/__init__.py - 69% coverage 
• samtranslator/model/naming.py - 71% coverage
• samtranslator/region_configuration.py - 79% coverage

Removing deprecated and unused `samtranslator/validator/validator.py`

vicheey:
Current usage analysis:
• No active imports of `SamTemplateValidator` in any source 
files
• No test files directly testing `SamTemplateValidator`

What IS being used:
• `samtranslator.validator.value_validator.sam_expect` - Used extensively across 12+ files
• Direct `jsonschema.validate()` calls in tests (instead of the deprecated SAM wrapper)

```
--------------------------------
TOTAL   9640    314   3402    220    95%

Required test coverage of 95% reached. Total coverage: 95.28%

=============== 4118 passed in 86.15s (0:01:26) ================
```

samtranslator/model/__init__.py
```diff
@@ -582,6 +687,71 @@ def _check_tag(self, reserved_tag_name, tags):  # type: ignore[no-untyped-def]
...
+                # If any property in the group is present, then all properties in the group must be present
+                # Check if some but not all properties from the group are present
+                missing_some_properties = 0 < len(present) < len(properties)
+                if missing_some_properties:
+                    error_messages.append(f"Properties must be used together: {self._combine_string(properties)}.")
+
+            elif (
+                rule_type == ValidationRule.CONDITIONAL_REQUIREMENT
+                and self._get_property_value(properties[0], validated_model) is not None
+                and self._get_property_value(properties[1], validated_model) is None
```
roger-zhangg: you can only set 2 prop in one rule?

samtranslator/model/__init__.py
```diff
@@ -582,6 +687,71 @@ def _check_tag(self, reserved_tag_name, tags):  # type: ignore[no-untyped-def]
...
+            return
+
+        rules = self.__validation_rules__
+        validated_model = (
+            self.validate_properties_and_return_model(schema_class, collect_all_errors) if schema_class else None
+        )
+
+        error_messages = []
+
+        for rule_type, properties in rules:
```
roger-zhangg: Can we do something like `[ rule.run() for rules in self.__validation_rules__]` so we can save all the if /elif below?

samtranslator/model/__init__.py
```diff
@@ -582,6 +687,71 @@ def _check_tag(self, reserved_tag_name, tags):  # type: ignore[no-untyped-def]
...
+            present = [prop for prop in properties if self._get_property_value(prop, validated_model) is not None]
```
roger-zhangg: is there a better name for `present`? It's hard to understnad

samtranslator/model/__init__.py
```diff
@@ -582,6 +687,71 @@ def _check_tag(self, reserved_tag_name, tags):  # type: ignore[no-untyped-def]
...
+            if rule_type == ValidationRule.MUTUALLY_EXCLUSIVE:
+                # Check if more than one property exists
+                if len(present) > 1:
+                    prop_names = [f"'{p}'" for p in present]
+                    error_messages.append(f"Cannot specify {self._combine_string(prop_names)} together.")
+
+            elif rule_type == ValidationRule.MUTUALLY_INCLUSIVE:
```
roger-zhangg: Is this like an antipattern? User need to constantly set 2 or more params to achieve a single action. Is there an example for this `MUTUALLY_INCLUSIVE`? Can we optimize it to be `CONDITIONAL_REQUIREMENT`?

samtranslator/model/__init__.py
```diff
@@ -582,6 +687,71 @@ def _check_tag(self, reserved_tag_name, tags):  # type: ignore[no-untyped-def]
                 "input.",
             )
 
+    def validate_before_transform(self, schema_class: Optional[Type[RT]], collect_all_errors: bool = False) -> None:
```
roger-zhangg: should collect_all_errors and schema_class be a class attr? So it's like `this.collect_all_errors` and `this. schema_class`

samtranslator/model/__init__.py
```diff
@@ -582,6 +687,71 @@ def _check_tag(self, reserved_tag_name, tags):  # type: ignore[no-untyped-def]
...
+            ):
+                # First property requires second property
+                error_messages.append(f"'{properties[0]}' requires '{properties[1]}'.")
+
+        # If there are any validation errors, raise a single exception with all error messages
+        if error_messages:
+            raise InvalidResourceException(self.logical_id, "\n".join(error_messages))
+
+    def _combine_string(self, words: List[str]) -> str:
+        return ", ".join(words[:-1]) + (" and " + words[-1] if len(words) > 1 else words[0] if words else "")
```
roger-zhangg: why? why not `", ".join(words)`

samtranslator/model/__init__.py
```diff
@@ -553,6 +624,40 @@ def _construct_tag_list(
         # customer's knowledge.
         return get_tag_list(sam_tag) + get_tag_list(additional_tags) + get_tag_list(tags)
 
+    @staticmethod
+    def propagate_tags_combine(
```
roger-zhangg, line 625: This is not mentioned in PR title or description

samtranslator/model/__init__.py
```diff
@@ -582,6 +687,71 @@ def _check_tag(self, reserved_tag_name, tags):  # type: ignore[no-untyped-def]
...
+                error_messages.append(f"'{properties[0]}' requires '{properties[1]}'.")
+
+        # If there are any validation errors, raise a single exception with all error messages
+        if error_messages:
+            raise InvalidResourceException(self.logical_id, "\n".join(error_messages))
+
+    def _combine_string(self, words: List[str]) -> str:
+        return ", ".join(words[:-1]) + (" and " + words[-1] if len(words) > 1 else words[0] if words else "")
+
+    def _get_property_value(self, prop: str, validated_model: Any = None) -> Any:
```
roger-zhangg: I would suggest to update `validate_properties_and_return_model` return a custom dict that allow you to directly use validated_model["a.b.c"]. Like
```
class DotDict(dict):
    def __getitem__(self, path):
        value = self
        for key in path.split('.'):
            value = dict.__getitem__(value, key)
        return value
```
`validated_model["a.b.c"]` would have a much better readability than `_get_property_value("a.b.c", validated_model)`

tests/test_model.py
```diff
@@ -391,3 +396,393 @@ class MyResource(Resource):
...
+            self.TestSettingProperties = parent.TestSettingProperties
+            self.TestResource = parent.TestResource
+
+        def test_validate_before_transform_all_rules_pass(self):
+            """Test successful validation when all rules are satisfied"""
+
+            class Resource(self.TestResource):
+                __validation_rules__ = [
+                    (self.ValidationRule.CONDITIONAL_REQUIREMENT, ["ConditionalVar1", "ConditionalVar2"]),
+                    (self.ValidationRule.CONDITIONAL_REQUIREMENT, ["ConditionalVar2", "ConditionalVar3"]),
```
roger-zhangg: from actually seeing this, does `Rule('ConditionalVar2').require('ConditionalVar1')` sounds better?
This can even extend to like `Rule('ConditionalVar2').require('ConditionalVar1==123')`

tests/test_model.py
```diff
@@ -391,3 +396,393 @@ class MyResource(Resource):
...
+                    (
+                        self.ValidationRule.CONDITIONAL_REQUIREMENT,
+                        ["NestedSetting1.NestedVar1", "NestedSetting2.NestedVar2"],
+                    ),
+                    (self.ValidationRule.MUTUALLY_EXCLUSIVE, ["ExclusiveVar1", "ExclusiveVar2"]),
```
roger-zhangg: `Rule('ConditionalVar2').reject('ConditionalVar1')`

tests/test_model.py
```diff
@@ -391,3 +396,393 @@ class MyResource(Resource):
...
+                    (self.ValidationRule.MUTUALLY_EXCLUSIVE, ["ExclusiveVar2", "ExclusiveVar3"]),
+                    (
+                        self.ValidationRule.MUTUALLY_EXCLUSIVE,
+                        ["NestedSetting1.NestedVar1", "NestedSetting2.NestedVar1"],
+                    ),
+                    (self.ValidationRule.MUTUALLY_EXCLUSIVE, ["NestedSetting1.NestedVar2", "ExclusiveVar3"]),
+                    (self.ValidationRule.MUTUALLY_INCLUSIVE, ["InclusiveVar1", "ExclusiveVar1"]),
```
roger-zhangg: `Rule('ConditionalVar1', 'ConditionalVar2')`
Just some thoughts

samtranslator/model/__init__.py
```diff
@@ -582,6 +687,71 @@ def _check_tag(self, reserved_tag_name, tags):  # type: ignore[no-untyped-def]
                 "input.",
             )
 
+    def validate_before_transform(self, schema_class: Optional[Type[RT]], collect_all_errors: bool = False) -> None:
+        if not hasattr(self, "__validation_rules__"):
+            return
+
+        rules = self.__validation_rules__
```
roger-zhangg: One concern, currently there's no type hint on `self. __validation_rules__`, so we need to look at document/ previous implementation everytime when we need to use this function

samtranslator/model/__init__.py
vicheey: Yes. This is to simplify the rule validation. You can as many rule as you want.

tests/test_model.py
vicheey: That's a good suggestion. Would you like the `reject()` to accept array instead of a string as well?

samtranslator/model/__init__.py
vicheey: Updated to using override method instead of `__validation_rules__`

tests/test_model.py
vicheey: Updated. Please take a look at the second revision.

tests/test_model.py
vicheey: Updated in new commit

samtranslator/model/__init__.py
vicheey: The validate_properties_and_return_model is used by other classes. It is risky to modify the method.

samtranslator/model/__init__.py
vicheey: Updated discription.

samtranslator/model/__init__.py
vicheey: This is to allow: `A, B, and C`, `A and B`, and `A`

samtranslator/model/__init__.py
vicheey: Adding a new variable requires adding the class to `_keyword`. I prefer function parameter for that matter.

samtranslator/model/__init__.py
vicheey: Updated commit allows mutually inclusive, mutually exclusive, conditionally inclusive and conditionally exclusive.

samtranslator/model/__init__.py
vicheey: Outdated.

samtranslator/model/__init__.py
vicheey: Updated code should support multiple element in the second param.

samtranslator/model/__init__.py
vicheey: I'll update.

samtranslator/model/sam_resources.py
```diff
@@ -252,6 +253,21 @@ class SamFunction(SamResourceMacro):
...
+    # return (PropertyRules(FunctionProperties)
+    #     .addMutuallyExclusive("ImageUri", "InlineCode", "CodeUri")
+    #     .addConditionalInclusive("DeploymentPreference", ["AutoPublishAlias"])
+    #     .addConditionalInclusive("ProvisionedConcurrencyConfig", ["AutoPublishAlias"])
+    #     .addConditionalInclusive("PackageType=Zip", ["Runtime", "Handler"])
+    #     .addConditionalInclusive("PackageType=Image", ["ImageUri"])
+    #     .addConditionalExclusive("PackageType=Zip", ["ImageUri", "ImageConfig"])
+    #     .addConditionalExclusive("PackageType=Image", ["Runtime", "Handler", "Layers"]))
+    # return None
+
```
valerena, line 270: Do we plan to leave this here, commented out?

samtranslator/model/sam_resources.py
```diff
@@ -274,6 +290,11 @@ def to_cloudformation(self, **kwargs):  # type: ignore[no-untyped-def] # noqa: P
         conditions = kwargs.get("conditions", {})
         feature_toggle = kwargs.get("feature_toggle")
 
+        # TODO: Skip pass schema_class=aws_serverless_function.Properties to skip schema validation for now.
+        # - adding this now would required update error message in error error_function_*_test.py
+        # - add this when we can verify that changing error message would not break customers
+        # self.validate_before_transform(schema_class=aws_serverless_function.Properties)
```
valerena, line 296: what is this `validate_before_transform` that appears here? Is it defined somewhere else?

samtranslator/validator/property_rule.py
```diff
@@ -0,0 +1,140 @@
...
+
+    def _validate_conditional_inclusive(
+        self, validated_model: RT, source_props: List[str], target_props: List[str]
+    ) -> Optional[str]:
+        source_present = any(self._check_property_condition(validated_model, prop) for prop in source_props)
+        if source_present:
+            missing = [prop for prop in target_props if not self._check_property_condition(validated_model, prop)]
+            if missing:
+                missing_props = ", ".join(f"'{p}'" for p in missing)
+                return f"'{source_props[0]}' requires all of: {missing_props}."
```
valerena, line 106: is `source_props` only of length 1?

If so, can you change line 101 to just take the first one from `source_props` instead of loop over the array?

Or if we want to still keep it extensible, we would have to change this error here to point to the all the `source_props` (or the right one), instead of just using `source_props[0]`

valerena: same for `conditional_exclusive`

tests/test_model.py
```diff
@@ -171,10 +175,13 @@ def test_resource_must_override_runtime_attributes(self):
...
             resource_type = "foo"
             property_types = {}
-            runtime_attrs = {"attr1": Mock(), "attr2": Mock()}
 
-            runtime_attrs["attr1"].return_value = "value1"
-            runtime_attrs["attr2"].return_value = "value2"
+            mock_attr1 = Mock()
+            mock_attr2 = Mock()
+            mock_attr1.return_value = "value1"
+            mock_attr2.return_value = "value2"
```
valerena, line 182: Not necessary (and I don't know how much it's used in other places, but you can also set the return value right away:
 
```
  mock_attr1 = Mock(return_value="value1")
  mock_attr2 = Mock(return_value="value2")
```

samtranslator/validator/property_rule.py
```diff
@@ -0,0 +1,140 @@
...
+    ) -> Optional[str]:
+        if rule_type == RuleType.MUTUALLY_EXCLUSIVE:
+            return self._validate_mutually_exclusive(validated_model, source_props)
+        if rule_type == RuleType.CONDITIONAL_EXCLUSIVE:
+            return self._validate_conditional_exclusive(validated_model, source_props, target_props)
+        if rule_type == RuleType.MUTUALLY_INCLUSIVE:
+            return self._validate_mutually_inclusive(validated_model, source_props)
+        if rule_type == RuleType.CONDITIONAL_INCLUSIVE:
+            return self._validate_conditional_inclusive(validated_model, source_props, target_props)
+        return None
```
roger-zhangg, line 69: consider making the rule a class, so we can do
```suggestion
        return self.rule.validate()
```

PR #3807, opened 2025-08-26
test: delete created change sets in companion stack if empty

reedham-aws:
### Issue #, if available
N/a

### Description of changes
In the integration tests, the companion stack being created is created with a change set. Often times, this change set is empty and ignored; after enough test runs, the stack hits its limit of 1000 changesets, causing create failures in the subsequent runs.

This change will now delete the created change set if it is empty.

In my personal account, I ran the tests a couple time to build up a couple empty change sets. I then made the change, and continued to run `make prepare-companion-stack`, manually verifying that there were no new changesets being kept on each test run.

### Checklist

    - [ ] Using correct values
    - [ ] Using wrong values

### Examples?

### DIFF
--- a/integration/helpers/deployer/deployer.py
+++ b/integration/helpers/deployer/deployer.py
@@ -375,7 +375,10 @@ class Deployer:
         except deploy_exceptions.ChangeEmptyError:
             try:
                 # Delete the most recent change set that failed to create because it was empty
-                changeset = sorted(self._client.list_change_sets(StackName=stack_name).get("Summaries"), key=lambda c: c["CreationTime"])[-1]
+                changeset = sorted(
+                    self._client.list_change_sets(StackName=stack_name).get("Summaries"),
+                    key=lambda c: c["CreationTime"],
+                )[-1]
                 if changeset.get("Status") == "FAILED" and changeset.get("ExecutionStatus") == "UNAVAILABLE":
                     self._client.delete_change_set(ChangeSetName=changeset["ChangeSetId"], StackName=stack_name)
             except Exception as ex:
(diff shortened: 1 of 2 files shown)


roger-zhangg (approved):
LGTM

integration/helpers/deployer/deployer.py
```diff
@@ -373,6 +373,16 @@ def create_and_wait_for_changeset(
             self.describe_changeset(result["Id"], stack_name)
             return result
         except deploy_exceptions.ChangeEmptyError:
+            try:
+                # Delete the most recent change set that failed to create because it was empty
+                changeset = sorted(
+                    self._client.list_change_sets(StackName=stack_name).get("Summaries"),
+                    key=lambda c: c["CreationTime"],
```
seshubaws, line 380: Can we clean up more than failed changesets here, like if we say delete for changesets whose creationtime is more than 1 week ago, then we can delete all old changesets and not have this problem anymore?

reedham-aws: I originally was considering deleting all changesets that were failed, but I was worried that it might take too miuch time/get throttled if there were hundreds of change sets in the account. 

In theory, deleting the empty ones as they're created should prevent this from happening again, and the only time it would fill up would be if the change sets are being created with actual changes, in which case I don't know if we want to delete them automatically.

seshubaws: I believe this is where changesets get created on every run of the integ tests, so we do want to delete old ones because they will eventually pile up and cause the limit exceeded exception. But we can also implement cleaners in every test account if we're worried about this throttling (think we have an internal tracking ticket to do that)

seshubaws (approved):

PR #3808, opened 2025-08-28
fix: Add intrinsic support for AWS::Serverless::Api BasePath property

reedham-aws:
### Issue #, if available
N/a
### Description of changes
The `BasePath` attribute of `Aws::Serverless::Api` does not currently accept `!Sub` as a parameter, instead leaving it as blank or returning `None`. As an example,
```yaml
Resources:
  ApiGatewayApi:
    Type: AWS::Serverless::Api
    Properties:
      StageName: prod
      CacheClusterEnabled: true
      CacheClusterSize: '0.5'
      MethodSettings:
        - ResourcePath: /
          HttpMethod: GET
          CachingEnabled: true
          CacheTtlInSeconds: 300
      Domain:
        BasePath: !Sub ${ Service }-api
        CertificateArn: !Ref CertificateArn
        DomainName: example.myinstance.com
        Route53:
          HostedZoneId: !Ref HostedZoneId
```
Transforms to the below, without the actual BasePath being added.

```json
"ApiGatewayApiBasePathMapping": {
    "Type": "AWS::ApiGateway::BasePathMapping",
    "Properties": {
        "DomainName": {
            "Ref": "ApiGatewayDomainName619a1fb284"
        },
        "RestApiId": {
            "Ref": "ApiGatewayApi"
        },
        "Stage": {
            "Ref": "ApiGatewayApiprodStage"
        }
    }
}
```

#### After updates
With the updates to allow intrinsics for this property, the output now looks like:

```json
"Parameters": {
    "Service": {
        "Type": "String",
        "Default": "NewServ"
    }
},
...
"TestApiBasePathMapping": {
    "Type": "AWS::ApiGateway::BasePathMapping",
    "Properties": {
        "BasePath": {
            "Fn::Sub": "NewServ-api"
         },
        "DomainName": {
            "Ref": "ApiGatewayDomainName54e3d11755"
        },
        "RestApiId": {
            "Ref": "TestApi"
        },
        "Stage": {
            "Ref": "TestApiprodStage"
        }
    }
}
```

Manually ran the transform with and without the updated changes and compared the templates. 

Updated the transform tests to include this new case. I didn't add any new tests because there are other locations that have `BasePath` without the intrinsics, so I figured it was fine to add this in the `api_with_basic_custom_domain_intrinsics` tests. 

Also confirmed that the added tests failed with the older transform version.

### Checklist

    - [x] Using correct values
    - [ ] Using wrong values

### Examples?

### DIFF
--- a/samtranslator/model/api/api_generator.py
+++ b/samtranslator/model/api/api_generator.py
@@ -535,13 +535,13 @@ class ApiGenerator:
 
         self._set_optional_domain_properties(domain)
 
-        basepaths: Optional[List[str]]
+        basepaths: Optional[List[Any]]
         basepath_value = self.domain.get("BasePath")
         # Create BasepathMappings
-        if self.domain.get("BasePath") and isinstance(basepath_value, str):
+        if isinstance(basepath_value, str) or is_intrinsic(basepath_value):
             basepaths = [basepath_value]
-        elif self.domain.get("BasePath") and isinstance(basepath_value, list):
-            basepaths = cast(Optional[List[Any]], basepath_value)
+        elif isinstance(basepath_value, list):
+            basepaths = basepath_value
         else:
             basepaths = None
 
@@ -554,13 +554,18 @@ class ApiGenerator:
             basepath_mapping = self._create_basepath_mapping(api_domain_name, rest_api, None, None)
             basepath_resource_list.extend([basepath_mapping])
         else:
-            sam_expect(basepaths, self.logical_id, "Domain.BasePath").to_be_a_list_of(ExpectedType.STRING)
             for basepath in basepaths:
-                # Remove possible leading and trailing '/' because a base path may only
-                # contain letters, numbers, and one of "$-_.+!*'()"
-                path = "".join(e for e in basepath if e.isalnum())
-                mapping_basepath = path if normalize_basepath else basepath
-                logical_id = "{}{}{}".format(self.logical_id, path, "BasePathMapping")
+                if is_intrinsic(basepath):
+                    mapping_basepath = basepath
+                    logical_id = self.logical_id + "BasePathMapping"
+                else:
+                    sam_expect(basepaths, self.logical_id, "Domain.BasePath").to_be_a_list_of(ExpectedType.STRING)
+                    # Remove possible leading and trailing '/' because a base path may only
+                    # contain letters, numbers, and one of "$-_.+!*'()"
+                    path = "".join(e for e in basepath if e.isalnum())
+                    mapping_basepath = path if normalize_basepath else basepath
+                    logical_id = "{}{}{}".format(self.logical_id, path, "BasePathMapping")
+
                 basepath_mapping = self._create_basepath_mapping(
                     api_domain_name, rest_api, logical_id, mapping_basepath
                 )
@@ -810,7 +815,7 @@ class ApiGenerator:
         api_domain_name: PassThrough,
         rest_api: ApiGatewayRestApi,
         logical_id: Optional[str],
-        basepath: Optional[str],
+        basepath: Optional[Any],
     ) -> ApiGatewayBasePathMapping:
 
         basepath_mapping: ApiGatewayBasePathMapping
--- a/samtranslator/model/api/api_generator.py
+++ b/samtranslator/model/api/api_generator.py
@@ -554,12 +554,13 @@ class ApiGenerator:
             basepath_mapping = self._create_basepath_mapping(api_domain_name, rest_api, None, None)
             basepath_resource_list.extend([basepath_mapping])
         else:
+            if not is_intrinsic(basepaths[0]):
+                sam_expect(basepaths, self.logical_id, "Domain.BasePath").to_be_a_list_of(ExpectedType.STRING)
             for basepath in basepaths:
                 if is_intrinsic(basepath):
                     mapping_basepath = basepath
                     logical_id = self.logical_id + "BasePathMapping"
                 else:
-                    sam_expect(basepaths, self.logical_id, "Domain.BasePath").to_be_a_list_of(ExpectedType.STRING)
                     # Remove possible leading and trailing '/' because a base path may only
                     # contain letters, numbers, and one of "$-_.+!*'()"
                     path = "".join(e for e in basepath if e.isalnum())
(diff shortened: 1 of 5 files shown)


samtranslator/model/api/api_generator.py
```diff
@@ -535,13 +535,13 @@ def _construct_api_domain(  # noqa: PLR0912, PLR0915
...
+        basepaths: Optional[List[Any]]
         basepath_value = self.domain.get("BasePath")
         # Create BasepathMappings
-        if self.domain.get("BasePath") and isinstance(basepath_value, str):
+        if isinstance(basepath_value, str) or is_intrinsic(basepath_value):
             basepaths = [basepath_value]
-        elif self.domain.get("BasePath") and isinstance(basepath_value, list):
-            basepaths = cast(Optional[List[Any]], basepath_value)
+        elif isinstance(basepath_value, list):
+            basepaths = basepath_value
```
seshubaws, line 544: should basepath_value not be a list here?

reedham-aws: Maybe misunderstanding your comment, but it's already a list based on the condition of the `elif`.

seshubaws: ah yeah, sorry missed that

seshubaws (dismissed):

samtranslator/model/api/api_generator.py
```diff
@@ -554,13 +554,18 @@ def _construct_api_domain(  # noqa: PLR0912, PLR0915
...
-                # Remove possible leading and trailing '/' because a base path may only
-                # contain letters, numbers, and one of "$-_.+!*'()"
-                path = "".join(e for e in basepath if e.isalnum())
-                mapping_basepath = path if normalize_basepath else basepath
-                logical_id = "{}{}{}".format(self.logical_id, path, "BasePathMapping")
+                if is_intrinsic(basepath):
+                    mapping_basepath = basepath
+                    logical_id = self.logical_id + "BasePathMapping"
+                else:
+                    sam_expect(basepaths, self.logical_id, "Domain.BasePath").to_be_a_list_of(ExpectedType.STRING)
```
Vandita2020: This validation should either use `basepath` (the current item) instead of `basepaths`, or move the list validation outside the loop to avoid redundant validation on every iteration. Currently, we're validating the entire list multiple times unnecessarily.

reedham-aws: Good callout, it should be `basepath`

reedham-aws: Fixed this to be run outside of the loop, but only when the object inside is not intrinsic

samtranslator/model/api/api_generator.py
```diff
@@ -535,13 +535,13 @@ def _construct_api_domain(  # noqa: PLR0912, PLR0915
 
         self._set_optional_domain_properties(domain)
 
-        basepaths: Optional[List[str]]
+        basepaths: Optional[List[Any]]
```
valerena, line 538: This should probably be `Optional[List[Intrinsicable[str]]]` instead of `Optional[List[any]]`.

from [samtranslator/utils/types.py](https://github.com/aws/serverless-application-model/blob/4421f575e5921918d0ee587f0b794e3c7e3cd7dc/samtranslator/utils/types.py#L7) (it's already imported on this file)

Vandita2020 (approved):

samtranslator/model/api/api_generator.py
```diff
@@ -554,13 +554,19 @@ def _construct_api_domain(  # noqa: PLR0912, PLR0915
...
+            if not is_intrinsic(basepaths[0]):
+                sam_expect(basepaths, self.logical_id, "Domain.BasePath").to_be_a_list_of(ExpectedType.STRING)
             for basepath in basepaths:
-                # Remove possible leading and trailing '/' because a base path may only
-                # contain letters, numbers, and one of "$-_.+!*'()"
-                path = "".join(e for e in basepath if e.isalnum())
-                mapping_basepath = path if normalize_basepath else basepath
-                logical_id = "{}{}{}".format(self.logical_id, path, "BasePathMapping")
+                if is_intrinsic(basepath):
+                    mapping_basepath = basepath
```
valerena, line 561: So the parameter `NormalizeBasePath` is not used anymore if the `BasePath` is passed as an intrinsic?

samtranslator/model/api/api_generator.py
```diff
@@ -554,13 +554,19 @@ def _construct_api_domain(  # noqa: PLR0912, PLR0915
...
+                    logical_id = self.logical_id + "BasePathMapping"
```
valerena, line 562: what happens if there are different base paths all defined with intrinsics? Will they just have the same logical id? Should we add some sort of hash for the intrinsic values here, so we actually create different resources?

samtranslator/model/api/api_generator.py
```diff
@@ -810,7 +816,7 @@ def _create_basepath_mapping(
         api_domain_name: PassThrough,
         rest_api: ApiGatewayRestApi,
         logical_id: Optional[str],
-        basepath: Optional[str],
+        basepath: Optional[Any],
```
valerena, line 819:
```suggestion
        basepath: Optional[Intrinsicable[str]],
```

tests/translator/input/api_with_basic_custom_domain_intrinsics.yaml
```diff
@@ -58,7 +58,7 @@ Resources:
         DomainName: !Sub 'example-${AWS::Region}.com'
         CertificateArn: !Ref MyDomainCert
         EndpointConfiguration: !Ref EndpointConf
-        BasePath: [/get, /fetch]
+        BasePath: !Sub '${AWS::Region}-api'
```
valerena, line 61: Can you still keep the list, but do with intrinsics instead? 

Something like 
```suggestion
        BasePath:
            - !Sub '${AWS::Region}-get'
            - !Sub '${AWS::Region}-fetch'
```

samtranslator/model/api/api_generator.py
```diff
@@ -535,13 +535,13 @@ def _construct_api_domain(  # noqa: PLR0912, PLR0915
 
         self._set_optional_domain_properties(domain)
 
-        basepaths: Optional[List[str]]
+        basepaths: Optional[List[Any]]
         basepath_value = self.domain.get("BasePath")
         # Create BasepathMappings
-        if self.domain.get("BasePath") and isinstance(basepath_value, str):
+        if isinstance(basepath_value, str) or is_intrinsic(basepath_value):
```
valerena, line 541: Thoughts on adding this same support on [http_api_generator.py](https://github.com/aws/serverless-application-model/blob/develop/samtranslator/model/api/http_api_generator.py#L328-L334) file?

samtranslator/model/api/api_generator.py
reedham-aws: Fixed

reedham-aws:
Closed due to realization that we don't actually want to support intrinsics for `BasePath`. More reasoning can be found in #2533

PR #3815, opened 2025-09-09
fix(integration): resolve stack cleanup issues in connector tests

bnusunny:
### Description of changes
fix(integration): resolve stack cleanup issues in connector tests

- Remove @retry_once decorators from test methods to preserve unittest lifecycle
- Implement separate verification methods with robust retry logic using tenacity
- Apply repository-standard retry configuration (5 attempts, exponential backoff + jitter)
- Ensure tearDown() always executes for proper CloudFormation stack cleanup
- Fix inconsistent stack removal in connector_httpapi_to_function.yaml tests

This resolves the issue where failed test retries prevented tearDown() execution, leaving orphaned stacks in CREATE_COMPLETE state while only successful retry stacks were cleaned up.

### Checklist

    - [ ] Using correct values
    - [ ] Using wrong values

### DIFF
--- a/integration/combination/test_connectors.py
+++ b/integration/combination/test_connectors.py
@@ -1,19 +1,31 @@
+import logging
 from time import sleep
 from unittest import SkipTest
 from unittest.case import skipIf
 
 from parameterized import parameterized
-from tenacity import retry, retry_if_exception, stop_after_attempt
+from tenacity import (
+    after_log,
+    retry,
+    retry_if_exception,
+    stop_after_attempt,
+    wait_exponential,
+    wait_random,
+)
 
 from integration.config.service_names import SCHEDULE_EVENT
 from integration.conftest import clean_bucket
 from integration.helpers.base_test import S3_BUCKET_PREFIX, BaseTest, nonblocking
 from integration.helpers.resource import current_region_does_not_support, generate_suffix
 
-retry_once = retry(
-    stop=stop_after_attempt(2),
-    # unittest raises SkipTest for skipping tests
+LOG = logging.getLogger(__name__)
+
+retry_with_backoff = retry(
+    stop=stop_after_attempt(5),
+    wait=wait_exponential(multiplier=1, min=4, max=16) + wait_random(0, 1),
     retry=retry_if_exception(lambda e: not isinstance(e, SkipTest)),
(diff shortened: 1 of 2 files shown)


reedham-aws:
Ah, so refactoring the retry decorator to a separate method will allow the `tearDown()` to run on every created stack created under the top level test, successful or not?

integration/combination/test_connectors.py
```diff
@@ -31,19 +43,18 @@ class TestConnectorsWithEventBus(BaseTest):
             ("combination/connector_function_to_eventbus_write",),
         ]
     )
-    @retry_once
     def test_connector_by_invoking_a_function_with_eventbus(self, template_file_path):
         self.create_and_verify_stack(template_file_path)
```
roger-zhangg, line 47: are we assuming stack creation won't fail so we can safely remove the retry?

integration/combination/test_connectors.py
```diff
@@ -88,46 +108,26 @@ def tearDown(self):
...
-        self.assertEqual(response.get("StatusCode"), 200)
-        self.assertEqual(response.get("FunctionError"), None)
+        self.verify_lambda_invocation(lambda_function_name)
 
     @parameterized.expand(
         [
             ("combination/connector_function_to_location_place_index",),
         ]
     )
-    @retry_once
```
roger-zhangg, line 113: same, what if stack creation is failed?

bnusunny:
Yes. 

> Ah, so refactoring the retry decorator to a separate method will allow the `tearDown()` to run on every created stack created under the top level test, successful or not?

integration/combination/test_connectors.py
bnusunny: The stack creation can fail and the test fails. The tearDown method will run even if the stack creation failed. Looking at the other integration tests in the repo, it follows the same pattern to allow the test fail if the stack creation fails.

roger-zhangg: So we can possibly still retry here in case the stack creation failed. We need to make tear down work consistently in this case. I think this would be even better

bnusunny: This is actually covered by the pipeline, which will rerun failed tests one more time.

PR #3822, opened 2025-10-04
fix: add `Enables` to MSK model. Use this property for MSK events in integ tests

valerena:
MSK tests are created more resources when they start polling (and the Topic we're connecting from doesn't even exist), so let's create the events with `Enabled: false`, so the polling doesn't actually start.

### Issue #, if available

Problems seen in our integration testing infrastructure.

### Description of changes

- Add `Enabled: false` to MSK templates in integration tests
- Add `Enabled` to the schema for MSK. Even though this property is supported, it wasn't part of the schema.
- Adding `s3:putObject` permissions to the test with an S3 destination for its event (`function_with_msk_trigger_and_s3_onfailure_events_destinations.yaml`).

- `pytest --no-cov integration/combination/test_function_with_msk.py -vv`

### Checklist

    - [ ] Using correct values
    - [ ] Using wrong values

### Examples?

### DIFF
--- a/integration/resources/templates/combination/function_with_msk_trigger_and_s3_onfailure_events_destinations.yaml
+++ b/integration/resources/templates/combination/function_with_msk_trigger_and_s3_onfailure_events_destinations.yaml
@@ -27,6 +27,12 @@ Resources:
               logs:CreateLogStream, logs:PutLogEvents, s3:ListBucket]
             Effect: Allow
             Resource: '*'
+          - Action: s3:PutObject
+            Effect: Allow
+            Resource:
+              Fn::GetAtt:
+              - PreCreatedS3Bucket
+              - Arn
       ManagedPolicyArns:
       - !Sub arn:${AWS::Partition}:iam::aws:policy/service-role/AWSLambdaBasicExecutionRole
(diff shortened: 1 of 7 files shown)


reedham-aws (approved):

reedham-aws:
Approved pending reformat

integration/resources/templates/combination/function_with_msk_trigger_and_s3_onfailure_events_destinations.yaml
```diff
@@ -27,6 +27,11 @@ Resources:
               logs:CreateLogStream, logs:PutLogEvents, s3:ListBucket]
             Effect: Allow
             Resource: '*'
+          - Action: [s3:PutObject, s3:ListBucket]
```
seshubaws, line 30: was accessing s3 failing before?

valerena: Yes, it was. The whole class is marked as `@non_blocking` so even when the test failing, it was failing the complete tests.

PR #3736, opened 2025-03-20
Allow AWS::NoValue to omit Role property in if - FeatureRequest#3728

SherrryX:
### Checklist

- [X] Adheres to the [development guidelines](https://github.com/aws/serverless-application-model/blob/develop/DEVELOPMENT_GUIDE.md#development-guidelines)
- [X] Add/update [transform tests](https://github.com/aws/serverless-application-model/blob/develop/DEVELOPMENT_GUIDE.md#unit-testing-with-multiple-python-versions)
    - [X] Using correct values
    - [X] Using wrong values
- [ ] Add/update [integration tests](https://github.com/aws/serverless-application-model/blob/develop/INTEGRATION_TESTS.md)

### Examples?

Please reach out in the comments if you want to add an example. Examples will be 
added to `sam init` through [aws/aws-sam-cli-app-templates](https://github.com/aws/aws-sam-cli-app-templates).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #3842, opened 2025-11-10
feat: support conditional AWS::NoValue in SAM Function's IAM Role

madsid:
### Issue #, if available
https://github.com/aws/serverless-application-model/pull/3736

### Description of changes
Support `AWS::NoValue` conditional for Role property in `AWS::Serverless::Function`

Role is an optional string value for `AWS::Serverless::Function`. An execution role is auto generated [code](https://github.com/aws/serverless-application-model/blob/9f0d08c414018bf492b97a77e2aa7c7a683b3a6d/samtranslator/model/sam_resources.py#L347) when the value is None. This change adds support for `AWS::NoValue`. 

Use cases : 
1. Role string passed though as is. 
2. Role If conditional with both `AWS::NoValue` creates an execution role for both true and false case and replaces `AWS::NoValue`
3. Role if conditional with 1 `AWS::NoValue` creates an execution role and replaces `AWS::NoValue` 
4. Role value with `Fn::GetAtt` gets passed through as is.  

- Added unit tests 
- Hand tested `sam-translate.py` with templates and verified translated templates

### Checklist

    - [x] Using correct values
    - [x] Using wrong values

### Examples?


tests/translator/output/function_with_conditional_iam_role_if.json
```diff
@@ -0,0 +1,85 @@
...
+        "Tags": [
+          {
+            "Key": "lambda:createdBy",
+            "Value": "SAM"
+          }
+        ]
+      },
+      "Type": "AWS::Lambda::Function"
+    },
+    "MinimalFunctionRole": {
```
valerena, line 61: I see that this role is always created. Even when it's not actually used (when the parameter is passed). 

The ideal case (but much harder), would be to add add a Condition: https://docs.aws.amazon.com/AWSCloudFormation/latest/TemplateReference/aws-attribute-condition.html 

This is conceptually simple **until** you realize that Conditions can only be used in positive. Therefore, if the `AWS::NoValue` is on the `else`, that means that the role needs to be created only if the condition is `not True` , which requires to create a new "not" condition to be able to use that one instead.

This is pretty much where we were blocked before because it's a more complicated implementation that whoever sent a PR before didn't have the time to try it out. 

do you think you could see if you can implement something like that?

madsid: Ack, Added the conditional for generated Role in the new rev

samtranslator/model/sam_resources.py
```diff
@@ -415,6 +412,51 @@ def to_cloudformation(self, **kwargs):  # type: ignore[no-untyped-def] # noqa: P
...
+
+        if lambda_role is None:
+            resources.append(execution_role)
+            lambda_function.Role = execution_role.get_runtime_attr("arn")
+
+        if is_intrinsic_if(lambda_role):
+            resources.append(execution_role)
+
+            # We need to create and if else condition here
+            role_resolved_value = intrinsics_resolver.resolve_parameter_refs(self.Role)
```
valerena: To be consistent, let's use the same `lambda_role` instead of `self.Role`. I think they're the same at this point, but it looks more logical to not use `self` (it's not being used anywhere else on this method)

samtranslator/model/sam_resources.py
```diff
@@ -415,6 +412,51 @@ def to_cloudformation(self, **kwargs):  # type: ignore[no-untyped-def] # noqa: P
...
+        resources: List[Any],
+        conditions: Dict[str, Any],
+    ) -> None:
+        lambda_role = lambda_function.Role
+
+        if lambda_role is None:
+            resources.append(execution_role)
+            lambda_function.Role = execution_role.get_runtime_attr("arn")
+
+        if is_intrinsic_if(lambda_role):
```
valerena: I don't know how much we save in time, but I think for clarity, this could be an `elif ` to make it clear that it's a different case that the one right above.

samtranslator/model/sam_resources.py
```diff
@@ -415,6 +412,51 @@ def to_cloudformation(self, **kwargs):  # type: ignore[no-untyped-def] # noqa: P
...
+            resources.append(execution_role)
+
+            # We need to create and if else condition here
+            role_resolved_value = intrinsics_resolver.resolve_parameter_refs(self.Role)
+            role_list = role_resolved_value.get("Fn::If")
+
+            is_both_intrinsic_no_values = is_intrinsic_no_value(role_list[1]) and is_intrinsic_no_value(role_list[2])
+
+            # both are none values, we need to create a role
```
valerena: just to emphasize I would add `always`

```suggestion
            # both are none values, we always need to create a role
```

samtranslator/model/sam_resources.py
```diff
@@ -415,6 +412,51 @@ def to_cloudformation(self, **kwargs):  # type: ignore[no-untyped-def] # noqa: P
...
+        conditions: Dict[str, Any],
+    ) -> None:
+        lambda_role = lambda_function.Role
+
+        if lambda_role is None:
+            resources.append(execution_role)
+            lambda_function.Role = execution_role.get_runtime_attr("arn")
+
+        if is_intrinsic_if(lambda_role):
+            resources.append(execution_role)
```
valerena: This is not exactly correct here. In theory we need to create this role only if **one of the values is `AWS::NoValue`**. 

If the Role field has an `If` with two different roles, then we shouldn't create the new one. (so, move this `resources.append` to inside a new condition "if 1 or 2 are intrinsic_no_value"

samtranslator/model/sam_resources.py
```diff
@@ -415,6 +412,51 @@ def to_cloudformation(self, **kwargs):  # type: ignore[no-untyped-def] # noqa: P
...
+
+            # We need to create and if else condition here
+            role_resolved_value = intrinsics_resolver.resolve_parameter_refs(self.Role)
+            role_list = role_resolved_value.get("Fn::If")
```
valerena: You can save the values of this into their specific variables so we don't need to be `role_list[n]` everywhere.

Maybe something like 

```
role_condition, role_if, role_else = role_resolved_value.get("Fn::If")
```
(but feel free to decide better names)

tests/model/test_sam_resources.py
```diff
@@ -741,6 +741,119 @@ def test_function_datasource_set_with_none():
...
+
+        self.assertEqual(len(generated_roles), 1)  # Should create execution role
+
+    def test_role_explicit_arn_no_execution_role(self):
+        test_role = "arn:aws:iam::123456789012:role/existing-role"
+        self.function.Role = test_role
+
+        cfn_resources = self.function.to_cloudformation(**self.kwargs)
+        generated_roles = [x for x in cfn_resources if isinstance(x, IAMRole)]
+        lambda_function = next(r for r in cfn_resources if r.resource_type == "AWS::Lambda::Function")
```
valerena, line 775: You can do the same than for the roles with the functions: `generated_functions = [x for x in cfnResources if isinstance(x, LambdaFunction)]`
and then you can also confirm that there's only one created. 

But it's not a big deal. We can probably change it in the future if needed.

tests/model/test_sam_resources.py
```diff
@@ -741,6 +741,119 @@ def test_function_datasource_set_with_none():
...
+
+    def test_role_fn_if_no_aws_no_value_keeps_original(self):
+        role_conditional = {
+            "Fn::If": ["Condition", "arn:aws:iam::123456789012:role/existing-role", {"Ref": "iamRoleArn"}]
+        }
+        self.function.Role = role_conditional
+
+        template = {"Conditions": {"Condition": True}}
+        kwargs = dict(self.kwargs)
+        kwargs["conditions"] = template.get("Conditions", {})
```
valerena: You don't use the variable `template` anywhere else, so you could just remove it and define the conditions below. (same in other tests)

```suggestion
        kwargs = dict(self.kwargs)
        kwargs["conditions"] = {"Condition": True}
```

tests/model/test_sam_resources.py
```diff
@@ -741,6 +741,119 @@ def test_function_datasource_set_with_none():
...
+    """
+    Tests for resolving IAM role property values in SamFunction
+    """
+
+    def setUp(self):
+        self.function = SamFunction("foo")
+        self.function.CodeUri = "s3://foobar/foo.zip"
+        self.function.Runtime = "foo"
+        self.function.Handler = "bar"
+        self.template = {"Conditions": {}}
```
valerena: You don't need this `self.template` variable since it's not being used. Right below you use it to get `Conditions` from here, but it will be just `{}`, so you can just do that (`"conditions": {}`)

valerena (dismissed):

samtranslator/model/sam_resources.py
```diff
@@ -415,6 +412,52 @@ def to_cloudformation(self, **kwargs):  # type: ignore[no-untyped-def] # noqa: P
...
+        resources: List[Any],
+        conditions: Dict[str, Any],
+    ) -> None:
+        lambda_role = lambda_function.Role
+
+        if lambda_role is None:
+            resources.append(execution_role)
+            lambda_function.Role = execution_role.get_runtime_attr("arn")
+
+        elif is_intrinsic_if(lambda_role):
```
roger-zhangg: you can avoid a indentation by early return
```suggestion
        if lambda_role is None:
            resources.append(execution_role)
            lambda_function.Role = execution_role.get_runtime_attr("arn")
            return
            
        if not is_intrinsic_if(lambda_role):
            return
```

samtranslator/model/sam_resources.py
```diff
@@ -415,6 +412,52 @@ def to_cloudformation(self, **kwargs):  # type: ignore[no-untyped-def] # noqa: P
 
         return resources
 
+    def _make_lambda_role(
+        self,
+        lambda_function: LambdaFunction,
```
roger-zhangg, line 429: this implementation has some unnecessary change IMO, we can keep the `if lambda_function.Role is None` part as is, then `_make_lambda_role` can return the created role, so we don't need to pass in `lambda_function` in this function.

samtranslator/model/sam_resources.py
```diff
@@ -388,16 +388,13 @@ def to_cloudformation(self, **kwargs):  # type: ignore[no-untyped-def] # noqa: P
...
-            )
-            lambda_function.Role = execution_role.get_runtime_attr("arn")
-            resources.append(execution_role)
+        execution_role = self._construct_role(
+            managed_policy_map,
+            event_invoke_policies,
+            intrinsics_resolver,
+            get_managed_policy_map,
+        )
+        self._make_lambda_role(lambda_function, intrinsics_resolver, execution_role, resources, conditions)
```
roger-zhangg: If we want to preserve the original code structure
```suggestion
        execution_role = self._construct_role(
            managed_policy_map,
            event_invoke_policies,
            intrinsics_resolver,
            get_managed_policy_map,
        )
        if lambda_function.Role is None:
            lambda_function.Role = execution_role.get_runtime_attr("arn")
            resources.append(execution_role)
        elif is_intrinsic_if(lambda_role):
            lambda_function.Role = self._make_lambda_role(lambda_function, intrinsics_resolver, execution_role, resources, conditions)
```

samtranslator/model/sam_resources.py
```diff
@@ -415,6 +412,52 @@ def to_cloudformation(self, **kwargs):  # type: ignore[no-untyped-def] # noqa: P
...
+            if is_both_intrinsic_no_values:
+                lambda_function.Role = execution_role.get_runtime_attr("arn")
+
+            # first value is none so we should create condition ? create : [2]
+            # create a condition for IAM role to only create on if case
+            elif is_intrinsic_no_value(role_if):
+                lambda_function.Role = make_conditional(
+                    role_condition, execution_role.get_runtime_attr("arn"), role_else
+                )
+                execution_role.set_resource_attribute("Condition", f"{role_condition}")
```
roger-zhangg: Seems we are modifying the execution role but it seems not used afterwards?

samtranslator/model/sam_resources.py
roger-zhangg: One other concern is we are modifying values for lambda_function, execution_role, resources, conditions ( which are the function's input) inside this function which returns None, but all these values might have changed after running this function. This might make this part of the code harder to review in the future. If we need to change all these variable's value in place, I would perfer to implement this directly in `to_cloudformation` instead of this subfunction. If you really prefer to have this subfunction. Maybe we can return the new value instead of modifying them inplace. So future reviewers could easily understand these variables are modified in this subfunction.

samtranslator/model/sam_resources.py
```diff
@@ -388,16 +388,30 @@ def to_cloudformation(self, **kwargs):  # type: ignore[no-untyped-def] # noqa: P
...
+            role_changes = self._make_lambda_role(lambda_function, intrinsics_resolver, execution_role)
+
+            if role_changes["should_append_role"]:
+                resources.append(execution_role)
+
+            if role_changes["lambda_role_value"] is not None:
+                lambda_function.Role = role_changes["lambda_role_value"]
+
+            if role_changes["execution_role_condition"] is not None:
+                execution_role.set_resource_attribute("Condition", role_changes["execution_role_condition"])
```
valerena: You can change this a little bit so you return a `role_changes["execution_role_resource"]` that it's either `execution_role` or `None`. And you set the `Condition` attribute inside the function not here outside. 

so then you have
```
if role_changes["execution_role_resource"] is not None:
    resources.append(role_changes["execution_role_resource"])
```

and that role comes ready to be added, you can change its `resource_attribute("Condition,` inside the function instead of here.

samtranslator/model/sam_resources.py
```diff
@@ -415,6 +429,54 @@ def to_cloudformation(self, **kwargs):  # type: ignore[no-untyped-def] # noqa: P
...
+        """
+        lambda_role = lambda_function.Role
+        execution_role_arn = execution_role.get_runtime_attr("arn")
+
+        result = {
+            "should_append_role": False,
+            "lambda_role_value": None,
+            "execution_role_condition": None,
+            "new_condition": None,
+        }
```
valerena: Another recommendation. Work with the simple variables first. And them to the result only at the very end for returning. Using the `result` variable for all the following manipulations makes the code look so much worse.

samtranslator/model/sam_resources.py
```diff
@@ -415,6 +429,54 @@ def to_cloudformation(self, **kwargs):  # type: ignore[no-untyped-def] # noqa: P
...
+
+        # We need to create and if else condition here
+        role_resolved_value = intrinsics_resolver.resolve_parameter_refs(lambda_role)
+        role_condition, role_if, role_else = role_resolved_value.get("Fn::If")
```
valerena, line 450: Any reason to remove the case where both where `NoValue` that you had in previous revisions?

madsid: Missed it in the commit, fixed

samtranslator/model/sam_resources.py
madsid: We add a condition to the created role in if else case. Without the condition we will end up creating an IAM role when it's not required. For Example if the customer has ` if (ExistingCondition) { AWS::NoValue } else { !Ref <existing Role> }` we should only create the role if `ExistingCondition` is true.

samtranslator/model/sam_resources.py
madsid: Fixed in next rev.

samtranslator/model/sam_resources.py
```diff
@@ -388,16 +388,25 @@ def to_cloudformation(self, **kwargs):  # type: ignore[no-untyped-def] # noqa: P
...
-                get_managed_policy_map,
-            )
             lambda_function.Role = execution_role.get_runtime_attr("arn")
             resources.append(execution_role)
+        elif is_intrinsic_if(lambda_function.Role):
+            role_changes = self._make_lambda_role(lambda_function, intrinsics_resolver, execution_role)
+
+            if role_changes["lambda_role_value"] is not None:
+                lambda_function.Role = role_changes["lambda_role_value"]
+                resources.append(execution_role)
```
valerena: The intention here should be that this new execution role still comes in the response from the `make_lambda_role` function. It could be just the same `execution_role`, but it *could be* that it changed. We can't rely on the method changing the parameter, so if we expect a potentially modified role, we should use the response from the function.

```suggestion
                resources.append(role_changes["iam_role_resource"])
```
(you need to add this `iam_role_resource` to the response from the method too)

roger-zhangg (approved):

PR #3844, opened 2025-11-17
Fix duplicate OPTIONS method error for paths with trailing slash

### LINKED ISSUE #3816: Bug: Duplicate method 'OPTIONS' error
SAM CLI, version 1.139.0
Linux

When there are both POST and PUT, api events in AWS::Serverless::Function resource + 
Cors section in an AWS::Serverless::Api resource. 

I get an error:

```
Resource handler returned message: "Errors found during import:
Duplicate method 'OPTIONS' on resource at path '/datasets'
(Service: ApiGateway, Status Code: 400, Request ID:
xxx) (SDK Attempt Count: 1)"
(RequestToken: xxx,
HandlerErrorCode: InvalidRequest)
```

This is the template file :
```yaml
Transform:
  - AWS::Serverless-2016-10-31
  - AWS::LanguageExtensions

Globals:
  Function:
    Timeout: 60
    MemorySize: 512
    Tracing: Active
    Tags:
      GitRev: !Ref GitRev
    Environment:
      Variables:
        DbName: !Ref DbName
        MongoSecretName: !Ref MongoSecretName
        StackName: !Sub ${AWS::StackName}
        POWERTOOLS_SERVICE_NAME: !Sub ${AWS::StackName}
        POWERTOOLS_LOG_LEVEL: INFO
        Environment: !Ref Environment
        GitRev: !Ref GitRev

Parameters:
  DbName:
    Type: String
    Description: Mongodb database name
  MongoSecretName:
    Type: String
    Description: AWS secret name containing Mongodb server secrets
  VPCEndpointId:
    Type: String
    Description: vpc endpoint associated with said APIGW
    AllowedPattern: ^vpce-[A-Za-z0-9]+$
  GitRev:
    Type: String
    Default: ""
  Environment:
    Type: String
    Default: test
    AllowedValues:
      - test
      - prod

Resources:
  AppBackend:
    Properties:
      MemorySize: 2048
      PackageType: Image
      Policies:
        - AWSSecretsManagerGetSecretValuePolicy:
            SecretArn: !Sub arn:${AWS::Partition}:secretsmanager:${AWS::Region}:${AWS::AccountId}:secret:${MongoSecretName}*
      Events:
        ListDatasets:
          Type: Api
          Properties:
            Path: /datasets
            Method: post
            RestApiId: !Ref MyRestApi
            Auth:
              Authorizer: MyLambdaRequestAuth
        GetDataset:
          Type: Api
          Properties:
            Path: /datasets/{dataset_id}
            Method: get
            RestApiId: !Ref MyRestApi
            Auth:
              Authorizer: MyLambdaRequestAuth
        PutDataset:
          Type: Api
          Properties:
            Path: /datasets/
            Method: put
            RestApiId: !Ref MyRestApi
            Auth:
              Authorizer: MyLambdaRequestAuth
        DeleteDataset:
          Type: Api
          Properties:
            Path: /datasets/{dataset_id}
            Method: delete
            RestApiId: !Ref MyRestApi
            Auth:
              Authorizer: MyLambdaRequestAuth
#        ListStudies:
#          Type: Api
#          Properties:
#            Path: /studies
#            Method: post
#            RestApiId: !Ref MyRestApi
#            Auth:
#              Authorizer: MyLambdaRequestAuth
#        GetStudy:
#          Type: Api
#          Properties:
#            Path: /studies/{study_id}
#            Method: get
#            RestApiId: !Ref MyRestApi
#            Auth:
#              Authorizer: MyLambdaRequestAuth
#        PutStudy:
#          Type: Api
#          Properties:
#            Path: /studies/
#            Method: put
#            RestApiId: !Ref MyRestApi
#            Auth:
#              Authorizer: MyLambdaRequestAuth
#        DeleteStudy:
#          Type: Api
#          Properties:
#            Path: /studies/{study_id}
#            Method: delete
#            RestApiId: !Ref MyRestApi
#            Auth:
#              Authorizer: MyLambdaRequestAuth
    Metadata:
      Dockerfile: Dockerfile
      DockerContext: ./functions/be


  MsAuthorizer:
    Properties:
      MemorySize: 2048
      CodeUri: functions/authorizer/src
      Handler: app.handler
      Runtime: python3.12

  MyRestApi:
    Type: AWS::Serverless::Api
    Properties:
#      BinaryMediaTypes:
#        - application/octet-stream
      Cors:
        AllowOrigin: '''*'''
        AllowHeaders: '''Content-Type,Authorization,X-Amz-Date'''
        MaxAge: '''300'''
      StageName: v1
      TracingEnabled: true
      GatewayResponses:
        MISSING_AUTHENTICATION_TOKEN:
          ResponseTemplates:
            application/json: '{"message": "missing authentication token error OR
              unsupported API method or resource" }'
      MethodSettings:
        - HttpMethod: '*'
          LoggingLevel: INFO
          ResourcePath: /*
          DataTraceEnabled: true
          CachingEnabled: false
          MetricsEnabled: true
      EndpointConfiguration:
        Type: PRIVATE
        VPCEndpointIds:
          - !Ref VPCEndpointId
      Auth:
        Authorizers:
          MyLambdaRequestAuth:
            FunctionPayloadType: TOKEN
            Identity:
              Header: Authorization
            FunctionArn:
              !GetAtt MsAuthorizer.Arn
        ResourcePolicy:
          CustomStatements:
            - Effect: Allow
              Action: execute-api:Invoke
              Resource: !Sub "arn:aws:execute-api:${AWS::Region}:${AWS::AccountId}:*/*/*/*"
              Principal: '*'
              Condition:
                StringEquals:
                  aws:SourceVpce: !Ref VPCEndpointId
```

abhishektang:
## Description

Fixes #3816

This PR fixes the issue where enabling CORS on an API with paths that differ only by a trailing slash (e.g., `/datasets` and `/datasets/`) causes API Gateway to throw an error:
```
Duplicate method 'OPTIONS' found for resource /datasets. Only one is allowed.
```

## Problem

When CORS is enabled, SAM adds OPTIONS methods to all paths. However, API Gateway treats `/path` and `/path/` as the same resource. If both paths exist in the template, SAM was adding OPTIONS to both separately, causing a duplicate method error.

## Solution

The fix normalizes paths before adding CORS OPTIONS methods:
- Paths are normalized by removing trailing slashes (except for root `/`)
- Normalized paths are tracked in a set to skip duplicates
- Ensures only one OPTIONS method is added per normalized path

## Changes

**Modified files:**
- `samtranslator/model/api/api_generator.py`: Added path normalization in `_add_cors()` method
- `tests/model/api/test_api_generator.py`: Added unit test `test_add_cors_with_trailing_slash_paths`

## Testing

✅ **Unit Tests**
- Added new test `test_add_cors_with_trailing_slash_paths` that verifies:
  - Paths `/datasets` (POST) and `/datasets/` (PUT) are correctly handled
  - Only one OPTIONS method is added (not duplicated)
  - Test passes successfully

✅ **Integration Tests**
- All 68 existing CORS integration tests pass without regression
- Verified no breaking changes to existing functionality

✅ **Real-world Validation**
- Tested with the production template from issue #3816
- Confirmed the fix resolves the duplicate OPTIONS error
- Template now transforms successfully

## Related Issues

Closes #3816

---

**Checklist:**
- [x] Fixed on `develop` branch
- [x] Added unit test for the fix
- [x] All tests pass locally
- [x] Tested with real-world production template
- [x] No regressions in existing CORS tests
- [x] Clear commit message
- [x] Code follows project conventions

### DIFF
--- a/samtranslator/model/api/api_generator.py
+++ b/samtranslator/model/api/api_generator.py
@@ -996,7 +996,7 @@ class ApiGenerator:
 
             try:
                 editor.add_cors(  # type: ignore[no-untyped-call]
-                    normalized_path,
+                    path,
                     properties.AllowOrigin,
                     properties.AllowHeaders,
                     properties.AllowMethods,


samtranslator/model/api/api_generator.py
```diff
@@ -975,7 +975,20 @@ def _add_cors(self) -> None:
...
+            
+            # Skip if we've already processed this normalized path to avoid duplicate OPTIONS methods
+            if normalized_path in normalized_paths_processed:
+                continue
+            
+            normalized_paths_processed.add(normalized_path)
+            
             try:
                 editor.add_cors(  # type: ignore[no-untyped-call]
                     path,
```
reedham-aws: Should we be using `normalized_path` here? It seems like there could be a situation in which you have both `/path` and `/path/`, and whichever one comes first in the list is the one that is added.

abhishektang: Yes, you are right, that should be normalized_path instead of path. I have made the required changes, please review them and let me know if there are any further changes required.

samtranslator/model/api/api_generator.py
```diff
@@ -980,10 +980,23 @@ def _add_cors(self) -> None:
...
+            # Skip if we've already processed this normalized path to avoid duplicate OPTIONS methods
+            if normalized_path in normalized_paths_processed:
+                continue
+
+            normalized_paths_processed.add(normalized_path)
+
             try:
                 editor.add_cors(  # type: ignore[no-untyped-call]
-                    path,
+                    normalized_path,
```
valerena: We should continue using `path` here (instead of `normalized_path`).

From an API Gateway perspective it's equivalent, but by still using `path` we keep backwards compatibility, in the sense that existing templates (without duplicated paths) continue generating exactly the same resources before and after this change (and changing this won't remove the fact that we will skip the duplicated normalized paths, because they're handled above)

abhishektang: I see the edge case you are talking about here as I didn't consider the API Gateway treats them equally, but don't you think using path is non-deterministic and using normalized_path is deterministic. 

Would you be open to a **sorted iteration** approach instead? It gives determinism without breaking the path lookup:

for path in sorted(editor.iter_on_path(), key=lambda p: (p.rstrip("/"), p)):

This guarantees `/path` is always iterated before `/path/`, so `OPTIONS` always lands on the non-trailing-slash path, or should I be worried about the latency of the code by introducing the sort method. Happy to go with whatever direction you prefer.

valerena: I'm thinking mainly on what happens in the case when there are **no** duplicated options really. The cors configuration will be added to a normalized path, which may be different than what the user has configured. Again, in theory it shouldn't affect the behavior in the end because for API GW it's the same, but it's just *different* than what currently happens before this change.

It's usually frowned upon to change SAM's behavior (obviously except on previously failing scenarios), because even if the final state is equivalent, changing the generated resources can cause CFN to "redeploy" a resource without the user making changes, which can confuse users. (and I guess we don't have an existing test case with this, that would have changed)
Same thing happens with a "sorted" iteration. After that, the generated API GW will have the paths in a different order than before this change. 

Also, my understanding is that this will still be consistent, where the item added will be "the first one" configured, since `iter_on_path` should return the items in the same order (`dict.items()` in `iter_in_path` returns elements in the order of insertion)

valerena: Let me know if that makes sense to you, because I'm still open to be convinced if you have a strong argument.

abhishektang: Yes, that clarifies everything for me. I have made the change and pushed the new code, please review the code and kindly let me know if there are any further changes required. Thanks a lot @valerena for explaining the whole scenario!

Issue #3855, opened 2025-12-02
CapacityProvider includes null values for optional ScalingConfig property


vicheey:
### Description
When using `AWS::Serverless::CapacityProvider` with optional `ScalingConfig` properties, the SAM translator incorrectly includes `None/null` values in the generated CloudFormation template, causing deployment failures or unexpected behavior.

### Steps to reproduce
1. Create a SAM template with `AWS::Serverless::CapacityProvider`
2. Define `ScalingConfig` with only `MaxVCpuCount` (without `AverageCPUUtilization`)
3. Or define InstanceRequirements with only some properties (e.g., only AllowedTypes)
4. Run `sam build` and `sam deploy`
5. The generated CloudFormation template includes null values for omitted properties

### Observed result
The generated CloudFormation template contains null/None values for optional properties that were not specified in the SAM template, potentially causing validation errors or unexpected behavior during deployment.

### Expected result
Optional properties that are not specified in the SAM template should be completely excluded from the generated CloudFormation template, not included with null values.

### Additional environment details
1. OS: macOS
2. If using the SAM CLI, `sam --version`: 1.149.0
3. AWS region: N/A

PR #3856, opened 2025-12-02
fix(lmi): exclude None values from CapacityProvider ScalingConfig and InstanceRequirements

vicheey:
### Issue #, if available
https://github.com/aws/serverless-application-model/issues/3855

### Description of changes
- Fixed CapacityProvider generator to properly handle None values in ScalingConfig and InstanceRequirements
- Changed dictionary membership checks from `in` operator to `.get()` method with explicit None checks
- Updated ScalingConfig serialization in sam_resources.py to use `exclude_none=True` parameter
- This ensures that when only MaxVCpuCount is specified (without AverageCPUUtilization), the ScalingMode correctly defaults to "Auto" instead of incorrectly setting "Manual" with null target values

- Added two new test cases (CapacityProviderTwo and PartialCapacityProviderThree) to `capacity_provider_full.yaml`
  - CapacityProviderTwo tests ScalingConfig with only AverageCPUUtilization (no MaxVCpuCount) → generates ScalingMode: "Manual"
  - PartialCapacityProviderThree tests ScalingConfig with only MaxVCpuCount (no AverageCPUUtilization) → generates ScalingMode: "Auto"
- Verified generated CloudFormation templates across all partitions (aws, aws-cn, aws-us-gov) correctly set ScalingMode and exclude unspecified properties
- All transform tests pass with correct output

### Checklist
- [x] Review the generative AI contribution guidelines
- [x] Adheres to the development guidelines
- [x] Add/update transform tests
  - [x] Using correct values
  - [x] Using wrong values
- [ ] Add/update integration tests

### Examples?

**Before this fix:**
```yaml
# SAM Template
Resources:
  MyCapacityProvider:
    Type: AWS::Serverless::CapacityProvider
    Properties:
      ScalingConfig:
        MaxVCpuCount: 10
        # AverageCPUUtilization intentionally omitted - should use Auto scaling
```
Generated CloudFormation would incorrectly produce:
```
"CapacityProviderScalingConfig": {
  "MaxVCpuCount": 10,
  "ScalingMode": "Manual",  // ❌ Wrong! Should be "Auto"
  "ScalingPolicies": [] // ❌ Wrong! Should not be included 
}
```

**After this fix:**
```
"CapacityProviderScalingConfig": {
  "MaxVCpuCount": 10,
  "ScalingMode": "Auto"  // ✅ Correct! Auto scaling when no CPU target specified
}
```

### DIFF
--- a/samtranslator/model/sam_resources.py
+++ b/samtranslator/model/sam_resources.py
@@ -1504,7 +1504,7 @@ class SamCapacityProvider(SamResourceMacro):
             instance_requirements=(
                 model.InstanceRequirements.dict(exclude_none=True) if model.InstanceRequirements else None
             ),
-            scaling_config=model.ScalingConfig.dict() if model.ScalingConfig else None,
+            scaling_config=model.ScalingConfig.dict(exclude_none=True) if model.ScalingConfig else None,
             kms_key_arn=passthrough_value(model.KMSKeyArn),
             depends_on=self.depends_on,
             resource_attributes=self.resource_attributes,
--- a/samtranslator/model/capacity_provider/generators.py
+++ b/samtranslator/model/capacity_provider/generators.py
@@ -133,13 +133,13 @@ class CapacityProviderGenerator:
         """
         instance_requirements = {}
 
-        if "Architectures" in self.instance_requirements:
+        if self.instance_requirements.get("Architectures") is not None:
             instance_requirements["Architectures"] = self.instance_requirements["Architectures"]
 
-        if "AllowedTypes" in self.instance_requirements:
+        if self.instance_requirements.get("AllowedTypes") is not None:
             instance_requirements["AllowedInstanceTypes"] = self.instance_requirements["AllowedTypes"]
 
-        if "ExcludedTypes" in self.instance_requirements:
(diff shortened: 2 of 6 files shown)


samtranslator/model/sam_resources.py
```diff
@@ -1504,7 +1504,7 @@ def to_cloudformation(self, **kwargs: Any) -> List[Resource]:
             instance_requirements=(
                 model.InstanceRequirements.dict(exclude_none=True) if model.InstanceRequirements else None
             ),
-            scaling_config=model.ScalingConfig.dict() if model.ScalingConfig else None,
+            scaling_config=model.ScalingConfig.dict(exclude_none=True) if model.ScalingConfig else None,
```
reedham-aws, line 1507: What does the `exclude_none` do here?

valerena: https://docs.pydantic.dev/1.10/usage/exporting_models/#modeldict

It removes from the returned dictionary all the fields that are `None`, so if a field is not set, then the field won't be exported here, instead of being there but having `None` as a value

valerena (approved):

Issue #3831, opened 2025-10-17
Support Python 3.14 / Update Pydantic


meeuw:
[Python 3.14](https://www.python.org/downloads/release/python-3140/) has been released and I've tried running the [build.yml](https://github.com/aws/serverless-application-model/blob/b0231f56f07b166f9a35669385089c4d66bb3f6c/.github/workflows/build.yml) workflow but it seems to fail with the following error message:

```
make pr
  shell: /usr/bin/bash -e {0}
  env:
    pythonLocation: /opt/hostedtoolcache/Python/3.14.0/x64
    PKG_CONFIG_PATH: /opt/hostedtoolcache/Python/3.14.0/x64/lib/pkgconfig
    Python_ROOT_DIR: /opt/hostedtoolcache/Python/3.14.0/x64
    Python2_ROOT_DIR: /opt/hostedtoolcache/Python/3.14.0/x64
    Python3_ROOT_DIR: /opt/hostedtoolcache/Python/3.14.0/x64
    LD_LIBRARY_PATH: /opt/hostedtoolcache/Python/3.14.0/x64/lib
# Checking latest schema was generated (run `make schema` if this fails)
mkdir -p .tmp
python -m samtranslator.internal.schema_source.schema --sam-schema .tmp/sam.schema.json --cfn-schema schema_source/cloudformation.schema.json --unified-schema .tmp/schema.json
/home/runner/work/serverless-application-model/serverless-application-model/samtranslator/compat.py:2: UserWarning: Core Pydantic V1 functionality isn't compatible with Python 3.14 or greater.
  from pydantic import v1 as pydantic
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/home/runner/work/serverless-application-model/serverless-application-model/samtranslator/internal/schema_source/schema.py", line 10, in <module>
    from samtranslator.internal.schema_source import (
    ...<10 lines>...
    )
  File "/home/runner/work/serverless-application-model/serverless-application-model/samtranslator/internal/schema_source/aws_serverless_api.py", line 5, in <module>
    from samtranslator.internal.schema_source.aws_serverless_connector import EmbeddedConnector
  File "/home/runner/work/serverless-application-model/serverless-application-model/samtranslator/internal/schema_source/aws_serverless_connector.py", line 16, in <module>
    class ResourceReference(BaseModel):
    ...<7 lines>...
[...]

PR #3859, opened 2025-12-12
fix: partially support Python3.14

valerena:
Allow the transform to be used in Python3.14, except GraphQLApi and Capacity Providers functionalities.

`make pr` still doesn't work, but the transform can be used successfully in Python 3.14 (if not using the above features)

```
bin/sam-translate.py --template ./template.yaml
```

### Issue #, if available
#3831 

### Description of changes

Add a try/except when importing the model classes. These classes are only used for Capacity Provider and GraphQLApi, so those items are still not supported when using Python3.14. But at least for now the rest of the transform works correctly. 

I also removed a variable `resource_property_schema` that was being initialized for the Lambda Capacity Provider class definition, but it never changed value and it was used only once. So I replaced it with the value imported directly instead of saving it in an intermediate variable.

I also updated `jsonschema` version because the previous version wasn't supported in python3.14. I also removed `cloudformation-cli` from dev dependencies, because it doesn't support latest versions of Python (and it's not actually used for anything)

Manual tests
`make pr` is still failing, because of course all the Capacity Provider and GraphQLApi tests that now are not working. And also a couple of other tests that import the model too, that are failing with pydantic v1. That's one of the reasons why we can't count python3.14 as really "supported" yet.

### Checklist

    - [ ] Using correct values
    - [ ] Using wrong values

### Examples?

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -3,5 +3,6 @@ jsonschema<5,>=4.23
 typing_extensions>=4.4 # 3.8 doesn't have Required, TypeGuard and ParamSpec
 
 # resource validation & schema generation
-# 1.10.15 and 1.10.17 included breaking change from pydantic, more info: https://github.com/aws/serverless-application-model/issues/3617
-pydantic>=1.8,<3,!=1.10.15,!=1.10.17
+# we should remove support for Python 3.8 soon (and 3.9), but supporting it for now
+pydantic>=2.10.6; python_version>="3.8"
+pydantic~=2.12.5; python_version>="3.9"
--- a/requirements/dev.txt
+++ b/requirements/dev.txt
@@ -23,7 +23,7 @@ black==24.3.0
 ruamel.yaml==0.17.21  # It can parse yaml while perserving comments
 
 # type check
-mypy~=1.3.0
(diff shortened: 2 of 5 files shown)


roger-zhangg (approved):
LGTM

samtranslator/model/sam_resources.py
```diff
@@ -34,11 +36,19 @@
...
-)
+
+# Pydantic 1 doesn't support Python 3.14 so these imports will fail until we migrate to v2
+try:
+    from samtranslator.internal.schema_source import (
+        aws_serverless_capacity_provider,
+        aws_serverless_function,
+        aws_serverless_graphqlapi,
+    )
+except Exception:
```
roger-zhangg: small thing
```suggestion
except ImportError:
```

valerena: Actually I tried this initially, but the exception here is not an `ImportError`. It's not that the library doesn't exist.. This actually just sends a `RuntimeException` coming from somewhere in pydantic code. And at that point I imagined that it's not really relevant where the exception came from.

seshubaws (approved):

PR #3861, opened 2025-12-15
chore: clean up cfn-lint ignored files

valerena:
When we have new features, cfn-lint might not have support to all those features yet, so we usually just ignore some test files until that's supported.. but then we rarely actually remove that so those test files just remain ignored in cfn-lint forever. So this is an overall cleanup to remove all the ignores that are not needed.

The next step would be to confirm why all these files are ignored, and eventually fix some that make sense to fix (except the ones that have some errors on purpose).

### Checklist

    - [ ] Using correct values
    - [ ] Using wrong values

### Examples?


vicheey:
I assume you ran `samdev validate --lint` with the changes and succeed, correct?

vicheey (approved):

valerena:
> I assume you ran `samdev validate --lint` with the changes and succeed, correct?

I didn't use SAM CLI. But here in the SAM-T repo I ran `make pr`, which runs `cfn-lint` on all the tests files

PR #3868, opened 2026-01-03
test: update custom domain integration tests

valerena:
The tests were outdated, so we updated domains and hosted zones, so we need to update the expected values here.

### Issue #, if available
Integration tests were failing.

### Description of changes
Updated the hosted zone and domain name for the tests, so the value of generated hashes for ApiGW Domain (depends on the domain name) and RecordSetGroup (depends on the hosted zone id) had to be updated.

```
$ pytest --no-cov --reruns 0 integration/combination/test_custom_rest_api_domains.py --prefix SamTranslatorLambda --internal --parameter-values template_parameter_values_new.yaml --stage prod
$ pytest --no-cov --reruns 0 integration/combination/test_custom_http_api_domains_test.py --prefix SamTranslatorLambda --internal --parameter-values template_parameter_values_new.yaml --stage prod
```
### Checklist

    - [ ] Using correct values
    - [ ] Using wrong values

### Examples?


seshubaws:
Looks like a linting error is causing the tests to fail

integration/combination/test_custom_rest_api_domains.py
```diff
@@ -46,7 +46,7 @@ def test_custom_rest_api_domains_regional(self):
         if "FeatureToggle" in self.pipeline_prefix:
             self.assertEqual("ftl.sam-gamma-regional.com", result["domainName"])
         else:
-            self.assertEqual("sam-gamma-regional.com", result["domainName"])
+            self.assertEqual(f"sam-gamma-regional.{CUSTOM_DOMAIN_TOP_LEVEL}", result["domainName"])
```
vicheey, line 49: Nit: For the test, shouldn't we use `tooling.lambda.aws.dev` instead of the variable? So that if we mistakenly modify the `CUSTOM_DOMAIN_TOP_LEVEL`, the test should be able catch it.

valerena: That variable is only used here for the tests and not anywhere in the code. The only purpose is to not have to repeat the full name everywhere, and we can have a single place with the domain that we use for all tests.

PR #3893, opened 2026-03-17
chore: drop Python 3.9 support for running SAM, add Python 3.12, 3.13 to CI

roger-zhangg:
## What

Drop Python 3.9 as a supported version for **running** the SAM translator itself, and extend CI coverage to Python 3.12, 3.13, and 3.14.

## Important distinction

This change only drops Python 3.9 as a version for running SAM. **Python 3.9 as a Lambda runtime in SAM templates is still fully supported.** Test fixtures that use `Runtime: python3.9` are intentionally left unchanged.

## Changes

- **setup.py** — `python_requires` bumped to `>=3.10`, classifiers updated to 3.10–3.14
- **.github/workflows/build.yml** — test matrix now runs 3.10, 3.11, 3.12, 3.13, 3.14(after pydantic v1 support added)
- **pyproject.toml** — Black `target_version` updated to py310–py314
- **ruff.toml / tests/ruff.toml / integration/ruff.toml** — `target-version` bumped to `py310`
- **requirements/base.txt** — simplified pydantic dependency (removed 3.8/3.9 conditional pins)
- **DEVELOPMENT_GUIDE.md** — updated supported versions and pyenv install instructions


integration/combination/test_function_with_cwe_dlq_generated.py
```diff
@@ -36,7 +36,7 @@ def test_function_with_cwe(self):
 
         # checking policy action
         actions = dlq_policy_statement["Action"]
-        action_list = actions if isinstance(actions, list) == list else [actions]
+        action_list = actions if actions is list else [actions]
```
valerena: I don't think this is correct. 

`type(actions) is list` would work, but `actions is list` will never be True (and it's not the same that we had before).

This is probably why you had to add an exception for this in the `ruff.toml` file, but `ruff` complaining about this usage of `is` is exactly what we want on this line

integration/ruff.toml
```diff
@@ -24,4 +24,5 @@ lint.select = [
 "**/*.py" = [
     "S101", # Use of `assert` detected
     "PLR", # pylint-refactor
+    "E721", # Use `is` and `is not` for type comparisons
```
valerena: I added another comment mentioning that we might not want to add this (at least another part when there's actually an error that should have been caught by this). Are there other parts that are failing with this rule?

integration/combination/test_function_with_cwe_dlq_generated.py
valerena: wait, I'm reading this code again. The previous code was very weird too! 
`isinstance(actions, list) == list` will always be false too. Because `isinstance` returns a boolean!

Probably that's why this new code is so weird too. I guess we never go in the `if` and apparently the test works well in that case. (because it's waiting only one element in the next line). 

Let's still fix this, but at least I see that the implementation had problems before

Allankw-dev (commented):
Including Python 3.13 is great for forward compatibility, but you may want to ensure dependencies fully support it to avoid unstable builds.

Allankw-dev:
Consider adding dependency caching (e.g., pip cache) to speed up workflow execution across runs and also It might be helpful to document what( make init) and ( make pr ) do, or reference them in the README for better clarity.

ruff.toml
```diff
@@ -38,6 +38,13 @@ lint.ignore = [
     "RUF012",
     # Logging statement uses f-string
     "G004",
+    # `import` should be at the top-level of a file
+    # Used intentionally to avoid circular imports
+    "PLC0415",
```
valerena: Can we just ignore this specifically where we use it? I really think imports should be at the top level, and AI agents are really good at adding inline imports, which we should try to avoid

ruff.toml
```diff
@@ -38,6 +38,13 @@ lint.ignore = [
...
+    # Object does not implement `__hash__` method
+    # Intentional in py27hash_fix.py
+    "PLW1641",
+    "PTH208",
```
valerena: Let's add a comment for this one too, so we maybe try to remove it later.

ruff.toml
```diff
@@ -52,6 +59,10 @@ lint.ignore = [
     # print() is allowed in bin/ as they are dev tools.
     "T201",
 ]
+"samtranslator/internal/schema_source/aws_serverless_function.py" = [
+    # E501: Line too long - unavoidable due to large Union type
+    "E501",
+]
```
valerena: Is this really needed? I see a `Union` spread across different lines (around 630), and also the `aws_serverless_statemachine` file has much longer Unions (this commits actually puts together in a single line a `Union` that used to be separate). So if anything, that file would have to have the ignore instead of the `serverless_function`. So it's surprising to me.

valerena (commented):
If we're not adding Python3.14, let's update the title/description to reflect that.

ruff.toml
roger-zhangg: I remember this is the fix we've done for pydantic v1. I can add the ignore there

ruff.toml
roger-zhangg: I can take another look

roger-zhangg: Humm, this seems comming from the first attempt, I reverted other code but not this one. Will delete

ruff.toml
roger-zhangg: removed

ruff.toml
roger-zhangg: removed

tests/translator/output/error_function_invalid_event_type.json
```diff
@@ -4,20 +4,20 @@
     "Number of errors found: 3. ",
     "Resource with id [FunctionApiTypeError] is invalid. ",
     "Event with id [ApiEvent] is invalid. ",
-    "Resource dict has missing or invalid value for key Type. ",
-    "Event Type is: API. ",
+    "Resource dict has missing or invalid value for key type. ",
+    "Event type is: API. ",
```
valerena: We shouldn't change the `Type`s in this file

samtranslator/model/__init__.py
```diff
@@ -120,7 +121,7 @@ def __init__(self) -> None:
 
 
 class Resource(ABC):
-    """A Resource object represents an abstract entity that contains a Type and a Properties object. They map well to
+    """A Resource object represents an abstract entity that contains a type and a Properties object. They map well to
```
valerena: This refers to the CFN `Type` so it shouldn't change to lower case. Same with other `type` in this file that were converted from `Type` in error messages and comments.

DEVELOPMENT_GUIDE.md
```diff
@@ -26,7 +26,7 @@ Environment setup
 -----------------
 ### 1. Install Python versions
 
-Our officially supported Python versions are 3.8, 3.9 and 3.10. 
+Our officially supported Python versions are 3.10, 3.11, 3.12, 3.13 and 3.14.
```
valerena: We shouldn't add 3.14 to this doc if we're not *really* supporting it yet.

samtranslator/model/__init__.py
roger-zhangg: Thanks for catching!

DEVELOPMENT_GUIDE.md
roger-zhangg: thanks! removed

tests/translator/output/error_function_invalid_event_type.json
roger-zhangg: reverted this file, thanks for catching

valerena (approved):

samtranslator/plugins/api/implicit_api_plugin.py
```diff
@@ -54,22 +54,22 @@ def __init__(self) -> None:
...
         # {api_id: {path: {method: condition_name_or_None}}}
-        self.api_conditions: Dict[str, Any] = {}
-        self.api_deletion_policies: Dict[str, Any] = {}
-        self.api_update_replace_policies: Dict[str, Any] = {}
+        self.api_conditions: dict[str, Any] = {}
+        self.api_deletion_policies: dict[str, Any] = {}
+        self.api_update_replace_policies: dict[str, Any] = {}
 
     @abstractmethod
-    def _process_api_events(  # noqa: PLR0913
```
vicheey, line 65: Nice.

vicheey (approved):

Issue #3866, opened 2025-12-18
Feature Request: Support SecurityPolicy and EndpointAccessMode in AWS::Serverless::Api


jth08527:
## Description

On **Nov 19th, 2025**, API Gateway added new endpoint security capabilities, including updated TLS policies and strict endpoint access modes. These features are documented here:

🔗 https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-security-policies.html

CloudFormation supports these features through the `AWS::ApiGateway::RestApi` properties `SecurityPolicy` and `EndpointAccessMode`. However, AWS SAM currently does **not** expose these properties in `AWS::Serverless::Api`.

As a result, deployments fail when APIs use modern security policies. SAM manages `EndpointConfiguration` but cannot include the required `SecurityPolicy`, leading to errors such as:

```
Unable to update EndpointType. SecurityPolicy is required for EndpointType update.
```

This has forced me to stop using `AWS::Serverless::Api` and instead define raw CloudFormation resources for RestApi, Stage, and Deployment.

## Request

Add support for the following properties to `AWS::Serverless::Api`:

```yaml
SecurityPolicy: SecurityPolicy_TLS13_1_3_FIPS_2025_09
EndpointAccessMode: STRICT
```

These should map directly to the corresponding CloudFormation fields on `AWS::ApiGateway::RestApi`.

List of supported security policies for the different endpoint types:

🔗 https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-security-policies-list.html

For `EndpointAccessMode`, the only supported values currently are `STRICT` and `BASIC`.

## Why this matters

This would allow users to continue using newer and more secure policies, while still allowing use of the much simpler `AWS::Serverless::Api` abstraction.

jth08527:
@iph and @bnusunny thank you both for working on PR #3872.

I see that this was merge and closed some time ago, and I had hoped it would be included in the release that followed a few days later. I hate to be that guy, but any idea when this will be released? My orgs InfoSec team would like me to enforce a higher TLS version and I would prefer to not have to create some mechanism outside of SAM or CloudFormation to do this.

Any hint as to when this might drop would be appreciated, but I also understand if you can't.

reedham-aws:
Hello @jth08527, this should be released in the next version of SAM coming the week of 2026-03-16. Thank you for bringing this back to our attention!

jth08527:
Hello @reedham-aws, thank you very much for the update!

[...]

PR #3898, opened 2026-03-25
feat: Support EndpointAccessMode property for AWS::Serverless::Api

wandora58:
### Issue #, if available

https://github.com/aws/serverless-application-model/issues/3866

### Description of changes

The `SecurityPolicy` property was added in a recent [change](https://github.com/aws/serverless-application-model/pull/3872). 

When specifying a policy that begins with SecurityPolicy_ in the `SecurityPolicy` property, the `EndpointAccessMode` property must also be specified. This change is to pass through the EndpointAccessMode property directly to the API.

https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-security-policies.html#apigateway-security-policies-endpoint-access-mode

```
Endpoint access mode is an additional parameter that you must specify for any REST API or custom domain name that uses an enhanced security policy that begins with SecurityPolicy_. You do this when you create your resource or if you change the security policy from a legacy policy to an enhanced policy.
```

unit & integration test

### Checklist

    - [x] Using correct values
    - [ ] Using wrong values


valerena (approved):
It looks good. We're checking with other people from the API Gateway service team to confirm that everything is okay.

integration/single/test_api_with_endpoint_access_mode.py
```diff
@@ -0,0 +1,18 @@
...
+class TestApiWithEndpointAccessMode(BaseTest):
+    """
+    Tests for AWS::Serverless::Api with EndpointAccessMode property
+    """
+
+    def test_api_with_endpoint_access_mode(self):
+        self.create_and_verify_stack("single/api_with_endpoint_access_mode")
+
+        rest_api_resources = self.get_stack_resources("AWS::ApiGateway::RestApi")
+        self.assertEqual(len(rest_api_resources), 1)
```
iph: Thoughts on adding some more asserts here to check for the strict+security policy?

samtranslator/internal/schema_source/sam-docs.json
```diff
@@ -723,6 +723,7 @@
...
       "Domain": "Configures a custom domain for this API Gateway API.  \n*Type*: [DomainConfiguration](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-property-api-domainconfiguration.html)  \n*Required*: No  \n*CloudFormation compatibility*: This property is unique to AWS SAM and doesn't have an CloudFormation equivalent.",
+      "EndpointAccessMode": "The endpoint access mode for the RestApi.  \n*Type*: String  \n*Required*: No  \n*AWS CloudFormation compatibility*: This property is passed directly to the [`EndpointAccessMode`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-apigateway-restapi.html#cfn-apigateway-restapi-endpointaccessmode) property of an `AWS::ApiGateway::RestApi` resource.",
```
iph, line 727: \<deep sigh> we really need to update our cloudformation docs, as this doesn't really describe what that does 🤣

Note: No change required here, just saddened by our docs.

samtranslator/model/api/api_generator.py
```diff
@@ -281,8 +282,9 @@ def __init__(  # noqa: PLR0913
         self.feature_toggle = feature_toggle
         self.policy = policy
         self.security_policy = security_policy
+        self.endpoint_access_mode = endpoint_access_mode
 
-    def _construct_rest_api(self) -> ApiGatewayRestApi:
+    def _construct_rest_api(self) -> ApiGatewayRestApi:  # noqa: PLR0912
```
iph, line 287: What's this?

tests/translator/input/api_with_endpoint_access_mode.yaml
```diff
@@ -0,0 +1,17 @@
...
+    Type: AWS::Serverless::Api
+    Properties:
+      StageName: Prod
+
+  # Top-level overrides Globals
+  ApiTopLevelOverride:
+    Type: AWS::Serverless::Api
+    Properties:
+      StageName: Prod
+      EndpointAccessMode: STRICT
```
iph, line 18: We should add a tls policy here, because I don't believe there is such a configuration where you can set EndpointAccessMode without the TLS Policy.

iph (changes requested):
Thanks for the PR! Some minor suggestions here and there around testing (and the random comment :) ).

I'm fine with it being in a separate PR (I can handle it after), but this doesn't solve for all use cases of EndpointAccessMode.

Essentially, you can add tls policies in 2 places:
- Rest APIs
- Custom Domains

hence, those are also the places you can add EndpointAccessMode. This covers Rest APIs but does not cover Custom Domains.

samtranslator/model/api/api_generator.py
valerena: This is to ignore a [`ruff` check](https://docs.astral.sh/ruff/rules/too-many-branches/). We have this in similar "construct_resource" functions because of just many conditional for different properties.

We probably want a better fix, but probably outside of the scope of this PR.

integration/single/test_api_with_endpoint_access_mode.py
valerena: To help doing this, here's an example of similar checks for other properties: 

https://github.com/aws/serverless-application-model/blob/develop/integration/single/test_api_with_ipaddresstype.py#L21-L25

rowanu:
Keen to have this so I can avoid adding a lot of CloudFormation 🙇 🙏 Thanks!

wandora58:
support Custom Domains and enhance the unit and integration test

integration/single/test_api_with_endpoint_access_mode.py
wandora58: enhanced unit and integration test

samtranslator/model/api/api_generator.py
wandora58: I'll leave this as-is for now — the same pattern exists across multiple construct methods, and refactoring it would touch a wide surface area beyond the scope of this PR

tests/translator/input/api_with_endpoint_access_mode.yaml
wandora58: enhanced

samtranslator/model/api/api_generator.py
iph: Sounds good -- agree with you both, had just never seen this before now. I'm newer to the codebase, more here as apigw representation. Sounds good to keep :)

iph (approved):

iph:
lgtm!

valerena (approved):

reedham-aws (approved):

valerena:
Thanks for the contribution! This still needs to go through a deployment through the different AWS regions, so it will still take some time to be fully available.

PR #3909, opened 2026-04-14
refactor(test): Pre-create MSK/MQ in companion stack for testing

vicheey:
### What

Pre-create long-provisioning resources (MSK cluster, MQ broker) in the companion stack instead of creating them inline per-test. Also adds an API Gateway CloudWatch logging role to the companion stack.

### Why

MSK clusters take 15-30 min to provision and MQ brokers take 15-20 min. When tests create these resources inline, a single test can take over 20 minutes. Pre-creating them in the companion stack lets tests reference existing resources via ARN parameters, reducing test execution time significantly.

The API Gateway logging role ensures `test_api_settings` tests don't fail with "CloudWatch Logs role ARN must be set in account settings" in environments where the role isn't pre-configured.

### Changes

**New file:**
- `integration/setup/companion-stack.yaml` — adds MSK cluster (kafka.t3.small), MQ broker (mq.t3.micro) with networking/secrets, API Gateway logging role + account resource. Uses CFN Conditions (`CreateMskCluster`, `CreateMqBroker`) so resources can be disabled in regions that don't support them.

**Refactored templates (6 files):**
- 4 MSK templates — removed inline `AWS::MSK::Cluster`, added `PreCreatedMskClusterArn` parameter
- 2 MQ templates — removed inline broker/VPC/subnet/security group/secret, added `PreCreatedMqBrokerArn` + `PreCreatedMqBrokerSecretArn` parameters

**Updated expected outputs (6 files):**
- Removed expected resources for inline MSK/MQ that no longer exist in templates

**Updated test code (6 files):**
- `conftest.py` — pass companion stack outputs as parameters to test stacks; region-aware disabling of MSK/MQ parameters
- `test_function_with_msk.py`, `test_function_with_mq.py` — simplified to use companion stack ARNs instead of creating resources inline
- `helpers/base_test.py` — support `parameters` in `create_and_verify`. Also moves `output_dir` to `/tmp/` for compatibility with read-only filesystem environments. Existing test reporting is unaffected since `output_dir` is only used for intermediate CFN changeset files, not test reports.
- `helpers/resource.py` — add `get_s3_uri` helper
- `helpers/stack.py` — pass parameters through `create_or_update` to `_deploy_stack`

### Risk

Low. The only difference is where the underlying MSK/MQ resources come from (companion stack parameter vs inline creation). Existing test workflows continue to work because the companion stack is already exit.

    - [ ] Using correct values
    - [ ] Using wrong values

### Examples?

### DIFF
--- a/integration/conftest.py
+++ b/integration/conftest.py
@@ -64,7 +64,7 @@ def clean_all_integ_buckets():
 
 @pytest.fixture()
 def setup_companion_stack_once(tmpdir_factory, get_prefix):
-    # When COMPANION_STACK_NAME is set via env var, the stack is pre-created (e.g. Hydra bootstrap stack)
+    # When COMPANION_STACK_NAME is set via env var, the stack is pre-created by the test platform
     if os.environ.get("COMPANION_STACK_NAME"):
         return
     tests_integ_dir = Path(__file__).resolve().parents[1]
--- a/integration/conftest.py
+++ b/integration/conftest.py
@@ -1,4 +1,5 @@
 import logging
+import os
 from pathlib import Path
 
 import boto3
@@ -6,6 +7,7 @@ import botocore
 import pytest
 from botocore.exceptions import ClientError
 
+from integration.config.service_names import MSK, MQ
 from integration.helpers.base_test import S3_BUCKET_PREFIX
 from integration.helpers.client_provider import ClientProvider
 from integration.helpers.deployer.exceptions.exceptions import S3DoesNotExistException, ThrottlingError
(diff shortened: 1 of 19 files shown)


roger-zhangg (commented):
Removed as not valid

roger-zhangg (approved):

integration/conftest.py
```diff
@@ -62,14 +64,29 @@ def clean_all_integ_buckets():
 
 @pytest.fixture()
 def setup_companion_stack_once(tmpdir_factory, get_prefix):
+    # When COMPANION_STACK_NAME is set via env var, the stack is pre-created (e.g. Hydra bootstrap stack)
+    if os.environ.get("COMPANION_STACK_NAME"):
```
licjun, line 68: This fixture will only be run by TOD path, right? Do we really need this check here?

vicheey: Yes, this fixture runs in both paths. When `COMPANION_STACK_NAME` is set via env var, the companion stack is pre-created by the test platform, so we skip creation here and skip teardown at line 175. When it's not set (local/CI runs), the fixture creates and manages the stack lifecycle itself. The same branching at line 201 resolves the correct stack name when fetching outputs.

licjun (approved):

PR #3912, opened 2026-04-16
Security: Insecure temporary file creation pattern in shared temp directory

tuanaiseo:
## Problem

`mktempfile` manually constructs a filename in the system temp directory and opens it with `open(..., 'w+')`. This pattern is vulnerable to race/symlink attacks in multi-user environments because creation is not done with secure exclusive flags. If this utility is used in privileged contexts, an attacker could potentially influence file targets.

**Severity**: `medium`
**File**: `integration/helpers/deployer/utils/artifact_exporter.py`

## Solution

Use `tempfile.NamedTemporaryFile` or `tempfile.mkstemp` (with restrictive permissions) to ensure atomic creation and avoid symlink/race vulnerabilities. Example: `fd, path = tempfile.mkstemp(); os.fdopen(fd, 'w+')`.

## Changes

- `integration/helpers/deployer/utils/artifact_exporter.py` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced

### DIFF
--- a/integration/helpers/deployer/utils/artifact_exporter.py
+++ b/integration/helpers/deployer/utils/artifact_exporter.py
@@ -18,7 +18,6 @@ This was ported over from the sam-cli repo
 
 import os
 import tempfile
-import uuid
 from contextlib import contextmanager
 from urllib.parse import parse_qs, urlparse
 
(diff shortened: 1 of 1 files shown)


reedham-aws (commented):
I don't think this is really a security fix because
1. It's in the integration test utils
2. The fact that we're using a uuid makes the race condition of generating a symlink pretty difficult

PR #3919, opened 2026-04-28
fix: surface user-facing errors for malformed Fn::If on Role and CodeUri

roger-zhangg:
### Issue #, if available

N/A

### Description of changes

Three malformed-template inputs crashed the SAM transform with unhandled
exceptions, which CloudFormation surfaced as:

> Transform AWS::Serverless-2016-10-31 failed with: Internal transform failure.

The customer had no way to know which resource or property was wrong.

**1. Malformed `Fn::If` on `AWS::Serverless::Function.Role`.** `Fn::If`
must have exactly 3 elements, but `_make_lambda_role` unpacked directly
without arity validation, so a 2- or 4-element list raised
`ValueError: not enough values to unpack (expected 3, got 2)`.
Now validates via the existing `validate_intrinsic_if_items` helper and
raises `InvalidResourceException` with the logical id.

**2. Malformed `Fn::If` on event-invoke destinations.**
`EventInvokeConfig.DestinationConfig.{OnSuccess,OnFailure}.Destination`
also supports `Fn::If`, and `_get_or_make_condition` indexed
`dest_list[1]` / `dest_list[2]` without arity validation. A 2-element
list raised `IndexError: list index out of range`. Same guard is now
applied here.

**3. Malformed `CodeUri` URIs.** Since the CVE-2024-11168 fix (Python
3.9.17+/3.10.12+/3.11.4+/3.12/3.13/3.14), `urllib.parse.urlparse`
validates bracketed-host segments against the IPv6/IPv4 grammars. CDK
sometimes synthesizes unresolved `IResolvable` tokens like
`s3://[TOKEN.25]/key` into template strings, which now raises
`ValueError` from inside `urlparse`. `parse_s3_uri` now catches that and
returns `None`, so `construct_s3_location_object` raises the existing
friendly `InvalidResourceException`.

After the fix, users see the same class of message SAM emits for other
malformed properties:

```
Resource with id [MyFunction] is invalid. Malformed 'Role' property:
Fn::If requires 3 arguments. Expected form:
Fn::If: [Condition, TrueValue, FalseValue].
```
```
Resource with id [MyFunctionEventInvokeConfig] is invalid. Malformed
'Destination' property: Fn::If requires 3 arguments. Expected form:
Fn::If: [Condition, TrueValue, FalseValue].
```
```
Resource with id [MyFunction] is invalid. 'CodeUri' is not a valid S3
Uri of the form 's3://bucket/key' with optional versionId query parameter.
```

- Added translator error fixtures (auto-discovered by
  `test_transform_invalid_document`):
  `tests/translator/input/error_function_role_fn_if_malformed.{yaml,json}`,
  `tests/translator/input/error_function_event_destination_fn_if_malformed.{yaml,json}`,
  `tests/translator/input/error_function_codeuri_unresolved_token.{yaml,json}`.
- Added unit tests for `parse_s3_uri` / `construct_s3_location_object`
  in `tests/model/s3_utils/test_uri_parser.py`.
- Added two cases (too-few and too-many `Fn::If` args) to
  `tests/model/test_sam_resources.py::TestSamFunctionRoleResolver`.
- Full suite: **4503 passed, 0 failed** (Python 3.13).

### Checklist

    - [x] Using correct values
    - [x] Using wrong values
- [ ] Add/update [integration tests](https://github.com/aws/serverless-application-model/blob/develop/INTEGRATION_TESTS.md) *(not needed — all three failure modes are deterministic and covered by translator error tests)*


tests/model/s3_utils/test_uri_parser.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def test_unresolved_cdk_token_returns_none(self):
+        """Bracketed host segments that are not valid IPv4/IPv6 raise ValueError
+        from urllib (see CVE-2024-11168); parse_s3_uri should treat the input as
+        "not a valid S3 URI" and return None so callers can raise a friendly
+        InvalidResourceException instead of crashing the transform.
+        """
+        self.assertIsNone(parse_s3_uri("s3://[TOKEN.25]/my/key"))
+        self.assertIsNone(parse_s3_uri("https://[TOKEN.25]/path"))
+        self.assertIsNone(parse_s3_uri("s3://bucket-[TOKEN.25]/key"))
```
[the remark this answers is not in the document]
roger-zhangg: `urllib.parse.urlsplit` invokes `_check_bracketed_netloc` as soon as it sees any `[` or `]` in the netloc (not only when `[` is at position 0). Mid-string brackets fail the balance/positioning check and raise `ValueError: Invalid IPv6 URL` — a different message from the leading-bracket case, but still a `ValueError` and still
  caught by the new `try/except` in `parse_s3_uri`. So the assertion `self.assertIsNone(parse_s3_uri("s3://bucket-[TOKEN.25]/key"))` is correct and passes. This case is worth keeping because it guards against a future regression where someone might narrow the `except` to only the `_check_bracketed_host` path.

reedham-aws (approved):

samtranslator/model/sam_resources.py
```diff
@@ -452,7 +453,17 @@ def _make_lambda_role(
...
-        role_condition, role_if, role_else = role_resolved_value.get("Fn::If")
+        if_items = role_resolved_value.get("Fn::If")
+        try:
+            validate_intrinsic_if_items(if_items)
+        except ValueError as e:
+            # Surface a user-facing error instead of letting a ValueError propagate
+            # all the way out of the transform macro as "Internal transform failure".
+            raise InvalidResourceException(
+                self.logical_id,
+                f"Malformed 'Role' property: {e!s}. Expected form: Fn::If: [Condition, TrueValue, FalseValue].",
```
valerena: Maybe minor comment about this message: The underlying error already says 
`Fn::If requires {_NUM_ARGUMENTS_REQUIRED_IN_IF} arguments `

https://github.com/aws/serverless-application-model/blob/develop/samtranslator/model/intrinsics.py#L187

Do we need to add the `Expected form: Fn::If: [Condition, TrueValue, FalseValue].",` ?

It's OK to add the `Malformed 'Role' property` message for context, but outside of that we should just show whatever the underlying method returns. In theory there could be another reason why the `validate_intrinsic_if_items` fails, but we don't know.

valerena: (of course, same comment for the other place where we're adding this)

tests/model/test_sam_resources.py
```diff
@@ -847,6 +847,63 @@ def test_role_get_att_no_execution_role(self):
...
+
+        with pytest.raises(InvalidResourceException) as excinfo:
+            self.function.to_cloudformation(**self.kwargs)
+
+        msg = str(excinfo.value)
+        self.assertIn("foo", msg)  # logical id
+        self.assertIn("Role", msg)
+        self.assertIn("Fn::If", msg)
+
+    def test_role_fn_if_too_many_arguments_raises_invalid_resource_exception(self):
```
valerena: Ideally we could `parameterize` these two Role tests instead of writing two separate tests, in case in the future there are other cases we want to test.

(same below, making a single test for `DestinationConfig` tests)

valerena (approved):

PR #3930, opened 2026-05-26
feat: async test stack cleanup — remove waiter + add sweeper Lambda

vicheey:
## Changes

- Remove CloudFormation waiter for stack deletion
- Add sweeper Lambda for async test stack cleanup
- Update permissions and refactor code

## Commits
- feat: async test stack cleanup — remove waiter + add sweeper Lambda
- update permission and refactor code

### DIFF
--- a/integration/setup/companion-stack.yaml
+++ b/integration/setup/companion-stack.yaml
@@ -278,7 +278,6 @@ Resources:
           - Effect: Allow
             Action:
             - cloudformation:DeleteStack
-            - cloudformation:UpdateTerminationProtection
             Resource: !Sub "arn:${AWS::Partition}:cloudformation:${AWS::Region}:${AWS::AccountId}:stack/*sam-integ-stack-*"
           - Effect: Allow
             Action:
@@ -303,7 +302,7 @@ Resources:
           - Effect: Allow
             Action:
             - logs:DeleteLogGroup
-            Resource: !Sub "arn:${AWS::Partition}:logs:${AWS::Region}:${AWS::AccountId}:log-group:/aws/lambda/*sam-integ-*:*"
+            Resource: !Sub "arn:${AWS::Partition}:logs:${AWS::Region}:${AWS::AccountId}:log-group:*sam-integ-*"
           - Effect: Allow
             Action: '*'
             Resource: '*'
@@ -315,7 +314,7 @@ Resources:
     Type: AWS::Lambda::Function
     Properties:
       FunctionName: !Sub "${AWS::StackName}-test-stack-sweeper"
-      Runtime: python3.10
+      Runtime: python3.13
       Handler: index.handler
       Timeout: 900
       MemorySize: 256
@@ -327,7 +326,7 @@ Resources:
 
           STACK_PATTERN = 'sam-integ-stack-'
           IAM_PATTERN = 'sam-integ-'
-          STATUSES = [
+          ELIGIBLE_STATUSES = [
             'CREATE_COMPLETE', 'ROLLBACK_COMPLETE', 'ROLLBACK_FAILED',
             'REVIEW_IN_PROGRESS', 'DELETE_FAILED', 'UPDATE_FAILED',
             'UPDATE_COMPLETE', 'UPDATE_ROLLBACK_COMPLETE',
@@ -350,30 +349,26 @@ Resources:
             _sweep_log_groups(logs, cutoff, ctx)
 
           def _sweep_stacks(cfn, iam, cutoff, ctx):
-            deleted = 0
-            for page in cfn.get_paginator('list_stacks').paginate(StackStatusFilter=STATUSES):
+            deleted = []
+            for page in cfn.get_paginator('list_stacks').paginate(StackStatusFilter=ELIGIBLE_STATUSES):
               for stack in page['StackSummaries']:
                 if not _has_time(ctx):
-                  print(f"Deleted: {deleted}")
+                  print(f"Attempt to delete ({len(deleted)}) stacks: {deleted}")
                   return
                 name = stack['StackName']
                 if not _is_test(name, strict=True):
                   continue
                 if stack['CreationTime'].replace(tzinfo=timezone.utc) >= cutoff:
                   continue
-                try:
-                  cfn.update_termination_protection(EnableTerminationProtection=False, StackName=name)
-                except Exception:
-                  pass
                 if stack['StackStatus'] == 'DELETE_FAILED':
                   _fix_and_retry(cfn, iam, name)
                 try:
                   cfn.delete_stack(StackName=name)
-                  deleted += 1
+                  deleted.append(name)
                   time.sleep(1)
                 except Exception as e:
                   print(f"delete_stack {name}: {e}")
-            print(f"Stacks: {deleted} deleted")
+            print(f"Attempt to delete ({len(deleted)}) stacks: {deleted}")
 
           def _fix_and_retry(cfn, iam, stack_name):
             try:
@@ -420,7 +415,7 @@ Resources:
           def _sweep_log_groups(logs, cutoff, ctx):
             cutoff_ms = int(cutoff.timestamp() * 1000)
             deleted = 0
-            for page in logs.get_paginator('describe_log_groups').paginate(logGroupNamePrefix='/aws/lambda/'):
+            for page in logs.get_paginator('describe_log_groups').paginate():
               for log_group in page['logGroups']:
                 if not _has_time(ctx):
                   return
--- a/integration/setup/companion-stack.yaml
+++ b/integration/setup/companion-stack.yaml
@@ -309,8 +309,8 @@ Resources:
             Resource: '*'
             Condition:
               ForAnyValue:StringEquals:
-                "aws:CalledVia": cloudformation.amazonaws.com
-          
+                aws:CalledVia: cloudformation.amazonaws.com
+
   TestStackSweeperFunction:
     Type: AWS::Lambda::Function
     Properties:
@@ -438,7 +438,7 @@ Resources:
   TestStackSweeperSchedule:
     Type: AWS::Events::Rule
     Properties:
-      ScheduleExpression: 'rate(6 hours)'
+      ScheduleExpression: rate(6 hours)
       State: ENABLED
       Targets:
       - Arn: !GetAtt TestStackSweeperFunction.Arn
--- a/integration/setup/companion-stack.yaml
+++ b/integration/setup/companion-stack.yaml
@@ -256,6 +256,213 @@ Resources:
     DependsOn: ApiGatewayLoggingRole
     Properties:
       CloudWatchRoleArn: !GetAtt ApiGatewayLoggingRole.Arn
+
+  TestStackSweeperRole:
+    Type: AWS::IAM::Role
+    Properties:
+      AssumeRolePolicyDocument:
+        Version: '2012-10-17'
+        Statement:
+        - Effect: Allow
+          Principal:
+            Service: lambda.amazonaws.com
+          Action: sts:AssumeRole
+      ManagedPolicyArns:
+      - !Sub "arn:${AWS::Partition}:iam::aws:policy/service-role/AWSLambdaBasicExecutionRole"
+      - !Sub "arn:${AWS::Partition}:iam::aws:policy/ReadOnlyAccess"
+      Policies:
+      - PolicyName: TestStackSweeperPolicy
+        PolicyDocument:
+          Version: '2012-10-17'
+          Statement:
+          - Effect: Allow
+            Action:
+            - cloudformation:DeleteStack
+            - cloudformation:UpdateTerminationProtection
+            Resource: !Sub "arn:${AWS::Partition}:cloudformation:${AWS::Region}:${AWS::AccountId}:stack/*sam-integ-stack-*"
+          - Effect: Allow
+            Action:
+            - iam:DeleteRolePolicy
+            - iam:DetachRolePolicy
+            - iam:DeleteRole
+            - iam:DeletePolicyVersion
+            - iam:DeletePolicy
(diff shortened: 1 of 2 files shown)


integration/setup/companion-stack.yaml
```diff
@@ -256,6 +256,202 @@ Resources:
...
+
+            _sweep_stacks(cfn, iam, cutoff, ctx)
+            _sweep_log_groups(logs, cutoff, ctx)
+
+          def _sweep_stacks(cfn, iam, cutoff, ctx):
+            deleted = 0
+            for page in cfn.get_paginator('list_stacks').paginate(StackStatusFilter=ELIGIBLE_STATUSES):
+              for stack in page['StackSummaries']:
+                if not _has_time(ctx):
+                  break
```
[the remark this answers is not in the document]
vicheey: Fixed in 05e16ea — changed `break` to `return` so the time-budget guard exits both loops. Deleted count is printed before returning.

integration/setup/companion-stack.yaml
```diff
@@ -256,6 +256,202 @@ Resources:
...
+                name = stack['StackName']
+                if not _is_test_resource(name, strict=True):
+                  continue
+                if stack['CreationTime'].replace(tzinfo=timezone.utc) >= cutoff:
+                  continue
+                try:
+                  cfn.update_termination_protection(EnableTerminationProtection=False, StackName=name)
+                except:
```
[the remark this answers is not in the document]
vicheey: Fixed in 05e16ea — changed all bare `except:` to `except Exception:`. Regarding adding `print()` to all sites: the Lambda code is inline CFN `ZipFile` with a hard 4096-char limit — we're at 4095/4096 after these fixes. Adding print statements to all 4 sites would require ~120 chars we don't have. The critical paths (`delete_stack`, `fix_and_retry`) already log errors. The helper functions (`_force_delete_role`, `_force_delete_policy`, S3 emptying, log group deletion) intentionally treat failures as non-fatal: the sweeper retries every 6h, and not-found errors (already-deleted resources) are expected during concurrent cleanup.

integration/setup/companion-stack.yaml
```diff
@@ -256,6 +256,202 @@ Resources:
...
+
+            _sweep_stacks(cfn, iam, cutoff, ctx)
+            _sweep_log_groups(logs, cutoff, ctx)
+
+          def _sweep_stacks(cfn, iam, cutoff, ctx):
+            deleted = 0
+            for page in cfn.get_paginator('list_stacks').paginate(StackStatusFilter=ELIGIBLE_STATUSES):
+              for stack in page['StackSummaries']:
+                if not _has_time(ctx):
+                  break
```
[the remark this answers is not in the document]
vicheey: Duplicate of earlier comment — addressed in same commit (05e16ea).

integration/setup/companion-stack.yaml
```diff
@@ -256,6 +256,202 @@ Resources:
...
+                name = stack['StackName']
+                if not _is_test_resource(name, strict=True):
+                  continue
+                if stack['CreationTime'].replace(tzinfo=timezone.utc) >= cutoff:
+                  continue
+                try:
+                  cfn.update_termination_protection(EnableTerminationProtection=False, StackName=name)
+                except:
```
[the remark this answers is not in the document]
vicheey: Duplicate of earlier comment — addressed in same commit (05e16ea).

integration/setup/companion-stack.yaml
```diff
@@ -256,6 +256,203 @@ Resources:
...
+              print(f"fix_and_retry {stack_name}: {e}")
+
+          def _force_delete_role(iam, role_name):
+            try:
+              for p in iam.list_role_policies(RoleName=role_name)['PolicyNames']:
+                iam.delete_role_policy(RoleName=role_name, PolicyName=p)
+              for p in iam.list_attached_role_policies(RoleName=role_name)['AttachedPolicies']:
+                iam.detach_role_policy(RoleName=role_name, PolicyArn=p['PolicyArn'])
+              iam.delete_role(RoleName=role_name)
+            except Exception: pass
```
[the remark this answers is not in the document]
vicheey: Acknowledged — this is a valid trade-off we made due to the CFN ZipFile 4096-char hard limit (currently at 4095/4096). The two critical paths already log errors: `delete_stack` (line 374) and `fix_and_retry` (line 397). The 4 silent `except Exception: pass` sites are in helper functions where failures are expected and non-fatal:

- **S3 emptying**: bucket may already be empty or deleted by concurrent CFN deletion
- **`_force_delete_role`**: role may not exist (already deleted by previous sweep or CFN)
- **`_force_delete_policy`**: same — policy may already be gone
- **`delete_log_group`**: log group may not exist or may be deleted concurrently

In all cases, the sweeper retries every 6h. If the underlying issue is persistent (e.g., missing IAM permission), it surfaces at the `delete_stack` level which IS logged. If we gain char budget in the future (e.g., by moving to S3-hosted code), we'll add logging to all sites.

integration/setup/companion-stack.yaml
```diff
@@ -256,6 +256,203 @@ Resources:
...
+            Resource: '*'
+            Condition:
+              ForAnyValue:StringEquals:
+                aws:CalledVia: cloudformation.amazonaws.com
+
+  TestStackSweeperFunction:
+    Type: AWS::Lambda::Function
+    Properties:
+      FunctionName: !Sub "${AWS::StackName}-test-stack-sweeper"
+      Runtime: python3.10
```
valerena: Can we do it in a newer version of Python?

integration/setup/companion-stack.yaml
```diff
@@ -256,6 +256,203 @@ Resources:
...
+                if not _has_time(ctx):
+                  print(f"Deleted: {deleted}")
+                  return
+                name = stack['StackName']
+                if not _is_test(name, strict=True):
+                  continue
+                if stack['CreationTime'].replace(tzinfo=timezone.utc) >= cutoff:
+                  continue
+                try:
+                  cfn.update_termination_protection(EnableTerminationProtection=False, StackName=name)
```
valerena: Our integ tests stacks should never have termination protection enabled. If we have them with termination protection on, then we should change that. It's not a good practice to just "disable termination protection in a stack just to be able to delete it". If it's a stack that's expected to be deleted, then it shouldn't have it, and if a stack has it, then we probably shouldn't delete it.

integration/setup/companion-stack.yaml
```diff
@@ -256,6 +256,203 @@ Resources:
...
+                except Exception:
+                  pass
+                if stack['StackStatus'] == 'DELETE_FAILED':
+                  _fix_and_retry(cfn, iam, name)
+                try:
+                  cfn.delete_stack(StackName=name)
+                  deleted += 1
+                  time.sleep(1)
```
valerena, line 368: This doesn't really confirm that we deleted a stack, right? We're just counting the stacks that we "sent the delete", not the stacks that "were actually deleted" (because we're not checking that, right?) (I don't think we need a change, but just for my own understanding)

integration/setup/companion-stack.yaml
```diff
@@ -256,6 +256,203 @@ Resources:
...
+              for v in iam.list_policy_versions(PolicyArn=arn)['Versions']:
+                if not v['IsDefaultVersion']:
+                  iam.delete_policy_version(PolicyArn=arn, VersionId=v['VersionId'])
+              iam.delete_policy(PolicyArn=arn)
+            except Exception: pass
+
+          def _sweep_log_groups(logs, cutoff, ctx):
+            cutoff_ms = int(cutoff.timestamp() * 1000)
+            deleted = 0
+            for page in logs.get_paginator('describe_log_groups').paginate(logGroupNamePrefix='/aws/lambda/'):
```
valerena: We also have a ton of LogGroups with prefix `/aws/amazonmq/broker` and `/aws/appsync/apis/`.. We might want to delete those too (we might find other log groups after we delete those)

integration/setup/companion-stack.yaml
vicheey: No, it is just to count the stacks the lambda attempts to delete,.

integration/setup/companion-stack.yaml
vicheey: I'll verify that we do not have any test cases that have termination protection on, then I will remove  this step.

roger-zhangg (approved):

integration/setup/companion-stack.yaml
vicheey: Remove the step in [9d80135](https://github.com/aws/serverless-application-model/pull/3930/commits/9d8013556e27876c90f4cf064e5d1996528be8ec)

integration/setup/companion-stack.yaml
vicheey: Remove the step in [9d80135](https://github.com/aws/serverless-application-model/pull/3930/commits/9d8013556e27876c90f4cf064e5d1996528be8ec)

integration/setup/companion-stack.yaml
vicheey: update to 3.13 in [9d80135](https://github.com/aws/serverless-application-model/pull/3930/commits/9d8013556e27876c90f4cf064e5d1996528be8ec)

vicheey: The Lambda uses python3.10 to match the Hydra test runner runtime (Brazil builds native extensions against CPython310). We could use 3.12+ for the sweeper itself since it doesn't share the test package closure, but 3.10 keeps it consistent with the test infrastructure and avoids a separate runtime dependency. Happy to bump if there's a preference.

integration/setup/companion-stack.yaml
vicheey: Correct — this is defensive for legacy stacks. The companion stack itself has termination protection enabled (set by `stack.py:_udpate_termination_protection`). Test stacks created during integration tests do NOT have it. The `update_termination_protection` call is a no-op for stacks that don't have it enabled, and handles the edge case of old stacks from before this PR that may have been manually protected. I can gate it behind the `DELETE_FAILED` check if you prefer — only disable protection when we know the stack is stuck.

integration/setup/companion-stack.yaml
vicheey: Correct — `delete_stack` is fire-and-forget (returns HTTP 200 immediately, CFN processes deletion asynchronously). We're recording "stacks we issued deletion for", not "stacks confirmed deleted". The next sweep (6h later) will find any that failed and retry. The message wording `Attempt to delete` is intentional to communicate this.

integration/setup/companion-stack.yaml
vicheey: Good catch — will broaden the log group cleanup to also scan `/aws/amazonmq/broker` and `/aws/appsync/apis/` prefixes with the same `sam-integ-stack-` name filter. These are created by test stacks that use MQ and AppSync resources.

integration/setup/companion-stack.yaml
```diff
@@ -256,6 +256,198 @@ Resources:
...
+            'REVIEW_IN_PROGRESS', 'DELETE_FAILED', 'UPDATE_FAILED',
+            'UPDATE_COMPLETE', 'UPDATE_ROLLBACK_COMPLETE',
+            'UPDATE_ROLLBACK_FAILED']
+
+          def _has_time(ctx):
+            return ctx.get_remaining_time_in_millis() > 30000
+
+          def _is_test(name, strict=False):
+            pattern = STACK_PATTERN if strict else IAM_PATTERN
+            return pattern in name and 'companion' not in name
```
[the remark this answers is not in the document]
vicheey: It is fine to leave

integration/setup/companion-stack.yaml
```diff
@@ -256,6 +256,198 @@ Resources:
...
+            - s3:DeleteBucket
+            Resource:
+            - !Sub "arn:${AWS::Partition}:s3:::*sam-integ-stack-*"
+            - !Sub "arn:${AWS::Partition}:s3:::*sam-integ-stack-*/*"
+          - Effect: Allow
+            Action:
+            - logs:DeleteLogGroup
+            Resource: !Sub "arn:${AWS::Partition}:logs:${AWS::Region}:${AWS::AccountId}:log-group:*sam-integ-*"
+          - Effect: Allow
+            Action: '*'
```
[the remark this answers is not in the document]
vicheey: Deliberate choice — per-service enumeration was fragile (we hit 5+ services with incompatible ARN/tag scoping). `DeleteStack` is already scoped to `*sam-integ-stack-*` stacks, so CFN can only cascade-delete within test stacks. The CalledVia condition ensures these broad permissions are only usable by CFN, not by the Lambda directly.

integration/setup/companion-stack.yaml
vicheey: Already addressed in the latest commit — removed the `/aws/lambda/` prefix filter so the sweeper now scans all log groups and filters by `sam-integ-stack-` in name. This covers `/aws/amazonmq/broker`, `/aws/appsync/apis/`, and any other prefix.

valerena: Unfortunately, that's not true. The AppSync log groups are like `/aws/appsync/apis/fvbmd4bfi5afhlll7v2jyeasjq` and MQ are `/aws/amazonmq/broker/b-045d8edd-83ba-4c16-bd6f-3ac6d1552704/audit` and `/aws/amazonmq/broker/b-045d8edd-83ba-4c16-bd6f-3ac6d1552704/audit` (of course hashes vary).

I think the idea of filtering by the prefix was good, but we need to make more API calls for the different prefixes.

In theory, we could add the `sam-integ-stack` filter too only for Lambda log groups.. but for the other one we can definitely just delete all of them filtering by just creationTime.

bnusunny (approved):

valerena (approved):

PR #3940, opened 2026-06-11
fix Globals.Function Architectures override behavior

### LINKED ISSUE #3939: Globals.Function Architectures should be overridden by resource-level Architectures
### Description
When `Architectures` is set in `Globals.Function` and also set on an individual `AWS::Serverless::Function`, SAM merges the two lists by concatenating them. This produces an invalid Lambda `Architectures` value because Lambda functions support only one architecture value.

This is inconsistent with scalar function properties like `MemorySize`, where the resource-level value overrides the global value. 

### Steps to reproduce
Use a SAM template with `Architectures` set globally and overridden by a function:

```yaml

Globals:
  Function:
    Runtime: nodejs20.x
    Handler: index.handler
    CodeUri: src/
    Architectures:
      - x86_64

Resources:
  MyFunction:
    Properties:
      Architectures:
        - arm64
```

### Observed result
The resource-level Architectures value is merged with the global value as a list, producing both architectures:

```yaml
Architectures:
  - x86_64
  - arm64
```

This follows the generic Globals list merge behavior, but it is not valid for Lambda function architectures.

### Expected result
The resource-level Architectures value should override the global value, the same way resource-level primitive properties like MemorySize override Globals.Function.MemorySize.

Expected transformed value:
```yaml
Architectures:
  - arm64
```

### Additional environment details

1. OS: Any
2. If using the [SAM CLI](https://github.com/aws/aws-sam-cli), `sam --version`: Not using SAM CLI directly; observed in SAM Translator behavior
3. AWS region: N/A

allenheltondev:
### Issue #, if available

Fixes #3939 

### Description of changes

Updates `Globals.Function` merge behavior so resource-level `Architectures` overrides global `Architectures` instead of concatenating both lists.

`Architectures` is list-shaped, but Lambda functions support a single architecture value. This keeps the existing default list merge behavior for other properties, such as `Layers`, while handling `Architectures` like other overridable function properties.

Ran:

```bash
python -m pytest tests/plugins/globals/test_globals.py --no-cov
```
Result: 48 passed.

Also updated the globals_for_function translator fixture to cover resource-level Architectures overriding Globals.Function.Architectures across standard, aws-cn, and aws-us-gov outputs.

*Integration tests were not added because this change is limited to Globals merge behavior and is covered by unit tests plus translator output fixtures.*

### Checklist

    - [x] Using correct values
    - [ ] Using wrong values

### Examples?

### DIFF
--- a/samtranslator/plugins/globals/globals.py
+++ b/samtranslator/plugins/globals/globals.py
@@ -306,8 +306,12 @@ class Globals:
                         f"Must be one of the following values - {supported_displayed}",
                     )
 
+            override_properties = []
+            if resource_type == SamResourceType.Function.value:
+                override_properties = ["Architectures"]
+
             # Store all Global properties in a map with key being the AWS::Serverless::* resource type
-            _globals[resource_type] = GlobalProperties(properties)
+            _globals[resource_type] = GlobalProperties(properties, override_properties)
 
         return _globals
 
@@ -433,10 +437,16 @@ class GlobalProperties:
       ```
     (in other words, Deployments will be turned off for the Function)
 
+    **Overridable Properties**
+    Some top-level resource properties must replace the global value instead of following the default merge behavior.
+    For example, Function Architectures is a list, but Lambda accepts only one architecture value. If both global and
+    local Architectures are set, the local value must override the global value instead of concatenating lists.
+
     """
 
-    def __init__(self, global_properties) -> None:  # type: ignore[no-untyped-def]
+    def __init__(self, global_properties, override_properties=None) -> None:  # type: ignore[no-untyped-def]
         self.global_properties = global_properties
+        self.override_properties = override_properties or []
 
     def merge(self, local_properties):  # type: ignore[no-untyped-def]
         """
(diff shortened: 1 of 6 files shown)


samtranslator/plugins/globals/globals.py
```diff
@@ -433,18 +437,35 @@ class GlobalProperties:
...
+
+    def _drop_overridden_globals(self, global_properties, local_properties):  # type: ignore[no-untyped-def]
+        if not self.override_properties or not isinstance(global_properties, dict) or not isinstance(local_properties, dict):
+            return global_properties
+
+        global_properties = global_properties.copy()
+        for key in self.override_properties:
+            if key in global_properties and key in local_properties:
+                del global_properties[key]
+        return global_properties
```
reedham-aws, line 468: I think this approach makes sense and would work, but I my first instinct is that we could get rid of the loop/copy by checking the `override_properties` in `_merge_dict`. I left another comment in the place where that might work, but I could be missing something, let me know what you think.

samtranslator/plugins/globals/globals.py
reedham-aws, line 522: I think we could have a condition here to see if the key is in the overrides, and if it's not then we could enter the `else`. I think this would accomplish the same thing.

reedham-aws (commented):
Thank you for raising this!

Seems like the linter is failing as well, you can fix locally with `make format` or `make lint`.

PR #3948, opened 2026-06-29
fix: resolve issues #3911 and #3939

### LINKED ISSUE #3939: Globals.Function Architectures should be overridden by resource-level Architectures

Hardikrepo:
## Summary

- **#3911 - Security: Logical ID uniqueness check bypass**
  Normalized all \do_not_verify\ values in \erify_logical_id.py\ from plain strings to lists. Python's \in\ operator on a string performs substring matching, allowing a crafted resource type like \AWS::Serverless::Fun\ to bypass the uniqueness check. Using lists ensures exact element matching.

- **#3939 - Globals.Function Architectures concatenates instead of overrides**
  Added \_list_override_properties = frozenset({'Architectures'})\ to \GlobalProperties\ in \globals.py\. When \Architectures\ is set at both global and resource level, the resource-level value now replaces the global value entirely. Previously, SAM concatenated both lists (e.g. \[x86_64, arm64]\), which is invalid since Lambda only supports one architecture.

## Files Changed

| File | Change |
|------|--------|
| \samtranslator/translator/verify_logical_id.py\ | Normalize all \do_not_verify\ values to lists |
| \samtranslator/plugins/globals/globals.py\ | Add \_list_override_properties\ and update \_merge_dict\ |
| \	ests/translator/test_verify_logical_id.py\ | New test file - 9 tests for #3911 |
| \	ests/plugins/globals/test_globals.py\ | 2 new test cases for #3939 |

## Test Plan

- [x] 55/55 tests pass
- [x] Substring and superstring bypass scenarios explicitly tested
- [x] Architectures override and inheritance both tested
- [x] No existing tests regressed

Fixes #3911
Fixes #3939

### DIFF
--- a/samtranslator/plugins/globals/globals.py
+++ b/samtranslator/plugins/globals/globals.py
@@ -321,6 +321,11 @@ class GlobalProperties:
     Object holding the global properties of given type. It also contains methods to perform a merge between
     Global & resource-level properties. Here are the different cases during the merge and how we handle them:
 
+    **List Override Properties**
+    Some list properties only accept a single value (e.g. Architectures — Lambda supports exactly one
+    architecture). For these, a resource-level list replaces the global list entirely rather than being
+    concatenated with it.
+
     **Primitive Type (String, Integer, Boolean etc)**
     If either global & local are of primitive types, then we the value at local will overwrite global.
 
@@ -436,6 +441,9 @@ class GlobalProperties:
 
     """
 
+    # List properties where the resource-level value replaces the global value instead of being concatenated.
+    _list_override_properties: frozenset[str] = frozenset({"Architectures"})
+
     def __init__(self, global_properties) -> None:  # type: ignore[no-untyped-def]
         self.global_properties = global_properties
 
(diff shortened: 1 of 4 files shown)


vicheey:
Hi @Hardikrepo

Thanks for the contribution! The #3939 (Architectures concatenation) has been fixed in PR #3945. So that portion of this PR is now redundant.

The #3911 fix (logical ID uniqueness bypass) looks like a separate valid issue though. Would you mind splitting it into its own PR so it can be reviewed independently? That way we don't have to untangle the two changes.

Thanks again for diving into these!

Issue #3791, opened 2025-07-03
Feature: Support for `RetentionInDays` and `Retain` in `LoggingConfig` for `AWS::Serverless::Function`


aler95:
### **Description:**

Currently, when defining a `AWS::Serverless::Function`, there is no native way to configure the retention period of the associated CloudWatch Log Group. The only available workaround is to manually create a corresponding `AWS::Logs::LogGroup` resource with a matching name and desired retention period. This adds complexity and creates potential deployment issues (e.g., log group already existing or being created out of order).

It would be beneficial to support log group retention directly in the `LoggingConfig` property of the `AWS::Serverless::Function` resource.

---

### **Proposed Solution:**

Extend the `LoggingConfig` property to support the following fields:

```yaml
MyFunction:
  Type: AWS::Serverless::Function
  Properties:
    LoggingConfig:
      RetentionInDays: 7        # Optional: number of days to retain logs (default: none/infinite)
      Retain: true              # Optional: preserve log group on resource deletion (default: false)

**Comment by sanyamk23** [comment]
Hi! Thanks for raising this. Having native support for log group retention and retention policies in LoggingConfig would make things simpler and less error-prone.

Proposed approach:
I’d suggest extending the `LoggingConfig` property for `AWS::Serverless::Function` to support both `RetentionInDays` and `Retain`. This way, users could directly set how long logs are kept and whether the log group should be preserved on function deletion, all within the function’s resource definition. For example:

```
MyFunction:
  Type: AWS::Serverless::Function
  Properties:
    LoggingConfig:
      RetentionInDays: 7      # Optional: number of days to retain logs
      Retain: true            # Optional: retain log group on deletion
```

If these properties are set, the construct would configure the log group accordingly; if not, it would fall back to the default (infinite retention, no retain policy). This would avoid the need for manual `AWS::Logs::LogGroup` resources and reduce deployment headaches.

If no one else is working on this, I’d be happy to start working on a PR. Does this approach make sense, or is there anything else I should consider? Let me know if you’d like me to move forward or if there’s more context to keep in mind. Thanks!

seshubaws:
Hi @sanyamk23 thank you for looking into this! I think your design sounds good, we can just pass both new properties to through to the LogGroup resource to be set on create. Feel free to put up a PR for this!

aler95:
Is there any news on this?

vicheey:
Hi @aler95, thank you for this feature request.

[...]

Issue #2057, opened 2021-06-12
Ability to customize CloudWatch LogGroups for Lambda Functions


esteban1983cl:
### Describe your idea/feature/enhancement

Provide a clear description.

Provide definition to configure cloudwatch log grups in SAM Template
Example: 

```yaml
Resources:
  DevOpsTaggerFunction:
    Type: AWS::Serverless::Function # More info about Function Resource: https://github.com/awslabs/serverless-application-model/blob/master/versions/2016-10-31.md#awsserverlessfunction
    Properties:
      CodeUri: ../../tagger_function
      Handler: tagger/app.lambda_handler
      Runtime: python3.8
      CloudWatchLogGroup:
         Retention: 30d
         Name: CustomName
         Tags:
         - Key: Key1
            Value: Value1
```

Things to consider:
[x] The [SAM documentation](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-specification.html) will need to be updated

hawflau:
Hi @esteban1983cl , can you elaborate more about what you want to achieve?

Logs for a Lambda Function are pushed to the log group with name `/aws/lambda/<function name>`. It does not seem possible to specify which log group to push logs to from Lambda's perspective.

If you just want to change the retention period of the log group that the Lambda function pushes logs to, you can do something like this:
```
Resources:
  ServerlessFunction:
    Type: AWS::Serverless::Function 
    Properties:
      CodeUri: ../../tagger_function
      Handler: tagger/app.lambda_handler
      Runtime: python3.8
      Role:
        Fn::GetAtt:
        - LambdaRole
        - Arn

  ServerlessFunctionLogGroup:
    Type: AWS::Logs::LogGroup
    Properties:
      LogGroupName: !Sub "/aws/lambda/${ServerlessFunction}"
      RetentionInDays: 300

  LambdaRole:
    Type: AWS::IAM::Role
    Properties:
      AssumeRolePolicyDocument:
        Version: '2012-10-17'
        Statement:
        - Effect: Allow
          Principal:
            Service: lambda.amazonaws.com
          Action: sts:AssumeRole
      Policies:
      - PolicyDocument:
          Version: '2012-10-17'
          Statement:
          - Effect: Allow
          Action:
          - logs:CreateLogStream
          - logs:PutLogEvents
          Resource:
[...]

Issue #1216, opened 2019-10-23
Feature: Create CloudWatch Log groups for Functions


theburningmonk:
**Description:**

CloudWatch Log Groups are not deleted when a stack is deleted. Leaving behind lots of abandoned log groups in CloudWatch Logs.

This happens because SAM doesn't explicitly create the log groups in the generated CloudFormation stack, and rely on Lambda to generate them at runtime, so the log groups are not part of the stack and don't get deleted when the stack is deleted.

**Steps to reproduce the issue:**
1. Create a template with Lambda functions
2. Package
3. Deploy
4. Invoke the functions a few times
5. Delete stack

**Observed result:**
The log groups remain in the account.

**Expected result:**
The log groups should be deleted.

jlhood:
@theburningmonk Thanks for reporting this issue! We're looking at adding SAM support to give users more control over the log group associated with their Lambda function. We could add this as part of that feature although I think it would need to be an optional setting since some customers may want to leave log groups around even in the event of a function deletion for compliance reasons.

jlhood:
@53ningen Thanks for the workaround. There are some caveats to using it that I want to call out:

The way you have it, you're allowing CloudFormation to name your Lambda function and then referencing it when creating the CW Logs group. This opens a race condition where if the function is invoked prior to the CW Logs group being created by CloudFormation, the CloudFormation deployment will fail with a log group already exists error. We've seen this happen, for example, when adding a new Lambda function that listens on an existing DynamoDB/Kinesis stream, SNS topic, or SQS queue.

A way to close this race condition is to explicitly name the function and then add a DependsOn to the LogGroup to ensure it gets created first. This is not ideal since you now have to name your own function (which has its own issues), but it does ensure this race condition does not happen.

Again, not against people using the workaround, just wanted to call out pitfalls.

Internally, we're discussing a way for SAM to support this natively without these pitfalls. My ideal is to add logging properties to `AWS::Serverless::Function` that make it possible to configure properties on the underlying LogGroup and have SAM handle all of these race conditions auto-magically for you under the covers. Hoping to post an RFC issue for that in the next couple months.

brysontyrrell:
@jlhood I'd like to +1 on configuring logging as a property of the `AWS::Serverless::Function`.

```yaml
MyFunction:
  Type: AWS::Serverless::Function
  Properties:
    Logging:
[...]

PR #3957, opened 2026-07-14
feat: add LogGroup property to AWS::Serverless::Function for stack-managed CloudWatch log groups

GuidoNebiolo:
### Issue #, if available
Refs #2665, #3791, #1216, #2057

### Description of changes
Adds a new `LogGroup` property to `AWS::Serverless::Function` that generates a stack-managed `AWS::Logs::LogGroup`, so the log group is created/updated/deleted together with the stack and its retention is configurable.

Two naming mechanisms:
- **Default** (no `LogGroupName`): the group uses the conventional `/aws/lambda/<function-name>` name via `Fn::Sub`. `LoggingConfig` is left untouched and no extra IAM is needed — the name coincides with Lambda's default destination and `AWSLambdaBasicExecutionRole` already grants log writes.
- **Custom name** (`LogGroupName`): binds the function via its native `LoggingConfig.LogGroup`, merging with any existing `LoggingConfig`.

Supported sub-properties: `RetentionInDays` (pass-through), `DeletionPolicy` (`Delete` default, or `Retain`) applied to both `DeletionPolicy` and `UpdateReplacePolicy`, and `LogGroupName`. Inheritable via `Globals`. When the property is absent, the transform output is unchanged (backward compatible).

This makes retention configurable and deletes the log group with the stack, addressing orphaned log groups (#1216).

- Added transform tests across all partitions (aws, aws-cn, aws-us-gov): default name, custom name with LoggingConfig wiring, `DeletionPolicy: Retain`, `Condition` propagation, Globals inheritance/override, and an invalid `DeletionPolicy` error case.
- Added unit tests for the `LogGroup` resource model and `_construct_log_group`.
- `make pr` is green: ruff, mypy --strict, format-check (schema regenerated), full test suite passing with 95.64% coverage.

### AI usage disclosure
This contribution was developed with the assistance of a generative AI coding tool. All code and tests were reviewed and are fully understood by the author, who is able to answer maintainer questions during review. The generated code was held to the same quality standards as hand-written code, and the diff/tests were kept minimal to ease review.

### Checklist
- [x] Review the generative AI contribution guidelines
- [x] Adheres to the development guidelines
- [x] Add/update transform tests
    - [x] Using correct values
    - [x] Using wrong values
- [ ] Add/update integration tests


valerena:
I don't see any check or guardrails in case a user sets `LoggingConfig` and this new `LogGroup` at the same time. What happens if the log group names are different? 

My initial instinct would be to prefer adding the log group name and other information as part of `LoggingConfig`, something like:

```yaml
MyFunction:
  Properties:
    ..
    LoggingConfig:
       CreateLogGroup: true
       LogGroup: "..."
       AdditionalProperties:
         DeletionPolicy: ...
         ...
```

Although it is a little more verbose, but it would prevent users from having the log group information separate in two different places. 

There's also the option of just accomplishing the same by just manually defining the LogGroup outside of the function directly. 

```
MyFunction:
  Properties:
    ..
    LoggingConfig:
       LogGroup: !Ref MyLogGroup

MyLogGroup:
  Type: AWS::Logs::LogGroup
    Properties:
      LogGroupName: ...
      RetentionInDays: ...
```
which is fairly straightforward.

The only case where this is not that simple is when you want to continue using Lambda's default name `/aws/lambda/<function-name>`, where the log group would depend on the function name, so the function would be created first and you can't just make both resources depend on each other.

The other issue when you want to keep using Lambda's default name (in your solution too, and it's mentioned in a few of the linked GitHub issues) is the fact that if for some reason the function gets invoked before the log group gets created, then Lambda will still create the log group, and the CFN deployment would fail when it tries to create the log group. 
This is more noticeable if you want to "add the LogGroup property to an existing function". In this case, the function likely already has an existing log group, so the log group can't be created via CFN because it already exist (you would have to change the name of the log group, or "import the log group" to the stack, or delete the previous log group so CFN can recreate it).

The point is that there are some edge cases that makes this not "as simple" as adding a new property. We're very interested in making a good solution for customers, but we don't want to have a confusing experience that can make it harder for new customers. Let me know your opinions about the above issues, and we'll continue the discussion internally with the team too.

GuidoNebiolo:
Thanks @valerena , these are really helpful points. Going through them:

1. **No guardrail when both `LoggingConfig` and `LogGroup` are set**. You're right, this is a gap. Today a custom `LogGroupName` silently overwrites `LoggingConfig.LogGroup`, and with the default name I leave `LoggingConfig` untouched (so a user-set `LoggingConfig.LogGroup` would point elsewhere and the managed group would receive no logs). **I'll add validation that raises a clear error when both define the destination, so it's never set in two places.**

2. **Extending `LoggingConfig` vs a dedicated property**. I chose a dedicated property mainly because `LoggingConfig` is a pure pass-through to AWS::Lambda::Function, and guideline #3 (stay close to the underlying CFN properties) pushes us to keep pass-throughs 1:1 with CloudFormation. Injecting SAM-only keys like `CreateLogGroup/AdditionalProperties` would force SAM to parse and strip them before passing through, and could collide if Lambda later adds real fields there. That said, the "info split in two places" concern is valid, I'm happy to go the `LoggingConfig` route if the team prefers that trade-off. **Just wanted to flag the pass-through implication**.

3. **You can already define the log group manually**. Agreed, and I don't think we should reinvent the `!Ref MyLogGroup` path. The case it doesn't cover cleanly is keeping Lambda's default `/aws/lambda/<function-name>` name while managing the group in the stack (the dependency on the possibly auto-generated function name makes it awkward). That's the case the linked issues (#1216, #2665) are about, so I can scope the feature primarily around it.

4. **"Already exists" on first deploy / retrofit**. Agreed, this is inherent to managing the default-named group in any approach, and mostly affects retrofitting a function that already has a log group. **I don't think the transform can solve it**, but I can document the behavior and mitigations (rename, resource import, or delete the orphan once). New functions aren't affected.

Proposal: add the conflict validation from (1), lean the feature toward the default-name use case from (3), and document (4). On (2) I'm open, do you prefer extending LoggingConfig or a dedicated property?

Issue #3973, opened 2026-08-13
transform() mutates the caller's template; transforming the same template twice fails


Adityaj0:
### Description

`transform()` edits the template it is given. `Translator.translate()` documents the opposite:

> :returns: a **copy of the template** with SAM resources replaced with the corresponding CloudFormation

What actually happens, measured on `main` (`ab5503a2`) by deep-copying the input, transforming, and comparing:

| template | caller's dict after `transform()` |
|---|---|
| bare function | unchanged |
| function + `Globals` | **`Globals` section deleted; merged `Timeout` written into the resource's `Properties`** |
| function + `Api` event | **function properties rewritten in place** |
| function + explicit `AWS::Serverless::Api` | **generated `DefinitionBody` written into the `Api` resource** |
| function + explicit `AWS::Serverless::HttpApi` | **generated `DefinitionBody` written into the `Api` resource** |
| explicit `Api` alone, state machine | unchanged |

The consequence is worse than the mutation itself. Because the first transform leaves its own generated `DefinitionBody` behind in the caller's template, **transforming the same object a second time fails**:

```
InvalidDocumentException: Event with id [Api] is invalid.
API method "get" defined multiple times for path "/x".
```

This hits any caller that holds on to the template — transforming once per region or partition, tooling that shows the user their original template after transforming, or anything that simply calls `transform()` twice. The error message points at the user's event definition, which is not where the problem is.

I checked for cross-invocation state leakage as well (transform template A, then B, and compare B against a clean run): that is clean. The damage is confined to the object passed in.

### Steps to reproduce

```python
import copy
from unittest.mock import MagicMock, patch
from samtranslator.translator.transform import transform

template = {
    "Transform": "AWS::Serverless-2016-10-31",
    "Resources": {
        "Fn": {
            "Type": "AWS::Serverless::Function",
            "Properties": {
                "CodeUri": "s3://bucket/key", "Handler": "index.handler", "Runtime": "python3.11",
                "Events": {"Api": {"Type": "Api", "Properties": {
                    "Path": "/x", "Method": "get", "RestApiId": {"Ref": "Api"}}}},
            },
        },
        "Api": {"Type": "AWS::Serverless::Api", "Properties": {"StageName": "prod"}},
    },
}

before = copy.deepcopy(template)
loader = MagicMock()
loader.load.return_value = {}

with patch("boto3.session.Session.region_name", "us-east-1"):
    transform(template, {}, loader)
[...]

PR #3974, opened 2026-08-13
fix: do not mutate the caller's template during transform

Adityaj0:
### Issue #, if available

#3973

### Description of changes

`Translator.translate()` documents that it returns "a copy of the template", but the parser and plugins edit the template in place: `Globals` is deleted from it, merged global properties are written into resource properties, and generated API definition bodies are written into explicit `AWS::Serverless::Api`/`HttpApi` resources. A caller that transforms the same template twice gets `InvalidDocumentException: API method "get" defined multiple times for path "/x"` on the second call.

This takes a copy at the `transform()` boundary.

The copy is taken **before** `to_py27_compatible_template()` on purpose. That function installs `Py27Dict`/`Py27UniStr` wrappers whose hash-ordering state logical ID generation depends on, and `copy.deepcopy` does not preserve it — putting the copy at the top of `translate()` instead changes generated logical IDs, which `test_transform_feature_toggle_0_feature_toggle_api_open_api_version_override` catches. Copying while the template is still plain dicts and strings avoids that.

Scope: this fixes the public `transform()` entry point. `Translator.translate()` called directly still mutates its argument; correcting that needs `__deepcopy__` support on the Py27 types, which I've left alone.

Four tests in `tests/translator/test_transform_does_not_mutate_input.py`, one per mutation path (Globals, implicit API event, explicit `Api`, and the double-transform failure for both REST and HTTP APIs). All four fail on `develop` and pass with this change.

Existing suites: `tests/translator` 2174 passed, and `tests/translator tests/plugins tests/parser` 2577 passed. Five failures in `tests/plugins/application/test_serverless_app_plugin.py` are pre-existing on `develop` (timing-based SAR tests) and fail identically without this change.

Overhead of the copy on a 246 KiB, 200-function template: 3.2 ms against a 1924 ms transform, 0.2%.

### Checklist

    - [x] Using correct values
    - [ ] Using wrong values

### DIFF
--- a/samtranslator/translator/transform.py
+++ b/samtranslator/translator/transform.py
@@ -1,3 +1,4 @@
+import copy
 from functools import cache
 from typing import Any
 
@@ -22,6 +23,20 @@ def transform(
     :returns: the transformed CloudFormation template
     :rtype: dict
     """
+
+    # Work on our own copies: the parser and plugins mutate the template in place --
+    # Globals are merged into resource properties and the Globals section itself is
+    # removed -- and to_py27_compatible_template() below mutates parameter_values in
+    # place, replacing its values with Py27UniStr/Py27Dict/Py27LongInt wrappers.
+    # Callers hand us objects they may still need afterwards, or may transform more
+    # than once.
+    #
+    # Both copies are taken before to_py27_compatible_template() so that only plain
+    # dicts and strings are copied. The Py27Dict/Py27UniStr wrappers that function
+    # installs carry hash-ordering state that logical ID generation depends on, and
+    # deep-copying them does not preserve it.
+    input_fragment = copy.deepcopy(input_fragment)
+    parameter_values = copy.deepcopy(parameter_values)
 
     sam_parser = Parser()
     to_py27_compatible_template(input_fragment, parameter_values)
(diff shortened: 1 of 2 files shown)


Adityaj0:
Good catch — fixed. \`transform()\` now also deep-copies \`parameter_values\` before \`to_py27_compatible_template()\`, at the same boundary as \`input_fragment\`, matching what \`SamParameterValues.__init__\` already does for \`Translator.translate()\`.

Added \`test_parameter_values_are_not_modified\`, which passes a non-empty \`parameter_values\` dict (str, int, and nested dict values) through a template with an explicit API resource (the only path that exercises the mutation) and asserts both value equality and exact type (\`Py27UniStr\` subclasses \`str\`, so a plain \`==\` check wouldn't have caught this). Confirmed it fails on the previous commit and passes now.

Full suite: \`tests/translator tests/plugins tests/parser\` — 2578 passed (same 5 pre-existing, unrelated SAR-timing failures as before).

Adityaj0:
Addressed the formatting comment on `tests/translator/test_transform_does_not_mutate_input.py:10`: that dict literal was 130 chars, over the 120-char `black` limit in `pyproject.toml`. Confirmed `black --check` failed on `develop`-based checkout before this commit and passes after. Ran `black` on the file (only whitespace/line-break changes, no logic touched) and re-ran `tests/translator/test_transform_does_not_mutate_input.py` — all 5 tests still pass. Pushed as 647fb7d.

PR #3978, opened 2026-08-14
fix: local PropagateTags: false is not overridden by global PropagateTags: true

### LINKED ISSUE #3977: Implicit API: local PropagateTags: false is silently overridden by global PropagateTags: true
### Describe the bug

`ImplicitApiPlugin._add_tags_to_implicit_api_if_necessary()` decides whether to copy a resource's `Tags` onto its generated implicit `AWS::Serverless::Api`/`HttpApi` resource using:

```python
should_propagate_tags = resource.properties.get("PropagateTags") or globals_var.get("PropagateTags")
```

`PropagateTags` is documented as `bool | None` (see `samtranslator/internal/schema_source/aws_serverless_function.py`), where `None` means "not set" and `False` is a meaningful, explicit opt-out. Using `or` means a local `PropagateTags: false` is indistinguishable from "not set" — it silently falls through to the global value.

### Reproduction

```yaml
Globals:
  Function:
    PropagateTags: true
    Tags:
      Team: Data

Resources:
  MyFunction:
    Properties:
      PropagateTags: false        # explicit opt-out
      Runtime: python3.12
      Handler: index.handler
      CodeUri: s3://bucket/key
      Events:
        Api:
          Type: Api
          Properties:
            Path: /hello
            Method: get
```

### Expected behavior

The function explicitly disables `PropagateTags`, so the generated implicit `AWS::Serverless::Api` should NOT receive the `Team: Data` tag.

### Actual behavior

`resource.properties.get("PropagateTags")` returns `False` (falsy), so `or` evaluates the global value (`True`) instead, and the tag is propagated to the implicit API anyway — silently ignoring the resource-level override.

This is inconsistent with the standard Globals merge logic (`GlobalProperties._prefer_local` in `samtranslator/plugins/globals/globals.py`), which always prefers the local value when the key is present locally, regardless of its truthiness.

### Fix

PR incoming.

Adityaj0:
## Summary

Fixes #3977.

`ImplicitApiPlugin._add_tags_to_implicit_api_if_necessary()` decided whether to copy a resource's `Tags` onto its generated implicit `AWS::Serverless::Api`/`HttpApi` resource with:

```python
should_propagate_tags = resource.properties.get("PropagateTags") or globals_var.get("PropagateTags")
```

`PropagateTags` is `bool | None` (see the Pydantic schema in `samtranslator/internal/schema_source/aws_serverless_function.py`), where `None` means "not set" and `False` is a meaningful, explicit opt-out. `or` treats a local `False` the same as "not set," so it silently falls through to the global value — a function with an explicit `PropagateTags: false` still gets tags propagated to its implicit API if a global `PropagateTags: true` is set.

This checks `is not None` instead, so an explicit local value (including `False`) always wins over the global, consistent with the "local wins when present" semantics `GlobalProperties._prefer_local` already uses for the standard Globals merge (`samtranslator/plugins/globals/globals.py`).

## Test plan

- [x] Added `tests/translator/input/function_with_local_propagate_tags_false_overrides_global_true.yaml` (local `PropagateTags: false` + global `PropagateTags: true` + global `Tags`) with expected output for all three partitions (`aws`, `aws-cn`, `aws-us-gov`), asserting the implicit API resource has no `Tags`.
- [x] Verified the new test fails against the pre-fix code (tags leak through) and passes with the fix.
- [x] `python -m pytest tests/translator tests/plugins tests/parser` — 2581 passed (same 5 pre-existing, unrelated SAR-timing/region failures as on `develop`).
- [x] `ruff check` / `black --check` clean on the changed file.

### DIFF
--- a/samtranslator/plugins/api/implicit_api_plugin.py
+++ b/samtranslator/plugins/api/implicit_api_plugin.py
@@ -112,7 +112,7 @@ class ImplicitApiPlugin(BasePlugin, Generic[T], metaclass=ABCMeta):
         should_propagate_tags = (
             local_propagate_tags if local_propagate_tags is not None else globals_var.get("PropagateTags")
         )
-        tags_properties = resource.properties.get("Tags") or globals_var.get("Tags")
+        tags_properties = {**(globals_var.get("Tags") or {}), **(resource.properties.get("Tags") or {})}
 
         if implicit_api_resource and tags_properties and should_propagate_tags:
             # This makes an assumption that the SAM resource has 'Tags' property and is a dictionary.
--- a/samtranslator/plugins/api/implicit_api_plugin.py
+++ b/samtranslator/plugins/api/implicit_api_plugin.py
@@ -108,7 +108,10 @@ class ImplicitApiPlugin(BasePlugin, Generic[T], metaclass=ABCMeta):
 
         implicit_api_resource = template.get(self.IMPLICIT_API_LOGICAL_ID)
         globals_var = template.get_globals().get(SamResourceType(resource.type).name) or {}
-        should_propagate_tags = resource.properties.get("PropagateTags") or globals_var.get("PropagateTags")
+        local_propagate_tags = resource.properties.get("PropagateTags")
+        should_propagate_tags = (
+            local_propagate_tags if local_propagate_tags is not None else globals_var.get("PropagateTags")
+        )
         tags_properties = resource.properties.get("Tags") or globals_var.get("Tags")
 
         if implicit_api_resource and tags_properties and should_propagate_tags:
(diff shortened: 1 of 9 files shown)


Adityaj0:
Agreed, and fixed rather than split out — the two lines were left inconsistent otherwise. Line 115 now does:

`tags_properties = {**(globals_var.get("Tags") or {}), **(resource.properties.get("Tags") or {})}`

Verified with an independent repro (Globals.Function.Tags={env: prod}, resource Tags={app: foo}, PropagateTags: true): before this change ServerlessRestApi got only {app: foo}; after, it gets {env: prod, app: foo} — matching the merge the resource itself already gets from Globals. Added tests/translator/input/function_with_local_and_global_tags_merge.yaml with expected output for all three partitions; new tests pass, plus the existing 33 propagate_tags tests and black/ruff both clean. Pushed as f8fe3c7.

PR #3990, opened 2026-09-02
feat: improve AuthType validation for WebSocketApi

valerena:
### Issue #, if available

Invalid values that looked like correct values but with the wrong casing, were previously being ignored, because their value was passing validation but it wasn't recognized as valid, and eventually being translated to `NONE`. 

### Description of changes

The templates that had `AuthType` values with the wrong casing now will explicitly fail instead of silently being ignored. A consequence of this is that templates that were using `None` or `none`, which were having "the expected behavior" of having "no auth", will now also fail (just like other mismatched casing of valid values).

This is a backwards-incompatible change, which we accept in favor of stricter validation. Existing deployed stacks won't change on their own, but the next transform will fail until the user corrects the invalid value.

We also **improved the error message returned**, so any customer can clearly see the specific value that they passed that was invalid.

The WebSocketApi resource was launched earlier this year, and its usage is low compared to other resources. This situation with `AuthType` is a niche issue that is limited to a very small portion of customers, and that's why accept the backwards incompatibility.

- `make pr`
- Add extra tests.

### Checklist

    - [ ] Using correct values
    - [x] Using wrong values

### Examples?


valerena:
From the bot comment:
> This is not a regression — previously auth_type.upper() raised AttributeError on the same input, which is worse 

This is correct. We've never supported intrinsics on this field (`CUSTOM` requires more properties to be passed, so it would be hard to use intrinsics only for the value of `AuthType`). So for now it's okay. We haven't seen data about anyone running into this issue, but we might support it in the future if we hear from customers.

roger-zhangg (approved):

licjun (approved):

PR #3991, opened 2026-09-04
fix: include API stage variables in the deployment hash

### LINKED ISSUE #3703: ApiGatewayDeployment.make_auto_deployable does not account for API Stage Variables
Stage variable updates will cause an API Gateway UpdateStage call during SAM API stack deployment. However, since API Stage variables are not considered as hash input to ApiGatewayDeployment.make_auto_deployable, an update to a SAM API Stage can be made without a new deployment. 

https://github.com/aws/serverless-application-model/blob/17000c195d65a41ee830173e7110ad2c36078a7a/samtranslator/model/apigateway.py#L95-L103

This means 

1. The API Stage variables updates are not 'deployed', 
2. UpdateStage will reset the Active API Deployment to it's last known good API Deployment ID. **If you have made other non-SAM managed deployments since your last SAM managed deployment, a latest change with just API Stage Variable changes will revert the active deployment Id to the last known good deployment**. This effectively reverts your latest changes made outside of the SAM managed changes.

----
My team uses scripts to adjust the API Resource Policy (to onboard new clients), then deploy those authz changes using create-deployment, The above issue 2) recently caused an issue where our latest onboarding Resource Policy changes were no longer active after an API Stage Variable only change SAM deployment

REPRO: 
1. Deploy SAM Api
2. Manually update the API Resource Policy
3. Manually deploy the API
4. Update SAM definition to add / change an API Stage Variable
5. Deploy the updated SAM definition

OBSERVE:
1. Step 5) creates an CFN Changeset with the update to the API Stage variables.
2. Processing that change set produces a Cloud Trail UpdateState Event where you can see both the new variable details, ,but also the 'reset' of the deploymnent value to something old.

IMPACT:
Your manual policy update step 2) is no longer active.

warren830:
## Which issue(s) does this change fix?

Fixes #3703

## Why is this change necessary?

`ApiGatewayDeployment.make_auto_deployable` builds its hash from the swagger body, the OpenAPI version, the custom domain and the associated function names — but not from the stage's `Variables`. Since `AWS::ApiGateway::Stage` does carry `Variables`, a template change that only touches stage variables produces the **same** deployment logical id, so CloudFormation creates no new `AWS::ApiGateway::Deployment`.

That has two effects, both described in the issue:

1. The stage-variable change is never actually deployed.
2. The `UpdateStage` call points the stage back at the deployment it already references. If any deployment was made outside SAM since the last SAM deployment, the stage silently reverts to the SAM-known deployment id — losing the out-of-band changes.

## How does it address the issue?

Stage variables are appended to `hash_input`, following the existing `if openapi_version:` / `if domain:` pattern.

They are appended **only when set**, which is the important part for backwards compatibility: templates that define no stage variables keep their current deployment logical id and see no churn. `sort_keys=True` keeps the hash stable regardless of key order in the template.

## What side effects does this change have?

Templates that **do** set stage variables get a new deployment logical id once, so the next deploy creates a new deployment. That is the intended behaviour — it is exactly the deployment that was previously being skipped — but it is a one-time redeploy for those stacks, so calling it out explicitly.

The 24 updated files under `tests/translator/output/` are that same effect in the checked-in expected outputs: only the deployment logical id and its `RestApi deployment id: <sha>` description changed. Eight testcases are affected (`explicit_api`, `explicit_api_openapi_3`, `api_with_openapi_definition_body_no_flag`, `api_with_swagger_and_openapi_with_auth`, `function_with_alias_and_event_sources`, `global_handle_path_level_parameter`, `globals_for_api`, `intrinsic_functions`) across the three partition folders. I rewrote just those identifiers rather than regenerating the files, so the diff stays reviewable.

## Testing

Verified on `develop` at `d24401d`, Python 3.12:

- Added three tests to `tests/translator/test_api_resource.py`:
  - stage variables are part of the hash input;
  - two stages differing **only** in variables produce different deployment logical ids (this is the regression test — it fails before the change);
  - a stage with no variables produces the same logical id as one with `{}`, which is the no-churn guard.
- Before the change: the first two fail. After: all pass.
- `tests/translator/` — 2423 passed, 0 failed (42 failed before the fixtures were refreshed).
- Whole suite: 268 failed / 4397 passed **with** the change versus 268 failed / 4394 passed on an otherwise clean tree — same failures either way (they are pre-existing environment failures in this sandbox, e.g. `test_region_configuration`, and need network/endpoint data), plus the 3 new passing tests. No regressions.
- `ruff check` (pinned `ruff~=0.15.6`) and `black --check` clean on both changed Python files; `git diff --check` clean.

One existing test needed a one-line change: `test_make_auto_deployable_with_swagger_dict` uses `stage = MagicMock()`, whose `.Variables` is a truthy `MagicMock` rather than the unset value a real `ApiGatewayStage` has, so it now sets `stage.Variables = None` to express "no stage variables".

### DIFF
--- a/samtranslator/model/apigateway.py
+++ b/samtranslator/model/apigateway.py
@@ -131,6 +131,14 @@ class ApiGatewayDeployment(Resource):
             hash_input.append(str(openapi_version))
         if domain:
             hash_input.append(json.dumps(domain))
+        # Stage variables are applied with an UpdateStage call, which points the stage back at the
+        # deployment it already references. If they are not part of this hash, a variables-only
+        # change reuses the existing deployment: the new variables are never deployed, and any
+        # deployment made outside SAM since the last SAM deployment is reverted.
+        # Only added when set, so templates without stage variables keep their existing hash.
+        stage_variables = getattr(stage, "Variables", None)
+        if stage_variables:
+            hash_input.append(json.dumps(stage_variables, sort_keys=True))
         function_names = redeploy_restapi_parameters.get("function_names") if redeploy_restapi_parameters else None
         # The deployment logical id is <api logicalId> + "Deployment"
         # The keyword "Deployment" is removed and all the function names associated with api is obtained
--- a/tests/translator/output/api_with_openapi_definition_body_no_flag.json
+++ b/tests/translator/output/api_with_openapi_definition_body_no_flag.json
@@ -55,9 +55,9 @@
       },
       "Type": "AWS::ApiGateway::RestApi"
     },
-    "ExplicitApiDeployment9252467a1e": {
+    "ExplicitApiDeploymente24cb02185": {
       "Properties": {
-        "Description": "RestApi deployment id: 9252467a1edc49ba35cb258640f5e3734cc9fab1",
+        "Description": "RestApi deployment id: e24cb0218583daf7c5e537a4e639dfb3a2971fdf",
         "RestApiId": {
           "Ref": "ExplicitApi"
         },
@@ -70,7 +70,7 @@
         "CacheClusterEnabled": true,
         "CacheClusterSize": "1.6",
         "DeploymentId": {
-          "Ref": "ExplicitApiDeployment9252467a1e"
+          "Ref": "ExplicitApiDeploymente24cb02185"
         },
         "RestApiId": {
           "Ref": "ExplicitApi"
@@ -230,9 +230,9 @@
       },
       "Type": "AWS::ApiGateway::RestApi"
     },
-    "ServerlessRestApiDeploymentdb4b9da82a": {
+    "ServerlessRestApiDeployment62318f9715": {
       "Properties": {
-        "Description": "RestApi deployment id: db4b9da82adc6031fcd32bf3a4954485464fc009",
+        "Description": "RestApi deployment id: 62318f9715b2b5932b597575f6a4a0eafd0cf7e3",
         "RestApiId": {
           "Ref": "ServerlessRestApi"
         },
@@ -245,7 +245,7 @@
         "CacheClusterEnabled": true,
         "CacheClusterSize": "1.6",
         "DeploymentId": {
-          "Ref": "ServerlessRestApiDeploymentdb4b9da82a"
+          "Ref": "ServerlessRestApiDeployment62318f9715"
         },
         "RestApiId": {
           "Ref": "ServerlessRestApi"
--- a/tests/translator/output/api_with_swagger_and_openapi_with_auth.json
+++ b/tests/translator/output/api_with_swagger_and_openapi_with_auth.json
@@ -54,9 +54,9 @@
       },
       "Type": "AWS::ApiGateway::RestApi"
     },
-    "ExplicitApiDeployment7c4f7dda23": {
+    "ExplicitApiDeployment8e20079165": {
       "Properties": {
-        "Description": "RestApi deployment id: 7c4f7dda23acd71e4a653861510d82ad7809e562",
+        "Description": "RestApi deployment id: 8e20079165a6421b14f28fc3f812a282e9ca3d90",
         "RestApiId": {
           "Ref": "ExplicitApi"
         },
@@ -69,7 +69,7 @@
         "CacheClusterEnabled": true,
         "CacheClusterSize": "1.6",
         "DeploymentId": {
-          "Ref": "ExplicitApiDeployment7c4f7dda23"
+          "Ref": "ExplicitApiDeployment8e20079165"
         },
         "RestApiId": {
           "Ref": "ExplicitApi"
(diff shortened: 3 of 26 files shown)


warren830:
Good catch — this was a real gap and I have fixed it in `a8fee86e`. (Replying here as a top-level comment rather than inline under the review thread, for tooling reasons on my side; it addresses the `[BUG]` note on `samtranslator/model/apigateway.py:139`.)

I reproduced it before changing anything: transforming a template whose stage variable is `{"Ref": "EndpointUri"}` with two different parameter values produced **the same** deployment logical id (`GetHtmlApiDeployment...`), so the original fix did nothing for the parameter-driven case, exactly as you described.

What changed:

- `SamApi.to_cloudformation` now resolves the stage variables for hashing purposes and passes them to `make_auto_deployable` as a new keyword-only `stage_variables` argument.
- As you recommended, `self.Variables` is **not** reassigned the way `Domain` is. `resolve_parameter_refs` mutates its argument and inlines values, so it resolves a `copy.deepcopy` and the resolved form is used only for the hash; the emitted `AWS::ApiGateway::Stage` keeps the customer's `Ref`.
- `make_auto_deployable` falls back to the stage's own variables when the argument is not supplied, so direct callers keep their current behaviour.

Tests added in `tests/translator/test_api_resource.py`:

- `test_redeploy_when_parameter_driven_stage_variable_changes` — two parameter values must give different deployment ids, and the same value must stay stable so an unchanged stack does not churn. It fails on the pre-fix code with `assert first != second` / "Both sets are equal", and passes after.
- `test_parameter_driven_stage_variables_are_not_inlined_into_stage` — asserts the emitted Stage still holds `{"EndpointUri": {"Ref": "EndpointUri"}}`, guarding the mutation hazard you flagged.

Fixtures: your point is already covered by the repo's own corpus — `tests/translator/input/explicit_api.yaml` sets `EndpointUri: {Ref: something}`, so the `GetHtmlApi` deployment id changed in the six `explicit_api` / `explicit_api_openapi_3` expected outputs. I substituted only the logical id, its `Ref`, and the `Description` digest in each file rather than regenerating them, so the diff shows only those three lines per fixture — and the Stage's `Variables` block is visibly untouched, which is the inlining check in fixture form.

Verification: `tests/translator/` is 2425/2425. Full suite compared against the branch point is 268 failed / 4399 passed versus 268 failed / 4397 passed at baseline — the failure set is identical line for line (pre-existing, `NoRegionFound` on a machine with no configured region) and the two extra passes are these tests. `ruff` (pinned 0.15.6), `black` and `mypy --strict` are clean, with mypy reporting the same 15 pre-existing errors as baseline and none in the changed files.

Two notes on judgement calls, in case you would rather they went the other way:

1. `make_auto_deployable` crossed the `PLR0913` argument limit, so it now carries `# noqa: PLR0913`, matching the 13 existing uses of that idiom in `samtranslator/`. The alternative was threading the value through the translator-level `redeploy_restapi_parameters` dict, which is shared across APIs and seemed worse than one more argument.
2. I pinned region resolution inside the new test helper rather than inheriting the ambient session, because the two pre-existing transform-based tests in this same file fail with `NoRegionFound` when the whole suite runs on a machine without a configured region. The new tests do not have that dependency.

Also worth flagging what I did **not** change: `HttpApi` and `WebSocketApi` go through `ApiV2Generator` and are untouched, so this stays scoped to the REST path in #3703.

warren830:
Review nudge — this PR is now fully green and has no open feedback.

- The automated `lambda-tooling pr reviewer` check went from **failure to success**: it reports *"No issues found. The changes look good."* on the current head. Its earlier failure was a genuine `[BUG]` finding, not infrastructure noise, and I have addressed it (see the comment above) rather than arguing it away.
- Branch is **up to date with `develop`**, and `mergeable` is true.
- Scope: 28 files, +264/-109 — 2 source files, 1 test file, and 25 expected-output fixtures where only the deployment logical id, its `Ref` and the `Description` digest changed.

It fixes #3703 (`type/bug`, `stage/bug-repro`, maintainer-reproduced): the API deployment hash omitted the stage `Variables`, so a variables-only change reused the existing deployment — the new variables were never deployed, and `UpdateStage` pointed the stage back at the deployment SAM already knew about, reverting anything deployed outside SAM since. The follow-up commit extends that to parameter-driven variables, which hashed identically for every parameter value.

Verification: `tests/translator/` is 2425/2425. Against the branch point the full suite is 268 failed / 4399 passed versus 268 failed / 4397 passed at baseline — the failure set matches line for line (pre-existing, `NoRegionFound` without a configured region) and the two extra passes are the new tests. Both new tests fail on the pre-fix code and pass after. `ruff` (pinned 0.15.6), `black` and `mypy --strict` are clean, mypy reporting the same 15 pre-existing errors as baseline and none in the changed files.

Two things worth a maintainer's eye, since they are judgement calls rather than mechanics:

1. **Stacks that already set API stage variables will get one deployment on the next `sam deploy`**, because their deployment logical id changes. That is inherent to including the variables in the hash — it is the fix — but it is a visible one-time effect and you may want it called out in release notes. Templates with no stage variables keep their existing logical id, and there is a test pinning that.
2. `make_auto_deployable` crossed the `PLR0913` argument limit and now carries `# noqa: PLR0913`, matching the 13 existing uses in `samtranslator/`. The alternative was routing the value through the shared `redeploy_restapi_parameters` dict; I judged one more argument cleaner, but I will restructure it if you prefer.

Happy to rebase, split the fixture churn into its own commit, or adjust anything else that would make review easier.

